PR #3, opened 2015-07-02
Change variable names to reduce confusion

ckimes89:
The `resolve` function for `friends` on the `droid` type has `human` as a variable name. This is somewhat confusing, since it will be called on a droid.


leebyron:
Good catch

PR #7, opened 2015-07-02
Proposition to change README examples to ES2015 style 

enaqx:
Due to GraphQL is also written in ES2015.


leebyron:
Yeah, I wondered about this, defaulting to something easily consumable. But I agree that people interested in this are pretty likely to know ES2015

enaqx:
Thanks for your work. GraphQL is awesome.

PR #8, opened 2015-07-02
Extract ESLint rules in config file

enaqx:
It should improve readability both package.json and eslint rules.


leebyron:
Indeed it does, great suggestion

Gerst20051:
:+1:

PR #10, opened 2015-07-02
Don't resolve when invalid

RStankov:
You are calling `execute` and double resolving even when document validation has failed.


leebyron:
Thanks! This integration point was a last minute addition and a rare undertested part of the library. Thank you for catching this, I'll follow up with better tests.

valentinzberea:
Honest question: is it possible to resolve a promise twice?

The way I understand promises is that once a promise is resolved, any subsequent calls to `resolve()` won't have any effect.

leebyron:
They don't have effect, no :)

PR #11, opened 2015-07-02
Misc comment typos

enaqx:


leebyron:
Where's my eslint rule for common spelling mistakes. Clearly we need that! Thanks again for fixing as you read :)

PR #12, opened 2015-07-02
Add headers where missed and update .gitignore 

enaqx:
Adds IntelliJ Webstorm directory exception.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ lib
 node_modules
 npm-debug.log
 coverage
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,9 @@
 *~
 .*.haste_cache.*
 .DS_Store
+.idea
+
 lib
 node_modules
 npm-debug.log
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ lib
 node_modules
 npm-debug.log
 coverage
+.idea
```
petrbela: you'd better put this into your `~/.gitignore` https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

enaqx: Yeah. I know about this feature. But lot of people use IntelliJ products and it can be convenient. React has Sublime in ignore list, for example https://github.com/enaqx/react/blob/master/.gitignore#L24

leebyron:
Can you move it to the top near the other IDE resource files in this list?

enaqx:
@leebyron Pls, look if I understood correctly about .gitignore.

leebyron:
Well done, thanks

PR #15, opened 2015-07-03
omit => emit

sevki:


leebyron:
Thank you, that would be confusing otherwise wouldn't it!

sevki:
indeed it would :smiley:

Issue #17, opened 2015-07-03
Broken promise behaviour when types used in list have async field resolve functions.


latentflip:
Okay, I'm still digging into this, and trying to simplify my examples, but it looks like something funky is going on with field resolve behaviour when promises are returned.

Here's example code which demonstrates the behaviour (as well as the results) https://gist.github.com/latentflip/f1c9520ac0d83b5237fc note how in the results, projectPromises differs from projectResults.

Okay, to explain further, the basic setup I have is:

```
type Org {
  id: String
  projects: [Project]
}

type Project {
  id: String
  versions: ProjectVersion
}

type ProjectVersion {
  id: String
}
```

and my root query just returns one org, so the query I'm executing is:

```
query Foo {
  org {
    id,
    projects {
      versions
    }
  }
}
```

which should return the org, all it's projects, and all their versions.

However, it seems if my `projects.versions` resolve function returns a promise, instead of immediately returning an array, something breaks. To demonstrate, in the actual code linked above, I've got a ProjectReturnType which just [returns an array for it's versions](https://gist.github.com/latentflip/f1c9520ac0d83b5237fc#file-input-js-L25-L30) and a ProjectPromiseType which [returns the same array, but wrapped in a `Promise.resolve()`](https://gist.github.com/latentflip/f1c9520ac0d83b5237fc#file-input-js-L47-L53). The latter seems to break the `projectPromises` in the response completely, and it just returns `{}`.

As I understand it, a resolve function returning a value, or a resolve function returning a promise of that same value, should be exactly equivalent, right?

latentflip:
Alright, well, I think I have a reasonable handle on the codebase now, thanks to a lot of node-inspectoring ;).

Here's a much simpler proof of the bug:

```
import {
  graphql,
  GraphQLSchema,
  GraphQLObjectType,
  GraphQLString,
  GraphQLNonNull,
  GraphQLList,
} from 'graphql';

var orgType = new GraphQLObjectType({
  name: 'Org',
  fields: () => ({
    id: {
      type: new GraphQLNonNull(GraphQLString),
      resolve: () => {
        return Promise.resolve(Math.random + '');
      }
    }
  })
});

var schema = new GraphQLSchema({
  query: new GraphQLObjectType({
    name: 'RootQueryType',
    fields: {
      orgs: {
        type: new GraphQLList(orgType),
        resolve: () => {
          return [{ id: 1 }]
        }
      }
    }
  }),
});


var query = `
query Foo {
[...]

PR #18, opened 2015-07-03
Fix async resolve promises not being resolved properly in list types.

latentflip:
Fixes: #17 

Made one of the fields in the tests return a promise to demonstrate the bug in #17, and added code as discussed in #17 to fix it.

`completeField` seemed to be the best place to put this logic, but let me know if it makes sense somewhere else and I can move it.


latentflip:
Signed the CLA.

PR #24, opened 2015-07-05
build: use && instead of &

vkurchatkin:
Using `&` looks like a typo. It causes `rm` to run in background, which introduces race condition


leebyron:
Good catch! Lucky our machines are faster at deleting folders than babel is at transpiling! 

Thanks for helping us improve

PR #29, opened 2015-07-06
Test recursion avoidance correctly

dittos:
`DataType` is unknown type, so `...Frag` fragment spread was ignored and recursion was not tested correctly.

Additional question: is it intended to just ignore a fragment on the unknown type?


leebyron:
Great catch, thanks!

leebyron:
Yes, that's correct. This is testing the behavior of *non-verified* queries, to ensure behavior is correct in the case a query is executed before being validated. 

If validated, the fragment recursion would have been caught before execution, as would the mistyped type information.

PR #31, opened 2015-07-06
Reworking of Directives

leebyron:
I think we have gotten directives slightly wrong. Since we're still in working draft, I think we should fix this:

Old:

```
{
  foo @if: true
  bar @unless: true
}
```

New:

```
{
  foo @include(if: true)
  bar @skip(if: true)
}
```

Now directives are just as powerful as fields in their ability to accept multiple argument values. More importantly, this is a *simplification* as directive arguments and field arguments are syntactically *identical*, and semantically extremely similar, allowing us to reuse spec, grammar, and impl for both.

This simplification is really valuable to fix some issues with directives today. Directive value nullability is poorly (and probably incorrectly) defined in the current form. What does a non-null type on a directive today mean? Directives cannot be required. Now the rules are the same as field arguments.

This also helps us support future directives that have no arguments or optional arguments, or directives that have multiple arguments. `@defer` alone is perfectly fine, and `@defer(priority: 3)` is much easier to read than `@defer: 3`.

So if we do this, then `@if` and `@unless` are replaced by `@include` and `@skip` respectively. I think this is helpful as it describes more accurately what the directives' jobs are.

So grammar becomes:

```
Directive : Name Arguments?
```

And the directive value validation rule is merged into the argument validation rule.


leebyron:
I'll start discussion here in the impl, and once everything is covered, I'll ensure the spec matches this behavior.

dschafer:
Yeah, I think this is a big improvement; I like the `skip` and `include` names too, I think that's clearer than `if` and `unless`.

leebyron:
I'll wait to land to give @schrockn and perhaps others a chance to look, and will get started on a PR for respective spec changes

schrockn:
This is an excellent change in terms of spec. Love it

leebyron:
Cool, everyone in favor - in it goes. Spec text next.

PR #35, opened 2015-07-08
Prevent registering more than one types on the same name

dittos:
Registering two different types with the same name was silently passing.

### DIFF
--- a/src/type/__tests__/schema.js
+++ b/src/type/__tests__/schema.js
@@ -1,0 +1,42 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import {
+  GraphQLSchema,
+  GraphQLObjectType,
+} from '../';
+
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+
+describe('Type System: Schema', () => {
+  it('does not allow more than one type of the same name', () => {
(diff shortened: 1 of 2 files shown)


devknoll:
I think it would be a little more natural if an error was thrown when `GraphQLSchema` was constructed, not later when `getTypeMap` is called

schrockn:
Agree with @devknoll .ctor seems a bit more deterministic/natural

dittos:
Changed to throw in the constructor, by building type map eagerly.

src/type/__tests__/schema.js
```diff
@@ -0,0 +1,42 @@
...
+import {
+  GraphQLSchema,
+  GraphQLObjectType,
+} from '../';
+
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+
+describe('Type System: Schema', () => {
+  it('does not allow registering more than one types on the same name', () => {
```
leebyron: "does not allow more than one type of the same name"

src/type/schema.js
```diff
@@ -85,11 +82,26 @@ type GraphQLSchemaConfig = {
...
     return typeMapReducer(map, type.ofType);
   }
-  if (!type || map[type.name]) {
+  if (!type) {
+    return map;
+  }
+  var prevType = map[type.name];
+  if (prevType) {
+    if (prevType !== type) {
+      throw new Error('Already registered type: ' + type.name);
```
leebyron: This should be more descriptive of the problem at hand. I recommend:

``` `Schema cannot contain more than one type named ${type.name}.` ```

leebyron:
Thanks for raising this, @dittos. I pulled in this PR in d3188d39b5ae1ae36987e035d6d166ff540f7e6a and added a wider suite of tests.

PR #37, opened 2015-07-09
Add missing `description` to field args

albertstill:
As the schema sais [__InputValue](http://facebook.github.io/graphql/#sec-Schema-Introspection) can have a `description`. When I tried to set a description inline (Not using an `INPUT_OBJECT `) it did not work. It seems the code in `definition.js` was missing the `description` in a map functions return object.

I've have fixed this an added a Star Wars spec to go with it.

Also a side note: `starWarsSchema.js` was setting [`name`](https://github.com/graphql/graphql-js/blob/master/src/__tests__/starWarsSchema.js#L246) values on the args but it serves no purpose because the name of the arg is set to the value of the key [here](https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L302). So I removed the two `name` declarations in the `#human` and `#droid` field args section in `starWarsSchema.js`.

### DIFF
--- a/src/__tests__/starWarsIntrospectionTests.js
+++ b/src/__tests__/starWarsIntrospectionTests.js
@@ -298,6 +298,83 @@ describe('Star Wars Introspection Tests', () => {
       return testQuery(query, expected);
     });
 
+    it('Allows querying the schema for field args', () => {
+      var query = `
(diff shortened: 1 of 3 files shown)


leebyron:
Thank you!

Issue #26, opened 2015-07-06
Database optimizations and graphql


freiksenet:
Hello!

In our system we took GraphQL query and translated it to ReQL (RethinkDB query language) for execution. This allowed us to optimize the query, eg use built-in joins in ReQL instead of querying the database two times. I can see how the similar system can be used for, eg, SQL queries. We also did some further optimizations, like returning only the fields that are requested by GraphQL, reducing the response payload size for tables with many keys.

I see a way to do it graphql-js by manually going through selection sets in resolve, but that feels very clumsy and hacky, as the library does this walking already. Some ways to do it would be to allow children `resolve` to modify root or to split collect/resolve/complete steps into independent parts that can be used separately.

Do you think deferred execution and query translation is a good idea? Should we stick to current model and do multiple requests to the database? Is there some solution to that, which I don't see?

Thanks!

freiksenet:
Related to this - many database systems has a concept of connection, currently I need to pass that connection on schema definition type, so that resolve function can run some query against the database. Separating `collect`, `resolve` and `complete` steps would help for this too.

freiksenet:
# Mapping vs reducing executor

Discussed this with @dschafer in IRC a bit, he asked to provide some concrete 
examples of what we would like to be able to do with executor.

As I understood, the current executor was made with the assumption that it's 
okay to make multiple round trips, so it's built from top down. If resolve is 
defined on some lower level fields, they get access to parent result and can 
query backend once more. One can say current executor uses `map` over tree, 
calling appropriate `resolve` function on each node.

In our system, we assumed that extra round-trips are bad and that, as our 
system allows it, we would like to be able to express GraphQL queries with one 
query, when it is possible. We also figured we can do optimizations along the 
way, like requesting only the selection set fields in the query. Our executor 
collected results of what is similar to `resolve` in graphql-js and then 
combined them together. One can say we did `reduce` over the tree, calling the 
`resolve` function along the way and then combining the result.

Let's imagine schema:

```
type User {
  id: String!
  name: String
  posts: [Post]
}

type Post {
  id: String!
  text: String
  createdAt: DateTime
  author: User
}
```

## Mapping executor

We assume we have fields in `query` to get `User` and `Post` by id. Resolve for
one of them would look something like:

```js
resolve(parent, {id}, {db, conn}) {
  return db.table('User').get(id).run(conn); // Promise
}
```

For relations (like `author` or `posts`) we'll have something like

```js
[...]

PR #39, opened 2015-07-10
Example of collect step

freiksenet:
So this is hacky and not really working attempt on adding collect step to executor, as we've discussed with @leebyron and @dschafer.

This is obviously not supposed to be merged, it's merely a way to continue discussion in #26.

Currently it's broken for many cases because of (lack of) type resolution in `collectFieldsDeep` step. I guess something similar to what `completeValue` does should also happen here.


freiksenet:
I'll close it as it served its demonstration purposes.

PR #42, opened 2015-07-12
Directive name is not-null

OlegIlyenko:
According to the spec, directive name should be not-null in the introspection API:

http://facebook.github.io/graphql/#sec-Schema-Introspection


OlegIlyenko:
Done

dschafer:
Thanks!

PR #46, opened 2015-07-14
Rename GraphQLFieldArgument to GraphQLArgument

andimarek:
This PR renames GraphQLFieldArgument to GraphQLArgument to make it more clear, that it's not only an argument type for Fields, but a more generic type (currently also used in directives)


leebyron:
Most excellent, thank you

PR #54, opened 2015-07-15
[RFC] standard printer uses no comma between fields

leebyron:
A couple issues have opened up about inconsistency in rendering graphql queries in the spec and in the ref impl - especially with respect to whitespace and commas.

Here I'm proposing that it's easier to read queries where every field is always on it's own line (eg, no `image { width, height }` one-liners) and that lines are split only by line-terminators and not also commas.

If there's agreement around this proposal, the second step will be to apply the printer to all graphql examples in the spec.

### DIFF
--- a/src/language/__tests__/printer.js
+++ b/src/language/__tests__/printer.js
@@ -49,12 +49,12 @@ describe('Printer', () => {
     expect(printed).to.equal(
   `query queryName($foo: ComplexType, $site: Site = MOBILE) {
   whoever123is: node(id: [123, 456]) {
-    id,
+    id
     ... on User @defer {
       field2 {
-        id,
+        id
         alias: field1(first: 10, after: $foo) @include(if: $foo) {
-          id,
+          id
           ...frag
         }
       }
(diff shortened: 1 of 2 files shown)


schrockn:
Makes sense to me.

My only request would be a "russHanneman" flag where the delimiter is tres comas (",,,") rather than a single comma ","

schrockn:
We also need to do a pass in the spec to make all examples conform to that

leebyron:
> ,,,

dschafer:
Yeah, agreed; I think this is better.

I do like the one-liners in some cases, but I don't think the standard printer should use them by any means.

leebyron:
I'm waiting on spec cleanup to make sure everyone's cool with the format :)

leebyron:
@dschafer I considered making the printer smarter but couldn't think about simple enough rules of when inlining is okay.

Within the spec I'll go for fairly aggressive consistency, elsewhere I agree - inlining is terse and nice

schrockn:
You may merge at will.

leebyron:
Firing away, captain

leebyron:
Follow up on the spec can be found here https://github.com/facebook/graphql/commit/188cb06d5e0c800739ce0a57af810bf227b06a17

PR #57, opened 2015-07-16
use isCompositeType function instead of manual checking

andimarek:
Make it more DRY: Use the appropriate helper function instead of repeating all composite types


leebyron:
Thank you!

Issue #58, opened 2015-07-16
Add tests for "context sensitivity" around on/fragment/query/mutation non-keywords


swolchok:
Name is context-sensitive in that, for example, the string "on" is a `Name` unless it appears in a place in the grammar that requires on specifically (e.g., right after an ellipsis, as in `InlineFragment`, or two tokens after fragment, as in `FragmentDefinition`). Naive application of tools like flex and bison seems likely to produce a system that doesn't allow apparent "keywords" like `on`/`fragment`/`query`/`mutation` to be used as `Name`s (e.g., as the names of fields). We should add test cases in `src/language/__tests__/parser.js` that demonstrate that these are legal Names.

PR #59, opened 2015-07-16
Add parser tests for accepting non-keywords

swolchok:
Resolves issue #58. There is one special case because fragments named `on` may be defined according to the current specification (`fragment on on SomeType { someField }`), but may not be referenced.

(If you try to reference your fragment named `on`, it's at best a syntax error, as in `{ ...on }`, which is the start of an inline fragment definition that doesn't have a type.)

### DIFF
--- a/src/language/__tests__/parser.js
+++ b/src/language/__tests__/parser.js
@@ -119,6 +119,34 @@ fragment MissingOn Type
     expect(() => parse(kitchenSink)).to.not.throw();
   });
 
+  it('allows non-keywords anywhere a Name is allowed', () => {
+    const nonKeywords = [
+      'on',
+      'fragment',
+      'query',
+      'mutation',
+      'true',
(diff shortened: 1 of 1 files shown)


leebyron:
Since https://github.com/graphql/graphql-js/commit/8ae5bd7e4d733c99dbbbece347eb15c4bcded951 just landed, maybe you want to revisit this slightly to not include "on" as a viable fragment name? If you rebase, this test should no longer pass

swolchok:
> If you rebase, this test should no longer pass

It currently passes; `on` is a legal query name, type name,  field name, directive name, argument name, and variable name, and the test is careful not to use it as a fragment name.

leebyron:
> It currently passes; on is a legal query name, type name, field name, directive name, argument name, and variable name, and the test is careful not to use it as a fragment name.

Nice work

PR #60, opened 2015-07-16
Babel async await

schrockn:
This PR adds the async/await babel polyfill and then changes src/__tests__ (the Star Wars tests) to use them. Using this transform makes life a lot easier.

1) The syntax is nicer
2) No need to use the promise-friendly chai from the chai-as-promised module. I dislike these directives because it makes it awkward to add console logging in a test in between the executed test code and then test of the result. I also think its often desirable to do multiple expects with more granular error messaging.
3) No more forgetting to return the promise and then have false-passes because the code just doesn't execute.
4) From brief testing the errors messages appear more reasonable. 

I think we should just get rid of chai-as-promised and switch to this.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,6 @@
 {
   "parser": "babel-eslint",
   "arrowFunctions": true,
-  "asyncFunctions": true,
-  "es7.asyncFunctions": true,
   "blockBindings": true,
   "classes": true,
   "defaultParams": true,
@@ -15,7 +13,6 @@
   "objectLiteralShorthandProperties": true,
   "spread": true,
   "templateStrings": true,
-  "experimental" : true,
   "env": {
     "node": true,
     "es6": true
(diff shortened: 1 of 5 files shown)


leebyron:
Sweeeeeeet. 

Are the changes to package.json necessary? As long as the async/await is only used within tests they might not be. Only curious because of concern that babel might include more polyfills that aren't being used in the built and npm published module

leebyron:
How are you seeing error messages change? Exciting to hear that you're seeing clearer messages, I'm curious what that looks like

schrockn:
It doesn't appear so. I tested npm run build with --ignore __tests__ commented out and it worked.

schrockn:
Lee I might be confusing it with your change but I can look into that more carefully. One sec.

leebyron:
Also, I agree that this syntax is nicer than chai as promised. I would love to see this pattern replace that dependency

leebyron:
Also make sure npm run cover works as well. That might be the thing in package.json that actually might require a change

schrockn:
Yeah I think I got confused there on the error reporting.

schrockn:
There are no magical flags in package.json. Do you mean .eslintrc?

PR #61, opened 2015-07-16
Bugfix TypeInfo: Missing/Wrong cleanup in leave

andimarek:
VariableDefinition needs to cleanup the inputTypeStack and not argument
Argument needs to cleanup inputTypeStack too


leebyron:
Indeed! Thank you for catching this

PR #64, opened 2015-07-16
Mention that Flow is used.

gabrielf:
To avoid confusion for developers new to the project that haven't heard of Flow.


leebyron:
Thanks

PR #65, opened 2015-07-16
Add Type System Printer

schrockn:
This adds a printer to translate the result of an introspection query
to the grammar that is currently used as shorthand in the specification.
This should be treated as the de facto specification until a formalized
grammar is produced.

The next step is to do the transformation in the other
direction.

A tool that should be built on top of this is one that executes
the introspection query against an arbitrary GraphQL HTTP endpoint
(specified per the upcoming graphql-http spec) and then performs
acceptable tests with that tool.

In the interim we hope that this will be useful for implementors
building and debugging their own type systems.

### DIFF
--- a/src/type/__tests__/introspection.js
+++ b/src/type/__tests__/introspection.js
@@ -36,110 +36,12 @@ describe('Introspection', () => {
       await graphql(EmptySchema, introspectionQuery)
     ).to.deep.equal({
       data: {
-        schemaType: {
-          __typename: '__Type',
-          enumValues: null,
-          fields: [
-            {
-              __typename: '__Field',
-              args: [],
-              deprecationReason: null,
-              isDeprecated: false,
-              name: 'types',
-              type: {
-                __typename: '__Type',
-                kind: 'NON_NULL',
-                name: null,
-                ofType: {
-                  __typename: '__Type',
-                  kind: 'LIST',
-                  name: null,
-                  ofType: {
-                    __typename: '__Type',
-                    kind: 'NON_NULL',
-                    name: null,
-                    ofType: {
-                      __typename: '__Type',
-                      kind: 'OBJECT',
-                      name: '__Type',
-                    }
-                  }
-                }
-              }
-            },
-            {
-              __typename: '__Field',
-              args: [],
-              deprecationReason: null,
-              isDeprecated: false,
-              name: 'queryType',
-              type: {
-                __typename: '__Type',
-                kind: 'NON_NULL',
-                name: null,
-                ofType: {
-                  __typename: '__Type',
-                  kind: 'OBJECT',
-                  name: '__Type',
-                  ofType: null,
-                },
-              },
-            },
-            {
-              __typename: '__Field',
-              args: [],
-              deprecationReason: null,
-              isDeprecated: false,
-              name: 'mutationType',
-              type: {
-                __typename: '__Type',
-                kind: 'OBJECT',
-                name: '__Type',
-                ofType: null,
-              },
-            },
-            {
-              __typename: '__Field',
-              args: [],
-              deprecationReason: null,
-              isDeprecated: false,
-              name: 'directives',
-              type: {
-                __typename: '__Type',
-                kind: 'NON_NULL',
-                name: null,
-                ofType: {
-                  __typename: '__Type',
-                  kind: 'LIST',
-                  name: null,
-                  ofType: {
-                    __typename: '__Type',
-                    kind: 'NON_NULL',
-                    name: null,
-                    ofType: {
-                      __typename: '__Type',
-                      kind: 'OBJECT',
-                      name: '__Directive',
-                    },
-                  },
-                },
-              },
-            },
-          ],
-          interfaces: [],
-          kind: 'OBJECT',
-          name: '__Schema',
-        },
-        queryRootType: {
-          __typename: '__Type',
-          enumValues: null,
-          fields: [],
-          interfaces: [],
-          kind: 'OBJECT',
-          name: 'QueryRoot',
-        },
         __schema: {
           __typename: '__Schema',
+          mutationType: null,
+          queryType: {
+            name: 'QueryRoot',
+          },
           types: [
             {
               __typename: '__Type',
@@ -147,7 +49,8 @@ describe('Introspection', () => {
               name: 'QueryRoot',
               fields: [],
               interfaces: [],
-              enumValues: null
+              enumValues: null,
+              possibleTypes: null,
             },
             {
               __typename: '__Type',
@@ -241,7 +144,8 @@ describe('Introspection', () => {
                 }
               ],
               interfaces: [],
-              enumValues: null
+              enumValues: null,
+              possibleTypes: null,
             },
             {
               __typename: '__Type',
@@ -446,7 +350,8 @@ describe('Introspection', () => {
                 }
               ],
               interfaces: [],
-              enumValues: null
+              enumValues: null,
+              possibleTypes: null,
             },
             {
               __typename: '__Type',
@@ -503,7 +408,8 @@ describe('Introspection', () => {
                   isDeprecated: false,
                   deprecationReason: null
                 }
-              ]
+              ],
+              possibleTypes: null,
             },
             {
               __typename: '__Type',
@@ -511,7 +417,8 @@ describe('Introspection', () => {
               name: 'String',
               fields: null,
               interfaces: null,
-              enumValues: null
+              enumValues: null,
+              possibleTypes: null,
             },
             {
               __typename: '__Type',
@@ -519,7 +426,8 @@ describe('Introspection', () => {
               name: 'Boolean',
               fields: null,
               interfaces: null,
-              enumValues: null
+              enumValues: null,
+              possibleTypes: null,
             },
             {
               __typename: '__Type',
@@ -635,7 +543,8 @@ describe('Introspection', () => {
                 }
               ],
               interfaces: [],
-              enumValues: null
+              enumValues: null,
+              possibleTypes: null,
             },
             {
               __typename: '__Type',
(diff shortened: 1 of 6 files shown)


src/type/__tests__/introspection.js
```diff
@@ -36,110 +36,12 @@ describe('Introspection', () => {
       await graphql(EmptySchema, introspectionQuery)
     ).to.deep.equal({
       data: {
-        schemaType: {
```
leebyron: Is this an unrelated change?

src/type/printer.js
```diff
@@ -0,0 +1,150 @@
...
+  ENUM: 'ENUM',
+  INPUT_OBJECT: 'INPUT_OBJECT',
+  LIST: 'LIST',
+  NON_NULL: 'NON_NULL',
+};
+
+// e.g Int, Int!, [Int!]!
+function printTypeDecl(type) {
+  if (type.ofType === undefined || type.ofType === null) {
+    return type.name;
```
dschafer: Why not just do this at the end instead of the Error?

src/type/printer.js
```diff
@@ -0,0 +1,150 @@
...
+function printArgs(field) {
+  if (field.args.length === 0) {
+    return '';
+  }
+  return '(' + field.args.map(printArg).join(', ') + ')';
+}
+
+function printFields(fields) {
+  return fields.map(f => '  ' + printField(f)).join('\n');
+  // var strs = [];
```
dschafer: rm

src/type/definition.js
```diff
@@ -102,6 +102,14 @@ export function isCompositeType(type: ?GraphQLType): boolean {
   );
 }
 
+export function typeHasFields(type: ?GraphQLType): boolean {
```
leebyron: this already exists as `isCompositeType` in this file

schrockn: That has different semantics. It does not include input types and includes unions

src/type/__tests__/introspection.js
schrockn: No. The Pre-existing introspection query was incorrect.

src/type/introspection.js
```diff
@@ -108,8 +109,7 @@ var __Type = new GraphQLObjectType({
         includeDeprecated: { type: GraphQLBoolean, defaultValue: false }
       },
       resolve(type, {includeDeprecated}) {
-        if (type instanceof GraphQLObjectType ||
-            type instanceof GraphQLInterfaceType) {
+        if (typeHasFields(type)) {
```
leebyron: This is not a safe change. Unions do not define any fields, I don't think it defines `getFields()`

src/type/printer.js
schrockn: √

src/type/printer.js
```diff
@@ -0,0 +1,150 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+var TypeKind = {
```
dschafer: This feels weird to me, not sure why.

src/type/printer.js
schrockn: Not following

src/type/introspection.js
schrockn: Look at the implementation of typeHasFields. It does not include unions

src/type/printer.js
schrockn: more specific?

src/type/definition.js
```diff
@@ -102,6 +102,14 @@ export function isCompositeType(type: ?GraphQLType): boolean {
...
+  return (
+    type instanceof GraphQLObjectType ||
+    type instanceof GraphQLInterfaceType ||
+    type instanceof GraphQLInputObjectType
```
dschafer: This is a change in behavior; `GraphQLInputObjectType` shouldn't have `fields`, it should have `inputFields`, right?

src/type/definition.js
leebyron: Ah, my mistake

src/type/introspection.js
leebyron: Misread the previous, but this is still probably not safe as now InputObject is going to present fields as the wrong type and in the wrong place in introspection

src/type/printer.js
```diff
@@ -0,0 +1,150 @@
...
+  UNION: 'UNION',
+  ENUM: 'ENUM',
+  INPUT_OBJECT: 'INPUT_OBJECT',
+  LIST: 'LIST',
+  NON_NULL: 'NON_NULL',
+};
+
+// e.g Int, Int!, [Int!]!
+function printTypeDecl(type) {
+  if (type.ofType === undefined || type.ofType === null) {
```
leebyron: I recommend doing instanceof checks instead of directly checking the value of `ofType` - we dont use this pattern elsewhere

src/type/introspection.js
schrockn: Yup I was wrong on that count. Will fix. Forgot about inputFields

src/type/definition.js
schrockn: yup. will fix.

src/type/printer.js
leebyron: sorry not instanceof - Kind checks.


```
function printTypeDecl(type) {
  if (type.kind === TypeKind.LIST) {
    return '[' + printTypeDecl(type.ofType) + ']';
  } else if (type.kind === TypeKind.NON_NULL) {
    return printTypeDecl(type.ofType) + '!';
  }
  return type.name;
}
```

schrockn: √

src/type/printer.js
dschafer:
```
function printTypeDecl(type) {
  if (type.kind === TypeKind.LIST) {
    return '[' + printTypeDecl(type.ofType) + ']';
  } else if (type.kind === TypeKind.NON_NULL) {
    return printTypeDecl(type.ofType) + '!';
  }
  return type.name;
}
```

src/type/printer.js
dschafer: I guess it feels weird to define this struct only to use every value once, compared to just using the strings inline, or reusing an existing struct if it exists.

src/type/printer.js
```diff
@@ -0,0 +1,150 @@
...
+function printFields(fields) {
+  return fields.map(f => '  ' + printField(f)).join('\n');
+  // var strs = [];
+  // fields.forEach(field => {
+  //   strs.push('  ' + printField(field));
+  // });
+  // return strs.join('\n');
+}
+
+function printIfaceList(type) {
```
leebyron: I would prefer spelling out, perhaps `printInterfaces`

src/type/printer.js
```diff
@@ -0,0 +1,150 @@
...
+  return type.name.startsWith('__');
+}
+
+function isBuiltIn(type) {
+  var name = type.name;
+  if (isIntrospectionType(type)) {
+    return true;
+  }
+
+  return name === 'String' || name === 'Boolean' || name === 'Int';
```
leebyron: Float?

leebyron: ID?

src/type/printer.js
schrockn: lolz

src/type/printer.js
schrockn: there's no existing struct. we should probably move this to a common locale

src/type/printer.js
schrockn: going to go with printInterfaceUsages to distinguish from printing the definition

src/type/printer.js
schrockn: Da

leebyron: I actually think I prefer not excluding these at all - for the same reason we don't exclude them from introspection.

src/type/printer.js
```diff
@@ -0,0 +1,150 @@
...
+  return values.map(v => '  ' + v.name).join('\n');
+}
+
+function printEnum(type) {
+  return `enum ${type.name} {
+${printEnumValues(type.enumValues)}
+}`;
+}
+
+function printType(type) {
```
leebyron: Should this also be exported? It seems useful to print only one type in the same way it's useful to introspect on only one type with `__type(name:)`

leebyron:
This is so awesome. Two high-level points of feedback:

1) I'd recommend splitting this into two PRs. One that address any changes or bug fixes to the introspection system and type definition API itself. It was a little hard for me to follow what additions were necessary for type printing and what additions were ancillary.

2) I think we should reflect on the grammar for types and schema a little more:

* It's a bit odd that every type starts with a keyword for the kind of type it is, expect object types. Now that I'm seeing this all come together, I think we should consider the leading keyword for object types be `object`.

```diff
interface Type { ... }
enum Enum { ... }
- type Type { ... }
+ object Type { ... }
```

* We need some way to describe the initial types for query and mutation. I don't have any immediately useful suggestions here, but I think this will be a point of confusion in printed schema.

* Scalars are a little weird in that they don't have definitions, just declarations. I suppose we could expand on this later.

src/type/printer.js
```diff
@@ -0,0 +1,150 @@
...
+    case TypeKind.INTERFACE:
+      return printInterface(type);
+    case TypeKind.INPUT_OBJECT:
+      return printInputObject(type);
+    case TypeKind.SCALAR:
+      return printScalar(type);
+    case TypeKind.ENUM:
+      return printEnum(type);
+    default:
+      throw new Error('Invalid kind: ' + type.kind);
```
leebyron: Maybe add a test illustrating the reason for this?

src/type/printer.js
schrockn: The only thing I find strange is how they are only included when you use them. I do not feel strongly however.

src/type/printer.js
schrockn: not following this comment

src/type/printer.js
schrockn: reason for what?

PR #66, opened 2015-07-16
Introspection query changes

schrockn:
This PR makes the introspection query more complete. We simply did not
query everything we needed to for a complete type system.

I also removed the __typename field everywhere. We should have a
different query for an introspection acceptance test. This query
is for tool-building.

### DIFF
--- a/src/type/introspection.js
+++ b/src/type/introspection.js
@@ -16,7 +16,7 @@ import {
   GraphQLEnumType,
   GraphQLInputObjectType,
   GraphQLList,
-  GraphQLNonNull
+  GraphQLNonNull,
 } from './definition';
 import { GraphQLString, GraphQLBoolean } from './scalars';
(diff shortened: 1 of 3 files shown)


leebyron:
LGTM Land at will!

src/type/introspection.js
```diff
@@ -16,7 +16,7 @@ import {
   GraphQLEnumType,
   GraphQLInputObjectType,
   GraphQLList,
-  GraphQLNonNull
+  GraphQLNonNull,
```
schrockn: don't we like trailing commas?

I added an import during the process of doing stuff. Didn't feel compelled to remove it.

leebyron: We like trailing commas to avoid unattributed changes to lines later, but haven't been picky about them.

I just thought it was hilarious that the only change to this file was to add a comma.

PR #67, opened 2015-07-16
Add Schema Printer

schrockn:
This adds a printer to translate the result of an introspection query
to the grammar that is currently used as shorthand in the specification.
This should be treated as the de facto specification until a formalized
grammar is produced.

The next step is to do the transformation in the other
direction.

A tool that should be built on top of this is one that executes
the introspection query against an arbitrary GraphQL HTTP endpoint
(specified per the upcoming graphql-http spec) and then performs
acceptable tests with that tool.

In the interim we hope that this will be useful for implementors
building and debugging their own type systems.

### DIFF
--- a/src/type/printer.js
+++ b/src/type/printer.js
@@ -1,0 +1,155 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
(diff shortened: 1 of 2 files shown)


src/type/printer.js
```diff
@@ -0,0 +1,155 @@
...
+}
+
+function printArgs(field) {
+  if (field.args.length === 0) {
+    return '';
+  }
+  return '(' + field.args.map(printArg).join(', ') + ')';
+}
+
+function printInterfaceUsages(type) {
```
leebyron: `printImplementedInterfaces` perhaps?

leebyron:
Nice

PR #68, opened 2015-07-16
Test that exponents on integers like `1e6` do not parse

swolchok:
I didn't see a test for this. It's legal JSON but intentionally illegal GraphQL.


leebyron:
For posterity: we decided to make legal JSON numbers legal GraphQL numbers and updated the spec to reflect this. Tests were added in ca5749621b122b1cf92101f43a01a446b922b57f which cover this particular case (but it now parses, rather than errors).

PR #70, opened 2015-07-16
Allow an episode to be passed to hero

dschafer:
It was odd in our demo that `hero` returned a `Character`, but it always returned R2-D2, who is a `Droid`. Change it so this takes an optional `episode`, and returns Luke as the hero for Episode V, demonstrating more clearly why this field returns an Interface, and the value of querying for `__typename` to find the concrete type.

### DIFF
--- a/src/__tests__/starWarsIntrospectionTests.js
+++ b/src/__tests__/starWarsIntrospectionTests.js
@@ -34,6 +34,9 @@ describe('Star Wars Introspection Tests', () => {
               name: 'Query'
             },
             {
+              name: 'Episode'
+            },
+            {
               name: 'Character'
             },
             {
@@ -41,9 +44,6 @@ describe('Star Wars Introspection Tests', () => {
             },
             {
               name: 'String'
-            },
-            {
-              name: 'Episode'
             },
             {
               name: 'Droid'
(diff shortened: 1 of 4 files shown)


src/__tests__/starWarsIntrospectionTests.js
```diff
@@ -327,7 +327,20 @@ describe('Star Wars Introspection Tests', () => {
             fields: [
               {
                 name: 'hero',
-                args: []
+                args: [
+                  {
+                    defaultValue: null,
+                    description: 'If omitted, returns the hero of the whole ' +
```
leebyron: nice

src/__tests__/starWarsSchema.js
```diff
@@ -233,7 +233,21 @@ var queryType = new GraphQLObjectType({
...
+                       'provided, returns the hero of that particular episode.',
+          type: episodeEnum
+        }
+      },
+      resolve: (root, {episode}) => {
+        if (episode === 5) {
+          // Luke is the hero of Episode V.
+          return luke;
+        }
+        // Artoo is the hero otherwise.
```
leebyron: lol

leebyron:
awesome, land away

src/__tests__/starWarsSchema.js
```diff
@@ -233,7 +233,21 @@ var queryType = new GraphQLObjectType({
...
+      args: {
+        episode: {
+          description: 'If omitted, returns the hero of the whole saga. If ' +
+                       'provided, returns the hero of that particular episode.',
+          type: episodeEnum
+        }
+      },
+      resolve: (root, {episode}) => {
+        if (episode === 5) {
+          // Luke is the hero of Episode V.
```
wincent: I think you misspelled "Lando" here.

PR #71, opened 2015-07-17
Bugfix getArgumentValues: DefaultValue not used

andimarek:
When no argument was provided, the defaultValue should be used as argument value


leebyron:
Thanks! Could you add a test illustrating this?

andimarek:
I added a test but it's much more unit-test than the usual test cases. For example I create an argument object (representing a argument literal) directly instead letting the parser do that.

So I don't know if this fits in the general project philosophy.

leebyron:
The unit test is nice, but it's also helpful to have more of an integration test with the executor to ensure that the use case in question is covered.

leebyron:
there are examples of this in executor/__tests__/variables.js

andimarek:
I added a more like integration test. The first unit-test could be removed, but also kept ... both are fine options. (imho)

andimarek:
ok ... now the tests also pass :-)

leebyron:
Just merged this in 08559470ab5b7bd4621128add4b1f7a5afde876c after squashing the commits and making some minor amendments.

Thanks again for investigating and fixing this!

PR #74, opened 2015-07-20
Schema Definition Language Parser

schrockn:
This is a parser for the ad-hoc type definition language included in the
spec. This should also serve as a de-facto spec for the language.
Clearly once this is stable we should include it in the actual spec as
well.

This should end up being useful in a few contexts, including
client-side types and also a DSL for producing the JS type definitions
which can be tedious to write, and a way to specify acceptance tests.
(This would be a clear next step for this project)

### DIFF
--- a/src/language/parser.js
+++ b/src/language/parser.js
@@ -9,8 +9,7 @@
 
 import { Source } from './source';
 import { syntaxError } from '../error';
-import { lex, TokenKind, getTokenKindDesc, getTokenDesc } from './lexer';
-import type { Token } from './lexer';
+import { TokenKind } from './lexer';
 import type {
   Name,
   Variable,
@@ -35,7 +34,7 @@ import type {
   Directive,
 
   Type,
-  NamedType
+  NamedType,
 } from './ast';
 
 import {
@@ -69,24 +68,19 @@ import {
   NON_NULL_TYPE,
 } from './kinds';
 
-/**
- * Configuration options to control parser behavior
- */
-type ParseOptions = {
-  /**
-   * By default, the parser creates AST nodes that know the location
-   * in the source that they correspond to. This configuration flag
-   * disables that behavior for performance or testing.
-   */
-  noLocation?: boolean,
-
-  /**
-   * By default, the parser creates AST nodes that contain a reference
-   * to the source that they were created from. This configuration flag
-   * disables that behavior for performance or testing.
-   */
-  noSource?: boolean,
-}
+import {
+  makeParser,
+  peek,
+  skip,
+  loc,
+  any,
+  many,
+  expect,
+  unexpected,
+  expectKeyword,
+  advance,
+  ParseOptions,
+} from './parserCore';
 
 /**
  * Given a GraphQL source, parses it into a Document.
@@ -102,161 +96,9 @@ export function parse(
 }
 
 /**
- * Returns the parser object that is used to store state throughout the
- * process of parsing.
- */
-function makeParser(source: Source, options: ParseOptions) {
-  var _lexToken = lex(source);
-  return {
-    _lexToken,
-    source,
-    options,
-    prevEnd: 0,
-    token: _lexToken(),
-  };
-}
-
-/**
- * Returns a location object, used to identify the place in
- * the source that created a given parsed object.
- */
-function loc(parser, start: number) {
-  if (parser.options.noLocation) {
-    return null;
-  }
-  if (parser.options.noSource) {
-    return {
-      start: start,
-      end: parser.prevEnd
-    };
-  }
-  return {
-    start: start,
-    end: parser.prevEnd,
-    source: parser.source
-  };
-}
-
-/**
- * Moves the internal parser object to the next lexed token.
- */
-function advance(parser): void {
-  var prevEnd = parser.token.end;
-  parser.prevEnd = prevEnd;
-  parser.token = parser._lexToken(prevEnd);
-}
-
-/**
- * Determines if the next token is of a given kind
- */
-function peek(parser, kind: string): boolean {
-  return parser.token.kind === kind;
-}
-
-/**
- * If the next token is of the given kind, return true after advancing
- * the parser. Otherwise, do not change the parser state and return false.
- */
-function skip(parser, kind: string): boolean {
-  var match = parser.token.kind === kind;
-  if (match) {
-    advance(parser);
-  }
-  return match;
-}
-
-/**
- * If the next token is of the given kind, return that token after advancing
- * the parser. Otherwise, do not change the parser state and return false.
- */
-function expect(parser, kind: string): Token {
-  var token = parser.token;
-  if (token.kind === kind) {
-    advance(parser);
-    return token;
-  }
-  throw syntaxError(
-    parser.source,
-    token.start,
-    `Expected ${getTokenKindDesc(kind)}, found ${getTokenDesc(token)}`
-  );
-}
-
-/**
- * If the next token is a keyword with the given value, return that token after
- * advancing the parser. Otherwise, do not change the parser state and return
- * false.
- */
-function expectKeyword(parser, value: string): Token {
-  var token = parser.token;
-  if (token.kind === TokenKind.NAME && token.value === value) {
-    advance(parser);
-    return token;
-  }
-  throw syntaxError(
-    parser.source,
-    token.start,
-    `Expected "${value}", found ${getTokenDesc(token)}`
-  );
-}
-
-/**
- * Helper function for creating an error when an unexpected lexed token
- * is encountered.
- */
-function unexpected(parser, atToken?: ?Token): Error {
-  var token = atToken || parser.token;
-  return syntaxError(
-    parser.source,
-    token.start,
-    `Unexpected ${getTokenDesc(token)}`
-  );
-}
-
-/**
- * Returns a possibly empty list of parse nodes, determined by
- * the parseFn. This list begins with a lex token of openKind
- * and ends with a lex token of closeKind. Advances the parser
- * to the next lex token after the closing token.
- */
-function any<T>(
-  parser,
-  openKind: number,
-  parseFn: (parser: any) => T,
-  closeKind: number
-): Array<T> {
-  expect(parser, openKind);
-  var nodes = [];
-  while (!skip(parser, closeKind)) {
-    nodes.push(parseFn(parser));
-  }
-  return nodes;
-}
-
-/**
- * Returns a non-empty list of parse nodes, determined by
- * the parseFn. This list begins with a lex token of openKind
- * and ends with a lex token of closeKind. Advances the parser
- * to the next lex token after the closing token.
- */
-function many<T>(
-  parser,
-  openKind: number,
-  parseFn: (parser: any) => T,
-  closeKind: number
-): Array<T> {
-  expect(parser, openKind);
-  var nodes = [parseFn(parser)];
-  while (!skip(parser, closeKind)) {
-    nodes.push(parseFn(parser));
-  }
-  return nodes;
-}
-
-/**
  * Converts a name lex token into a name parse node.
  */
-function parseName(parser): Name {
+export function parseName(parser): Name {
   var token = expect(parser, TokenKind.NAME);
   return {
     kind: NAME,
@@ -264,7 +106,6 @@ function parseName(parser): Name {
     loc: loc(parser, token.start)
   };
 }
-
 
 // Implements the parsing rules in the Document section.
 
@@ -603,7 +444,7 @@ function parseDirective(parser): Directive {
 /**
  * Handles the Type: NamedType, ListType, and NonNullType parsing rules.
  */
-function parseType(parser): Type {
+export function parseType(parser): Type {
   var start = parser.token.start;
   var type;
   if (skip(parser, TokenKind.BRACKET_L)) {
@@ -627,7 +468,7 @@ function parseType(parser): Type {
   return type;
 }
 
-function parseNamedType(parser): NamedType {
+export function parseNamedType(parser): NamedType {
   var start = parser.token.start;
   return {
     kind: NAMED_TYPE,
--- a/src/language/parserCore.js
+++ b/src/language/parserCore.js
@@ -1,0 +1,185 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import { lex, TokenKind, getTokenKindDesc, getTokenDesc } from './lexer';
+
+import { Source } from './source';
+import { syntaxError } from '../error';
+import type { Token } from './lexer';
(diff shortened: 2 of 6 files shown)


src/language/ast.js
```diff
@@ -242,3 +242,52 @@ export type NonNullType = {
   loc?: ?Location;
   type: NamedType | ListType;
 }
+
+export type TypeDefinition = {
```
leebyron: These should probably go into another file, clientAST.js or something like that.

src/language/ast.js
```diff
@@ -242,3 +242,52 @@ export type NonNullType = {
...
+  kind: 'InterfaceDefinition';
+  loc?: ?Location;
+  name: Name;
+  fields: Array<FieldDefinition>;
+}
+
+export type FieldDefinition = {
+  kind: 'FieldDefinition';
+  loc?: ?Location;
+  name: Name;
```
leebyron: missing `args`

src/language/ast.js
```diff
@@ -242,3 +242,52 @@ export type NonNullType = {
...
+  type: Type;
+}
+
+export type EnumDefinition = {
+  kind: 'EnumDefinition';
+  loc?: ?Location;
+  name: Name;
+  values: Array<Name>;
```
leebyron: `EnumValueDefinition` should be it's own AST node probably.

src/language/ast.js
```diff
@@ -242,3 +242,52 @@ export type NonNullType = {
...
+  kind: 'ScalarDefinition';
+  loc?: ?Location;
+  name: Name;
+}
+
+export type InputObjectDefinition = {
+  kind: 'InputObjectDefinition';
+  loc?: ?Location;
+  name: Name;
+  fields: Array<FieldDefinition>;
```
leebyron: `InputObjectFieldDefinition`?

src/language/kinds.js
```diff
@@ -47,3 +47,14 @@ export const DIRECTIVE = 'Directive';
 export const NAMED_TYPE = 'NamedType';
 export const LIST_TYPE = 'ListType';
 export const NON_NULL_TYPE = 'NonNullType';
+
+// Schema
```
leebyron: Similarly, should move to another file - clientKinds.js or something.

Or maybe this all goes into a subfolder? `language/client/*.js`

src/language/parser.js
```diff
@@ -635,3 +634,19 @@ function parseNamedType(parser): NamedType {
     loc: loc(parser, start)
   };
 }
+
+export var ParseOps = {
```
leebyron: Interesting - perhaps all these fn's should get extracted to their own file and exported directly. That file could then support both of these?

src/language/schemaParser.js
```diff
@@ -0,0 +1,246 @@
...
+      return parseScalarDefinition(parser);
+    default:
+      throw ParseOps.unexpected(parser);
+  }
+}
+
+function parseScalarDefinition(parser): ScalarDefinition {
+  var start = parser.token.start;
+  ParseOps.expectKeyword(parser, 'scalar');
+  var name = ParseOps.parseName(parser);
```
leebyron: put these inline

src/language/schemaParser.js
```diff
@@ -0,0 +1,246 @@
...
+  var definitions = [];
+  do {
+    if (!ParseOps.peek(parser, TokenKind.NAME)) {
+      throw ParseOps.unexpected(parser);
+    }
+    definitions.push(parseNextSchemaDefinition(parser));
+  } while (!ParseOps.skip(parser, TokenKind.EOF));
+
+  return {
+    kind: DOCUMENT,
```
leebyron: probably want to define this something other than DOCUMENT so the AST isn't overloaded.

src/language/schemaParser.js
```diff
@@ -0,0 +1,246 @@
...
+    definitions,
+    loc: ParseOps.loc(parser, start)
+  };
+}
+
+function parseInterfaceDefinition(parser): InterfaceDefinition {
+  var start = parser.token.start;
+  ParseOps.expectKeyword(parser, 'interface');
+  var name = ParseOps.parseName(parser);
```
leebyron: inline name and fields

leebyron:
This all seems pretty solid to me. 

I think there's just some relatively minor code isolation issues to fix up, perhaps moving most of this stuff into `language/schema/*.js` rather than altering individual files.

I also think you can terse up the parser a bit by putting the parse steps inline in the object definition - that pattern is pretty pervasive through the current parser.

leebyron:
One higher-level question is how this relates to client parsers (read: FB's graphql-data) which interleave GraphQL query and GraphQL schema for the purposes of code-generation, and isolate fragments into separate files and need to refer to them in some way.

It seems like either this schema should be merged with that use case, or there is a 3rd variety of parser we will want to eventually support.

src/language/parser.js
schrockn: yeah i could do that. i just think it's really annoying to have to export that many things. I would rather have a parser object with methods. no semantic changes. the parser object is mutable anyway and gets mutated anyways so I don't really see how its that different

src/language/schemaParser.js
schrockn: no. i do not put side effects in constructor arguments.

src/language/schemaParser.js
schrockn: cool

schrockn:
re: your last comment definitely a third type with interleaving IMO

src/language/parser.js
```diff
@@ -63,16 +58,28 @@ import {
...
+  any,
+  many,
+  parseName,
+  expect,
+  parseType,
+  parseNamedType,
+  unexpected,
+  expectKeyword,
+  advance,
+} from './parserCore';
```
leebyron: The role of this is a little muddy. I wonder if there's a better place for `parseName` and the `parseType` group to exist? They're definitely specific parser rules not parser core functionality.

For the short term, and for the sake of maintaining the clarity of this document relative to the spec, I would suggest (deep breath) copying these few functions directly in `schema/parser.js`. I think DRY is working against us here.

src/language/parser.js
```diff
@@ -63,16 +58,28 @@ import {
...
+
 /**
  * Configuration options to control parser behavior
  */
-type ParseOptions = {
+export type ParseOptions = {
```
leebyron: move to parserCore?

src/language/parserCore.js
```diff
@@ -0,0 +1,265 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+// export var ParseOps = {
```
leebyron: rm

src/language/schema/__tests__/parser.js
```diff
@@ -0,0 +1,450 @@
...
+          kind: 'TypeDefinition',
+          name: nameNode('Hello', loc(5, 10)),
+          interfaces: [typeNode('World', loc(22, 27))],
+          fields: [],
+          loc: loc(0, 31),
+        }
+      ],
+      loc: loc(0, 31),
+    };
+    expect(printJson(doc)).to.equal(printJson(expected));
```
leebyron: `expect(doc).to.deep.equal(expected)`

I suggest leaving JSON printing out of it.

src/language/schema/ast.js
```diff
@@ -0,0 +1,84 @@
+/**
```
leebyron: make sure to add `/*@flow*/ here

src/language/schema/__tests__/parser.js
schrockn: I think we should do this as a standard. The error reporting is much, much better.

src/language/schema/ast.js
```diff
@@ -0,0 +1,84 @@
...
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import {
```
leebyron: For similar clarity reasons, I would suggest duplicating these few rules here instead of trying to isolate them.

src/language/schema/parser.js
```diff
@@ -0,0 +1,271 @@
...
+}
+
+function parseInputObjectDefinition(parser): InputObjectDefinition {
+  var start = parser.token.start;
+  expectKeyword(parser, 'input');
+  var name = parseName(parser);
+  var fields = any(
+    parser,
+    TokenKind.BRACE_L,
+    parseFieldDefinition,
```
leebyron: it should be a parse error for input object fields to take arguments

src/language/parser.js
```diff
@@ -597,13 +425,24 @@ function parseDirective(parser): Directive {
   };
 }
 
+/**
```
leebyron: remain in current position

src/language/parser.js
```diff
@@ -9,10 +9,11 @@
 
 import { Source } from './source';
 import { syntaxError } from '../error';
-import { lex, TokenKind, getTokenKindDesc, getTokenDesc } from './lexer';
-import type { Token } from './lexer';
+import { TokenKind } from './lexer';
 import type {
   Name,
+  NamedType,
```
leebyron: still need these?

src/language/schema/__tests__/parser.js
leebyron: Let's ultimately define a helper for mocha then in mocha-bootstrap, and perhaps send it upstream to mocha.

src/language/schema/ast.js
```diff
@@ -0,0 +1,84 @@
...
+  Location,
+  Name,
+  Type,
+} from '.';
```
leebyron: Is this correct location to import from?

src/language/schema/ast.js
```diff
@@ -0,0 +1,84 @@
...
+  loc?: ?Location;
+  definitions: Array<SchemaDefinition>;
+}
+
+export type SchemaDefinition =
+  TypeDefinition |
+  InterfaceDefinition |
+  EnumDefinition |
+  ScalarDefinition |
+  InputObjectDefinition;
```
leebyron: nit: no semi needed after defining a union (similar kind of nit to trailing comma)

src/language/schema/ast.js
```diff
@@ -0,0 +1,84 @@
...
+
+export type TypeDefinition = {
+  kind: 'TypeDefinition';
+  loc?: ?Location;
+  name: Name;
+  interfaces?: ?Array<Name>;
```
leebyron: should this be `Array<NamedType>` ?

src/language/schema/ast.js
```diff
@@ -0,0 +1,84 @@
...
+  kind: 'ScalarDefinition';
+  loc?: ?Location;
+  name: Name;
+}
+
+export type InputObjectDefinition = {
+  kind: 'InputObjectDefinition';
+  loc?: ?Location;
+  name: Name;
+  fields: Array<FieldDefinition>;
```
leebyron: InputObjectFieldDefinition should be reflected here as well

src/language/schema/parser.js
```diff
@@ -0,0 +1,274 @@
...
+
+function parseArgumentDefs(parser): Array<ArgumentDefinition> {
+  if (!peek(parser, TokenKind.PAREN_L)) {
+    return [];
+  }
+  return many(
+    parser,
+    TokenKind.PAREN_L,
+    parseArgumentDef,
+    TokenKind.PAREN_R);
```
leebyron: formatting nit: `);` on own unindented line

src/language/schema/parser.js
```diff
@@ -0,0 +1,274 @@
...
+  var name = parseName(parser);
+  var location = loc(parser, start);
+  return {
+    kind: SCALAR_DEFINITION,
+    name: name,
+    loc: location,
+  };
+}
+
+function parseSchemaDocument(parser): SchemaDocument {
```
leebyron: just in terms of ease of reading this file, I'd suggest moving this fn between `parseSchema` and `parseNextSchemaDefinition` (which I may also suggest naming just `parseSchemaDefinition`)

src/language/schema/parser.js
```diff
@@ -0,0 +1,274 @@
...
+  var start = parser.token.start;
+  var definitions = [];
+  do {
+    if (!peek(parser, TokenKind.NAME)) {
```
leebyron: I'd suggest moving this `if` block into `parseNextSchemaDefinition`

src/language/schema/__tests__/parser.js
```diff
@@ -0,0 +1,450 @@
...
+  return {
+    kind: 'FieldDefinition',
+    name: name,
+    type: type,
+    arguments: args,
+    loc: loc,
+  };
+}
+
+describe('Schema Parser', () => {
```
leebyron: How about a couple tests for things which should syntax error?

```
query Foo { field }
```

```
{ field }
```

```
input Type {
  field(arg: Boolean)
}
```

src/language/schema/parser.js
schrockn: #goodcatch

src/language/schema/parser.js
schrockn: don't really understand why but don't care

src/language/parser.js
schrockn: yup

leebyron: can you move to where they were defined before instead of moving them in this rev? They are in the same order as defined in the spec for readability reasons

src/language/parser.js
```diff
@@ -65,28 +63,23 @@ import {
   DIRECTIVE,
 
   NAMED_TYPE,
-  LIST_TYPE,
   NON_NULL_TYPE,
+  LIST_TYPE,
```
leebyron: same here

src/language/schema/ast.js
```diff
@@ -0,0 +1,91 @@
...
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import {
+  Location,
+  Name,
+  Type,
+} from '../';
```
leebyron: I believe this line should read:

```
import type { Location, Name, Type } from '../ast.js'
```

I'm surprised flow didn't catch this

src/language/schema/ast.js
```diff
@@ -0,0 +1,91 @@
...
+export type FieldDefinition = {
+  kind: 'FieldDefinition';
+  loc?: ?Location;
+  name: Name;
+  type: Type;
+  arguments: Array<ArgumentDefinition>;
+}
+
+export type InputFieldDefinition = {
+  kind: 'FieldDefinition';
```
leebyron: kind: 'InputFieldDefinition';

src/language/schema/ast.js
leebyron: `import type {`

PR #76, opened 2015-07-21
fix directive inspection according to spec

martinandert:
The spec states that

```
type __Directive {
  # ...
  onOperation: Boolean!
  onFragment: Boolean!
  onField: Boolean!
}
```

but the implementation defines it as

```
type __Directive {
  # ...
  onOperation: Boolean
  onFragment: Boolean
  onField: Boolean
}
```


dschafer:
Thanks!

PR #77, opened 2015-07-21
Schema Materializer

schrockn:
Adds the schema materializer. This components takes a schema definition
AST and produces a set of type definition objects corresponding to the
schema definitions. This allows a user to specify schema in a far
more literate fashion

The next tool that I'm going to write is one that takes the
schema definition DSL and outputs the result of the introspection
query against that schema. This will be really useful for tests, etc.

### DIFF
--- a/src/language/schema/__tests__/materializer.js
+++ b/src/language/schema/__tests__/materializer.js
@@ -1,0 +1,277 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+*  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+import { parseSchema } from '../parser';
+import { materializeSchema } from '../materializer';
+import { printSchema } from '../../../type/printer';
+import { introspectionQuery } from '../../../type/introspectionQuery';
+import { graphql } from '../../../';
+
+// 80+ char lines are useful in describe/it, so ignore in this file.
+/*eslint-disable max-len */
+
+function printForTest(result) {
+  return '\n' + printSchema(result) + '\n';
+}
+
+async function getOutput(body, queryType) {
+  var doc = parseSchema(body);
+  var schema = materializeSchema(doc, queryType);
+  var result = await graphql(schema, introspectionQuery);
+  return await printForTest(result);
+}
+
+describe('Schema Materializer', () => {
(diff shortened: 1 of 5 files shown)


src/language/schema/__tests__/materializer.js
```diff
@@ -0,0 +1,277 @@
...
+
+type WorldTwo {
+  str: String
+}
+`;
+    var output = await getOutput(body, 'Root');
+    expect(output).to.equal(body);
+  });
+
+  it('Simple Union', async () => {
```
dschafer: Should be named "Scalar"

src/language/schema/materializer.js
```diff
@@ -0,0 +1,263 @@
...
+    });
+  }
+
+  // This actually constructs all the types by iterating over the schema
+  // This makes it so that errors actually get caught before this function
+  // exits.
+  schema.getTypeMap();
+  return schema;
+
+  function makeSchemaDef(def) {
```
dschafer: Why all these all inner functions?

src/language/schema/materializer.js
```diff
@@ -0,0 +1,263 @@
+/**
```
dschafer: I wish this file was @flow, I think it would clarify what the various `def` variables are.

src/language/schema/materializer.js
```diff
@@ -0,0 +1,263 @@
...
+      String: GraphQLString,
+      Int: GraphQLInt,
+      Boolean: GraphQLBoolean,
+      ID: GraphQLID,
+    };
+
+    return (typeAST) => {
+      var typeName = getInnerTypeName(typeAST);
+      if (!nullish(innerTypeMap[typeName])) {
+        return buildWrappedType(innerTypeMap[typeName], typeAST);
```
dschafer: Why don't you memoize the wrapped type?

dschafer:
Overall, seems reasonable to me.

src/language/schema/materializer.js
schrockn: It's so they can have access to the memoization maps without having to thread them around.

src/language/schema/materializer.js
schrockn: Good idea.

src/language/schema/materializer.js
dschafer: Hm, this structure feels odd to me. Does `getTypeDefProducer` have to be an inner function as well? I think it's the combo of "code block, inline function, code block, chunk of inline functions" that feels tough to follow for me?

src/language/schema/materializer.js
schrockn: I guess we could construct some key scheme based on the ! + [] notation but it seemed unnecessary

PR #79, opened 2015-07-23
Print Schema Tool

schrockn:
This tool accepts an input graphql schema definition file and
outputs the corresponding introspection query result for that file

### DIFF
--- a/src/language/schema/printer.js
+++ b/src/language/schema/printer.js
@@ -1,0 +1,19 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import { parseSchema } from './parser';
+import { materializeSchema } from './materializer';
+import { introspectionQuery } from './../../type/introspectionQuery';
+import { graphql } from '../../';
+
+export async function getIntrospectionResult(body, queryType) {
+  var doc = parseSchema(body);
+  var schema = materializeSchema(doc, queryType);
+  return await graphql(schema, introspectionQuery);
+}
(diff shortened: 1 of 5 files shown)


src/language/schema/printer.js
```diff
@@ -0,0 +1,19 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
```
dschafer: Why is this file named `printer.js`?

src/tools/print-schema/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+  return await JSON.stringify(result, null, 2);
+}
+
+var helpString = `
+This tool consumes GraphQL schema definition files and outputs the
+introspection query result from querying that schema.
+
+Required:
+
+--file <path>: The path to the input schema definition file.
```
dschafer: I kinda wish these were optional, and it would assume the first type in the input schema was the query type, and read from stdin if no file was required. But that can be in a follow-up if we ever need it.

src/tools/print-schema/print-schema.js
```diff
@@ -0,0 +1,21 @@
...
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+require('babel/register')({
+  optional: ['runtime', 'es7.asyncFunctions']
+});
+
+var executeTool = require('./src').executeTool;
+
+executeTool().then(function (out) {
```
dschafer: The separation between script and src feels odd to me; this does the `stdout` stuff but the src does the file stuff? I almost wish this was just one script, since the "guts" of the script are only three lines, everything else is managing params and output.

dschafer:
Overall, seems good, merge away after addressing inlines.

PR #87, opened 2015-07-28
Add visitor and printer utils to schema language

leebyron:
This adds a "visitor" and "printer" utilities for the schema language in the same form as those used for the document language.

### DIFF
--- a/src/language/printer.js
+++ b/src/language/printer.js
@@ -14,101 +14,111 @@ import { visit } from './visitor';
  * formatting rules.
  */
 export function print(ast) {
-  return visit(ast, {
-    leave: {
-      Name: node => node.value,
-      Variable: node => '$' + node.name,
+  return visit(ast, { leave: printDocASTReducer });
+}
 
-      // Document
+export var printDocASTReducer = {
+  Name: node => node.value,
+  Variable: node => '$' + node.name,
 
-      Document: node => join(node.definitions, '\n\n') + '\n',
-      OperationDefinition(node) {
-        var op = node.operation;
-        var name = node.name;
-        var defs = manyList('(', node.variableDefinitions, ', ', ')');
-        var directives = join(node.directives, ' ');
-        var selectionSet = node.selectionSet;
-        return !name ? selectionSet :
-          join([op, join([name, defs]), directives, selectionSet], ' ');
-      },
-      VariableDefinition: node =>
-        join([node.variable + ': ' + node.type, node.defaultValue], ' = '),
-      SelectionSet: node =>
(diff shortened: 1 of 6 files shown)


src/language/printer.js
```diff
@@ -14,101 +14,96 @@ import { visit } from './visitor';
...
 
-function manyList(start, list, separator, end) {
-  return length(list) === 0 ? null : start + join(list, separator) + end;
+export function many(start, list, separator, end) {
+  return length(list) ? start + join(list, separator) + (end || '') : '';
 }
 
-function length(maybeArray) {
-  return maybeArray ? maybeArray.length : 0;
+export function join(maybeArray, separator) {
```
schrockn: what's all this maybe stuff. very confusing.

leebyron: utilities for printing where the logic inline would be messy. AST has keys that are nullable or keys that can contain empty arrays, and this logic helps handle those nicely.

src/language/printer.js
```diff
@@ -14,101 +14,96 @@ import { visit } from './visitor';
...
+    var selectionSet = node.selectionSet;
+    return !name ? selectionSet :
+      join([op, join([name, defs]), directives, selectionSet], ' ');
+  },
+  VariableDefinition: node =>
+    join([node.variable + ': ' + node.type, node.defaultValue], ' = '),
+  SelectionSet: node => block(node.selections),
+  Field: node =>
+    join([
+      join([node.alias, node.name], ': ') +
```
schrockn: let's be consistent about the way we form alias: name strings. you do the more obvious construction in the Argument case, which I think is more clear and preferable.

src/language/printer.js
leebyron: I'll add some docblocks

src/language/printer.js
leebyron: the more obvious construction won't work here since alias is optional.

This pattern of making a list of nullable tokens and then joining is used through the printer

schrockn:
In general I think the emphasis on one-liners detracts from the readability of this. I think most of the joins that don't involve actual arrays can be replaced by template strings which results in more obvious and clear code. It seems like you created the join hammer and everything is a nail.

schrockn:
However a lot of those are just moves and outside the scope of this diff.

schrockn:
actually most of the ones for the schema def are just template strings. i was mostly responding to the moved codez.

leebyron:
Good feedback on the moved codes though, I'll look at improvements

schrockn:
land this. good work.

PR #89, opened 2015-07-29
Client Schema Builder

leebyron:
Given a full introspection result, this builds a GraphQLSchema instance that can be used for everything except execution, allowing clients to use it throughout tools, establishing GraphQLSchema as the common in-mem representation of the type system for use across both server and client.

This differs from the DSL materializer in a few important ways:

1. Input. This accepts introspection result, which is trivial for a client to produce given any server. This avoids requiring client tools to leverage the whole GraphQL executor locally to derive this information, which lets it be a sync operation.
2. Scope. This preserves everything that introspection represents including descriptions (and in a follow-up, deprecations).

### DIFF
--- a/src/utils/keyValMap.js
+++ b/src/utils/keyValMap.js
@@ -10,30 +10,28 @@
 
 /**
  * Creates a keyed JS object from an array, given a function to produce the keys
- * for each value in the array.
- *
- * This provides a convenient lookup for the array items if the key function
- * produces unique results.
+ * and a function to produce the values from each item in the array.
  *
  *     var phoneBook = [
  *       { name: 'Jon', num: '555-1234' },
  *       { name: 'Jenny', num: '857-6309' }
  *     ]
  *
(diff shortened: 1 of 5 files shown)


src/utils/keyValMap.js
```diff
@@ -0,0 +1,20 @@
...
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+export default function keyValMap<T, V>(
```
dschafer: Can we throw a docblock on this, with a quick example usage?

leebyron: fo sho

dschafer:
This looks really good. Merge away!

schrockn:
Dan thanks for reviewing. Agreed this is excellent.

PR #94, opened 2015-07-29
Print the schema directly, rather than the introspection result.

leebyron:
This allows schemaPrinter to be a sync rather than async function.

Also introduces the (meaty) "astFromValue" utility which accepts arbitrary JS values and outputs GraphQL AST.

### DIFF
--- a/src/utils/__tests__/astFromValue.js
+++ b/src/utils/__tests__/astFromValue.js
@@ -1,0 +1,187 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+import astFromValue from '../astFromValue';
+import {
+  GraphQLEnumType,
+  GraphQLInputObjectType,
+  GraphQLList
+} from '../../type/definition';
+import { GraphQLFloat } from '../../type/scalars';
+
+
+describe('astFromValue', () => {
+
+  it('converts boolean values to ASTs', () => {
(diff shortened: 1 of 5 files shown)


src/utils/__tests__/astFromValue.js
```diff
@@ -0,0 +1,265 @@
...
+  GraphQLInputObjectType,
+  GraphQLList
+} from '../../type/definition';
+import { GraphQLFloat } from '../../type/scalars';
+
+
+describe('astFromValue', () => {
+
+  it('converts boolean values to ASTs', () => {
+    expect(astFromValue(
```
schrockn: nit: for this whole file most of these could be just two lines. astFromValue(...) could be one line. and to.deep.equal(...) could be one line

src/utils/astFromValue.js
```diff
@@ -0,0 +1,129 @@
...
+
+  // Convert JavaScript array to GraphQL list. If the GraphQLType is a list, but
+  // the value is not an array, convert the value using the list's item type.
+  if (Array.isArray(value)) {
+    var itemType = type instanceof GraphQLList ? type.ofType : null;
+    return {
+      kind: LIST,
+      values: value.map(item => astFromValue(item, itemType))
+    };
+  } else if (type instanceof GraphQLList) {
```
schrockn: This is quite confusing to me. Is this because arguments have the list coercion rules?

src/utils/astFromValue.js
```diff
@@ -0,0 +1,129 @@
...
+ * | JSON Value    | GraphQL Value        |
+ * | ------------- | -------------------- |
+ * | Object        | Input Object         |
+ * | Array         | List                 |
+ * | Boolean       | Boolean              |
+ * | String        | String / Enum Value  |
+ * | Number        | Int / Float          |
+ *
+ */
+export default function astFromValue(
```
schrockn: I think this function is sufficiently complicated that it should have its own unit tests. Would be useful as documentation as well.

schrockn:
This is excellent. Land away.

PR #98, opened 2015-07-30
Move GraphQL-specific utils to tools

leebyron:


schrockn:
go for it. let's choose a new name as discussed!

Issue #100, opened 2015-08-02
Bug in buildASTSchema when mutations are passed


nemanja-stanarevic:
All, great work all around on graphql reference implementation, it's incredibly helpful.

I have been using the schema DSL parser and buildASTSchema to materialize the schema and noticed a small bug in utilities/buildASTSchema.  Specifically, when both queryTypeName and mutationTypeName are passed to the buildASTSchema, the schema is not constructed correctly - query is missing. 

I traced the bug down to the lines 150-157 of utilities/buildASTSchema, where if mutationTypeName is passed, schema object is constructed with the queryTypeName string instead of queryType object: 

```javascript
  if (isNullish(mutationTypeName)) {
    schema = new GraphQLSchema({query: queryType});
  } else {
    schema = new GraphQLSchema({
      query: queryTypeName,
      mutation: produceTypeDef(astMap[mutationTypeName]),
    });
  }
```

This should fix it - 

```javascript
  if (isNullish(mutationTypeName)) {
    schema = new GraphQLSchema({query: queryType});
  } else {
    schema = new GraphQLSchema({
      query: queryType,
      mutation: produceTypeDef(astMap[mutationTypeName]),
    });
  }
```

Note that this slipped since buildASTSchema tests don't include a case when both mutation and query are passed.  I am happy to fix this, add a test and submit a PR, if welcome.

dschafer:
Thanks for the report and PR! Just pushed 0.2.4 which should contain a fix for this issue!

PR #101, opened 2015-08-02
Fix issue with buildASTSchema when both query and mutation are passed

nemanja-stanarevic:
Added a test that confirms issue #100 and a simple fix (one liner).


dschafer:
Thanks!

Issue #104, opened 2015-08-05
Bug in buildASTSchema: Adding a Float scalar to type definition breaks the materializer


nemanja-stanarevic:
When a Float field is added to type definition, the materializer (`buildASTSchema`) breaks down with `Error: Type Float not found in document`.  Adding a float field to HelloScalars test confirms the issue.

Patch is simple, one-line tweak to getTypeDefProducer function in utilities/buildASTSchema.js to add `Float` to `innerTypeMap` map, as follows:

```javascript
  function getTypeDefProducer() {

    var innerTypeMap = {
      String: GraphQLString,
      Int: GraphQLInt,
++++  Float: GraphQLFloat,
      Boolean: GraphQLBoolean,
      ID: GraphQLID,
    };
```

With this, the new HelloScalars test passes, I'll submit a PR in a sec (updated test + patch).

dschafer:
Looks like #105 fixed this, thanks for the report and patch!

nemanja-stanarevic:
Dan, you are welcome; I'm looking forward to v0.2.5 on npm :)

Thanks again for all the great work you all are putting into this

PR #105, opened 2015-08-05
Fixes issue #104 with buildASTSchema and Float fields

nemanja-stanarevic:
Fixes issue #104 and allows buildASTSchema to materialize a type with a Float field. Adds tests for Float and ID to 'Simple type' test and for Float, ID, Int and Boolean to 'Single argument field' test.


dschafer:
Awesome, thanks!

PR #107, opened 2015-08-06
Ensure unreferenced types that implement reference interfaces are detected

dschafer:
Ensures that if a type isn't referenced directly, but implements an interface that is references, that type is included in the schema, with a unit test to verify.

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -136,6 +136,11 @@ export function buildASTSchema(
         throw new Error('Nothing constructed for ' + typeName);
       }
       innerTypeMap[typeName] = innerTypeDef;
+      if (astMap[typeName].kind === INTERFACE_DEFINITION) {
+        // Now that we've created and cached our interface, ensure
+        // we create all implementations to point to it.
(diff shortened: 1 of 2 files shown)


src/utilities/buildASTSchema.js
```diff
@@ -211,6 +216,16 @@ export function buildASTSchema(
     return def.interfaces.map(inter => produceTypeDef(inter));
   }
 
+  function makeInterfaceImplementations(inter: InterfaceDefinition) {
```
leebyron: This is `O(N^2)` since you need to filter down the AST and produceTypeDef for the implementing types for every interface you encounter, (so really `O(I*N)`)

Here's how I did it in https://github.com/graphql/graphql-js/blob/master/src/utilities/buildClientSchema.js#L302

On a side note, you may consider comparing this file to that file to see if there are any other gaps

dschafer:
Good call. Changed it to iterate through all types; now the `produceTypeMap` for the query will just hit cache (where before, it was actually constructing the types).

leebyron:
wonderful. I love that it ended up being a one-liner and the bulk of this is tests :) land it

PR #108, opened 2015-08-06
Add check for unknown mutation type

dschafer:
We weren't checking unknown mutation types (and we were double checking unknown query types).

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -105,6 +105,11 @@ export function buildASTSchema(
       ' not found in document.');
   }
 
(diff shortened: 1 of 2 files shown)


src/utilities/buildASTSchema.js
```diff
@@ -143,10 +148,6 @@ export function buildASTSchema(
 
   var produceTypeDef = getTypeDefProducer(ast);
 
-  if (isNullish(astMap[queryTypeName])) {
```
dschafer: This was duplicating the check on line 103.

leebyron:
Yeah. Obviwin

Issue #115, opened 2015-08-11
Coercing variables


freiksenet:
Currently Scalar types have two methods `coerce` and `coerceLiteral`. `coerceLiteral` is used to create a javascript value from ast, `coerce` is used both on variables and on data received from resolve in executors. 

Unfortunately that is not enough - variables usually come from some format like `json` which wouldn't always support all the new scalar types and most likely will return them as string. On the other hand, eg, databases might return types in javascript format already. Therefore `coerce` can't support both variable coercion and execution coercion.

@leebyron suggested that for variable coercion, a new method is added to the Scalars - `coerceVariable`.

freiksenet:
Okay, so we cleared it with @leebyron. `coerce` is actually a function that should do serialization. We've agreed to rename `coerce` to `serialize`, rename `coerceLiteral` to `parseLiteral` and add `parseVariable`.

I'll try to make a PR asap.

leebyron:
Merged in! this will be in the next release.

Dashed:
These changes should reflect onto the docs: http://graphql.org/docs/api-reference-type-system/#graphqlscalartype

When playing with graphql-js, I got super confused on why `coerce` stuff was missing and not working.

PR #116, opened 2015-08-11
Add parseVariable and move from coerce terminology

freiksenet:
* Add parseVariable to scalar, to separate parsing of values from
  their serialization
* Rename coerce to serialize and coerceLiteral to parseLiteral

As per #115. I am not 100% sure anymore about terminology change, needs more discussion.

### DIFF
--- a/src/execution/__tests__/variables.js
+++ b/src/execution/__tests__/variables.js
@@ -21,15 +21,38 @@ import {
   GraphQLList,
   GraphQLString,
   GraphQLNonNull,
+  GraphQLScalarType
 } from '../../type';
 
+var TestComplexScalar = new GraphQLScalarType({
+  name: 'ComplexScalar',
+  serialize(value) {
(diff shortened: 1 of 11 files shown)


leebyron:
Merged in a623fc6a75c03c336bb0447161c52d0018388343 - I made some minor touch ups, but also decided that `parseValue` was more clear than `parseVariable` since it's used for more than just parsing variable values and it could be misinterpreted as parsing the actual variable nodes.

kamek-pf:
Hey guys,
The doc here : http://graphql.org/docs/api-reference-type-system/#graphqlscalartype
Doesn't reflect those changes.

leebyron:
Thanks for the report.

leebyron:
https://github.com/graphql/graphql.github.io/commit/7a87a9217d1c4517d34f9ee661de19c4be5e1a0e

PR #117, opened 2015-08-11
Fix some input objects missing from the type map

fson:
Input object types were not included in the type map when they were used as a field type in other input objects, because `typeMapReducer` did not recurse into those fields.

Modify `typeMapReducer` to also handle GraphQLInputObjectType and add a test case that checks also nested input object types are handled.


leebyron:
Thank you for diagnosing and fixing!

PR #118, opened 2015-08-11
[RFC] Do not format errors from graphql()

leebyron:
This changes the implementation of `graphql()` to not format any returned errors. Most importantly, this allows access to the `stack` property of each error, along with richer information contained in GraphQLError like the affected AST nodes and source document.

This means that the responsibility for formatting errors must be handled elsewhere, likely by an HTTP service like `express-graphql`.

### DIFF
--- a/src/graphql.js
+++ b/src/graphql.js
@@ -12,8 +12,7 @@ import { Source } from './language/source';
 import { parse } from './language/parser';
 import { validate } from './validation/validate';
 import { execute } from './execution/execute';
-import { formatError } from './error';
-import type { GraphQLFormattedError } from './error/formatError';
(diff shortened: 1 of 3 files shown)


schrockn:
After this change what is printed out in a test case if you make a programming error in the core somewhere

leebyron:
This is not a full fix for that case, but a step in that direction. You would likely see a string-encoded form of the Error object on behalf of the mocha framework (eg \n instead of actual newlines) but that Error object includes a stack.

A separate PR could introduce a test utility for clearer messaging when there is an inadvertent error.

schrockn:
not pretty but better than nothing I guess

Issue #103, opened 2015-08-03
Pass rootValue from execution context to resolveType


charypar:
It would be quite useful to get the `rootValue` in `resolveType` the same way you get it in field resolvers. 

In my case I need to dynamically chose which fetching backend implementation to use and I'm passing it in in the rootValue. The backend also does list filtering by various predicates, one of which is the item type - the backend therefore needs to be able to resolve the item type the same way the interface does. 

The natural place for this logic is the backend itself (all backend-specific things sit in one place), but since I only know the right implementation at query run time the type detection logic is now duplicated.

I can imagine other use-cases for the `rootValue` in resolveType - e.g. using an authorisation system to resolve to basic or extended type (guess that is kinda contrived) or some sort of user-specified type resolution rules maybe.

leebyron:
This happens today, though the variable naming is probably confusing. We're using the terms `context` and `rootValue` interchangeably. I'll keep this task open to resolve the naming sloppiness.

leebyron:
These last two diffs should add the functionality you're looking for!

leebyron:
They will be released on npm in the near future.

Issue #96, opened 2015-07-30
Getting selectionSet from fieldASTs with external fragment


gyzerok:
How can I get fields requested in external fragment from fieldASTs?

Consider following query

```javascript
`
  query QueryWithFragment {
    todo(_id: "55a624bad009804e552eeea8") {
      ...TextFragment
    }
  }

  fragment TextFragment on Todo {
    text
  }
`
```
This query results in following, so there is no way to get those fields without directly parsing query string.
```javascript
{
  "kind": "Field",
  "alias": null,
  "name": {
    "kind": "Name",
    "value": "todo",
    "loc": {
      "start": 27,
      "end": 31,
      "source": {
        "body": "\n  query UseFragment {\n    todo(_id: \"55a624bad009804e552eeea8\") {\n      ...TextFragment\n    }\n  }\n\n  fragment TextFragment on Todo {\n    text\n  }\n",
        "name": "GraphQL request"
      }
    }
  },
  "arguments": [{
    "kind": "Argument",
    "name": {
      "kind": "Name",
      "value": "_id",
      "loc": {
        "start": 32,
        "end": 35,
        "source": {
          "body": "\n  query UseFragment {\n    todo(_id: \"55a624bad009804e552eeea8\") {\n      ...TextFragment\n    }\n  }\n\n  fragment TextFragment on Todo {\n    text\n  }\n",
          "name": "GraphQL request"
        }
      }
    },
    "value": {
      "kind": "StringValue",
      "value": "55a624bad009804e552eeea8",
      "loc": {
        "start": 37,
        "end": 63,
        "source": {
          "body": "\n  query UseFragment {\n    todo(_id: \"55a624bad009804e552eeea8\") {\n      ...TextFragment\n    }\n  }\n\n  fragment TextFragment on Todo {\n    text\n  }\n",
          "name": "GraphQL request"
        }
      }
    },
    "loc": {
      "start": 32,
      "end": 63,
      "source": {
        "body": "\n  query UseFragment {\n    todo(_id: \"55a624bad009804e552eeea8\") {\n      ...TextFragment\n    }\n  }\n\n  fragment TextFragment on Todo {\n    text\n  }\n",
        "name": "GraphQL request"
      }
    }
  }],
  "directives": [],
  "selectionSet": {
    "kind": "SelectionSet",
    "selections": [{
      "kind": "FragmentSpread",
      "name": {
        "kind": "Name",
        "value": "TextFragment",
        "loc": {
          "start": 76,
[...]

PR #119, opened 2015-08-11
[RFC] Expose more information to field resolver functions.

leebyron:
This exposes more info to field resolvers, but reduces the arity of these functions to a fixed 3 arguments: source, arguments, execution info.

This along with ES6 argument destructuring makes accessing this information more natural.


leebyron:
This adds `fragments` into scope which will solve #96

leebyron:
And clarifies the variable naming for #103

dschafer:
Looks good to me!

PR #120, opened 2015-08-12
Update starWarsQueryTests to mimic real database API

cletusw:
A back-end implementation with a real database wouldn't have direct access to the full dataset.


leebyron:
I like this, but will defer to @dschafer since he has some plans here.

dschafer:
This looks good to me, thanks!

cletusw:
@dschafer If you'd like, I can do the same to the examples in the main repo (https://github.com/facebook/relay/tree/master/examples).

PR #124, opened 2015-08-13
[RFC] move type validation to ctor invariants

### LINKED ISSUE #121: "Arg must supply type" error message seems to include field index rather than field name.
e.g., 

```
  "errors": [
    {
      "message": "Arg must supply type. createSite.0", ...
```

And this is the relevant portion of my _schemaConfig:

```
{ type: 
   { name: 'Site',
     description: undefined,
     _typeConfig: { name: 'Site', fields: [Function] },
     _interfaces: [] },
  args: 
   [ { name: [Object] },
     { address: [Object] },
     { owner: [Object] },
     { nodes: [Object] } ],
  resolve: [Function] }
```

[As you can see, the 'type' fields are indeed missing from the args yet a more useful error message would include the field name (and the source code `/node_modules/graphql/type/definition.js:259` does seem to attempt to print the field name)].

leebyron:
This removes the type validation system, and in its place puts invariants into the type ctors.

This ensures that errors are produced as types are created, rather than when (or if) they run through the validator. This should lead to faster error recovery and less errors.

There were a number of issues detected in our test files that these invariants picked up.

Fixes #121


leebyron:
Good feedback, I'll up the SNR in this test

leebyron:
Updated. Good suggestion.

PR #125, opened 2015-08-13
Improve error message on null-values for non-nullable fields

rexxars:
I was implementing a GraphQL schema today and encountered this error: `Cannot return null for non-nullable type.`

I thought it left something to be desired in terms of figuring out which field/type was causing this issue. This PR adds a more descriptive error message, including the field name and parent type name to make this easier to track down.

Let me know if this needs any changes.


leebyron:
This is awesome, thanks!

PR #127, opened 2015-08-13
[RFC] schema enforced interface adherance.

leebyron:
This adds an invariant step in the Schema ctor for enforcing that all Object types in the Schema properly adhere to the Interface types they claim to implement.

This effectively implements http://facebook.github.io/graphql/#sec-Object-type-validation

### DIFF
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -66,6 +66,16 @@ export class GraphQLSchema {
       this.getMutationType(),
       __Schema
     ].reduce(typeMapReducer, {});
+
+    // Enforce correct interface implementations
+    Object.keys(this._typeMap).forEach(typeName => {
+      var type = this._typeMap[typeName];
+      if (type instanceof GraphQLObjectType) {
+        type.getInterfaces().forEach(
+          iface => assertObjectImplementsInterface(type, iface)
+        );
+      }
+    });
   }
 
   getQueryType(): GraphQLObjectType {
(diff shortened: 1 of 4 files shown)


src/type/schema.js
```diff
@@ -147,3 +157,61 @@ function typeMapReducer(map: TypeMap, type: ?GraphQLType): TypeMap {
...
+    });
+  });
+}
+
+function isEqualType(typeA: GraphQLType, typeB: GraphQLType): boolean {
+  if (typeA instanceof GraphQLNonNull) {
+    if (typeB instanceof GraphQLNonNull) {
+      return isEqualType(typeA.ofType, typeB.ofType);
+    }
+    return false;
```
dschafer: Hm, this feels a bit tricky to me, since we return false here when A is NonNull and B is not, but we don't return false here when B is NonNull and A is not.

Perhaps:

```js
if (typeA instanceof GraphQLNonNull && typeB instanceof GraphQLNonNull) {
  return isEqualType(typeA.ofType, typeB.ofType);
}
if (typeA instanceof GraphQLList && typeB instanceof GraphQLList) {
  return isEqualType(typeA.ofType, typeB.ofType);
}
return typeA === typeB;
```

would be clearer?

leebyron: Sure thing

dschafer:
How do we want to handle it if the interface specifies a field returns String, and the implementation returns non-null string? Similarly, if the interface specifies the arg must be a string, and the implementation specifies the arg must be a non-null string?

"rejects an Object with a differently modified Interface field type" describes the first one as invalid, we should probably do the same for arguments. It's not immediately clear to me why those are invalid, though.

leebyron:
Let's defer that for this particular PR. Keeping this invariant preserves option value to introduce co/contravariance like that in the future.

In fact, there are similar concerns with introducing non-null arguments on an implementing object field. I'll edit this to make the set of arguments invariant as well to preserve the same.

I actually really like at least the idea of nullability co/contravariance I think that this is a subtlety that people will eventually encounter and be surprised by. Let's tackle it in a separate PR.

leebyron:
Just updated with the invariance change.

We had an existing validation test case that assumed this invariance did not exist, the spec does not claim this invariance either - so we should update it.

dschafer:
Looks good!

PR #129, opened 2015-08-14
Add enum coercion test for mutation.

johanatan:
...

### DIFF
--- a/src/type/__tests__/enumType.js
+++ b/src/type/__tests__/enumType.js
@@ -59,7 +59,18 @@ describe('Type System: Enum Values', () => {
     }
   });
 
-  var schema = new GraphQLSchema({ query: QueryType });
+  var MutationType = new GraphQLObjectType({
+    name: 'Mutation',
+    fields: {
+      favoriteEnum: {
+        type: ColorType,
+        args: { color: { type: ColorType } },
+        resolve(value, { color }) { return color; }
+      }
+    }
+  });
+
+  var schema = new GraphQLSchema({ query: QueryType, mutation: MutationType });
 
   it('accepts enum literals as input', async () => {
     expect(
@@ -101,6 +112,19 @@ describe('Type System: Enum Values', () => {
       ]
     });
   });
+
+  it('accepts enum literals as input arguments to mutations',
+    async () => {
+      expect(
+        await graphql(schema,
+          'mutation x($color:Color!) { favoriteEnum(color:$color) }',
+            null, {color: 'GREEN'})
+      ).to.deep.equal({
+        data: {
+          favoriteEnum: 'GREEN'
+        }
+      });
+    });
 
   it('does not accept incorrect internal value', async () => {
     expect(


src/type/__tests__/enumType.js
```diff
@@ -30,6 +30,13 @@ describe('Type System: Enum Values', () => {
     }
   });
 
+  var FavoriteType = new GraphQLObjectType({
```
leebyron: This type seems unnecessary for the test. Should `Mutation.favoriteEnum`'s type just be `ColorType`?

src/type/__tests__/enumType.js
```diff
@@ -59,7 +66,31 @@ describe('Type System: Enum Values', () => {
...
   });
 
-  var schema = new GraphQLSchema({ query: QueryType });
+  var MutationType = new GraphQLObjectType({
+    name: 'Mutation',
+    fields: {
+      favoriteEnum: {
+        type: FavoriteType,
+        args: { color: { type: ColorType } },
+        resolve(value, color) { return color; }
```
leebyron: the second arg here is a little misleading to call `color`. It's the object of all provided arguments, you should be explicit and use `resolve(value, { color })` to destructure the color key from the arguments object.

johanatan: That actually did not work when I tried it-- and let me to believe that if a field accepts only one argument (as in this case) then it will not be embedded within an object. There may be a deeper issue here?

src/type/__tests__/enumType.js
```diff
@@ -59,7 +66,31 @@ describe('Type System: Enum Values', () => {
...
+      }
+    }
+  });
+
+  var schema = new GraphQLSchema({ query: QueryType, mutation: MutationType });
+
+  it('accepts enum literals as input arguments to mutations',
```
leebyron: Could you move this below to the equivalent test ensuring variables are coerced correctly for query operations on line 135?

leebyron:
Thanks for adding this test! Just a few suggestions for improvement inline.

src/type/__tests__/enumType.js
johanatan: Here's the error I get when destructuring the `color` from the arguments object:

```
  1 failing

  1) Type System: Enum Values accepts enum literals as input arguments to mutations:

      AssertionError: expected { Object (data) } to deeply equal { Object (data) }
      + expected - actual

       {
         "data": {
           "favoriteEnum": {
      -      "color": [null]
      +      "color": "GREEN"
           }
         }
       }
      
      at Object.callee$1$0$ (src/type/__tests__/enumType.js:88:17)
```

src/type/__tests__/enumType.js
johanatan: I tried that too and it worked. I only created this `FavoriteType` to more closely match the problematic case I had in my original code. I can remove it if you prefer (but in my opinion, the more complex the case, the better).

src/type/__tests__/enumType.js
johanatan: Sure.

src/type/__tests__/enumType.js
leebyron: This test is currently working because the `color` variable you're returning here has the value `{"color":"GREEN"}`. Since the output type is `FavoriteType` which is an Object with a field "color", it accepts this value, and the default resolver accesses the "color" property of that object and returns "GREEN".

However, I believe this is a little more verbose than necessary for the case you're testing.

I think you the "FavoriteType" Object type is unnecessary for testing that mutations will properly coerce variables. I would suggest not returning the entire arguments object, but only the color key, so the output of resolve is instead `"GREEN"`, and then change the type of this field to "ColorType".

src/type/__tests__/enumType.js
leebyron: For unit tests, I prefer as simple a case as possible to illustrate the condition to be tested. Complexity can often obscure what is actually being tested.

johanatan:
Ok, I've made the requested changes. Thx!

leebyron:
Great, thanks!

PR #131, opened 2015-08-17
Add lib directory to .gitignore

enaqx:


enaqx:
Ops, sorry. Confused with dist. My fault.

PR #132, opened 2015-08-17
Do not mutate passed field map in ObjectType and ObjectInput

freiksenet:
Types used to mutate the fields passed to it, which can be very confusing and prevents reuse (parts) of field definitions between multiple types. Changed it so that fields is first shallow copied and new object is returned as a field map.


leebyron:
Good idea, thanks!

PR #139, opened 2015-08-19
Add explanation of bad number character

dschafer:
When debugging https://github.com/facebook/relay/issues/121, I realized it would be useful to have the bad character listed in this case as well.


leebyron:
Let me take a look at the error message, I think I can improve further

PR #142, opened 2015-08-20
Use Math.trunc for Int coercion

enaqx:
ES6 support truncate without polyfilling:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc

### DIFF
--- a/src/type/scalars.js
+++ b/src/type/scalars.js
@@ -20,10 +20,9 @@ var MIN_INT = -9007199254740991;
 function coerceInt(value) {
   var num = Number(value);
   if (num === num && num <= MAX_INT && num >= MIN_INT) {
-    return (num < 0 ? Math.ceil : Math.floor)(num);
+    return Math.trunc(num);
   }
   return null;
-
 }
 
 export var GraphQLInt = new GraphQLScalarType({


leebyron:
We explicitly don't use trunc (or any of the other ES6-only methods) to avoid requiring users of graphql-js to include a global Math polyfill.

enaqx:
Got it. But it is compiled using babel. It must insert this polyfill from core-js anyway
https://github.com/zloirock/core-js/blob/master/modules/es6.math.trunc.js

leebyron:
Babel does not install global polyfill functions like `Math.trunc` or `String.startsWith` during transpilation, instead you must remember to also include the `babel/polyfill` submodule in your runtime. We decided that it was too much to ask that all runtime environments include that submodule, or that we would bundle that submodule in ourselves. That would mean including the `graphql` module would alter the global types in your javascript environment, which seemed overbearing.

enaqx:
Thanks for clarifying. Do you think it will be nice to mention this in Contribution Guide somehow?

leebyron:
Does this suffice? f9cd233e9fa70f4d932859fc394e33fe104ec890

enaqx:
Yup. Thanks. I believe it may prevent from similar mistakes. :+1:

leebyron:
No problem, I appreciate you combing through!

PR #145, opened 2015-08-21
Check for valid field arg names

jmandel:
If I've understood the intent of https://facebook.github.io/graphql/#sec-Names, field arg names are restricted to `/^[_a-zA-Z][_a-zA-Z0-9]*$/`.

Currently the js implementation allows the creation of non-conformant field arg names, which means you can create a field arg that a client can't actually reference in a query.

This PR adds a test for field arg name validity.

### DIFF
--- a/src/type/__tests__/validation.js
+++ b/src/type/__tests__/validation.js
@@ -353,6 +353,47 @@ describe('Type System: Objects must have fields', () => {
 });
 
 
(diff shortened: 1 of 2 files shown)


leebyron:
Fantastic suggestion. Would you mind applying the same to the fields themselves?

jmandel:
Applied.

leebyron:
Great, thank you!

PR #147, opened 2015-08-23
Minor change: remove unnecessary update of local variable

philix:
I'm implementing a GraphQL compiler in another language and noticed this line can be removed while I was reading the code.


dschafer:
Thanks!

PR #162, opened 2015-08-28
[RFC] RequestDocument

leebyron:
This proposes renaming `Document` to `RequestDocument` - change in spec would follow - to make room for a `ClientDocument` which contains references that have no executable semantic but do have code-gen or other client-side utility.

PR #163, opened 2015-08-29
[RFC] Expand GraphQL language

leebyron:
This proposes merging the schema language into the GraphQL language.

After some time of thinking about this, I believe this is the best path. There is only one "GraphQL language" and thus only one spec for a parser. 

I believe this will have substantial simplification forces on GraphQL server libs and client tools and the APIs they expose. As an example, GraphiQL will never need a "client mode" to parse and operate on "client GraphQL" and other IDE tools built by the community will only have one language to target.

This also means the validation phase can be designed to accept both "client GraphQL" and "server GraphQL" rather than two separate utilities with two separate AST and rule structures. Instead we're just likely to want to introduce standard rule sets for each environment. 

The drawback is explaining the semantics. This diff primarily adds type definition syntax to GraphQL, and it should be made clear that *at present* GraphQL servers do not have semantics for these and in fact will not execute queries which contain them.

Note: this is a counterproposal to #162


schrockn:
Note I think we can do "best-of-both-worlds" that do a simple pass to lint/check for constructs that are meaningless in specific contexts. This would layer a really simple piece of software on top instead of bifurcating a core component.

leebyron:
> Note I think we can do "best-of-both-worlds" that do a simple pass to lint/check for constructs that are meaningless in specific contexts.

Totally agree. The spec should describe that executing a query that contains anything other than Operation or Fragment definitions is illegal.

This diff now includes https://github.com/graphql/graphql-js/pull/163/files#diff-7216927a33b2b307f325a8190e6fbf9bR178 which enforces this.

devknoll:
Now that this is closed, I feel less bad for cluttering up the PR :wink: @leebyron or @schrockn can you clarify what the purpose of the schema language is? Who's the consumer? When/why is it generated?

Sorry, haven't poked around this side of GraphQL much yet :smile:

schrockn:
In the short term we simply wanted to define formally this language which appears in the spec.

Our client tooling often needs a representation of the schema in order to perform various operations. This is human readable as opposed to the result of a schema query.

We also anticipate a number of future uses like client-defined schema over a typeless backend. Things like that.

devknoll:
Would a client-defined schema not use `resolve` functions/be defined like a GraphQL-js schema? I guess they might not have to...

leebyron:
> Would a client-defined schema not use resolve functions/be defined like a GraphQL-js schema? I guess they might not have to...

Correct. Client-side schema are not responsible for execution, so they don't define resolve functions.

This addition is an [IDL](https://en.wikipedia.org/wiki/Interface_description_language) which is just a grammatic visualization of the same information you can derive from Schema introspection. It's useful for it's human readability, and can be parsed back into a Schema instance to use for many purposes.

PR #165, opened 2015-08-31
[RFC] Type extension syntax

leebyron:
This proposal introduces a syntactic addition for client-side tools that perform code generation. It allows those tools to describe "extensions" to the server-supplied GraphQL schema. Tools may then use this information to generate types/classes/objects in the client language which include these client-side-only extensions.

The syntax I'm proposing is:

```
TypeExtensionDefinition : `extend` ObjectDefinition
```

An example of this might look like:

```
type User {
  name: String
  birthday: Int
}

extend type User {
  age: Int
}
```

While I could have proposed simply `extend User { ... }`, I chose to fully reuse the ObjectDefinition rule to preserve the future ability to expand this syntax to include extending interfaces, enums, and other types.

### DIFF
--- a/src/language/ast.js
+++ b/src/language/ast.js
@@ -55,6 +55,7 @@ export type Node = Name
                  | EnumTypeDefinition
                  | EnumValueDefinition
                  | InputObjectTypeDefinition
+                 | TypeExtensionDefinition
 
 // Name
 
@@ -75,6 +76,7 @@ export type Document = {
 export type Definition = OperationDefinition
                        | FragmentDefinition
                        | TypeDefinition
+                       | TypeExtensionDefinition
 
 export type OperationDefinition = {
   kind: 'OperationDefinition';
@@ -322,4 +324,12 @@ export type InputObjectTypeDefinition = {
   loc?: ?Location;
   name: Name;
   fields: Array<InputValueDefinition>;
+}
+
+// Type Extention
+
+export type TypeExtensionDefinition = {
+  kind: 'TypeExtensionDefinition';
+  loc?: ?Location;
+  definition: ObjectTypeDefinition;
 }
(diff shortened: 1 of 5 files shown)


freiksenet:
Are semantics intentionally undefined?

toulouse:
Can you elaborate on what you had in mind for this functionality? Derived properties client-side? Cached client-side values scoped to their attached generated model object?

My initial impression is that it strikes me as a slight complication to GraphQL that probably won't be burdensome to implement, but I think the story, so to speak, that justifies its usage does not yet sound clear.

fson:
I guess this is going to be useful for in-memory and device data in Relay (https://github.com/facebook/relay/issues/114).

toulouse:
Ah, makes sense. I can definitely see 'seen states' as a very compelling application for client-only properties.

Some brainstorming:
If the application of this is client-side code-generation, then I would suggest standardizing an `@transient` directive of some sort to indicate that a client-side property shouldn't be persisted (i.e. Core Data transient properties) to its client-side cache or backing store.

Additionally, if one can define client-only properties, could this concept be extended to client-side-only types? Perhaps with some sort of directives on the schema, i.e. `@scope(availability: "client")`, `@scope(availability: "server")`?

leebyron:
There are intentionally no execution semantics here - this is an addition to the GraphQL IDL which only describes a type system but does not describe how it is executed. In fact, The executor will fail if you provide it with a query that includes an extension or type definition.

> Additionally, if one can define client-only properties, could this concept be extended to client-side-only types? 

Yes, in fact the entire IDL language is intended to be used for client-side-only types.

leebyron:
At Facebook, we use this syntax in our iOS and Android applications to describe additions to model classes that we auto generate. These additions contain data that purely exists on the client.

josephsavona:
This is great and will likely be necessary for facebook/relay#114 as @fson noted. A couple questions:
- Does it make sense to allow non-nullable types in extensions? There's no way to guarantee that client data exists for a server record that hasn't been seen before, for example (or would a non-null client-only type mean that the client implementation just has to provide a default value?).
- Does this allow adding client-only root fields and mutations (via extending the query and mutation types)?

cc @yungsters @yuzhi @wincent

Issue #170, opened 2015-09-06
Inject Orphan Types in schema


Sandreu:
Hello,

I noticed in GraphiQL that I had an unrecognized fragment type, I understood that my type wasn't in the schema introspection, but my request works fine.

I was in this situation

```
Interface Foo { a, b }
obj Bar interfaces Foo {a, b, c }
```

and I tried to query
```
Foo {
    a,
    b,
   ... on Bar {c}
}
```
I will not have queries or graph relationship with Bar directly... so my introspection is partial...
 
I took a look and I saw that push the new type in `schema._typeMap` works fine and I can't see any side effect...

If you agree to a schema.injectType() function, I can PR.

Regards

Sandreu:
My bad #171

PR #171, opened 2015-09-08
Add Schema.injectType method #170

Sandreu:
Hello here is an implementation of injectType.

I didn't add tests since the constructor test code is already testing it :)

Regards.

### DIFF
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -60,15 +60,23 @@ export class GraphQLSchema {
     );
     this._schemaConfig = config;
 
-    // Build type map now to detect any errors within this schema.
-    this._typeMap = [
-      this.getQueryType(),
-      this.getMutationType(),
-      __Schema
-    ].reduce(typeMapReducer, {});
+    this._typeMap = {};
 
-    // Enforce correct interface implementations
-    Object.keys(this._typeMap).forEach(typeName => {
+    // Build type map now to detect any errors within this schema.
+    this.injectType(this.getQueryType());
+    this.injectType(this.getMutationType());
+    this.injectType(__Schema);
+  }
+
+  injectType(newType: ?GraphQLType): void {
+    var oldKeys = Object.keys(this._typeMap);
+
+    this._typeMap = typeMapReducer(this._typeMap, newType);
+
+    // Enforce correct interface implementations for new keys
+    Object.keys(this._typeMap).filter(typeName => {
+      return oldKeys.indexOf(typeName) === -1;
+    }).forEach(typeName => {
       var type = this._typeMap[typeName];
       if (type instanceof GraphQLObjectType) {
         type.getInterfaces().forEach(


leebyron:
Can you clarify the purpose of this PR? I'm not sure I understand it from the code alone.

Sandreu:
The situation is this one :

```
Interface Foo {
  a,
  b
}

Bar interfaces Foo {
  a,
  b,
  c
}

query {
  allFoos {...}
}
```

and the query
```
query {
  allFoos {
    a,
    b,
   ... on Bar {c}
  }
}
```

Then will not be able to solve `Bar` since the schema doesn't know it.

It works if you put a `resolveType` in `Foo` but if you want to deal with `isTypeOf`, it's not.
And even if you use `resolveType`, your introspection will not be aware of `Bar`

In fact, it may be a way to hide a branch of your tree from introspection ! ;)

So with this PR, you can do a `schema.injectType(barType)` if you don't want to hide it or use `isTypeOf` :)

I put an issue #170 about this.

Another way to solve this should be to add an option to the constructor to declare an array of orphan types... Perhaps it's even a better way... 

```
type GraphQLSchemaConfig = {
  query: GraphQLObjectType;
  mutation?: ?GraphQLObjectType;
  types?: ?Array;
}
```

Sandreu:
I see it's not clear to put #170 in title :)

dittos:
Why not just use `interfaces` field when creating GraphQLObjectType?

```
import {graphql, GraphQLInterfaceType, GraphQLObjectType, GraphQLString, GraphQLList, GraphQLSchema} from 'graphql';

var foo = new GraphQLInterfaceType({
  name: 'Foo',
  fields: {
    a: { type: GraphQLString },
    b: { type: GraphQLString },
  }
});

var bar = new GraphQLObjectType({
  name: 'Bar',
  fields: {
    a: { type: GraphQLString },
    b: { type: GraphQLString },
    c: { type: GraphQLString },
  },
  interfaces: [foo], // <=== this
  isTypeOf: () => true
});

var query = new GraphQLObjectType({
  name: 'Query',
  fields: {
    allFoos: { type: new GraphQLList(foo) }
  }
});

var schema = new GraphQLSchema({
  query: query
});

graphql(schema, `
{
  allFoos {
    __typename
    a,
    b,
    ... on Bar { c }
  }
}
`, {allFoos: [{a: 'a', b: 'b', c: 'c'}]}).then(result => {
    console.log(result.data.allFoos);
    // prints [ { __typename: 'Bar', a: 'a', b: 'b', c: 'c' } ]
});
```

Sandreu:
You're definitively right... I don't know what I screwed up! I did have the interfaces field set but I think my GraphiQL windows hadn't an updated schema and I had a wrong red underline (because my request answer was all right).

Ok so, sorry for making you waste your time !

PR #172, opened 2015-09-10
[Proposal] POC "JSON/Raw/Unchecked/Free/WhateverYouWantAsName" Field Objects

Sandreu:
Hello,

Here is an implementation of a concept that doesn't exists in specs... The idea is to be able to use raw objects as leaf type.

Sometimes your server has to handle data that it doesn't care of, and sometimes theses data are objects... Either your client has to serialize/deserialize it himself, or Raw objects allow that...

This PR allow things like that :

```javascript
import gql, {
  graphql,
  GraphQLString,
  GraphQLSchema,
  GraphQLObjectType,
  GraphQLRawObjectType
} from './src';

var test = new GraphQLRawObjectType({
  name: 'Hello',
});

var Queries = new GraphQLObjectType({
  name: 'Query',
  fields: () => ({
    foo: {
      type: GraphQLString,
      resolve: () => ('Ho !')
    },
    bar: {
      type: test,
      resolve: () => {
        return { val:'Hi !'}
      }
    }
  }),
});

var Schema = new GraphQLSchema({
  query: Queries,
});

graphql(Schema, '{ foo, bar }').then(r => {
  console.log(r);
  // this prints { data: { foo: 'Ho !', bar: { val: 'Hi !' } } }
});
```

/!\ This is only a POC to discuss the subject ! Not an "as it is" Pull Request.

Another way could be to serialize the object on a scalar type, but this would be much less pretty! :D

Let's talk !!

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -663,8 +663,10 @@ function completeValue(
     var serializedResult = returnType.serialize(result);
     return isNullish(serializedResult) ? null : serializedResult;
   }
-  
-  if (returnType instanceof GraphQLRawObjectType) return result;
+
+  if (returnType instanceof GraphQLRawObjectType) {
+    return result;
+  }
 
   // Field type must be Object, Interface or Union and expect sub-selections.
   var objectType: ?GraphQLObjectType;
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -18,6 +18,7 @@ import { getVariableValues, getArgumentValues } from './values';
 import {
   GraphQLScalarType,
   GraphQLObjectType,
+  GraphQLRawObjectType,
   GraphQLEnumType,
   GraphQLList,
   GraphQLNonNull,
@@ -662,6 +663,8 @@ function completeValue(
     var serializedResult = returnType.serialize(result);
     return isNullish(serializedResult) ? null : serializedResult;
   }
+  
+  if (returnType instanceof GraphQLRawObjectType) return result;
 
   // Field type must be Object, Interface or Union and expect sub-selections.
   var objectType: ?GraphQLObjectType;
--- a/src/index.js
+++ b/src/index.js
@@ -17,6 +17,7 @@ export { GraphQLSchema } from './type/schema';
 export {
   GraphQLScalarType,
   GraphQLObjectType,
+  GraphQLRawObjectType,
   GraphQLInterfaceType,
   GraphQLUnionType,
   GraphQLEnumType,
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -87,7 +87,8 @@ export function isOutputType(type: ?GraphQLType): boolean {
     namedType instanceof GraphQLObjectType ||
     namedType instanceof GraphQLInterfaceType ||
     namedType instanceof GraphQLUnionType ||
-    namedType instanceof GraphQLEnumType
+    namedType instanceof GraphQLEnumType ||
+    namedType instanceof GraphQLRawObjectType
   );
 }
 
@@ -102,6 +103,7 @@ export function isLeafType(type: ?GraphQLType): boolean {
   var namedType = getNamedType(type);
   return (
     namedType instanceof GraphQLScalarType ||
+    namedType instanceof GraphQLRawObjectType ||
     namedType instanceof GraphQLEnumType
   );
 }
@@ -174,6 +176,22 @@ export function getNamedType(type: ?GraphQLType): ?GraphQLNamedType {
   return unmodifiedType;
 }
 
+export class GraphQLRawObjectType {
+  name: string;
+  description: ?string;
+
+
+  constructor(config: GraphQLObjectTypeConfig) {
+    invariant(config.name, 'Type must be named.');
+    assertValidName(config.name);
+    this.name = config.name;
+    this.description = config.description;
+  }
+
+  toString(): string {
+    return this.name;
+  }
+}
 
 /**
  * Scalar Type Definition
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -34,7 +34,8 @@ export type GraphQLType =
   GraphQLEnumType |
   GraphQLInputObjectType |
   GraphQLList |
-  GraphQLNonNull;
+  GraphQLNonNull |
+  GraphQLRawObjectType;
 
 export function isType(type: any): boolean {
   return (
@@ -163,7 +164,8 @@ export type GraphQLNamedType =
   GraphQLInterfaceType |
   GraphQLUnionType |
   GraphQLEnumType |
-  GraphQLInputObjectType;
+  GraphQLInputObjectType |
+  GraphQLRawObjectType;
 
 export function getNamedType(type: ?GraphQLType): ?GraphQLNamedType {
   var unmodifiedType = type;
--- a/src/type/index.js
+++ b/src/type/index.js
@@ -27,6 +27,7 @@ export {
   // Definitions
   GraphQLScalarType,
   GraphQLObjectType,
+  GraphQLRawObjectType,
   GraphQLInterfaceType,
   GraphQLUnionType,
   GraphQLEnumType,


freiksenet:
I've seen this requested couple of times, but I am not sure it's such a good idea.

When is this kind of object preferred over a defined ObjectType? I think the great thing about graphql is that you can be sure that the data your get is in the shape you requested. With such arbitrary objects you are removing that benefit and removing all the typechecking related to it.

freiksenet:
Also if we do go this way, I'd say it should be a scalar type, because then one could ensure it's validity as JS object in parse/serialize methods.

Sandreu:
Hello, 

Sometimes you need your backend to save data that are absolute client's app data. It could be 
  - cards order/size on a dashboard,
  - filters samples,
  - visible columns list
  - ...

This kind of type is just to isolate UI data that the server hasn't to understand... when it just has to store and provide it...
If the client's team would like to change the way data is organised, it doesn't need the backend's team to review all their schema layer...

I agree that the intend is really near scalar types but you don't have to serialize/deserialize it to go through validation...
And give the object right away into the answer is really nicer to clients than a serialized object...

By the way, I'am starting to think that maybe a Raw Type may be Type free (I am thinking out loud) and leave validation to the real case implementations... In fact the serialization forces you to have a correct object to serialize... Leave validation to the one consuming it... the client... and the server just has to prevent malicious code according storage system... eventually...

I agree that devs have to be very cautious about using that kind of types, and should not miss the intend of such a type, but maybe graphql could provide a config interface forcing a validate method to warn the server dev that no validation is made on raw type... 

That's why I think scalar clarifies the intend... but to give object into answer is really nicer once the aim is well understood...

taion:
IMO the biggest use case for something like this is sending JSON blobs in a few cases where I actually do want to send shapeless data to my front end.

However, to my mind, that's much better handled by sending the data across the wire as a string blob and deserializing semantically in Relay, per: https://github.com/facebook/relay/issues/91

I feel like "raw" is very much a misnomer here because, while JS objects are primitives in JS, there's nothing _special_ about them except the JSON-ness. JSON objects are quite awkward in static type systems, while there's nothing special about strings as keys in many languages outside of JS.

What additional problems get solved by adding this to the GraphQL spec, rather than having clients that can do generic hydration of data into types beyond the GraphQL type system?

Sandreu:
You talk about JSON blobs as if it is a brand new concept to graphql. 
In fact answers are JSON, and that doesn't seem odd to me to get an object naturally represented in the answer... On the other hand JSON string representations into JSON represented object is much more weird...

But I agree about 2 things :
 - Raw may be a wrong name for this,
 - Untyped type may be a bad idea.

This doesn't solve much I agree, but at least the double JSON parsing that you have to deal with relay like implementations where extend scalar types is not quite simple.. (Not sure about that, but https://gist.github.com/taion/df126b9252825927b037 seems quite verbose for something that is already done at the very beginning...)

Sandreu:
By the way Scalar types are greatly handled in graphql-js!

Sandreu:
What do you think about NestedMaps as type name?

Sandreu:
But NestedMaps doesn't mean lists... so I prefer RawObject, but I don't really care... As you like...

magalhas:
+1

magalhas:
+1

josemarluedke:
:+1:

jhurliman:
I will leave our example use case for this feature here. We have a machine learning service that returns both a prediction and a "feature map", a JSON object containing keys and numeric values that were used as inputs to the prediction. This map is fairly large (currently around 120 key/value pairs), evolving, and not at all important to the API spec. From both the server and client perspective it is simply an opaque object that can be stored and optionally sent back to the server later for debugging purposes. We could certainly change the API to make this a serialized JSON string, but it feels like a workaround. I understand the argument about this opening the door to semi-opaque (and half-documented) APIs, but would ultimately still prefer the flexibility if possible.

taion:
@jhurliman 

I feel like your example is quite different from what @Sandreu is proposing. Among other things, you actually have something that is easily and usefully expressible in a type system as something like `map<string, double>`. Among other things this is a concept that can be represented in most statically typed languages, without calling out into some explicit library for supporting JSON.

taion:
My primary criticism is that calling this a "raw object" makes no sense, since it's only anything like a "raw object" in JS.

That's not to say there wouldn't be value in something like a `GraphQLJSONType` (`GraphQlJsonType` if you like Google-flavored camel-casing), but I think that gets more at the real question, which is whether JSON is some sort of special scalar type that needs this level of support, or whether we'd be better off handling this with something like https://github.com/facebook/relay/issues/91.

Sandreu:
Hello @jhurliman,

I understand your feeling, and I know that you are a tiny use case where this would help you, but I agree with you, If you want to use a JSON representation of a "scalar" data, GraphQL should be your friend since it is its native transport applicative protocol

However, your use case is quite easy to solve and as @taion said, the use of a list of key and value objects can easily do the job:
map : [{key, value}]

Still, I agree with you : change a JSON representation to another even if GraphQL knows JSON very well because of specs limitations is very frustrating !

Sandreu:
I already told how I feel about a stringified JSON object into a stringified JSON object... :)

freiksenet:
I don't quite get how @jhurliman case in different from @Sandreu case. I think both can be represented with a Map data type that @taion proposed. Do we really need unstructured row object or just a way to represent arbitrary Maps?

Sandreu:
Difference is nesting, in fact I propose the ability to use Javascript objects as a primitive type... without any field control.
@jhurliman case is just a basic map...

freiksenet:
I do feel Maps and Lists can work for the most things that you've mentioned above.

Note that not every language has unstructured nested JSON-like thing like JavaScript.

Sandreu:
Ok, I think we are saying the same thing :) JSON-like handles Maps but Maps are only a piece of what JSON-like brings :)

jhurliman:
There is discussion of adding a map type @ https://github.com/facebook/graphql/issues/101

clintwood:
Is there any movement on this?  Could really do with 'ad-hoc' JSON objects (Raw Objects)!  I'm currently stringifying to GraphQLString fields but that is not ideal.

taion:
:+1: to @freiksenet's point - a JSON object is not at all "raw" outside of JS.

To me, this really just seems like a potentially application-specific custom scalar type. Maybe the real question is one of having a better way to hydrate this on the client.

clintwood:
@taion, in the context of other languages I agree and the same requirement may surface for those language specific types too.  So maybe what we looking for is a way to:
- Mark a field as 'raw' (better word needed)
- Provide a custom serializer/deserialzer for this 'raw' field

freiksenet:
@clinton, at least on Graphql server side, this is already possible with custom scalar type. The only problem is Relay support.

clintwood:
@freiksenet, Ah!, ok, I'm not using Relay, I must have missed the custom scalar type - thanks for the heads-up!

taion:
Exactly - custom scalar types are IMO the way to go. This concept of "raw" is terribly unrobust.

clintwood:
@taion, @freiksenet if you referring to `GraphQLScalarType` then that totally solves this issue for me - awesome.

Sandreu:
Hello,

Ok, Raw may be a bad name, but we're not here to talk about the name.... And I agree that serialization is a good way to solve it, and it is the way I do today...
But again, the aim here is to avoid serialized JSON into serialized JSON... Which feels like a workaround !

And I don't really see why this concept is more "unrobust" than a simple string ! GraphQL is not here to protect you from malicious code... And if you're not carefull with strings you can have serious damages... And if you want to talk about the robustness of JSON... again GraphQL **IS** already talking JSON !

In fact I don't see what is **terribly unrobust** here...?

taion:
How do you think JSON is transmitted on the wire, if not as a serialized string?

Sandreu:
Title changed :D

Sandreu:
> But again, the aim here is to avoid serialized JSON into serialized JSON... Which feels like a workaround !

Or ... 
> (...) double JSON parsing that you have to deal with relay > like implementations where extend scalar types is not quite simple.. (Not sure about that, but https://gist.github.com/taion/df126b9252825927b037 seems quite verbose for something that is already  done at the very beginning...)

taion:
Right - the actual issue is more on https://github.com/facebook/relay/issues/91, where there isn't currently a great way to hydrate custom scalar fields in Relay.

Other clients may well not have the same limitation. Fixing https://github.com/facebook/relay/issues/91 also gives people a much better way to handle e.g. `Date` objects.

Sandreu:
Agree about this ticket ! https://github.com/facebook/relay/issues/91 should be fixed ! :)

But the point here, is, whatever your client library is :

serialized JSON int serialized JSON **IS** an awkward/workaroud/strange/nothingVeryPleasant feeling :D

to have to change a map to a `[{key, value}]` in a context speaking JSON **IS** an awkward/workaroud/strange/nothingVeryPleasant feeling :D

taion:
What do you mean? It'd be the same as for every custom scalar type.

You pull it off the wire and have whatever kind of deserialization logic you need. In JS it's just `JSON.stringify` and `JSON.parse`.

Sandreu:
Is it me or do I miss something, because I have the feeling that you're both trying to say something to the other but it seems that we don't understand what the other is trying to say :)

Sandreu:
I'm not trying to make a spec for every kind of data GraphQl may provide.

GraphQL **IS talking JSON** and nested JSON serialization feels odd ! I can't see how I can be simpler...

Moreover, multiple JSON parsing may, sometimes (as in relay) not to be as easy as it should be...

Perhaps we should drop this ticket... The concept may be too complicated... Or too specific to my use case...

sofish:
:+1:

luiscarli:
Hi, I found this issue because I'm looking for ways to send thousands of objects to the front-end, to be visualized. 
I does not seem optimal to fit those big arrays on graphql type system because of all the checks and mapping that they pass through on Relay.
For that reason it would be nice to have a unchecked leaf node with any type of json structure.

steveluscher:
Thanks for all of the discussion everyone!

Pursuant to @Sandreu's suggestion that we drop this PR in favor of serialization/deserialization of custom scalar values, I've spent the morning drafting up a specification here: https://github.com/facebook/relay/issues/91#issuecomment-228906920

Let's pull for that solution. We're not pursuing this internally, but if someone wants to pitch in with an implementation of what I've sketched out, we'd be happy to take a look at a PR!

taion:
@steveluscher It turned out that the special case of JSON scalar types "just worked" across both graphql-js and Relay as a custom scalar type with trivial serialization/deserialization handling on the GraphQL server side, and they "just work" in Relay as well. I believe @leebyron was considering adding that type to graphql-js, as it's a little bit unintuitive, but the existing solution for handling JSON scalar types is pretty good, though perhaps not broadly known.

PR #177, opened 2015-09-17
Replaced validator kind strings with constants

woodb:
There were a few places in the validation rules that were not using the
constants found in `src/language/kinds.js`. This commit replaces those strings
with their respective constants.

This can be considered a cleanup.


woodb:
:+1: @leebyron

leebyron:
Cool, thanks

PR #197, opened 2015-10-04
Define possibleTypes on InterfaceType

tgriesser:
Mutation of interfaces in type definition is problematic, as it allows for incomplete schema generation if not all types are loaded at runtime.

This adds an optional `possibleTypes` property on the InterfaceType definition. It may be an Array or thunk, resolving with an Array of types an interface may implement.

It also allows a thunk for interfaces to act as intended (lazily). Currently it is immediately invoked as a side-effect of the `addImplementationToInterfaces` call in constructor.

It's a bit of a big change, so let me know if you have any questions or need me to clarify anything about where I ran into this as an issue.


tgriesser:
Closing in favor of #199

PR #199, opened 2015-10-07
Add GraphQLSchema types field

tgriesser:
Per discussion with @leebyron - adds a "types" field on the `GraphQLSchema` config, requiring all `GraphQLObjectType`s implementing interfaces be specified on the Schema definition. `GraphQLUnionType`s may be specified in place of their possible `ObjectType`s.

Moves the `InterfaceType` mutation of implementations from the `ObjectType` constructor to the definition of the Schema, properly deferring a thunk provided as the "interfaces" prop on ObjectType configuration.

This helps prevent the situation where where types are required lazily, thus making it possible an incomplete schema definition could be generated due to runtime mutation of the schema.

### DIFF
--- a/src/__tests__/starWarsSchema.js
+++ b/src/__tests__/starWarsSchema.js
@@ -270,5 +270,6 @@ var queryType = new GraphQLObjectType({
  * type we defined above) and export it.
  */
 export var StarWarsSchema = new GraphQLSchema({
-  query: queryType
+  query: queryType,
+  types: [ humanType, droidType ]
 });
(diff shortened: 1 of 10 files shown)


leebyron:
Just wanted to let you know that I haven't forgotten about this - I'd like to do this. I believe this is a breaking change though, and want to wait until we've flushed out some remaining issues with the v0.4.x line before digging into this :)

tgriesser:
Awesome, thanks for the heads up! I figured it was a kind of bigger change and might require a bit more thought.

leebyron:
The time has come to make this happen! I just rebased atop recent work and went a little further down the rabbit hole in #327. I'm closing this PR in favor of that, but it's still heavily based on your work, @tgriesser so thanks again.

tgriesser:
Awesome, looks great! Thanks @leebyron!

PR #204, opened 2015-10-12
Make input objects and lists return better errors

freiksenet:
Add index of invalid element or field of invalid field and make it recurse
into children until faulty place is found.

### DIFF
--- a/src/execution/values.js
+++ b/src/execution/values.js
@@ -113,9 +113,8 @@ function getVariableValue(
   }
   var message = errors ? '\n' + errors.join('\n') : '';
   throw new GraphQLError(
-    `Variable "$${variable.name.value}" expected value of type ` +
-    `"${print(definitionAST.type)}" but got: ${JSON.stringify(input)}.` +
-    `${message}`,
+    `Variable "$${variable.name.value}" got invalid value ` +
+    `${JSON.stringify(input)}.${message}`,
     [ definitionAST ]
   );
 }
--- a/src/execution/values.js
+++ b/src/execution/values.js
@@ -94,7 +94,8 @@ function getVariableValue(
     );
   }
   var inputType: GraphQLInputType = (type: any);
-  if (isValidJSValue(input, inputType)) {
+  var errors = isValidJSValue(input, inputType);
+  if (!errors.length) {
     if (isNullish(input)) {
       var defaultValue = definitionAST.defaultValue;
       if (defaultValue) {
@@ -110,9 +111,11 @@ function getVariableValue(
       [ definitionAST ]
     );
   }
+  var message = errors ? '\n' + errors.join('\n') : '';
   throw new GraphQLError(
     `Variable "$${variable.name.value}" expected value of type ` +
-    `"${print(definitionAST.type)}" but got: ${JSON.stringify(input)}.`,
+    `"${print(definitionAST.type)}" but got: ${JSON.stringify(input)}.` +
+    `${message}`,
     [ definitionAST ]
   );
 }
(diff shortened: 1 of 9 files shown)


src/validation/rules/DefaultValuesOfCorrectType.js
```diff
@@ -27,10 +27,12 @@ export function defaultForNonNullArgMessage(
 export function badValueForDefaultArgMessage(
   varName: any,
   type: any,
-  value: any
+  value: any,
+  verboseErrors?: [any]
 ): string {
+  var message = verboseErrors ? [ '', ...verboseErrors ].join(' ') : '';
```
leebyron: rather than duplicating the verboseErrors array here, you could replace this with:

```js
var message = verboseErrors ? ' ' + verboseErrors.join(' ') : '';
```

src/execution/values.js
```diff
@@ -110,9 +111,11 @@ function getVariableValue(
       [ definitionAST ]
     );
   }
+  var message = errors ? [ '', ...errors ].join(' ') : '';
```
leebyron: Similar feedback here

src/type/definition.js
```diff
@@ -1055,3 +1055,9 @@ function assertValidName(name: string): void {
     `Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "${name}" does not.`
   );
 }
+
+// Validation errors type
+export type TypeValidationError = {
```
leebyron: This is a weird place for this, it doesn't really have anything to do with defining GraphQL types.

src/utilities/isValidJSValue.js
```diff
@@ -25,25 +25,43 @@ import type { GraphQLInputType } from '../type/definition';
  * accepted for that type. This is primarily useful for validating the
  * runtime values of query variables.
  */
-export function isValidJSValue(value: any, type: GraphQLInputType): boolean {
+export function isValidJSValue(
+  value: any, type: GraphQLInputType
```
leebyron: nit: each argument on its own line when using multiple lines

src/utilities/isValidJSValue.js
```diff
@@ -73,5 +110,7 @@ export function isValidJSValue(value: any, type: GraphQLInputType): boolean {
 
   // Scalar/Enum input checks to ensure the type can parse the value to
   // a non-null value.
-  return !isNullish(type.parseValue(value));
+  return {
+    isValid: !isNullish(type.parseValue(value))
```
leebyron: should this case also present a clear error message?

leebyron:
This is really awesome, thanks for putting this together!

I think it may simplify things to change the return type from `{ isValid: boolean, errors: Array<string> }` to just `Array<string>`, treating an empty array as valid.

I'd also really love to get even more specific with these errors to describe the type names expected when one is missing and to make the error verbiage a little easier to follow.

For example, when validating a literal value it would be great for the AST data in the GraphQLError to be as specific as possible, pointing to the exact sub-value that poses the problem.

freiksenet:
I've fixed the issues you underlined. Sorry for the squash, history started to look ugly %)

I would prefer to go on without adding exact locations yet, cause I am not sure how exactly to go on there. For instance, there can be multiple errors in one input object, what should be the location in that case? Also this would require passing error locations up to the top level, which is complicated too.

I've made \n separators for errors so that it looks good in GraphiQL.

graphiql

freiksenet:
Uh, damn, forgot to fix the test messages. :)

freiksenet:
Now it's ready.
graphiql

leebyron:
Awesome, thanks for these updates. I'll read carefully and merge as soon as I can!

freiksenet:
Thank you!

PR #227, opened 2015-11-12
extendSchema()

leebyron:
This new utility function produces a new Schema, provided an existing one and a document containing schema type definitions and extensions.

### DIFF
--- a/src/utilities/__tests__/extendSchema.js
+++ b/src/utilities/__tests__/extendSchema.js
@@ -1,0 +1,636 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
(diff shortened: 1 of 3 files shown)


jontonsoup:
@leebyron does this enable me to combine two schemas that have already been built? 

I tried using it and got this error: 

```
Error: Must provide valid Document AST
```

Thanks!

leebyron:
You could print one and then use that to extend the other, but otherwise no. This is for extending a schema for metadata only for use by client tools

jontonsoup:
@leebyron 

so something like this would work? 
https://github.com/graphql/graphql-js/blob/master/src/utilities/schemaPrinter.js#L28

```
printed_schema = printSchema(first_schema)
combined_schema = extendSchema(second_schema, printed_schema)
```

leebyron:
You need to parse the printed schema into a document AST first, but yes

jontonsoup:
@leebyron thanks! Whats the AST parser function? 

This one?

https://github.com/graphql/graphql-js/blob/2f10ad44e7e6e8e2865c83ac006105203001eae6/src/language/parser.js#L122

Issue #53, opened 2015-07-14
Querying same field w/ and w/o attributes for different union types results in error


mnylen:
Query:

    {
      curatedList {
         ...on Program {
            title(lang: "fi")
         }

         ...on Article {
            title
         }
      }
    }

Error:

    Fields title conflict because they have differing arguments.

Should it really work like that, or is this a bug?

Article and Program are both members of union type Content

leebyron:
This is currently working as designed, but feedback is welcome. 

The design goal here is to make the resulting response unambiguous. 

If for example, Article and Program were both Interfaces that one type adhered to, it would be unclear which variant of "title" was being requested.

schrockn:
This is very deliberate. Any time that two fields that "merge" can *potentially* return different results we do not allow this. 

Imagine if "title" returned "foo" and "title(language: fi" returned "bar" which value would the "title" key be?

schrockn:
If you wish to query for both you should alias at least one of the fields to be unambiguous.

leebyron:
I think we could probably improve this if two fields are known to never "merge". In this case, the two types are Object types and in no condition would both "title" fields be queried for the same object. 

We should be a bit smarter about this (likely common) case

mnylen:
Is there any possibility for confusion when using the `...on TypeName { ... }` query, even if interfaces were used instead of union types?

If there was a SearchResult interface that defined a field `pictureUrl(width: Int)` and a `title`. The interface is implemented by `Page` and `Image`. Wouldn't it be a valid use case to do something like this:

     ...on Page { title, pictureUrl(width: 120) }
     ...on Image { title, pictureUrl(width: 600) }

This would also result in error with current implementation.

mnylen:
Actually, even if the query was something like

    title,
    pictureUrl(width: 120),
    ...on Image { pictureUrl(width: 600) }

that'd still be pretty easy to reason about: select 120px picture url for anything that isn't an Image, and 600px for Image.

mnylen:
The only problem I can think of is with case where an interface defines a field with some set of args and it's implemented with different set of args. For example, if `Article` and `Program` implemented a `HasTitle` interface with a `title` field w/o args, it would clearly be an error for `Program` to have one with `language` arg.

In this case it would be pretty hard to respond to query like `...on HasField { title }`, because you can only apply that query to `Article` - `Program` needs the `language` arg.

But this should be a compile time check to ensure you can't create such schema in the first place. Instead of it happening on query time.

[...]

PR #229, opened 2015-11-13
[validation] Allow safe divergence

leebyron:
As pointed out in #53, our validation is more conservative than it needs to be in some cases. Particularly, two fields which could not overlap are still treated as a potential conflict.

This change loosens this rule, allowing fields which can never both apply in a frame of execution to diverge.


leebyron:
cc @schrockn @alangenfeld @mnylen

schrockn:
yeah this makes a lot of sense #shipit

PR #231, opened 2015-11-13
[Validation] perf improvements for fragment cycle detection

leebyron:
This converts the O(N^2) cycle detector to an O(N) detector and optimizes the visit to short-circuit where parts of the validation visitor were unnecessary.

On an internal very large stress test query this was previously ~30s and after this diff runs in 15ms.


leebyron:
In addition, also finds more cycle cases the previous algorithm did not. Win!

KyleAMathews:
👏🏻

PR #239, opened 2015-11-17
[Schema] Implementing interfaces with covariant return types.

leebyron:
This proposes loosening the definition of implementing an interface by allowing an implementing field to return a subtype of the interface field's return type.

This example would previously be an illegal schema, but becomes legal after this diff:

```graphql
interface Friendly {
  bestFriend: Friendly
}

type Person implements Friendly {
  bestFriend: Person
}
```

### DIFF
--- a/src/type/__tests__/validation.js
+++ b/src/type/__tests__/validation.js
@@ -1579,12 +1579,7 @@ describe('Objects must adhere to Interface they implement', () => {
         name: 'AnotherInterface',
         resolveType: () => null,
         fields: {
-          field: {
-            type: GraphQLString,
-            args: {
-              input: { type: GraphQLString },
-            }
-          }
+          field: { type: GraphQLString }
         }
       });
 
@@ -1592,12 +1587,7 @@ describe('Objects must adhere to Interface they implement', () => {
         name: 'AnotherObject',
         interfaces: [ AnotherInterface ],
         fields: {
-          field: {
-            type: SomeScalarType,
-            args: {
-              input: { type: GraphQLString },
-            }
-          }
+          field: { type: SomeScalarType }
         }
       });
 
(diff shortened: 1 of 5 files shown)


leebyron:
Implements https://github.com/facebook/graphql/pull/123

src/type/__tests__/validation.js
```diff
@@ -1711,14 +1809,61 @@ describe('Objects must adhere to Interface they implement', () => {
...
 
       return schemaWithFieldType(AnotherObject);
     }).to.throw(
       'AnotherInterface.field expects type "String" but ' +
-      'AnotherObject.field provides type "String!".'
+      'AnotherObject.field provides type "[String]".'
+    );
+  });
+
+  it('accepts an Object with a subset modified Interface field type', () => {
```
dschafer: I might mention null/non-null in this string.

src/utilities/typeComparators.js
```diff
@@ -0,0 +1,85 @@
...
+
+  // Otherwise the types are not equal.
+  return false;
+}
+
+/**
+ * Provided a parent type and a child type, return true if the second is either
+ * equal or a subset of the second (covariant).
+ */
+export function isSubType(parent: GraphQLType, child: GraphQLType): boolean {
```
dschafer: Hm, I wish there were some way to indicate the ordering here;

```js
isSubType(a, b)
```

doesn't make it immediately clear to me whether we're checking `a` as a subtype of `b` or vice versa.

dschafer:
Looks good to me, this is a nice cleanup of the `isEqualType` / `isSubType` implementations too.

leebyron:
Thanks! Some follow up hopefully makes the ordering more clear, but I agree it's a hard one to name.

PR #246, opened 2015-11-21
Use the same babel options in npm published package

plievone:
Followup commit to c61ccac where babel plugin options were inlined to command line. Brings npm published version and built version in sync.


leebyron:
good catch

PR #250, opened 2015-11-23
Fix tests by updating babel-eslint

plievone:
Tests started failing a few days ago, when escope (eslint dependency) updated v3.2.0 to v3.2.1, with updated dependencies. There may still be fixes coming in babel-eslint, escope, and their dependencies, but pumping babel-eslint for now seems to fix the tests.


leebyron:
Thanks for digging into this.

PR #252, opened 2015-11-29
add main to package.json

rtorr:
Adding main to package.json tells consumers what is the main eatery point (as im sure you all know) [https://docs.npmjs.com/files/package.json#main](https://docs.npmjs.com/files/package.json#main). I know consumers should fallback to index.js, but its is still nice to have. I should probably add this case on the thing I am building as well.


leebyron:
Thanks

PR #254, opened 2015-12-01
Bug report: visitInParallel doesn't skip subtrees when enter() returns false

igorcanadi:
Bug report comes with a test that currently fails. This is inconsistent with the comment above the function:

    [...] returning false to skip sub-branches is supported.


leebyron:
Sloppy mistake, thanks for the breaking test case. I included it along with the fix.

PR #255, opened 2015-12-01
Fix a bug where KnownTypeNames rule incorrectly fails

igorcanadi:
To better understand the bug, check out the new test in __tests__/KnownTypeNames.js.

What happens is some part of our client-side schema might be left
non-imported, because it's not referenced (since expandSchema works by
continiously expanding root queryTypes). However, when a non-imported
type refrences another non-imported type KnownTypeNames reports a failure.
To fix this, I added a rule to skip a subtree of a non-imported type in
KnownTypeNames.

I also had to fix validate.js because visitInParallel() doesn't support
skipping the subtree when enter() returns false. See bug report #254.

### DIFF
--- a/src/validation/rules/KnownTypeNames.js
+++ b/src/validation/rules/KnownTypeNames.js
@@ -21,9 +21,22 @@ export function unknownTypeMessage(type: any): string {
  *
  * A GraphQL document is only valid if referenced types (specifically
  * variable definitions and fragment conditions) are defined by the type schema.
+ *
+ * We ignore the types that are referenced by types that are not part of the
+ * schema.
  */
 export function KnownTypeNames(context: ValidationContext): any {
   return {
+    ObjectTypeDefinition(node) {
+      var typeName = node.name.value;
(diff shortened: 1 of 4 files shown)


src/validation/rules/KnownTypeNames.js
```diff
@@ -21,9 +21,22 @@ export function unknownTypeMessage(type: any): string {
  *
  * A GraphQL document is only valid if referenced types (specifically
  * variable definitions and fragment conditions) are defined by the type schema.
+ *
+ * We ignore the types that are referenced by types that are not part of the
+ * schema.
  */
 export function KnownTypeNames(context: ValidationContext): any {
   return {
+    ObjectTypeDefinition(node) {
```
leebyron: We shouldn't be validating client side type definitions at all, presently. I think you can safely return false without any other logic here. Similar will be necessary for other kinds of type definitions.

igorcanadi:
Thanks for fixing this so fast!

PR #256, opened 2015-12-03
Update Int type description to match implementation

benbjohnson:
This commits updates the `GraphQLInt` type description to match the changes made in 06f97b6.


leebyron:
Good catch, thank you!

PR #261, opened 2015-12-07
Fix the error message for unknown field

fson:
The error message for unknown field was literal:
"In field "${providedField}": Unknown field."

Add necessary back ticks to the template literal so the message
includes the actual field name. Also change the test that claims to
test this case to actually test it.


freiksenet:
Ship.

freiksenet:
Hahaha! :) Thought it's a PR in our internal repo :D Sorry :)

Aweary:
@leebyron is this going to be merged at some point? We're seeing this issue as well.

leebyron:
Nice find and fix! #clowntown

PR #264, opened 2015-12-14
Add directives to FieldDefinition

igorcanadi:
This pull requests adds a new feature to the language -- directives for fields in type definitions and extensions. For example, now you're able to do this:

    type Hello {
       world: String @mock(value: "hello")
    }

We need directives for fields to enable the mock functionality we're building -- we'd like product developers to quickly specify some mock data on top of which they can build UI. That way we decouple building UI and server endpoints.

We could add directives to fields in a query, but if the platform is using type models, it's easier to add @mock to fields in a type definition -- the auto-generated type model just needs to override the default field getter.

### DIFF
--- a/src/language/__tests__/kitchen-sink.graphql
+++ b/src/language/__tests__/kitchen-sink.graphql
@@ -55,3 +55,8 @@ fragment frag on Friend {
   unnamed(truthy: true, falsey: false),
   query
 }
+
+extend type User {
+  @iAmAnAnnotation(default: "Foo")
+  name: String
+}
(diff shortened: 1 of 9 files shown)


igorcanadi:
BTW this PR is just extending the language, AST and visitor. I'm still working on a separate PR that will enable directives in extendSchema(). Let me know if there's anything else I also need to implement.

igorcanadi:
I'll merge this PR with #265. Stacked diffs are a pain to handle in github.

PR #266, opened 2015-12-21
Changed use of apostrophes to back ticks so that string interpolation works

urbanappetite:
Simple fix


urbanappetite:
Didn't see the other pull request for this - can be rejected

Issue #133, opened 2015-08-18
GraphQL should allow mutating optional fields to 'null'.


johanatan:
Unless I'm overlooking something, I'm not seeing any way to pass `null` into a mutation for an 'optional' arg. It seems that the keys with `null` values get stripped out inside GraphQL-JS itself before making it to my `resolve` in the `obj` parameter.

leebyron:
In GraphQL input variables, there is no difference between `null` and `undefined`. The value `null` means "lack of a value".

While passing null as a sentinel value to mean "delete this value" is a pattern sometimes seen in JavaScript mutation APIs, it's not one that GraphQL is well suited to directly mirror at the moment.

Does this match what you're trying to do, or is there more information you could provide about your use case?

johanatan:
Yes, that's what I'm trying to do:  delete a pre-existing [optional] value. Of course, deleting a required (i.e., GraphQLNonNull) value should not be permitted.

Do you have any recommended workaround for this?

leebyron:
Since GraphQL does not have this sentinal value, I recommend a more explicit form for deleting things:

```
mutation myMutation {
  editThing(id: 3, additions: { foo: "foo" }, deletions: [ "bar" ]) { ... }
}
```

johanatan:
That would work. However, note that using `null` is also explicit. The above would be:

```
mutation x {
  editThing(id: 3, values: {foo: "foo", "bar": null }) { ... }
}
```

whereas just changing `foo` and leaving `bar` (as well as `baz` or whatever else may be defined) to [their] pre-existing value(s) would be:
```
mutation x {
   editThing(id: 3, values: {foo: "foo"}) { ... }
}
```

i.e., the presence of a key with a value of null is an explicit intention whereas the absence of key is an implicit one.

Any chance this could be supported at some point?

leebyron:
We do not plan on supporting this since most environments do not have different `null` and `undefined` sentinel values - that's a specific feature of JavaScript and GraphQL is designed to be supported by many more platforms.

> However, note that using null is also explicit.

I don't personally agree with this. Using `null` to implicitly mean "delete this value" is a JavaScript-ism that doesn't translate well to platform-agnostic systems. I personally much prefer the explicit `deletions: ` field.

I fear that if we embraced this concept, the implicit `null` would be misinterpreted by different servers and result in the accidental deletion of data.

toulouse:
[...]

PR #268, opened 2015-12-21
Allow graphql to pass through null values instead of ignoring them

lprhodes:
As set out in #133 and https://github.com/facebook/graphql/pull/83

### DIFF
--- a/src/execution/values.js
+++ b/src/execution/values.js
@@ -130,7 +130,7 @@ function coerceValue(type: GraphQLInputType, value: any): any {
     return coerceValue(nullableType, value);
   }
 
-  if (isNullish(value)) {
+  if (value === undefined) {
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


urbanappetite:
I'm not seeing any trailing whitespace - any ideas?

leebyron:
Thanks for investigating this. It looks like the linter is complaining about line 164 in values.js. I'd suggest configuring your IDE to trim trailing whitespace on save.

Right now it's very much intentional how null values are treated because of how the GraphQL spec is written. I'm still working on that spec change which will determine the runtime behavior changes.

src/execution/values.js
```diff
@@ -148,10 +148,9 @@ function coerceValue(type: GraphQLInputType, value: any): any {
     return Object.keys(fields).reduce((obj, fieldName) => {
       var field = fields[fieldName];
       var fieldValue = coerceValue(field.type, value[fieldName]);
-      if (isNullish(fieldValue)) {
+      if (fieldValue === undefined) {
         fieldValue = field.defaultValue;
-      }
-      if (!isNullish(fieldValue)) {
+      } else {
```
leebyron: this seems like a mistake - did your unit tests pass or are we missing a test for this condition? As I read it, if a fieldValue is coerced to be undefined, then we set `fieldValue` to the default value, however never actually set it on the response object.

wincent:
Closing this one due to staleness.

PR #273, opened 2016-01-07
Include fields for GraphQLError

### LINKED ISSUE #192: Name of a field inside errors:{locations:[]}
Can we add somehow the name of a field which has an invalid value? I would like to use this inside `parseLiteral` - that's where I do validation and where the GraphQLError with a custom message is thrown.

```
mutation RootMutation {
  addUser(
    email: "invalid.email",
  ){email}
}
```

``` js
    name: attrs.name,
    serialize: function(value) {
      return value;
    },
    parseValue: function(value) {
      return value;
    },
    parseLiteral: function(ast) {
      if (!isValid) {
        throw new GraphQLError(`The field ${ast.fieldName} is invalid.`, [ast]); // HERE
      }
      return ast.value;
    }
  });
```

```
{
  "errors": [
    {
      "message": "\"Email\" is invalid.",
      "locations": [
        {
          "line": 5,
          "column": 12,
          "field": "email" // HERE
        }
      ]
    }
  ]
}
```

CatTail:
Possible fix #192

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -340,7 +340,8 @@ function collectFields(
   runtimeType: GraphQLObjectType,
   selectionSet: SelectionSet,
   fields: {[key: string]: Array<Field>},
-  visitedFragmentNames: {[key: string]: boolean}
+  visitedFragmentNames: {[key: string]: boolean},
+  parentField: ?Field
 ): {[key: string]: Array<Field>} {
   for (var i = 0; i < selectionSet.selections.length; i++) {
     var selection = selectionSet.selections[i];
@@ -353,6 +354,7 @@ function collectFields(
         if (!fields[name]) {
           fields[name] = [];
         }
+        selection.parentField = parentField;
         fields[name].push(selection);
         break;
       case Kind.INLINE_FRAGMENT:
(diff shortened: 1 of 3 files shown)


src/execution/execute.js
```diff
@@ -353,6 +354,7 @@ function collectFields(
         if (!fields[name]) {
           fields[name] = [];
         }
+        selection.parentField = parentField;
```
leebyron: We want to ensure that we're not mutating the provided AST, as it may be used elsewhere.

leebyron:
Thanks so much for investigating this!

We should be careful not to mutate inputs but also careful to ensure the performance of query execution is not negatively impacted. One case I imagine would not work correctly here is if a query includes the same fragment in two locations, an error occurring in both locations. The `parentField` property would have been overridden by the second call, and the first error would have an incorrect execution path.

CatTail:
Add unittest for

> One case I imagine would not work correctly here is if a query includes the same fragment in two locations, an error occurring in both locations.

Maybe I am misunderstand?

And the problem of add `parentField` for AST, can you provide a better way to retain parent field information without modification of AST?

CatTail:
ping @leebyron

leebyron:
For example:


```
fragment X on T {
  x
}

query {
  a { ...X }
  b { ...X }
}
```

Where the field `x` throws an error. 

Currently by my understanding of your code, this will result in printing out two errors, where both errors will report `b.x` even though the first should state `a.x` - this occurs because of the mutation of the field nodes, and the fragment being used in more than one location.

PR #276, opened 2016-01-15
Better error for GraphQLList

nodkz:
screen shot 2016-01-15 at 12 15 14

In my Object about 8 list fields, and it is not possible to determine which exactly produces an error.

--- 

Error message now contains field name 
`User Error: expected iterable, but did not find one for field Resume.experience.`


leebyron:
Thanks, this is great

PR #281, opened 2016-02-01
Add printFineSchema() for printing hierarchy ordered types.

iamchenxin:
Add printFineSchema() for printing hierarchy ordered types,the ranks are determined by reference/extends between each type.
Not modify the original printSchema.
Something like :
```
C {  a:A }
B {  c:C }
A {}
```
will be ordered to their hierarchy , more details is in __tests__
```
A {}
C {  a:A }
B {  c:C }
```

### DIFF
--- a/src/utilities/__tests__/schemaPrinter-test.js
+++ b/src/utilities/__tests__/schemaPrinter-test.js
@@ -26,6 +26,7 @@ import {
   GraphQLBoolean,
   GraphQLList,
   GraphQLNonNull,
+  GraphQLID,
 } from '../../';
 
 
@@ -547,8 +548,7 @@ scalar Odd
 type Root {
   odd: Odd
 }
-`
-     );
+`);
   });
 
   it('Enum', () => {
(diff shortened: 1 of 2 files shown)


leebyron:
Thanks for assembling this, this is really awesome.

I'm curious why you think `printSchema` and `printFineSchema` should both exist? With the existence of this new method, when would you explicitly want to use `printSchema` instead?

leebyron:
Also, I'm curious about the primary motivation. Are you looking for `printSchema` to be stable as types evolve? If so, perhaps alphabetical may be better as a schema could add interfaces to types as it evolves which could thrash a file.

Or are you just looking to see these printed in a way that is easier to read in the output?

iamchenxin:
@leebyron I write `printFineSchema ` for  easier to read the output types. Im not sure if the `printSchema` have other usages other than output types for reading.  So i add an additional function to print readable types.  With lots of types, alphabetical will become hard to read.

iamchenxin:
@leebyron hello, is this suitable for merge? Cause im poor at english, if i have some misunderstanding your meaning,please forgive me.
I make the `printSchma` like [this](https://github.com/graphql/graphql-js/pull/281/files#diff-71ba52e9c625f826d2b0df2963c8633aR26), by default it print schema by 'alphabet', or explicitly indicate printing by 'hierarchy'(to get a readable schema).
```
// type printStyle = 'alphabet' | 'hierarchy';
export function printSchema(
  schema: GraphQLSchema,style: string = 'alphabet'): string {
  switch (style) {
    case 'hierarchy':
      return printFineSchema(schema);
    case 'alphabet':
    default:
      return printFilteredSchema(schema, isDefinedType);
  }
}
```

leebyron:
What I'm trying to understand is the purpose - the "why" of this method. Have you added it because it should definitely exist in parallel to the existing `printSchema` method, or simply out of precaution to have this diff be only additive? When would you use `printSchema` and when would you use `printFineSchema`?

iamchenxin:
Maybe it is some personal usage for me(just a few people need this). 
The `printFineSchema` output a  hierarchy ordered type. just for human reading .I should move these code to some package called `graphql-utils`.

PR #282, opened 2016-02-01
[validator] Add suggested types to incorrect field message

dschafer:
Update the error message for `undefinedField` to take in a list of suggested type names.

### DIFF
--- a/src/validation/__tests__/validation.js
+++ b/src/validation/__tests__/validation.js
@@ -65,9 +65,9 @@ describe('Validate: Supports full validation', () => {
     );
 
     expect(errors).to.deep.equal([
-      { message: 'Cannot query field "catOrDog" on "QueryRoot".' },
-      { message: 'Cannot query field "furColor" on "Cat".' },
-      { message: 'Cannot query field "isHousetrained" on "Dog".' }
+      { message: 'Cannot query field "catOrDog" on type "QueryRoot".' },
+      { message: 'Cannot query field "furColor" on type "Cat".' },
+      { message: 'Cannot query field "isHousetrained" on type "Dog".' }
     ]);
   });
 
--- a/src/validation/rules/FieldsOnCorrectType.js
+++ b/src/validation/rules/FieldsOnCorrectType.js
@@ -11,14 +11,31 @@
 import type { ValidationContext } from '../index';
 import { GraphQLError } from '../../error';
 import type { Field } from '../../language/ast';
-import type { GraphQLType } from '../../type/definition';
-
+import {
+  isAbstractType,
+  GraphQLAbstractType,
+  GraphQLObjectType,
+} from '../../type/definition';
 
 export function undefinedFieldMessage(
   fieldName: string,
-  type: GraphQLType
+  type: string,
+  suggestedTypes: Array<string>
 ): string {
-  return `Cannot query field "${fieldName}" on "${type}".`;
+  let message = `Cannot query field "${fieldName}" on type "${type}".`;
+  const MAX_LENGTH = 5;
+  if (suggestedTypes.length !== 0) {
+    let suggestions = suggestedTypes
+      .slice(0, MAX_LENGTH)
+      .map(t => `"${t}"`)
+      .join(', ');
+    if (suggestedTypes.length > MAX_LENGTH) {
+      suggestions += `, and ${suggestedTypes.length - MAX_LENGTH} other types`;
+    }
+    message += ` However, this field exists on ${suggestions}.`;
+    message += ` Perhaps you meant to use an inline fragment?`;
+  }
+  return message;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


dschafer:
Oops I opened this off a super old `master`. Will rebase and update.

dschafer:
Okay, rebased.

src/validation/__tests__/validation.js
```diff
@@ -65,9 +65,9 @@ describe('Validate: Supports full validation', () => {
     );
 
     expect(errors).to.deep.equal([
-      { message: 'Cannot query field "catOrDog" on "QueryRoot".' },
-      { message: 'Cannot query field "furColor" on "Cat".' },
-      { message: 'Cannot query field "isHousetrained" on "Dog".' }
+      { message: 'Cannot query field "catOrDog" on type "QueryRoot".' },
+      { message: 'Cannot query field "furColor" on type "Cat".' },
+      { message: 'Cannot query field "isHousetrained" on type "Dog".' }
```
leebyron: maybe these 3 should actually be calls to `undefinedField` to match the other tests?

src/validation/rules/FieldsOnCorrectType.js
```diff
@@ -11,14 +11,70 @@
...
+} from '../../type/definition';
 
 export function undefinedFieldMessage(
-  fieldName: string,
-  type: GraphQLType
-): string {
-  return `Cannot query field "${fieldName}" on "${type}".`;
+    field: string,
+    type: string,
+    suggestedTypes: Array<string>): string {
```
leebyron: minor style nit: the rest of the codebase drops the return type of >80ch functions to it's own line.

e.g.

```
export function undefinedFieldMessage(
  field: string,
  type: string,
  suggestedTypes: Array<string>
): string {

File: src/validation/rules/FieldsOnCorrectType.js [hunk 3]
```diff
@@ -11,14 +11,70 @@
...
+  GraphQLAbstractType,
+  GraphQLObjectType,
+} from '../../type/definition';
 
 export function undefinedFieldMessage(
-  fieldName: string,
-  type: GraphQLType
-): string {
-  return `Cannot query field "${fieldName}" on "${type}".`;
+    field: string,
```
leebyron: I suggest keeping `fieldName` instead of `field` since it's slightly more clear that you're dealing with the name of something rather than the thing itself (could be misinterpreted as a field AST or field definition). I suggest actually renaming type to `typeName` since the type is now string

src/validation/rules/FieldsOnCorrectType.js
```diff
@@ -11,14 +11,70 @@
...
+      }
+      if (acc[i.name] === undefined) {
+        acc[i.name] = 0;
+      }
+      acc[i.name] += 1;
+    });
+    return acc;
+  }, {});
+  return Object.keys(suggestedInterfaces)
+    .sort((a,b) => suggestedInterfaces[b] - suggestedInterfaces[a]);
```
leebyron: Minor nit, I think it's preferable to put functions which are only used internally below any that are `export`ed, so it's easier to see what functions are exported on first read.

src/validation/rules/FieldsOnCorrectType.js
```diff
@@ -11,14 +11,70 @@
...
+ * Go through all of the implementations of type, and find other interaces
+ * that they implement. If those interfaces include `field` as a valid field,
+ * return them, sorted by how often the implementations include the other
+ * interface.
+ */
+function getSiblingInterfacesIncludingField(
+    type: GraphQLAbstractType,
+    fieldName: string): Array<string> {
+  const implementingObjects = type.getPossibleTypes()
+    .filter(t => t instanceof GraphQLObjectType);
```
leebyron: This bit is the same as lines 46/47 above. Perhaps there's a way to only do this once?

Maybe you could set this up in three steps:

1. Collect possible types that define this field.
2. Collect overlapping interfaces (that types from (1) implement) that also define this field.
3. Combine: (2).concat((1).sort())

src/validation/__tests__/validation.js
dschafer: Probably; if we do so, I definitely need to add a dedicated test for `undefinedField` (I'll do that anyway).

src/validation/rules/FieldsOnCorrectType.js
dschafer: Yeah, I lost this in the rebase, will fix.

src/validation/rules/FieldsOnCorrectType.js
dschafer: Will do.

src/validation/rules/FieldsOnCorrectType.js
dschafer: Will do.

src/validation/__tests__/validation.js
dschafer: Actually, I'll keep this as a true "integration" test.

src/validation/rules/FieldsOnCorrectType.js
```diff
@@ -11,14 +11,70 @@
...
+
+  const suggestedInterfaces = implementingObjects.reduce((acc, t) => {
```
leebyron: if you wanted this to operate on objects instead of strings, you could do:


```js
const implCount = new Map();

implementingObjectTypes.forEach(objType => {
  objType.getInterfaces().forEach(interfaceType => {
    const count = implCount.get(interfaceType);
    if (count) {
      implCount.set(interfaceType, count + 1);
    } else if (interfaceType.getFields()[fieldName]) {
      implCount.set(interfaceType, 0);
    }
  })
})

return [ ...implCount.values() ].sort(
  (a, b) => implCount.get(b) - implCount.get(a)
);

leebyron:
This is awesome! Some minor feedback inline, but otherwise this is a huge upgrade on the error quality for this particular validator.

dschafer:
Updated with @leebyron feedback.

PR #286, opened 2016-02-05
Test schema instead of function

chapel:
The test was run against the function instead of the schema, a simple typo omission.


dschafer:
Thanks!

PR #287, opened 2016-02-05
Test schema instead of function 2

chapel:
Looks like there were plenty of these typos. I was in a rush to catch the train and didn't see them all.


leebyron:
Lol, yikes. Good thing tests still passed after you fixed these. Thanks!

Issue #288, opened 2016-02-08
Printer assumes anonymous operations are queries. 


BobertForever:
After opening graphql/graphiql#90, it was found by @asiandrummer that the printer assumes all unnamed operations are queries and removes the operation. Via this assumption, unnamed mutations get incorrectly printed. For example:
```
mutation {
  doSomething {
    [...]
  } ,
}
```
gets transformed to:
```
{
  doSomething {
    [...]
  } ,
}
```
Which will fail because doSomething is a mutation and not a query.

[This](https://github.com/graphql/graphql-js/blob/master/src/language/printer.js#L34) appears to be the offending line.

leebyron:
Thanks for the report, this has now been fixed in a patch release!

PR #289, opened 2016-02-08
Provides a correct OperationType without name in GraphQLPrinter

asiandrummer:
Addresses #288. Simply inserts another check to see OperationType is `query` and only omit the keyword in that case. For `mutation` (and `subscription`) the query string should still contain OperationType.

### DIFF
--- a/src/language/__tests__/printer.js
+++ b/src/language/__tests__/printer.js
@@ -34,6 +34,44 @@ describe('Printer', () => {
     );
   });
 
+  it('correctly prints non-query operations without name', () => {
+    const queryAstShorthanded = parse(`query { id, name }`);
+    expect(print(queryAstShorthanded)).to.equal(
+`{
+  id
+  name
+}
+`);
+
+    const mutationAst = parse(`mutation { id, name }`);
+    expect(print(mutationAst)).to.equal(
+`mutation {
+  id
+  name
+}
+`);
+
+    const queryAstWithArtifacts = parse(
+`query ($foo: TestType) @testDirective { id, name }`
+);
+    expect(print(queryAstWithArtifacts)).to.equal(
+`query ($foo: TestType) @testDirective {
+  id
+  name
+}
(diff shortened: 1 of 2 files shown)


src/language/__tests__/printer.js
```diff
@@ -34,6 +34,23 @@ describe('Printer', () => {
     );
   });
 
+  it('correctly prints non-query operations without name', () => {
```
asiandrummer: This test could potentially be within kitchen sink test, but I felt like I'd slow down the tests using it. Let me know if you think otherwise.

src/language/printer.js
```diff
@@ -31,7 +31,8 @@ const printDocASTReducer = {
     const defs = wrap('(', join(node.variableDefinitions, ', '), ')');
     const directives = join(node.directives, ' ');
     const selectionSet = node.selectionSet;
-    return !name ? selectionSet :
+    return !name ? op === `query` ?
+      selectionSet : join([ op, selectionSet ], ' ') :
```
leebyron: unnamed queries can have variable defs and directives.

I think you may want to add a test case like:

```
mutation ($var: String) @directive { field }
```

leebyron:
Sweet, thanks so much for digging into this!

src/language/printer.js
```diff
@@ -31,7 +31,8 @@ const printDocASTReducer = {
     const defs = wrap('(', join(node.variableDefinitions, ', '), ')');
     const directives = join(node.directives, ' ');
     const selectionSet = node.selectionSet;
-    return !name ? selectionSet :
+    return !name ? !directives && !defs && (!op || op === `query`) ?
```
asiandrummer: @leebyron thanks for the good catch - especially that `query` with directives/variable defs would ignore the artifacts and only show the selectionSet, treating it as a short-handed query. Added explicit checks for that exact condition, as well as tests for it.

PR #296, opened 2016-02-17
Validate: Unique variable names

jjergus:


dschafer:
Looks good to me; I might change the field name to a real field in the sample schema to make it more clear that the valid queries are truly valid (and that the invalid queries are only invalid because of this validation rule).

jjergus:
Sure. However, the queries are also invalid because the variables are unused. Should we address that as well? Other validation tests don't, as far as I can tell.

leebyron:
You need to also add this to the default list of validation rules in https://github.com/graphql/graphql-js/blob/master/src/validation/specifiedRules.js

leebyron:
Otherwise this looks great. @jjergus do you want to also draft a PR to the spec to include this?

src/validation/__tests__/UniqueVariableNames.js
```diff
@@ -0,0 +1,45 @@
...
+    expectPassesRule(UniqueVariableNames, `
+      query A($x: Int, $y: String) { __typename }
+      query B($x: String, $y: Int) { __typename }
+    `);
+  });
+
+  it('duplicate variable names', () => {
+    expectFailsRule(UniqueVariableNames, `
+      query A($x: Int, $x: Int, $x: String) { __typename }
+      query B($x: String, $x: Int) { __typename }
```
leebyron: perhaps also include a test like `query C($x: Int, $x: Int) { __typename }` just to illustrate that *only* duplicating a variable name is still not ok

jjergus:
> do you want to also draft a PR to the spec to include this?

https://github.com/facebook/graphql/pull/145

src/validation/__tests__/UniqueVariableNames.js
jjergus: Done.

leebyron:
Excellent

PR #298, opened 2016-02-20
Visitor does not support `enter` and `leave` for the root node

dminkovsky:
Opening an issue to discuss this apparent problem.

### DIFF
--- a/src/language/__tests__/visitor.js
+++ b/src/language/__tests__/visitor.js
@@ -20,6 +20,63 @@ import { getNamedType, isCompositeType } from '../../type';
 
 
 describe('Visitor', () => {
+
+  it('allows editing a node both on enter and on leave', () => {
+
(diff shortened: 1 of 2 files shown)


dminkovsky:
So, I added a proposed fix.

It seems like [this check](https://github.com/graphql/graphql-js/blob/75e7be00285a5aa3b1ef3397a4e544148d9b1660/src/language/visitor.js#L256-L258) exists exclusively for cases where the root node is edited, and if that's the case, then this fix might actually be correct. The 0th edit is the edit returned from `enter` and the 1st edit is the edit returned from `leave`. I used `edits.length` instead of `1`, though.

Also, my added tests are up top just for ease of access. Can move them anywhere, or consolidate them into one test.

leebyron:
Thanks so much for detecting and investigating a fix. I'll investigate to add some confidence to the fix, but from first look it appears good.

dminkovsky:
Thanks Lee. This visitor is really useful.

PR #300, opened 2016-02-25
Type checking under GraphQLList is too strict.

mwschall:
`Array.isArray` is being used before running `map`. This precludes, for example, [Backbone](http://backbonejs.org/)-style Collections from being considered valid iterables despite them having the required array-like functionality.

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -695,7 +695,7 @@ function completeValue(
   // If field type is List, complete each item in the list with the inner type
   if (returnType instanceof GraphQLList) {
     invariant(
-      Array.isArray(result),
+      typeof result.map === 'function',
       'User Error: expected iterable, but did not find one ' +
       `for field ${info.parentType}.${info.fieldName}.`
     );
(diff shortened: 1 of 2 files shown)


leebyron:
Thanks for bringing this up! I agree it's too strict, and as you can see there's a slightly aging TODO to fix this, so I'm glad you took a crack at it. However I don't think this is the right approach.

Firstly because there are many array-like iterable collections that do not provide a `map` method, and would continue to be ignored by this.

Also, there are array-like iterable collections for which `map` does not return an Array, but another instance of the same array-like iterable collection type. That could cause these methods to return the wrong type of thing and result in hard to track bugs.

I think the right approach is to use the [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols] protocol. There is a little bit more complication to this than just directly using this protocol because not all JavaScript runtimes support it (modern ones do) and so you still need to handle the Array case directly.

This would also mean that using non-Array collection would need to implement the iterator protocol. Many do today, but as far as I can tell Backbone Collection does not yet. (It wouldn't be that hard to add!)

PR #305, opened 2016-03-09
Add test for parseLiteral on ComplexScalar

tgriesser:
I was looking through to get a sense for how `parseLiteral` was used and realized there wasn't a test for its use here, so I added one.


leebyron:
yay! thank you!

PR #306, opened 2016-03-09
Expose more info to serialize

tgriesser:
Similar to #103, this PR adds the `GraphQLResolveInfo` as a trailing argument to `serialize` on `GraphQLScalarType`.

The use case is that for a situation like Relay where you'd want to have a generated key based on a combination of the `id` property and the object type. You could just use a custom defined Scalar, which would now know which type it's being serialized on.

Test case included to illustrate this example.

### DIFF
--- a/src/execution/__tests__/abstract.js
+++ b/src/execution/__tests__/abstract.js
@@ -13,6 +13,7 @@ import {
   graphql,
   GraphQLSchema,
   GraphQLObjectType,
+  GraphQLScalarType,
   GraphQLInterfaceType,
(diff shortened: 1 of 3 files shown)


JeffRMoore:
The work that serialize is doing in the test case feels more like resolving to me than serializing a resolved value.  My conception of resolve is that it fetches or computes values and serialize translates the value from one type system to another.  In this case, serialize is computing rather than translating.  Could this be done with a custom resolve function?

tgriesser:
Sounds good, thanks to both for the explanations.

PR #310, opened 2016-03-12
Add named type for user defined resolveType and isTypeOf functions

JeffRMoore:
The definition file uses a named type GraphQLFieldResolveFn to describe the user provided field resolving function, but uses inline types to describe the resolveType and isTypeOf functions.

This change brings symmetry by defining a named types for the other two resolving functions: resolveType and isTypeOf.  This adds named types for those, too: GraphQLTypeResolveFn and GraphQLIsTypeOfFn.

This is in the #304 PR, but stands on its own.  In that PR, each function of the three functions is changed to have a different type for the info parameter.  Having symmetrical definitions makes that change easier to follow. 

This was not done for serialize, parseValue and parseLiteral because they are not exposed on their object and the named type would have only been referenced in one place.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -301,7 +301,7 @@ export type GraphQLScalarTypeConfig<InternalType> = {
 export class GraphQLObjectType {
   name: string;
   description: ?string;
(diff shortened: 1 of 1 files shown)


leebyron:
Awesome, thanks!

PR #313, opened 2016-03-14
build client schema scalar type with serialise & parse etc

jackielii:
Is there any reason why we're not doing this?

My use case is to build argument values on the client side.

### DIFF
--- a/src/utilities/buildClientSchema.js
+++ b/src/utilities/buildClientSchema.js
@@ -207,13 +207,13 @@ export function buildClientSchema(
     return new GraphQLScalarType({
       name: scalarIntrospection.name,
       description: scalarIntrospection.description,
-      serialize: () => null,
+      serialize: (v) => v,
       // Note: validation calls the parse functions to determine if a
       // literal value is correct. Returning null would cause use of custom
       // scalars to always fail validation. Returning false causes them to
       // always pass validation.
-      parseValue: () => false,
-      parseLiteral: () => false,
+      parseValue: (v) => v,
+      parseLiteral: (ast) => ast.value,
     });
   }
 


leebyron:
Can you show some example code from the use case you're trying to enable?

jackielii:
@leebyron I'm normalising the result of a graphql query to "EntityMaps" for caching, and using `getArgumentValues` from `graphql/execution/values`. Using different argument to query would result in different fields. 

This a library we're working on that integrates redux and graphql. It's not open source yet until it's proven by a production app.

Here is the source of the normalisation bit:

```javascript
import invariant from 'invariant'
import { getNamedType } from 'graphql'
import { getArgumentValues } from 'graphql/execution/values'
import { buildExecutionContext, collectFields } from './execute'

import getWrappedObjectType from './getWrappedObjectType'
import toDocumentAST from './toDocumentAST'
import defaultGetId from './getIdFieldName'
import formatArguments from './formatArguments'

/*
 * `normalise` turns a data tree into an 3 level entitySet:
 * Map<typeName: Map<key, Map<fieldNameWithArgs, value>>>
*/
export default function normalise(schema, query, data, variables, getIdFieldName=defaultGetId) {
  const ast = toDocumentAST(query)
  const exeContext = buildExecutionContext(schema, ast, null, variables)
  const operation = exeContext.operation

  let objectType
  switch (operation.operation) {
    case 'query':
      objectType = schema.getQueryType()
      break
    case 'mutation':
      objectType = schema.getMutationType()
      break
    case 'subscription':
    default:
      invariant(`unknown operation type: "${operation.operation}", or not supported yet?`)
  }
  const result = {}
  entityFromType(result, exeContext, objectType, [ operation ], data, getIdFieldName)
  return result
}

/*
 * `entityFromType` recursively walk through data tree, and collect typed entities
*/
function entityFromType(result, exeContext, type, fieldASTs, data, getIdFieldName, idFieldName) {
  invariant(getNamedType(type), `type must be a composite type to normalise to an entity from data`)

  if (data === null || data === undefined) {
    return null
  }

  let fields = Object.create(null)
  const visitedFragmentNames = Object.create(null)
  for (let i=0; i<fieldASTs.length; i++) {
    const selectionSet = fieldASTs[i].selectionSet
    fields = collectFields(
      exeContext,
      type,
      selectionSet,
      fields,
      visitedFragmentNames
    )
  }

  const entity = Object.keys(fields).reduce((entity, responseName) => {
    const fieldASTs = fields[responseName]
    const fieldName = fieldASTs[0].name.value
    const fieldDef = type.getFields()[fieldName]
    const fieldType = fieldDef ? fieldDef.type : null

    invariant(fieldType, 'I need a field type to continue...')

    const property = typeof data[responseName] !== 'undefined' ? data[responseName] : null
    const fieldObjectType = getWrappedObjectType(fieldDef.type)

    const args = getArgumentValues(fieldDef.args, fieldASTs[0].arguments, exeContext.variableValues)
    const fieldNameWithArgs = fieldName + formatArguments(args)

    if (fieldObjectType) {
      const fieldObjectIdName = getIdFieldName(fieldObjectType)
      if (Array.isArray(property)) {
        entity[fieldNameWithArgs] = property.map((elem) => {
          const fieldEntity = entityFromType(result, exeContext, fieldObjectType, fieldASTs, elem, getIdFieldName, fieldObjectIdName)
          return fieldEntity ? fieldEntity[fieldObjectIdName] : null
        })
      } else {
        const fieldEntity = entityFromType(result, exeContext, fieldObjectType, fieldASTs, property, getIdFieldName, fieldObjectIdName)
        entity[fieldNameWithArgs] = fieldObjectIdName === fieldObjectType.name
          ? fieldObjectType.name
          : fieldEntity ? fieldEntity[fieldObjectIdName] : null
      }
    } else {
      entity[fieldNameWithArgs] = property
    }
    return entity
  }, Object.create(null))

  // side effect: populate entityMap
  const typeName = type.name
  if (typeof result[typeName] === 'undefined') {
    result[typeName] = Object.create(null)
  }
  const entityMap = result[typeName]
  if (!idFieldName) {
    idFieldName = getIdFieldName(type)
  }
  const idValue = idFieldName === typeName ? typeName : data[idFieldName]
  invariant(idValue, `id field "${idFieldName}" value must be in the selections on type ${typeName} to normalise`)
  entityMap[idValue] = entity

  return entity
}

```

leebyron:
The reason to not do this is that these functions fail to perform their purpose in this proposed form. If they just pass values through then they are not properly serializing or parsing values according to the types they proport. This could lead to accidental mistakes if used as is.

jackielii:
Thanks for the explanation. Although in my use case the data coming down always has the proper type, no coercing needed. Closing this.

PR #318, opened 2016-03-18
[RFC] Directives in schema language

leebyron:
This adds directives to schema language and to the utilities that use it (schema parser, and buildASTSchema). Directives are one of the few missing pieces from representing a full schema in the schema language.

Note: the schema language is still experimental, so there is no corresponding change to the spec yet.

DirectiveDefinition :
  - `directive` `@` Name ArgumentsDefinition? `on` DirectiveLocations

DirectiveLocations :
  - Name
  - DirectiveLocations `|` Name

Example:

```
directive @skip(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT

# Suggested formatting when there are many valid locations
directive @include(if: Boolean!)
  on FIELD
   | FRAGMENT_SPREAD
   | INLINE_FRAGMENT
```

Note: this depends on #317

### DIFF
--- a/src/language/__tests__/schema-kitchen-sink.graphql
+++ b/src/language/__tests__/schema-kitchen-sink.graphql
@@ -36,3 +36,10 @@ input InputType {
 extend type Foo {
   seven(argument: [String]): Type
 }
+
+directive @skip(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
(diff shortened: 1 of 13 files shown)


leebyron:
Merged into master in fdafe32724f2d6dac1ba360f1a440c0e633e75d9

Issue #27, opened 2015-07-06
Promise support in the resolve() function


rricard:
I started to work on this graphql implementation in order to integrate it at least partially in our systems. We would like to stay very close to this implementation as much as we can as it may become the first community-supported implementation.

A feature we would need is the ability to wait Promise results when resolving. It doesn't seem that the current implementation support it.

I would like to discuss with you if you are open to support the addition of Promise responses. In this case, I can handle the Pull-request.

So if I were to implement this, I would do it like this:

1. Write tests with promises (either mix promises and direct calls in the current test schemas or create a completely new test schema with Promises)
2. When resolving the query, wait for the promises when a resolve function returns a Promise. At the end, this doesn't need any API change since the result goes to a final promise when querying anyway.

We can discuss all of that if you wish of course!

freiksenet:
It should be possible to return promises in `resolve`. Can you provide code samples on what doesn't work?

rricard:
Oh great. Didn't got it either from the code or the examples but I'll
definitely take a look. I could write some tests on this. You can close
that,  thanks!

leebyron:
Yep! we support returning Promises or Arrays of Promises.

PR #321, opened 2016-03-22
Reference promises for resolve func

chrisbolin:
In response to https://github.com/graphql/graphql-js/issues/27 and SO questions. I don't believe there is "official" documentation (besides in issues and tests) of Promise support in the resolve function. Totally willing to be wrong :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,9 @@ var schema = new GraphQLSchema({
 ```
 
 This defines a simple schema with one type and one field, that resolves
-to a fixed value. A more complex example is included in the top level
-[tests](src/__tests__) directory.
+to a fixed value. The `resolve` function can return a value, a promise, 
+or an array of promises. A more complex example is included in the top 
+level [tests](src/__tests__) directory.
 
 Then, serve the result of a query against that type schema.
 


leebyron:
Our "official" documentation is pretty poor anyhow, but thanks for helping us move to correct that!

chrisbolin:
@leebyron I could help with graphql-js docs. are there specific topics / issues you'd like to document?

leebyron:
That would be awesome! I'd love any help you're willing to provide.

Here's my rough wish-list for improved documentation:

1. More in-depth prose and examples on using this library. http://graphql.org/docs/getting-started/ is a start, but there's not a lot of help beyond this for setting up anything more than this simple schema.

2. General higher quality version of all the "api reference" docs, filling in missing pieces and better examples and prose for each. It would be really awesome if we could automatically generate good documentation from the codebase itself. We've worked hard to maintain good flow type definitions and doc blocks within the codebase, but a reference docs site would be much easier to navigate.

3. Move the "api reference" documentation for graphql-js out of https://github.com/graphql/graphql.github.io/ (graphql.org) and into this repo so it's easier to maintain in sync with the rest of this codebase. (I'll probably eventually tackle this since it requires some gh-pages & travis-ci setup)

chrisbolin:
Quick philosophical question: where do docs about _`graphql-js`_ the library go vs. _GraphQL_ the spec? It sounds like you'd like to have graphql.org provide both, correct?

leebyron:
The current reality is that they exist as a sub-section of graphql.org - http://graphql.org/docs/api-reference-graphql/

Philosophically, graphql.org really should be about general GraphQL usage not specific to any one host language. However it should still have ample links out to GraphQL implementations including most importantly this one since it's the reference implementation.

I think the appropriate end state is that this repo host a subdir on graphql.org, so `graphql.org/graphql-js` can be 100% focused on graphql-js usage.

PR #326, opened 2016-03-23
[RFC] Add explicit context arg to graphql execution

leebyron:
This *BREAKING* change introduces a new argument to the GraphQL execution API which is presented to resolution functions: `context`.

This solves a long-standing point of confusion about the correct way to represent authentication or a "viewer context" in which a query is executed.

Previously, we suggested that the `rootValue` contain any authentication tokens, however this led to awkward code:

```
resolve: (val, args, { rootValue: { authToken } }) { ... }
```

Which can now be written as:

```
resolve: (val, args, authToken) { ... }
```

The `info` object is still created and provided to resolution functions, and the `rootValue` is still provided within it, however it is now the *fourth* argument rather than the *third*.

### DIFF
--- a/src/__tests__/starWarsQueryTests.js
+++ b/src/__tests__/starWarsQueryTests.js
@@ -181,7 +181,7 @@ describe('Star Wars Query Tests', () => {
           name: 'Luke Skywalker'
         }
       };
-      const result = await graphql(StarWarsSchema, query, null, params);
+      const result = await graphql(StarWarsSchema, query, null, null, params);
       expect(result).to.deep.equal({ data: expected });
     });
 
@@ -201,7 +201,7 @@ describe('Star Wars Query Tests', () => {
           name: 'Han Solo'
         }
       };
-      const result = await graphql(StarWarsSchema, query, null, params);
+      const result = await graphql(StarWarsSchema, query, null, null, params);
       expect(result).to.deep.equal({ data: expected });
     });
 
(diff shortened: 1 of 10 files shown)


freiksenet:
You probably meant

```js
resolve: (val, args, { authToken }) { ... }
```

freiksenet:
I like the change otherwise, it clears the slightly confusing "root value as context" pattern. Also destructuring in resolve became much easier.

JeffRMoore:
I think its a good idea, not for the de-structuring sugar, but to give each value a single responsiblity.

joewood:
Stupid question: - why wasn't the argument added after the AST `info` rather than before to avoid the breaking change?

leebyron:
The breaking change is undesirable, but the final state of the API is much better. The now-fourth info argument will only very rarely be accessed, and that also allows us some opportunities for future performance tuning.

leggsimon:
Apologies if this a stupid question, pretty new to graphQL and am updating our versions. Just to clarify,
This
```js
...
    technology: {
      type: Page,
      resolve: (root, _, { rootValue: { flags, backend = backendReal }}) =>
        backend(flags).capi.page(sources.technology.uuid)
    },
...
```
Now becomes 
```js
...
    technology: {
      type: Page,
      resolve: (root, _, { flags, backend = backendReal }) =>
        backend(flags).capi.page(sources.technology.uuid)
    },
...
```
Is that right?

PR #328, opened 2016-03-24
Move getTypeOf to execute.js and rename to defaultResolveTypeFn

JeffRMoore:
Default behavior for resolving fields is located in execute.js, while default behavior for resolving types is located in definition.js.  These two functions should be located in the same place.

Since type resolving is an execution phase activity, it more closely belongs in execute.js than in definition.js.  In #304, I ended up having to change this algorithm, also indicating that it changes when execution algorithms change and thus should be co-located with execution.

This also results in a cleaner definition.js for cases when execution is not needed.

getObjectType was removed, as field resolve has no such accessor. 

In more aggressive optimization experiments it turns out that knowing when the the default was to be used was valuable information.  The accessor for type resolving obscures this information.

I hope to eventually propose a similar change for serialize, except there the accessor method is doing more work and references "this," making symmetry harder to achieve.  The right approach for serialize eludes me at this time.  There is value in being able to know the underlying serialize function.  For example, `String` gets called a lot, cannot throw an error, cannot return null and cannot return a promise, all checks we must make for user defined serialization functions that are wasted effort for `String`.

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -798,7 +798,13 @@ function completeAbstractValue(
   info: GraphQLResolveInfo,
   result: mixed
 ): mixed {
-  const runtimeType = returnType.getObjectType(result, info);
+  let runtimeType: ?GraphQLObjectType;
+  if (returnType.resolveType) {
+    runtimeType = returnType.resolveType(result, info);
+  } else {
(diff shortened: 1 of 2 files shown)


leebyron:
Excellent suggestion.

PR #329, opened 2016-03-24
Remove unnecessary optionality in type resolve parameters

JeffRMoore:
What does it mean that the GraphQLResolveInfo parameter is optional for GraphQLTypeResolveFn and GraphQLIsTypeOfFn?  Does that mean when I'm implementing an execution engine I don't have to pass it?  Does that mean if I'm implementing a resolve function I might receive null?

info is not optional for GraphQLFieldResolveFn, so lets make the function signatures symmetric by making it required for GraphQLTypeResolveFn and GraphQLIsTypeOfFn.  The value can be ignored, but will clearly always be sent.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -470,12 +470,12 @@ type GraphQLFieldConfigMapThunk = () => GraphQLFieldConfigMap;
 
 export type GraphQLTypeResolveFn = (
   value: mixed,
-  info?: GraphQLResolveInfo
+  info: GraphQLResolveInfo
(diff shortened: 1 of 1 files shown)


leebyron:
Thanks Jeff! The mistake here was attempting to advertise that a type resolution function could *optionally* request that second parameter, but really what flow is typing is how the function can be *called* and in that case you're correct that since it's only ever called internally, that we should ensure we're always providing that second arg. Great catch.

PR #330, opened 2016-03-24
unnecessary call throwing errors

wenzowski:
Somehow it appears that `call` is becoming unbound, resulting in errors being thrown by `defaultResolveFn`.

Why not just call the function directly?

Affects mattkrick/meatier#123 and [trace can be seen on ci](https://circleci.com/gh/wenzowski/meatier/15)


wenzowski:
This doesn't have a `CLA Signed` tag yet, but I completed the form.

leebyron:
Thanks!

leebyron:
Oh wait, actually - I'm going to revert this and fix in another way and add tests. The first argument to `call` is the `this` context, not the first argument, so this change is not equivalent.

PR #332, opened 2016-03-24
Add tests and refine default resolve function.

leebyron:
This is a follow up to #330

### DIFF
--- a/src/execution/__tests__/resolve.js
+++ b/src/execution/__tests__/resolve.js
@@ -1,0 +1,114 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


src/execution/__tests__/resolve.js
```diff
@@ -0,0 +1,114 @@
...
+    });
+  });
+
+  it('default function calls methods', async () => {
+    const schema = testSchema({ type: GraphQLString });
+
+    const source = {
+      _secret: 'secretValue',
+      test() {
+        return this._secret;
```
leebyron: @wenzowski this is the test that would fail with #330.

wenzowski: beautiful. thanks @leebyron!

PR #340, opened 2016-03-31
Add schema sanity checks in buildASTschema

helfer:
Adds checks in buildASTschema to make sure there is only a single query, mutation and subscription operation defined in the schema. This will now throw an error:

```
schema {
  query: Query
  query: Hello
}

type Query {
  sayHello: Hello
}

type Hello {
  greeting: String
}
```
Same story for mutation and subscription.


helfer:
Pushed to this branch by accident, now the diff is completely useless. I'll close this pull-request and open a new one that contains only the changes the original pull request had.

PR #341, opened 2016-03-31
Add sanity checks for schema in buildASTschema

helfer:
Adds checks to make sure the AST contains exactly one query, and at most one mutation and subscription operation.

Here's the original pull-request: https://github.com/graphql/graphql-js/pull/340


leebyron:
Nice work

PR #342, opened 2016-04-01
Improve coercion error messages

JeffRMoore:
I had an error in my resolveType function which took me a long time to track down.  If resolveType returns null, even accidentally, completeAbstractValue just returns null and no error is noted.

It is reasonable for resolveType to return null if a type cannot be determined.  But, is it reasonable for execute to ignore when this happens?

This PR introduces an invariant to trigger an error message when the runtime type cannot be determined, or something weird is returned from resolveType.


leebyron:
Yeah, you're right. Getting `null` from `resolveType` shouldn't be okay.

Issue #303, opened 2016-03-03
Less strict validation of input objects


glan:
I'm wondering if we really need raise an error wen additional field is passed as part of an input object.

see: https://github.com/graphql/graphql-js/blob/master/src/utilities/isValidJSValue.js#L67-L72

Any additional fields could just be ignored or dropped and warnings rather than errors produced.

leebyron:
The primary case we're looking to highlight here is accidental typos, which happen time to time and can be hard to track down, especially if they're used for a variable or input field that's only occasionally used.

If you have the query `query ($foo: String) { field(arg: $foo) }` and the variables `{ "fooo": "abc" }`, we currently flag this as an error, but we could potentially miss this typo if we did not raise errors.

glan:
@leebyron Sorry I think you misunderstood my suggestion. I'll try and explain in more detail. This error is raised when type checking the contents of a var with and input schema types, for example if the following schema is created:

```js
new GQL.GraphQLInputObjectType({
  name: 'CoordsInput',
  fields: {
    lat: { type: new GQL.GraphQLNonNull(GQL.GraphQLFloat) },
    long: { type: new GQL.GraphQLNonNull(GQL.GraphQLFloat) }
  }
});
```

with this query: `query ($coords: CoordsInput) { field(coords: $coords) }`

and the input variable:

```
{
  "coords": {
    "lat": 37.5,
    "long": -122.4,
    "alt": 105.7
  }
}
```

Then that results in an Unknown field error since `alt` is not in the schema. I believe that we shouldn't throw an error here, or at least have some kind of flag which suppresses this error.

The reason I highlight this issue is that I think it's a case the robustness principle:

> Be conservative in what you send, be liberal in what you accept.

https://en.wikipedia.org/wiki/Robustness_principle

Which seems to me what GraphQL embraces.

Cheers

leebyron:
I'm open to making a change like you suggest, but I want to make sure we cover all the ramifications.

The robustness principle is a good one, however the competing issue is avoiding mistakes. 

The example you provide might be a good example of this, a possible misinterpretation by the client that the server supports an "alt" property. The client might write code which assumes that this property is being used, and then express surprise when they find the results seem to disregard it and assume it's a bug from the server when in fact it wasn't supported in the first place.

Another possible mistake that I think is far more likely to occur involves a slight modification to your example:

```js
new GraphQLInputObjectType({
  name: 'CoordsInput',
  fields: {
[...]

PR #343, opened 2016-04-01
[RFC] Add flag to tolerate additional properties on input objects.

glan:
Adding `tolerateAdditionalProps` property to a GraphQLInputObjectType schema will allow that type to tolerate additional properties which are not specified in the schema.

Example:

``` js
new GraphQLInputObjectType({
  name: 'TestFlexInputObject',
  tolerateAdditionalProps: true,
  fields: {
    a: { type: GraphQLString },
    b: { type: new GraphQLList(GraphQLString) },
    c: { type: new GraphQLNonNull(GraphQLString) },
    d: { type: TestComplexScalar },
  }
});
```

Will accept:

```
{a: "foo", b: ["bar"], c: "baz", dog: "dog"}
```

Note that the additional properties will still get filtered out but we will not get an error.

### DIFF
--- a/src/execution/__tests__/variables-test.js
+++ b/src/execution/__tests__/variables-test.js
@@ -56,6 +56,17 @@ const TestInputObject = new GraphQLInputObjectType({
   }
 });
 
+const TestFlexInputObject = new GraphQLInputObjectType({
+  name: 'TestFlexInputObject',
+  tolerateAdditionalProps: true,
+  fields: {
+    a: { type: GraphQLString },
+    b: { type: new GraphQLList(GraphQLString) },
+    c: { type: new GraphQLNonNull(GraphQLString) },
+    d: { type: TestComplexScalar },
+  }
+});
+
 const TestNestedInputObject = new GraphQLInputObjectType({
   name: 'TestNestedInputObject',
   fields: {
@@ -70,6 +81,11 @@ const TestType = new GraphQLObjectType({
     fieldWithObjectInput: {
       type: GraphQLString,
       args: { input: { type: TestInputObject } },
+      resolve: (_, { input }) => input && JSON.stringify(input)
+    },
+    fieldWithFlexObjectInput: {
+      type: GraphQLString,
+      args: { input: { type: TestFlexInputObject } },
       resolve: (_, { input }) => input && JSON.stringify(input)
     },
     fieldWithNullableStringInput: {
@@ -121,6 +137,7 @@ const TestType = new GraphQLObjectType({
   }
 });
 
+
 const schema = new GraphQLSchema({ query: TestType });
 
 describe('Execute: Handles inputs', () => {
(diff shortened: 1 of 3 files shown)


leebyron:
Thanks for testing out this idea. I'm still a bit worried about the pitfalls this produces, but I'm happy to leave this open for more comments. I'm sure @dschafer also would like to take a look.

My opinion is that we should probably pick a behavior and stick with it rather than making it a per-object opt-in or opt-out. If it is customized per object, it should also be exposed via introspection and this behavior would need to be reflected in the specification as well.

glan:
@leebyron 

I was just looking at alternative approach by creating an `Any` object type in my schema. @mlp5ab is this similar to what you guys were doing?

```js
const GraphQLAny = new GQL.GraphQLScalarType({
  name: 'Any',
  description: 'Any object',
  serialize(x) {
    return JSON.stringify(x);
  },
  parseValue(x) {
    return JSON.parse(x);
  },
  parseLiteral(ast) {
    let obj = {}
    ast.fields.forEach(x => obj[x.name.value] = x.value.value);
    return obj;
  }
});
```

erydo:
I'm also worried about this feature, but since I don't see much more discussion toward resolving this PR, here are some thoughts to get it rolling.

If a variation of this feature were to make sense, it may be better applied as an option to schema validation as a whole. I can see an argument for the following use-cases which would desire different schema behaviour:
- Servers gracefully accepting inputs with extra fields, in line with the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle), might want this enabled.
- Client validators (linters, etc.) strictly throwing errors on extra input fields when validating against a schema would want this disabled.

However, there are big arguments against allowing servers to accept unexpected properties:
- Unless the parser scrubs the input object for expected fields before passing to the resolver, there are potential security issues; for example when a developer uses the input object in a spread. And if those fields are scrubbed, then that's probably violating the intention of the client for those fields to be passed through or preserved.
- A client sending extra fields usually indicates a semantic error. What the server does with extra fields is undefined, but if the fields weren't important, they shouldn't have been included. If the fields were deprecated, there are other mechanisms for that. If the input is actually free-form, there are less complicated ways of handling that case too, for example custom scalars or JSON encoded strings.

Just my two cents!

leebyron:
Thanks for the thoughts!

Another case I've been thinking about is evolving input objects over time. It's non breaking to add new (optional) fields to an existing input object, however given the current behavior we could not deprecate and remove input fields. As far as input goes, if the server doesn't care about the input anymore, there's no reason to include the field for the client's benefit.

I think probably a version of this that makes the most sense is too not throw an error for additional fields, but also ignore those additional fields as they're not described in the server schema. 

The downside here is that runtime validation becomes a little weaker, and misspellings may be harder to debug.

glan:
As an alternative approach to having this flag it is possible to craft a schema to not introspect an input, see: https://github.com/graphql/graphql-js/issues/303#issuecomment-220134780 as an example of an Any input type.

It's kind of ugly though and completely lacks any document API.

erydo:
@leebyron - It seems that applying deprecation directives to specific fields would solve input type evolution more intuitively. (I think this is what https://github.com/facebook/graphql/issues/197 is suggesting). To deprecate a field one would add the directive to that field in the schema and make sure it's nullable. Am I missing something?

The benefit is that deprecation is explicitly documented, emits warnings, and isn't quite at the threshold of "unexpected" inputs from the perspective of a developer implementing resolvers. Implicitly accepting undocumented input fields seems to just open up a huge bug surface area for both client and server developers.

lisbakke:
What's the latest on this? I'm looking forward to using this option so that my service isn't as fragile.

wasauce:
I would love to use this with my service. Any update? Is there something I can do to help?

lisbakke:
Ping. Is this issue still alive?

racerxdl:
Ping. This will be merged?

glan:
It's been a while since I proposed this PR. I somewhat satisfied my own use case by defining `Any` input types. I'd comment on @leebyron [RFC] Ignore undefined input object fields #235

leebyron:
I'm going to close this PR since we should avoid changing behavior based on configuration flags in favor of consistent correct behavior. There's an ongoing conversation about this relative to the spec text with good arguments on both sides and no clear outcome yet.

PR #344, opened 2016-04-03
Attempt to fix the context test

zhaojunz:
It looks like the context test is testing the rootValue field if I interpret the test correctly. The fix is simply to pass and access the context field instead of the rootValue field. I might be wrongly interpreting the test since I am new to GraphQL.


leebyron:
This test is perhaps poorly named, but is correct and testing the behavior of a root value.

PR #346, opened 2016-04-06
RFC: Return type overlap validation

leebyron:
Implements https://github.com/facebook/graphql/pull/162

### DIFF
--- a/src/validation/rules/OverlappingFieldsCanBeMerged.js
+++ b/src/validation/rules/OverlappingFieldsCanBeMerged.js
@@ -20,15 +20,18 @@ import { FIELD, INLINE_FRAGMENT, FRAGMENT_SPREAD } from '../../language/kinds';
 import { print } from '../../language/printer';
 import {
   getNamedType,
+  isLeafType,
+  GraphQLList,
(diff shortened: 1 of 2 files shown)


src/validation/rules/OverlappingFieldsCanBeMerged.js
```diff
@@ -153,34 +185,43 @@ export function OverlappingFieldsCanBeMerged(context: ValidationContext): any {
...
+        )
+      ];
     }
   }
 
   return {
     SelectionSet: {
       // Note: we validate on the reverse traversal so deeper conflicts will be
-      // caught first, for clearer error messages.
+      // caught first, for correct calculation of mutual exclusivity and for
```
jjergus: A more intuitive approach might be to use the `parentFieldsAreMutuallyExclusive` flag as part of the memoization key, but this should work too.

leebyron: Yeah. I had the same thought. Maybe I'll add more comments inline near the memoization source to make this more clear. This was already initially here to get the best error messages, and it just happened to simplify memoization.

PR #350, opened 2016-04-09
Error logging for new interface type semantics

clayallsopp:
If you don't implement `schema.types`, the current error message is very unhelpful (`reduce is not a function on undefined`). this should help others in the future


leebyron:
Awesome, thank you

PR #355, opened 2016-04-12
Include possible field, argument, type names when validation fails

yuzhi:
Currently, if one misspelled something or flipped the order of words, GraphQL validation does not give any suggestions. Engineers would have to go through the schema to find the valid field/type/argument or check in GraphiQL. This should give useful suggestions for most of these cases so engineers can perform the proper fix much faster.

### DIFF
--- a/src/jsutils/__tests__/suggestionList-test.js
+++ b/src/jsutils/__tests__/suggestionList-test.js
@@ -23,6 +23,6 @@ describe('suggestionList', () => {
 
   it('Returns options sorted based on simularity', () => {
     expect(suggestionList('abc', [ 'a', 'ab', 'abc' ]))
-      .to.deep.equal([ 'abc', 'ab', 'a' ]);
+      .to.deep.equal([ 'abc', 'ab' ]);
   });
 });
--- a/src/jsutils/__tests__/suggestionList-test.js
+++ b/src/jsutils/__tests__/suggestionList-test.js
@@ -21,7 +21,7 @@ describe('suggestionList', () => {
     expect(suggestionList('input', [])).to.deep.equal([]);
   });
 
-  it('Returns options sorted based on simularity', () => {
+  it('Returns options sorted based on similarity', () => {
     expect(suggestionList('abc', [ 'a', 'ab', 'abc' ]))
       .to.deep.equal([ 'abc', 'ab' ]);
   });
--- a/src/jsutils/suggestionList.js
+++ b/src/jsutils/suggestionList.js
@@ -10,7 +10,7 @@
 
 /**
  * Given an invalid input string and a list of valid options, returns a filtered
- * list of valid options sorted based on their simularity with the input.
+ * list of valid options sorted based on their similarity with the input.
  */
 export function suggestionList(
     input: string,
--- a/src/jsutils/suggestionList.js
+++ b/src/jsutils/suggestionList.js
@@ -9,9 +9,8 @@
  */
 
 /**
- * Given a JavaScript value and a GraphQL type, determine if the value will be
- * accepted for that type. This is primarily useful for validating the
- * runtime values of query variables.
+ * Given an invalid input string and a list of valid options, returns a filtered
+ * list of valid options sorted based on their simularity with the input.
  */
 export function suggestionList(
     input: string,
@@ -20,10 +19,15 @@ export function suggestionList(
   let i;
   const d = {};
   const oLength = options.length;
+  const inputThreshold = input.length / 2;
   for (i = 0; i < oLength; i++) {
-    d[options[i]] = lexicalDistance(input, options[i]);
+    const distance = lexicalDistance(input, options[i]);
+    const threshold = Math.max(inputThreshold, options[i].length / 2, 1);
+    if (distance <= threshold) {
+      d[options[i]] = distance;
+    }
   }
-  const result = options.slice();
+  const result = Object.keys(d);
   return result.sort((a , b) => d[a] - d[b]);
 }
 
--- a/src/utilities/suggestionList.js
+++ b/src/utilities/suggestionList.js
@@ -1,0 +1,78 @@
+/* @flow */
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+/**
+ * Given a JavaScript value and a GraphQL type, determine if the value will be
+ * accepted for that type. This is primarily useful for validating the
+ * runtime values of query variables.
+ */
+export function suggestionList(
+    input: string,
+    options: Array<string>
+  ): Array<string> {
+  let i;
+  const d = {};
+  const oLength = options.length;
+  for (i = 0; i < oLength; i++) {
+    d[options[i]] = lexicalDistance(input, options[i]);
+  }
+  const result = options.slice();
+  return result.sort((a , b) => d[a] - d[b]);
+}
+
+/**
+ * Computes the lexical distance between strings A and B.
+ *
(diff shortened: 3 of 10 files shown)


src/utilities/suggestionList.js
```diff
@@ -0,0 +1,78 @@
+/* @flow */
```
leebyron: this file should go in the jsutils folder - that is where we keep the dependency-free JS utilities that are not specific to graphql, just happen to be used here.

src/utilities/suggestionList.js
```diff
@@ -0,0 +1,78 @@
...
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+/**
+ * Given a JavaScript value and a GraphQL type, determine if the value will be
+ * accepted for that type. This is primarily useful for validating the
+ * runtime values of query variables.
```
leebyron: this description doesn't seem right

src/utilities/suggestionList.js
```diff
@@ -0,0 +1,78 @@
...
+    options: Array<string>
+  ): Array<string> {
+  let i;
+  const d = {};
+  const oLength = options.length;
+  for (i = 0; i < oLength; i++) {
+    d[options[i]] = lexicalDistance(input, options[i]);
+  }
+  const result = options.slice();
+  return result.sort((a , b) => d[a] - d[b]);
```
leebyron: Maybe instead of or in addition to just sorting them all, we should also filter them down to those with a very low edit distance. Some of the suggestions in the texts seem like they're unlikely to be typos.

src/utilities/suggestionList.js
```diff
@@ -0,0 +1,78 @@
...
+ * insertion, deletion, or substitution of a single character, or a swap of two
+ * adjacent characters.
+ *
+ * This distance can be useful for detecting typos in input or sorting
+ *
+ * @param {string} a
+ * @param {string} b
+ * @return {int} distance in number of edits
+ */
+function lexicalDistance(a, b) {
```
leebyron: I wonder if we need a variation of this for the "flipped words" scenario you mentioned. Can you give an example of a common typo we would want to recommend a fix for?

src/validation/rules/FieldsOnCorrectType.js
```diff
@@ -63,8 +82,22 @@ export function FieldsOnCorrectType(context: ValidationContext): any {
               getImplementationsIncludingField(schema, type, node.name.value)
             );
           }
+          let suggestedFields = [];
+          if (type instanceof GraphQLObjectType ||
+              type instanceof GraphQLInterfaceType ||
+              type instanceof GraphQLInputObjectType) {
+            suggestedFields = suggestionList(
+              node.name.value,
+              Object.keys(schema.getType(type.name).getFields())
```
leebyron: `type.getFields()` here since `schema.getType(type.name) === type`

src/validation/rules/KnownArgumentNames.js
```diff
@@ -17,22 +17,43 @@ import {
...
-  return `Unknown argument "${argName}" on field "${fieldName}" of ` +
+  let message = `Unknown argument "${argName}" on field "${fieldName}" of ` +
     `type "${type}".`;
+  if (suggestedArgs.length) {
+    const suggestions = suggestedArgs
+      .map(t => `"${t}"`)
+      .join(', ');
+    message += ` Perhaps you meant ${suggestions}?`;
+  } else {
+    message += ' There is no known argument for this field.';
```
leebyron: This seems redundant with the rest of the message. I think if there are no suggested args, you can just leave it at that and not amend to the message.

leebyron:
This is looking awesome!

src/utilities/suggestionList.js
yuzhi: I've seen this couple of times when someone introduces a pretty long field name or mutation response type. They either flip two of the middle words or missed something like the word "Payload" completely. Because at a glance nothing looked misspelled, they thought it was more of a problem with updating the schema and ended up spending a lot of time trying to debug and rerun that.
I actually started with trying to split up the `input` into words and trying to match up `n-1` word, but I didn't think the extra complexity to get that number is going to be significantly better than just `lexicalDistance`. That's why I decided to keep it simple. We can always add the variation for that case when we have evidence that the `lexicalDistance` isn't good enough.

src/utilities/suggestionList.js
yuzhi: Because of the flipped word case below, what to you think about setting the filter to keep only suggestions that match `min(input.length/2)`?

src/jsutils/__tests__/suggestionList-test.js
```diff
@@ -0,0 +1,28 @@
...
+
+  it('Returns results when input is empty', () => {
+    expect(suggestionList('', [ 'a' ])).to.deep.equal([ 'a' ]);
+  });
+
+  it('Returns empty array when there are no options', () => {
+    expect(suggestionList('input', [])).to.deep.equal([]);
+  });
+
+  it('Returns options sorted based on simularity', () => {
```
wincent: s/simularity/similarity/

src/jsutils/suggestionList.js
```diff
@@ -0,0 +1,82 @@
...
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+/**
+ * Given an invalid input string and a list of valid options, returns a filtered
+ * list of valid options sorted based on their simularity with the input.
```
wincent: simularity -> similarity

src/validation/rules/FieldsOnCorrectType.js
```diff
@@ -34,6 +41,18 @@ export function undefinedFieldMessage(
...
+  if (suggestedFields.length !== 0) {
+    let suggestions = suggestedFields
+      .slice(0, MAX_LENGTH)
+      .map(t => `"${t}"`)
+      .join(', ');
+    if (suggestedFields.length > MAX_LENGTH) {
+      suggestions += `, or ${suggestedFields.length - MAX_LENGTH} other field`;
+    }
+    message +=
+      ` Did you mean to query ${suggestions}?`;
```
wincent: Seems this doesn't need to be split over two lines.

src/jsutils/suggestionList.js
```diff
@@ -0,0 +1,82 @@
...
+  return result.sort((a , b) => d[a] - d[b]);
+}
+
+/**
+ * Computes the lexical distance between strings A and B.
+ *
+ * The "distance" between two strings is given by counting the minimum number
+ * of edits needed to transform string A into string B. An edit can be an
+ * insertion, deletion, or substitution of a single character, or a swap of two
+ * adjacent characters.
```
wincent: Might be worth mentioning "Levenshtein" somewhere in here.

(Also, did you consider using Damerau–Levenshtein, which considers transpositions as well?)

KyleAMathews:
@yuzhi This will be helpful!

src/jsutils/suggestionList.js
yuzhi: This is copied pasta from https://github.com/graphql/codemirror-graphql/blob/master/src/utils/hintList.js#L40

I did not put too much though on which specific algorithm to use. I just found the one that will get us most of the way there. Improvements can always be made later when we understand the pitfalls of this one better. Looking at the wikipedia page for `Damerau–Levenshtein`, is "a transposition of two adjacent characters" different from "a swap of two adjacent characters" ?

wincent: Ah, no! I missed the word "swap" in there. It is exactly transposition. So this probably is   Damerau–Levenshtein already.

yuzhi:
@leebyron thoughts on the update?

leebyron:
Most excellent

PR #363, opened 2016-04-21
Improve validation error message when field names conflict

robzhu:
When I submit a query with overlapping field names like so:

_node {
          echo(val: "test"),
          echo(val: "foo")
        }_

I get the following validation error: 
**"Fields \"echo\" conflict because they have differing arguments."**

This PR modifies the error message to suggest using aliases when this occurs: 
**"Fields \"echo\" conflict because they have differing arguments. Use aliases on the fields to fetch both if this was intentional."**

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "babel-plugin-transform-runtime": "6.6.0",
     "babel-preset-es2015": "6.6.0",
     "chai": "3.5.0",
+    "chai-string": "1.2.0",
     "chai-subset": "1.2.2",
     "coveralls": "2.11.9",
     "eslint": "2.7.0",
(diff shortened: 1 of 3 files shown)


yuzhi:
@robzhu Update unit test? or add one if it doesn't exist yet? 
The unit test for this file should be in:
src/validation/__tests__/OverlappingFieldsCanBeMerged-test.js

leebyron:
Agree with @yuzhi - can you take a look at some of the output messages?

I patched this in to test. For example https://github.com/graphql/graphql-js/blob/master/src/validation/__tests__/OverlappingFieldsCanBeMerged-test.js#L308 now produces:

```
Fields \"field\" conflict because subfields \"deepField\" conflict because subfields \"x\" conflict because a and b are different fields. Use aliases on the fields to fetch both if this was intentional.. Use aliases on the fields to fetch both if this was intentional.. Use aliases on the fields to fetch both if this was intentional.
```

leebyron:
Ace

robzhu:
Thanks for the quick review and merge!

PR #368, opened 2016-04-25
Allow object types without fields.

dylanahsmith:
## Problem

graphql-js has an invariant that prevents the object type from having no fields, which could be useful for the result of mutation fields.  Using an object type allows fields to be added to the result later, so an object type would provide forward compatibility.  I would also like to remove the requirement for relay mutation payloads from having to include a clientMutationId, since I don't see a reason why it needs to be sent over the network.

The GraphQL specification has an [Object type validation](http://facebook.github.io/graphql/#sec-Object-type-validation) section, but doesn't have a restriction that prevents an object type from having no field.

The GraphQL specification does validate that a selection set on an object type must not be empty, but the __typename field can still be selected on an object type without any fields.  That field wouldn't be necessary for the mutation use case, but would be useful if an object type without fields was used in a union.

## Solution

Remove the invariant, since it doesn't seem necessary.

## Question

There is another invariant that prevents an input object type from being empty.  That is also a restriction that doesn't seem to be imposed by the GraphQL spec.  Should that be removed as well?

### DIFF
--- a/src/type/__tests__/validation-test.js
+++ b/src/type/__tests__/validation-test.js
@@ -355,38 +355,12 @@ describe('Type System: Objects must have fields', () => {
     );
   });
 
-  it('rejects an Object type with empty fields', () => {
-    expect(
-      () => schemaWithFieldType(new GraphQLObjectType({
-        name: 'SomeObject',
-        fields: {}
-      }))
-    ).to.throw(
-      'SomeObject fields must be an object with field names as keys or a ' +
-      'function which returns such an object.'
-    );
-  });
-
   it('rejects an Object type with a field function that returns nothing', () => {
     expect(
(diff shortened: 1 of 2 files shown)


leebyron:
Thanks for digging into this, unfortunately I don't think this change is what we should do.

Since GraphQL always requires you to select fields down to scalar values, an Object type without any defined fields cannot be accessed in any way in a query. This could be even more problematic for Input Objects where a required input object argument with no fields could result in a field that is impossible to query without producing an error.

In order to avoid these pitfalls, it's required that at least one field is defined. These error messages have the second-order valuable effect of catching mistakes and typos when defining a schema.

Thanks for pointing out the issue of the spec not making this clear - I can make an edit to the spec to ensure it includes that Objects, Input Objects and Interfaces must define at least one field to be valid.

PR #371, opened 2016-04-28
Remove all 'instanceof GraphQLSchema' checks

helfer:
This change removes the instanceof GraphQLSchema checks in validate and execute. This makes graphql-js play more nicely with schemas generated by other libraries, or even schemas created by other versions of graphql-js as long as they are compatible (a common issue with npm v2).

@leebyron Right now I just removed the checks and didn't replace them with anything, because it turned out that no tests were affected by the change. Let me know if you think we should have some duck typing checks in there, such as checking that getTypeMap, get Type, getQueryType, getDirectives etc. are functions.

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -117,11 +117,6 @@ export function execute(
   operationName?: ?string
 ): Promise<ExecutionResult> {
   invariant(schema, 'Must provide schema');
-  invariant(
-    schema instanceof GraphQLSchema,
-    'Schema must be an instance of GraphQLSchema. Also ensure that there are ' +
-    'not multiple versions of GraphQL installed in your node_modules directory.'
-  );
 
   // If a valid context cannot be created due to incorrect arguments,
   // this will throw an error.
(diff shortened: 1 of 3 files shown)


leebyron:
Yeah, those invariants were probably a little over-protective. I think this is fine.

leebyron:
See the revert - this actually caused problems :(

rdrey:
@helfer I just ran into issues with this when I `npm link`ed graphql-tools. I unlinked it and am running OK again. (I don't have any changes to graphql-tools, just linked it to debug something, so I don't need it linked permanently.) Just thought I'd mention that here.

Would linking the same `graphql` into both my repo and `graphql-tools` be a possible workaround?

helfer:
@rdrey yes, that should do the trick. But to be honest, sometimes I found it to be faster to use `npm publish` under a fake package name than to manually fix all the dependencies that `npm link` doesn't manage to get right.

helfer:
Also, FWIW: I would now solve the problem that this PR was supposed to help with by extending the base classes from graphql-js in my own package, if I needed to.

PR #384, opened 2016-05-07
Deprecated directive

leebyron:
This adds a new directive as part of the experimental schema language:

```
directive @deprecated(reason: String = "No longer supported") on FIELD_DEFINITION | ENUM_VALUE
```

It also adds support for this directive in the schemaPrinter and buildASTSchema.

Additionally exports a new helper `specifiedDirectives` which is encoured to be used when addressing the collection of all directives defined by the spec. The `@deprecated` directive is optimistically added to this collection. While it's currently experimental, it will become part of the schema definition language RFC.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -66,9 +66,14 @@ export {
   GraphQLBoolean,
   GraphQLID,
 
-  // Built-in Directives
+  // Built-in Directives defined by the Spec
+  specifiedDirectives,
   GraphQLIncludeDirective,
   GraphQLSkipDirective,
+  GraphQLDeprecatedDirective,
+
+  // Constant Deprecation Reason
+  DEFAULT_DEPRECATION_REASON,
 
   // Meta-field definitions.
   SchemaMetaFieldDef,
--- a/src/type/directives.js
+++ b/src/type/directives.js
@@ -13,7 +13,7 @@ import type {
   GraphQLFieldConfigArgumentMap,
   GraphQLArgument
 } from './definition';
-import { GraphQLBoolean } from './scalars';
+import { GraphQLString, GraphQLBoolean } from './scalars';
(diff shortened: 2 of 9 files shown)


leebyron:
This is a follow-up to #376 cc @helfer, @OlegIlyenko

leebyron:
Example of use:

```graphql
schema {
  query: Query
}

enum MyEnum {
  VALUE
  OLD_VALUE @deprecated
  OTHER_VALUE @deprecated(reason: "Terrible reasons")
}

type Query {
  field1: String @deprecated
  field2: Int @deprecated(reason: "Because I said so")
  enum: MyEnum
}
```

src/index.js
```diff
@@ -66,9 +66,14 @@ export {
   GraphQLBoolean,
   GraphQLID,
 
-  // Built-in Directives
+  // Specified Directives
+  specifiedDirectives,
```
robzhu: Is this redundant?

src/type/directives.js
```diff
@@ -139,3 +139,40 @@ export const GraphQLSkipDirective = new GraphQLDirective({
...
+export const DEFAULT_DEPRECATION_REASON = 'No longer supported';
+
+/**
+ * Used to declare element of a GraphQL schema as deprecated.
+ */
+export const GraphQLDeprecatedDirective = new GraphQLDirective({
+  name: 'deprecated',
+  description:
+    'Marks an element of a GraphQL schema as no longer supported.',
+  locations: [
```
robzhu: Do we intend to support more directive locations later?

leebyron: Possibly so. Right now these are the only two kinds of things in the schema which can be deprecated according to the spec.

robzhu:
Looks great! This will be a good example for building custom directives.

PR #386, opened 2016-05-10
Validation: improving overlapping fields quality

leebyron:
This improves the overlapping fields validation performance and improves error reporting quality by separating the concepts of checking fields "within" a single collection of fields from checking fields "between" two different collections of fields. This ensures for deeply overlapping fields that nested fields are not checked against each other repeatedly. Extending this concept further, fragment spreads are no longer expanded inline before looking for conflicts, instead the fields within a fragment are compared to the fields with the selection set which contained the referencing fragment spread.

e.g.

```graphql
{
  same: a
  same: b
  ...X
}

fragment X on T {
  same: c
  same: d
}
```

In the above example, the initial query body is checked "within" so `a` is compared to `b`. Also, the fragment `X` is checked "within" so `c` is compared to `d`. Because of the fragment spread, the query body and fragment `X` are checked "between" so that `a` and `b` are each compared to `c` and `d`. In this trivial example, no fewer checks are performed, but in the case where fragments are referenced multiple times, this reduces the overall number of checks (regardless of memoization).

**BREAKING**: This can change the order of fields reported when a conflict arises when fragment spreads are involved. If you are checking the precise output of errors (e.g. for unit tests), you may find existing errors change from `"a" and "c" are different fields` to `"c" and "a" are different fields`.

From a perf point of view, this is fairly minor as the memoization "PairSet" was already keeping these repeated checks from consuming time, however this will reduce the number of memoized hits because of the algorithm improvement.

From an error reporting point of view, this reports nearest-common-ancestor issues when found in a fragment that comes later in the validation process. I've added a test which fails with the existing impl and now passes, as well as changed a comment.

This also fixes an error where validation issues could be missed because of an over-eager memoization. I've also modified the `PairSet` to be aware of both forms of memoization. Also includes a test which fails before this patch.

### DIFF
--- a/src/validation/rules/OverlappingFieldsCanBeMerged.js
+++ b/src/validation/rules/OverlappingFieldsCanBeMerged.js
@@ -15,6 +15,7 @@ import type {
   SelectionSet,
   Field,
   Argument,
+  FragmentDefinition,
 } from '../../language/ast';
 import { FIELD, INLINE_FRAGMENT, FRAGMENT_SPREAD } from '../../language/kinds';
 import { print } from '../../language/printer';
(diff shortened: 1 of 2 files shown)


leebyron:
I've been staring at this for way too long. Someone tell me if I'm off my rocker or if this looks sane.

Also, this validation check is by far the most complicated one in the rule set, so please speak up if any part of the change is unclear or under-commented.

Folks who may want to review: @robzhu, @dschafer, @OlegIlyenko, @finneganh

src/validation/rules/OverlappingFieldsCanBeMerged.js
```diff
@@ -279,54 +554,61 @@ function doTypesConflict(
...
-  type1: ?GraphQLOutputType,
-  ast2: Field,
-  type2: ?GraphQLOutputType
-): ?AstAndDefCollection {
-  const selectionSet1 = ast1.selectionSet;
-  const selectionSet2 = ast2.selectionSet;
-  if (selectionSet1 && selectionSet2) {
-    const visitedFragmentNames = {};
-    let subfieldMap = collectFieldASTsAndDefs(
+  cache,
```
finneganh: Is it worth calling this `fieldAndFragmentNameCache` or somesuch throughout?

src/validation/rules/OverlappingFieldsCanBeMerged.js
```diff
@@ -97,28 +95,279 @@ type AstAndDef = [ GraphQLCompositeType, Field, ?GraphQLFieldDefinition ];
...
+      areMutuallyExclusive,
+      fieldMap2,
+      fragmentNames1[i]
+    );
+  }
+
+  // Also collect conflicts between any fragment names by the first and
+  // fragment names by the second.
+  for (let i = 0; i < fragmentNames1.length; i++) {
+    for (let j = 0; j < fragmentNames2.length; j++) {
```
finneganh: can this be `j = i + 1`?

src/validation/rules/OverlappingFieldsCanBeMerged.js
```diff
@@ -127,46 +376,60 @@ function findConflicts(
...
+  parentFieldsAreMutuallyExclusive: boolean,
+  fieldMap1: AstAndDefCollection,
+  fieldMap2: AstAndDefCollection
+): void {
+  Object.keys(fieldMap1).forEach(responseName => {
+    const fields2 = fieldMap2[responseName];
+    if (fields2) {
+      const fields1 = fieldMap1[responseName];
+      for (let i = 0; i < fields1.length; i++) {
+        for (let j = 0; j < fields2.length; j++) {
```
finneganh: `j = i + 1`?

finneganh: n/m, different arrays it's iterating over. (whoops)

finneganh:
I looked this over for a while. As far as I can tell it seems reasonable? I think I was able to follow the recursion enough that it was doing the right thing to the right values and would terminate.

The only quadratic memory growth now appears to be in fragment comparison, which is probably harder to get excessively large due to the raw bytes of syntax required to make each one.

The `cache` and `comparedSet` args could probably be given more specific names for clarity.

src/validation/rules/OverlappingFieldsCanBeMerged.js
leebyron: Great idea

src/validation/rules/OverlappingFieldsCanBeMerged.js
leebyron: I'll add some comments inline to clarify

leebyron:
Thanks for the review, this is great feedback. Yeah, the comparison memoization being between fragments instead of fields should make it much harder to get problematic quadratic memory growth, but still potentially an issue. I'll look into ways to restrict that.

PR #388, opened 2016-05-13
Update babel-cli and flow-bin package references

robzhu:
We currently get a bunch of graceful-fs deprecated warnings when running "npm install". This change fixes two such warnings by updating babel-cli and flow-bin.


leebyron:
Sweet

leebyron:
Sorry, had to revert this. It caused test runs to fail on travis. I'll investigate.

PR #395, opened 2016-06-01
Fix test assertions for validation test when using custom TypeInfo

sogko:
- Changes in #355, which improved validation messages, should had broken the test.
- But `expect().to.deep.equal()` checks for the right number of errors  but does not check for equality of error messages.

Notes:
- `npm run check` complained that the error messages were too long; broke each error into 2 lines.

PS: Can someone with travis access re-run the build? It failed for only one of the environment `nodejs 4`, seems like `flow` didn't get installed properly (`ENOENT` error)


leebyron:
Thanks for your help and careful reading!

PR #396, opened 2016-06-01
Errors thrown from resolvers have the execution path

Slava:
I tried my best to make changes as least disruptive as possible.

The execution path is different from the `location` on errors, it is the path on the actual data, rather than the query, so you can tell precisely which element's resolver caused the error.

This information is also propagated in the `info` argument to resolvers for an opportunity of better instrumentation for tools like [Apollo](apollostack.com).


helfer:
@leebyron If you have some time, could you take a quick look at this PR some time and give us feedback? This is something that would be really useful for Apollo and other projects for a couple of reasons:

1. It provides better error messages to developers
2. It lets GraphQL clients do smarter things with errors, such as discarding part of the result and not updating the store.
3. It lets UI frameworks display errors in the context where they make the most sense.
4. It's a step towards sending normalized partial responses. Also helpful for streaming.
5. It's also a step towards doing logging and profiling that knows about causal relationships (which resolver had to wait for which other resolver).

stealthybox:
I just wanted to say a huge thank you to @helfer and the meteor team for pushing this functionality in response to my questions on Slack.
The work on this is incredible, and Errors are much clearer now.

I would like to highlight that this PR is not just restricted to error handling.
@Slava mentioned that the `executionPath` is included in `info` for all resolvers.
The ability to leverage the `executionPath` in `resolve` functions allows for communication of that path to the client.
This is a powerful construct that allows is the center of memoization, query-minification, and subscription in projects like Falcor.

I have been playing with builds of this fork for the past few days now -- the functionality appears to work as expected with nested `Type` and `Field` level `resolve` functions when querying with redundant fields and relations.

This PR totally gets my thumbs up.
I'm super excited about it.

JeffRMoore:
I really like the idea of giving path information to the resolve function.  It feels like the right abstraction to me for describing location (vs AST).

However,  it seems like a lot of work and memory allocation, creating a path array for every value in the result.

Worth it?

Slava:
@JeffRMoore I think it is a fair question to ask, I was considering two possible implementations:

- copy the whole path whenever there is a need to extend the list (implemented in this PR)
- have an implementation of a persistent single-linked list (that looks like a tree, essentially)

I decided to go with a simpler implementation that behaves just like a regular JS array to begin with. Since we are talking about this information being useful in a context of instrumentation and profiling, it seems like it makes sense to sacrifice a couple of object allocations per request for better tooling to reveal a much bigger sink of time?

leebyron:
Sorry for the delay - this looks really good. We need this for some of the future proposals as well.

Major feedback is that we shouldn't need to introduce another kind of Error. The path should just be added to `GraphQLError` when necessary.

I'll add a few inline comments as well.

src/type/definition.js
```diff
@@ -483,6 +483,7 @@ export type GraphQLResolveInfo = {
   rootValue: mixed,
   operation: OperationDefinition,
   variableValues: { [variableName: string]: mixed },
+  executionPath: Array<string | number>
```
leebyron: This would be more accurately called `responsePath` since it describes the path the value occurs in the response, rather than the charted path in the executor. Though I think simply `path` is descriptive enough.

src/execution/execute.js
```diff
@@ -780,9 +823,17 @@ function completeListValue(
...
   const itemType = returnType.ofType;
   let containsPromise = false;
-  const completedResults = result.map(item => {
-    const completedItem =
-      completeValueCatchingError(exeContext, itemType, fieldASTs, info, item);
+  const completedResults = result.map((item, index) => {
+    const childExePath = exePath.slice();
+    childExePath.push(index);
+
+    const childInfo = Object.assign({}, info, {
```
leebyron: Is this necessary? This occurs after running `resolve`, and I'm concerned about the memory overhead of copying this object on each child.

src/execution/execute.js
```diff
@@ -594,7 +622,12 @@ function resolveOrError(
   } catch (error) {
     // Sometimes a non-error is thrown, wrap it as an Error for a
     // consistent interface.
-    return error instanceof Error ? error : new Error(error);
+    const wrappedError = error instanceof Error ? error : new Error(error);
+    const pathedError = new PathedError(
```
leebyron: I'm not sure if this is the correct place for this information to be added. It should probably instead be added where GraphQLError is constructed.

src/execution/execute.js
```diff
@@ -565,19 +582,30 @@ function resolveField(
...
+  const completed = completeValueCatchingError(
     exeContext,
     returnType,
     fieldASTs,
     info,
+    exePath,
     result
   );
+
+  if (isThenable(completed)) {
```
leebyron: This block shouldn't be necessary, since the `completeValueCatchingError` already does this work.

src/error/locatedError.js
```diff
@@ -17,7 +18,7 @@ import { GraphQLError } from './GraphQLError';
  * document responsible for the original Error.
  */
 export function locatedError(
```
leebyron: I believe the right change here is to add a third `path` argument to `locatedError`

src/__tests__/starWarsQuery-test.js
```diff
@@ -364,4 +364,107 @@ describe('Star Wars Query Tests', () => {
...
+      expect(result.data).to.deep.equal(expected);
+      expect(result.errors.map(e => e.message)).to.deep.equal(expectedErrors);
+      expect(
+        result.errors.map(e => e.originalError.executionPath)
+      ).to.deep.equal([ [ 'mainHero', 'story' ] ]);
+    });
+
+
+  });
 });
```
leebyron: Another important test to write, which you will need to write a custom schema for, is that the appropriate response path is included when fields are non-nullable:

```graphql
type A {
  nullableA: A
  nonNullA: A!
  throws: String!
}

query {
  nullableA {
    nullableA {
      nonNullA { 
        nonNullA { 
          throws
        }
      }
    }
  }
}
```

Should provide the response:

```js
{
  data: {
    nullableA: {
      nullableA: null
    }
  },
  errors: [
    {
       path: [ 'nullableA', 'nullableA', 'nonNullA', 'nonNullA', 'throws' ]
    }
  ]
}
```

src/execution/execute.js
```diff
@@ -322,18 +332,24 @@ function executeFields(
...
   fields: {[key: string]: Array<Field>}
 ): Object {
   let containsPromise = false;
 
   const finalResults = Object.keys(fields).reduce(
     (results, responseName) => {
       const fieldASTs = fields[responseName];
+
+      const childExePath = exePath.slice();
+      childExePath.push(responseName);
```
leebyron: Just a minor detail that the changes to the body of this function should match `executeFieldsSerially` as much as possible.

Slava: Thanks! I will work on this feedback immediately and will follow up with
fixes. Hopefully before the lunch time.
2016년 6월 10일 (금) 오전 8:32, Lee Byron <notifications@github.com>님이 작성:

src/execution/execute.js
Slava: I think it is necessary, since each child executor has a different `info` object with a different `exePath` property. Since the execution can run concurrently in async, the child executors can't share the same `info` object.

Does it seem fair? Another option that avoids copying objects around is to have a tree-like data structure and require anyone who wants to collect the `path` to walk the tree up to the root. Which complicates the interface and the implementation and only saves some amount of object allocations.

leebyron: Ah yes, I see now why you're correct. Child fields will still rely on this portion of the path to be added. My hope was that this intermediate `childInfo` object should not need to be constructed since it is never provided to a resolver function (instead, it is provided to the `resolveType` and `isTypeOf` functions) - Perhaps only the `childExePath` is necessary but the new `childInfo` is not?

Slava: You are right. I was confused by the chain of the execution. Now I see that the `info` object passed down from here is not actually every passed into resolvers. I updated the PR with an appropriate comment.

Slava:
@leebyron I have updated the PR responding to all your comments. Let me know if there is anything else I can do to improve this PR. If everything is clear, we can squash everything into 1 change.

src/execution/execute.js
```diff
@@ -565,19 +583,23 @@ function resolveField(
...
     variableValues: exeContext.variableValues,
+    path: exePath
   };
 
   // Get the resolve function, regardless of if its result is normal
   // or abrupt (error).
   const result = resolveOrError(resolveFn, source, args, context, info);
 
-  return completeValueCatchingError(
+  const completed = completeValueCatchingError(
```
leebyron: You can continue to return directly.

src/execution/execute.js
```diff
@@ -605,12 +627,14 @@ function completeValueCatchingError(
...
   info: GraphQLResolveInfo,
+  exePath: Array<string | number>,
   result: mixed
 ): mixed {
   // If the field type is non-nullable, then it is resolved without any
   // protection from errors.
   if (returnType instanceof GraphQLNonNull) {
-    return completeValue(exeContext, returnType, fieldASTs, info, result);
+    return completeValue(
+      exeContext, returnType, fieldASTs, info, exePath, result);
```
leebyron: small stylistic thing: typically when a function call doesn't fit in the 80c line, it has each argument presented on its own line for better legibility.

src/execution/execute.js
```diff
@@ -209,13 +209,16 @@ function buildExecutionContext(
     rawVariableValues || {}
   );
 
+  const executionPath = [];
```
leebyron: I don't believe this change is necessary. The value this returns is different from `info`, and an path is already created during initial execution on line 243

leebyron:
This is looking really nice, thanks for your quick follow-ups!

Slava:
updated

helfer:
@leebyron Could turn off the coveralls comments? They're really breaking up the flow. I believe there's a repo setting in coveralls for that. We've done it for all the Apollo repos. You still get the coverage in the summary of checks.

leebyron:
Good idea, @helfer, just did that for this repo, will do it for the others soon.

leebyron:
This is great work, @Slava!

PR #399, opened 2016-06-03
move babel config to the babelrc

lacker:
This moves babel config to the babelrc and unincludes it from the npm module.

For most use cases, this shouldn't make any difference. But, some build tools will look through node_modules and barf if there is a package.json with a babel config in there. In particular, jest has a  react-native-specific hack that does this. In particular, any project using jest, react-native, and the graphql module doesn't work at the moment. E.g. we had to do a similar thing to fix enzyme. https://github.com/airbnb/enzyme/pull/311

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,13 @@
+{
+  "presets": [
+    "es2015"
+  ],
+  "plugins": [
+    "syntax-async-functions",
+    "transform-class-properties",
+    "transform-flow-strip-types",
+    "transform-object-rest-spread",
+    "transform-regenerator",
+    "transform-runtime"
+  ]
+}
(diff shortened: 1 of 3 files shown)


lacker:
(see also fb-internal discussion at https://www.facebook.com/groups/jesting/permalink/1074292982619316/)

cpojer:
@leebyron can you merge this and publish a new release to unblock @lacker?

leebyron:
In Paris so off time zone, but will get this in soon. Sorry for the conflict, surprised this hasn't come up as an issue for anyone else before. 





On Fri, Jun 3, 2016 at 5:32 PM -0700, "Christoph Pojer" <notifications@github.com> wrote:












@leebyron can you merge this and publish a new release to unblock @lacker?

PR #401, opened 2016-06-05
export flow types in build

### LINKED ISSUE #400: Make flow types available by including source file in NPM release.
If the build system is updated to to just copy all the source files with the `.js.flow` extension next to the built files, flow will pick up the type definitions.

Further, this is the best way to provide flow type definitions as there is no need to maintain a separate type definition file and changes in the source code will automatically show for users using the library.

(Will try to do a PR)

nmn:
Updated the build script to export the source files with the `.flow ` extension in the dist folder so that anyone using flow gets the type definitions for free.

Fixes #400

`echo`s the path to all the test files as side-effect for now. I'm sure someone better at bash will be able to fix it.


KyleAMathews:
👍

nmn:
Closing in favor of #412
Thank @nodkz

Issue #109, opened 2015-08-07
Instrument GraphQL and expose data with built-in type


KyleAMathews:
Was just brainstorming how cool it'd be if GraphQL was self-instrumented. Imagine if GraphQL logged every query + resolve timings for each field AND exposed all this data as a built-in type.

On top of this you could build:
- Monitoring (e.g. if query times for this customer critical query goes above 100 ms, send an alert)
- Something analogous to DBs' slow query log
- Build tool similar to GraphiQL but for ops. Connect to your production system and watch a live log of queries. Show histogram of 20 slowest queries over the past week. Zoom in on trouble spots and prioritize speeding those queries.
- Audit logs for users.
- etc.

The default implementation would be in-memory but easily overridable for any serious install so logs could be persisted.

Thoughts? What does Facebook do for instrumenting/logging GraphQL?

This is potentially a _huge_ advantage to using GraphQL. Monitoring/alerting/auditing is freaking hard and this would standardize a big chunk of it.

KyleAMathews:
And the live updating monitor of queries would also be huge to use during development as it'd give a lot of insight to what's actually happening under the hood. Very helpful to those new to GraphQL (read everyone) to help understand how the system works and also very helpful when prototyping new types.

johanatan:
+1 to this.

rmosolgo:
Wow, that's a great idea! Even if it doesn't make it into the spec, seems like servers could implement it and expose it through the `extensions` key (http://facebook.github.io/graphql/#sec-Response-Format)

KyleAMathews:
@rmosolgo as I understand it, there wouldn't need to be any changes to the spec. The metrics gathering / logging would be handled internally and be exposed as built-in types e.g. `queryLogType`

KyleAMathews:
So you could do queries something like:

```
query slowQueries {
  queryLog(durationGreaterThan: "100", since: "2015-08-08T18:58:27+00:00") {
    type
    duration
    fields {
      name
      description
      duration
    }
  }
}
```

leebyron:
We do something really similar to what's proposed here!

We actually do not query the results of query executions using GraphQL itself, but we do have a logging interface throughout GraphQL which allows us to measure access and performance on a per-field basis.

KyleAMathews:
[...]

PR #402, opened 2016-06-09
Logging support

Slava:
This PR builds on top of #396 and provides a new option `logFn` that is called by GraphQL.js with the following events:
- Resolver starts execution
- Resolver finished execution
- Resolver threw an error
- Resolver's subtree starts execution (for a compound type)
- Resolver's subtree finished execution
- Resolver's subtree threw an error

This logging allows external users of the `graphql` module to tap into the execution time-line for profiling or logging/debugging.

Note that this is not intended to be a logging system available from resolver's code. For that, we expect users of the library to pass their logging mechanism through the `context` property.

### DIFF
--- a/src/execution/logging.js
+++ b/src/execution/logging.js
@@ -1,0 +1,32 @@
+/* @flow */
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+export const TAG = {
+  SUBTREE_START: 'SUBTREE_START',
+  SUBTREE_END: 'SUBTREE_END',
+  SUBTREE_ERROR: 'SUBTREE_ERROR',
+  RESOLVER_START: 'RESOLVER_START',
+  RESOLVER_END: 'RESOLVER_END',
+  RESOLVER_ERROR: 'RESOLVER_ERROR',
+};
+
+export type GraphQLLoggingTagType = $Keys<typeof TAG>; // eslint-disable-line
+
+export function passThroughResultAndLog(
+  logFn: any,
+  tag: GraphQLLoggingTagType,
+  payload: any,
+  info: any
+): any {
+  return function (result) {
+    logFn(tag, payload, info);
+    return result;
+  };
+}
--- a/src/graphql.js
+++ b/src/graphql.js
@@ -39,6 +39,10 @@ import type { GraphQLSchema } from './type/schema';
  *    The name of the operation to use if requestString contains multiple
  *    possible operations. Can be omitted if requestString contains only
  *    one operation.
+ * logFn:
+ *    The function that is called with a tag of an event, an event payload
+ *    and other execution information.
+ *    The examples include: errors thrown, before and after the resolver call.
  */
 export function graphql(
   schema: GraphQLSchema,
@@ -46,7 +50,8 @@ export function graphql(
   rootValue?: mixed,
   contextValue?: mixed,
   variableValues?: ?{[key: string]: mixed},
-  operationName?: ?string
+  operationName?: ?string,
+  logFn?: (tag: string, payload: mixed, info: mixed) => void
 ): Promise<GraphQLResult> {
   return new Promise(resolve => {
     const source = new Source(requestString || '', 'GraphQL request');
@@ -62,7 +67,8 @@ export function graphql(
           rootValue,
           contextValue,
           variableValues,
-          operationName
+          operationName,
+          logFn
         )
       );
     }
--- a/src/__tests__/starWarsQuery-test.js
+++ b/src/__tests__/starWarsQuery-test.js
@@ -11,6 +11,7 @@ import { expect } from 'chai';
 import { describe, it } from 'mocha';
 import { StarWarsSchema } from './starWarsSchema.js';
 import { graphql } from '../graphql';
+import { TAG } from '../execution';
 import {
   GraphQLObjectType,
   GraphQLNonNull,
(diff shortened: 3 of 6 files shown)


Slava:
The tests pass locally. Can anyone please help me to understand why eslint complains about Flowtype type hints on Travis CI?

leebyron:
This is pretty sweet, I'm not sure why the flow errors are being introduced.

leebyron:
Ah, it looks like when the flow version was recently upgraded, it started catching new errors which are live in the library. I'll investigate.

leebyron:
It's a bit hard to review this with the stacked PRs (I really wish Github had better support for this), but now that the previous has landed, perhaps you can rebase this?

leebyron:
I also landed a few minor follow-ups after your PR, so hopefully that doesn't cause too much rebase pain. Sorry for that in advance.

Slava:
Yeah, I am working on rebasing it. Will update soon.

Slava:
@leebyron I updated the PR. Similar to the previous time, the eslint errors on Travis CI don't make much sense to me. Tests pass locally.

leebyron:
The eslint errors look relevant actually. In a follow-up I unified the names of variables such that `exePath` is now `path`.

Slava:
@leebyron oh, sorry! Seems like I didn't fetch after your follow ups second time. Should be good now.

KyleAMathews:
Is info the same thing as the context object? Or will it be possible to pass a request or transaction id that gets passed to this logFn?

Slava:
@KyleAMathews no, info is not the same as the context object, it is the same as the `info` object passed to a resolver function.

In the code as it is now, there is no way to do it unless you pass a log function that is aware of some higher-level context.

helfer:
@KyleAMathews There may be several requests processed concurrently at any given time, so I think the best way to do what you ask for is to create a new log function for every request, and generate the ID when you create that function. That's what I've been doing for tracing and timing executions in apollo server.

KyleAMathews:
@helfer ah that sounds very doable. I implemented http://lightstep.com/ recently (which is 💯 ) and manually instumented some really important flows but this will let me instrument everything which is really exciting.

Been hoping for this for a while :-) thanks @Slava! https://github.com/graphql/graphql-js/issues/109

Slava:
The tests are failing on Travis because the `eslint` plugin was borked in recent versions (around 6 days ago). The workaround is to force a different version of the eslint plugin:

```npm install babylon@6.8.0```

**Comment by Joyeric2** [comment]
Dearest one
My name is Joy Eric how are you doing today and how about your health,hope you are fine and good.Any way after going through your profile in this site i became interested in you.I will like to have you as my good friend whom i will like to share my life experience with.Your sex, age, race nor distance does not matter to me rather what matters most in a relationship is the maturity,truth and honest that exist between friends.So i will like you to contact me back through my private mail address(joyeric688@gma!l.com )So that i can tell you more about my self and send you more photos of me.Yours Sincerely,

**Comment by Slava** [comment]
@leebyron rebased the PR!

**Comment by leebyron** [comment]
Thank you, @Slava! I haven't forgotten about this and want to give it proper focus very soon. 

Just for some context as to how I'd like to approach this, I'd like to take a step back and propose a handful of alternative logging APIs and have a discussion around the pros/cons of each.

**Comment by Slava** [comment]
@leebyron thanks for talking another look at this! I will be happy to look at alternative API designs.

**Comment by sibelius** [comment]
I think this should be outside graphql-js core

take a look on this resources: https://github.com/graphql-compose/graphql-perf/issues/2

you can wrap all your resolvers to do logging and performance metrics on them

what do you think @leebyron ?

**Comment by baer** [comment]
@sibelius Instrumenting resolvers can be really helpful but there are certain metrics, like knowing if a deprecated field is still being called by clients, that could be more easily accessed if GraphQL exposed some instrumentation APIs. For example, the Apollo Optics project accomplishes this by walking the schema and [instrumenting it field by field](https://github.com/apollographql/optics-agent-js/blob/master/src/Instrument.js#L240-L272) which would be made much easier if GraphQL just supported instrumentation natively.

**Comment by sibelius** [comment]
@baer what would be the benefit of "natively" approach from the apollo optics approach?

**Comment by baer** [comment]
Instrumenting field-by-field in the way they have to do it is not terribly graceful. It would be an improvement if GraphQl provided a generic way to add higher order functions all resolvers in this case I think that performance, timings and API usage data is so foundational that it may not make sense for this to be implemented in user-land.

File: src/execution/logging.js [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+export const TAG = {
```
baer: What's the purpose of logging `RESOLVER_ERROR` or `SUBTREE_ERROR ` if you've already got the `exeContext.errors` Array?

src/execution/logging.js
```diff
@@ -0,0 +1,32 @@
...
+  SUBTREE_START: 'SUBTREE_START',
```
baer: Isn't a `SUBTREE` just a non-leaf node? I don't understand what the reason to log here is since the `RESOLVER` will log immediately after. Seems redundant but I'm probably missing something.

src/graphql.js
```diff
@@ -39,14 +39,19 @@ import type { GraphQLSchema } from './type/schema';
...
  */
 export function graphql(
   schema: GraphQLSchema,
   requestString: string,
   rootValue?: mixed,
   contextValue?: mixed,
   variableValues?: ?{[key: string]: mixed},
-  operationName?: ?string
+  operationName?: ?string,
+  logFn?: (tag: string, payload: mixed, info: mixed) => void
```
baer: It seems like there are two main cases that this PR will need to account for:

- Timing information
- Inspection of poorly performing resolvers

_**Timings**_
Since the logging of `*_START` and `*_END` are separate, in order to measure resolver timings similar to what Apollo Optics does, this function should probably guarantee either a key like `query:<path>:<fieldName>` or enough information about a resolver to construct one. I see that it accepts path (in some places) already, but that is likely not quite enough, especially if multiple queries are running at once.

_**Inspection**_
My guess is that it will eventually be useful to folks to log all sorts of information about poorly performing resolvers, much more than just the path or key. For example, if an Array resolver has a search/filter argument, it would be helpful to know which filters are running slowly. Rather than a payload and info of type `mixed`, why not just pass the same or similar objects to what a resolver function receives?

chrisregnier:
This looks fantastic!

One thing I'd suggest though is that the terminology seems a little odd. I think 'logFn' is a little specific and makes me think I should be providing a logger. I'd suggest something like 'onSignal' since firing signals seems more appropriate for the purpose here.

ScionOfBytes:
Any updates on this?

prantlf:
An example using the instrumentation could implement an output compatible with https://github.com/apollographql/apollo-tracing.

langpavel:
I'm heavily for logging support :+1: but honestly,
I'm don't understand code changes introduced.
This PR looks like bigger and interesting goal…
Please, explain all in details and then this may be accepted.

vanthome:
This is desperately needed for debugging scenarios where single field resolvers fail not on the root level. Will this make it in?

PR #420, opened 2016-07-02
fix all errors with flow 0.28

iamchenxin:
Fix all errors(update Flow to 0.27), will write unit test later.(flow 0.27 errors motioned in #412 )
And keep `GraphQLType` to a normal type (not update it to a recursive type):
```
export type GraphQLType =
  GraphQLNamedType|
  GraphQLList |
  GraphQLNonNull;
```
cause in flow 0.27 `GraphQLList<GraphQLType>` was still resolved to `GraphQLList<any>`,seems its no difference between `GraphQLList` and  `GraphQLList<GraphQLType>`. I am asking this behaviors in [flow/issues/2024](https://github.com/facebook/flow/issues/2024), will test this tomorrow, the `any` type should be cleaned.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ npm-debug.log
 node_modules
 coverage
 dist
+.vscode
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "eslint-plugin-babel": "3.2.0",
     "eslint-plugin-flow-vars": "^0.4.0",
     "eslint-plugin-flowtype": "2.2.7",
-    "flow-bin": "0.26.0",
+    "flow-bin": "^0.28.0",
     "isparta": "4.0.0",
     "mocha": "2.5.3",
     "sane": "1.3.4"
--- a/src/jsutils/typecast.js
+++ b/src/jsutils/typecast.js
@@ -1,0 +1,42 @@
+/* @flow */
+/**
+ *  Copyright (c) 2016, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+// import invariant from './invariant';
+import {
+  castToNamedType,
+  castToInputType,
+  castToOutputType,
+} from '../type/definition';
+
+/**
+ * These functions are Dynamic checks which will give the checked value
+ * a right Flow Type.
+**/
+
+let dev = {
+  castToNamedType,
+  castToInputType,
+  castToOutputType
+};
+
+if (process.env.NODE_ENV !== 'dev') {
+  dev = {
+    castToNamedType: (v: any): any => v,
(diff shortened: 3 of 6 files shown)


iamchenxin:
Seems its a bug from Nuclide. will watch that type later.

leebyron:
Thanks so much for investigating this! Ideally we introduce as little runtime code as possible when fixing flow errors - I'm not sure this warrants the new utility functions you've added. Lets attempt to solve the type errors just with better types first.

nodkz:
Flow 0.28 landed
https://flowtype.org/blog/2016/07/01/New-Unions-Intersections.html

@iamchenxin may be you make another try with 0.28 and it help find simpler solution.

iamchenxin:
@leebyron Its okay, i will add a `process.env.NODE_ENV` switch for all of those dynamic test. So they just run in development environment. And will have not produce time-load(as light as Flow). If only using Flow to static resolve type,some of those code will need to rewrite code logic.

iamchenxin:
@nodkz let me take a look at 0.28

iamchenxin:
@leebyron @nodkz Seems 0.28 work well with polymorphic,just have 3 errors with 0.28. against 12 errors with 0.27.

iamchenxin:
@leebyron Update this PR with new master.
Maybe should consider not to use functions like `isInputType` . Those functions are pure runtime check(accept values but return a boolean,that will lead to Flow lose Type information.) 
----------------------------------------------------------------------------
Keep the GraphQLType still use any ,should update it to * later.
```
export type GraphQLType =
  GraphQLNamedType |
  GraphQLList<any> |
  GraphQLNonNull<any>;
```
if change 'GraphQLList<any>' to `GraphQLList<*>` will get additional 33 errors. I am not sure which code style are prefer to fix those errors.
For example: [buildASTSchema.js#286](https://github.com/graphql/graphql-js/commit/516dfb9674b4dbde1ee17f715e4c0b6bc24a9f92#diff-2691b670eb7e51187edc1f918c624a70R286) The Type get from `produceType` will finally need a check by code to get a certain type,can not do this by Flow, cause `produceType` produced Types from string value.(It produce Type in runtime.) Although logically, we can infer that if we input something it will return a certain data,but it not something about static.
So some of runtime checks should exist. like its not nessaray to split `buildFieldType` to `buildInputFieldType` and `buildOutputFieldType` in [extendSchema.js#521](https://github.com/graphql/graphql-js/commit/516dfb9674b4dbde1ee17f715e4c0b6bc24a9f92#diff-a2222d77ff884acdf97c5f295babf27cR521). Cause they are using a same code logic indeed, if some runtime data produced ,there should exsit some coding check to check logic errors or something happed when code are ran out of the Ideal environment.( Like unit test or dynamic check). So split or not , a coding check should always need.

leebyron:
Thanks again for looking into this. I think given the current state of things there should be an easier path to gaining flow 0.28 support, I'll investigate

leebyron:
Closing since the upgrade ended up being pretty straight forward after the other improvements to type checking were made elsewhere for v0.27

PR #425, opened 2016-07-08
 `GraphQLObjectTypeConfig<TSource>` to `GraphQLFieldResolveFn<TSource, TResult>`

iamchenxin:
`GraphQLObjectTypeConfig<TSource>` to `GraphQLFieldResolveFn<TSource, TResult>`
Make user can do a whole top-bottom Flow check between their resolvers.
With the Flow typed feature, user can ensure his data structure is correspond to schema . ex: [starWarsData.js L103](https://github.com/graphql/graphql-js/commit/046cbba2732be8bbbef74f988fffd04294b583c2#diff-e6e81fa96fbb4bdccb4e3f0042b5f1a3R103)

More details see the [src/**tests**/starWarsSchema.js
](https://github.com/graphql/graphql-js/commit/046cbba2732be8bbbef74f988fffd04294b583c2#diff-14d97c822b45992f98d875d53ca45626R112)
A simple demonstration:

```
  type DCPost;
  type DComment;
  type DUser;
    name: 'Post',
    fields: () => ({
      ...
      comment: {
        type: new GraphQLList(Comment),
        resolve: (src:DPost):DComment[] => { // Flow check DPost -> DComment[]
          return ...;
        } }
    }),
  });

  const Comment = new GraphQLObjectType({
    name: 'Comment',
    fields: () => ({
      ...
      user: {
        type: User,
        resolve: (src:DComment):DUser => { // DComment -> DUser
          return ...;
        }
      },
    }),
  });

  const User = new GraphQLObjectType({
      ...
  });
```

Note: Because in v0.61 ,in source code use a `any` type to force cast `TSource`. So it is user's responsibility to check the Data Type is not typo. In above code , if `resolve: (src:DPost):DComment[]` is typo to `resolve: (src:DPost):DSomeAnotherType[]`(and user return a wrong data), Flow will still pass. Flow will not check `DSomeAnotherType` is `DComment`,cause of `DSomeAnotherType` -> `any` ->  `DComment`

### DIFF
--- a/src/__tests__/starWarsData.js
+++ b/src/__tests__/starWarsData.js
@@ -1,10 +1,12 @@
+/* @flow */
+/* eslint quote-props: ["error", "as-needed",
+{ "keywords": true, "unnecessary": false }]*/
 /**
  *  Copyright (c) 2015, Facebook, Inc.
  *  All rights reserved.
  *  LICENSE file in the root directory of this source tree. An additional grant
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
-
 /**
  * This defines a basic set of data for our Star Wars Schema.
  *
@@ -12,6 +14,8 @@
  * fetching this data from a backend service rather than from hardcoded
  * JSON objects in a more complex demo.
  */
+
+import invariant from '../jsutils/invariant';
 
 const luke = {
   id: '1000',
@@ -81,9 +85,37 @@ const droidData = {
 };
 
 /**
+ * These are data types corresponding to the schema design.
+ * They represent what the input and output datas should be
+ * in schema's resolving hierarchy.
+ * If the schema is re-designed,these types should be re-designed too.
+ */
+export type Character = {
+  id: string,
+  name?: string,
+  friends?: Array<string>,
+  appearsIn?: Array<number>,
+};
+
+export type Human = {
+  id: string,
+  name?: string,
+  friends?: Array<string>,
+  appearsIn?: Array<number>,
+  homePlanet?: string,
+};
+
+export type Droid = {
+  id: string,
+  name?: string,
+  friends?: Array<string>,
+  appearsIn?: Array<number>,
+  primaryFunction?: string
+};
+/**
  * Helper function to get a character by ID.
  */
-function getCharacter(id) {
+function getCharacter(id): Promise<*> {
   // Returning a promise just to illustrate GraphQL.js's support.
   return Promise.resolve(humanData[id] || droidData[id]);
 }
@@ -91,14 +123,20 @@ function getCharacter(id) {
 /**
  * Allows us to query for a character's friends.
  */
-export function getFriends(character) {
+export function getFriends(character: Character): Array<*> {
+  // as you can see,with Flow
+  // If you choose a schema structure like {friends?: Array<*>}.
+  // You must check it ,or re-design your schema.
+  // Flow will keep you away from hiden bugs.
+  invariant(character.friends !== undefined,
+    'Character.friends should not be undefined');
   return character.friends.map(id => getCharacter(id));
 }
 
 /**
  * Allows us to fetch the undisputed hero of the Star Wars trilogy, R2-D2.
  */
-export function getHero(episode) {
+export function getHero(episode: number): Character {
   if (episode === 5) {
     // Luke is the hero of Episode V.
     return luke;
@@ -110,13 +148,13 @@ export function getHero(episode) {
 /**
  * Allows us to query for the human with the given id.
  */
-export function getHuman(id) {
+export function getHuman(id: string): Human {
   return humanData[id];
 }
 
 /**
  * Allows us to query for the droid with the given id.
  */
-export function getDroid(id) {
+export function getDroid(id: string): Droid {
   return droidData[id];
 }
--- a/src/__tests__/starWarsData.js
+++ b/src/__tests__/starWarsData.js
@@ -1,6 +1,4 @@
 /* @flow */
-/* eslint quote-props: ["error", "as-needed",
-{ "keywords": true, "unnecessary": false }]*/
 /**
  *  Copyright (c) 2015, Facebook, Inc.
  *  All rights reserved.
@@ -115,7 +113,7 @@ export type Droid = {
 /**
  * Helper function to get a character by ID.
  */
-function getCharacter(id): Promise<*> {
+function getCharacter(id) {
   // Returning a promise just to illustrate GraphQL.js's support.
   return Promise.resolve(humanData[id] || droidData[id]);
 }
@@ -123,7 +121,7 @@ function getCharacter(id): Promise<*> {
 /**
  * Allows us to query for a character's friends.
  */
-export function getFriends(character: Character): Array<*> {
+export function getFriends(character: Character): Array<Promise<Character>> {
   // as you can see,with Flow
   // If you choose a schema structure like {friends?: Array<*>}.
   // You must check it ,or re-design your schema.
--- a/src/__tests__/starWarsSchema.js
+++ b/src/__tests__/starWarsSchema.js
@@ -1,3 +1,4 @@
+/* @flow */
 /**
  *  Copyright (c) 2015, Facebook, Inc.
  *  All rights reserved.
@@ -18,6 +19,8 @@ import {
 } from '../type';
 
 import { getFriends, getHero, getHuman, getDroid } from './starWarsData.js';
+
+import type {Character, Human, Droid} from './starWarsData.js';
 
 /**
  * This is designed to be an end-to-end test, demonstrating
@@ -132,7 +135,7 @@ const characterInterface = new GraphQLInterfaceType({
     },
   }),
   resolveType: character => {
-    return getHuman(character.id) ? humanType : droidType;
+    return getHuman((character: any).id) ? humanType : droidType;
   }
 });
 
@@ -164,7 +167,7 @@ const humanType = new GraphQLObjectType({
       type: new GraphQLList(characterInterface),
       description: 'The friends of the human, or an empty list if they ' +
                    'have none.',
-      resolve: human => getFriends(human),
+      resolve: (human: Human): Array<Promise<Character>> => getFriends(human),
     },
     appearsIn: {
       type: new GraphQLList(episodeEnum),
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@ npm-debug.log
 node_modules
 coverage
 dist
+.vscode
```
leebyron: could you remove this to keep this PR single-purpose?

src/__tests__/starWarsData.js
```diff
@@ -1,10 +1,11 @@
-/**
+/** @flow
```
leebyron: please match the style of other flow files

src/__tests__/starWarsData.js
```diff
@@ -1,10 +1,11 @@
...
  *  Copyright (c) 2015, Facebook, Inc.
  *  All rights reserved.
  *  LICENSE file in the root directory of this source tree. An additional grant
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
-
+/* eslint quote-props: ["error", "as-needed",
```
leebyron: This seems unnecessary and distracts from the purpose of the PR, as does changing the quoted props below. Was this necessary?

src/__tests__/starWarsData.js
```diff
@@ -76,29 +79,39 @@ const artoo = {
...
 }
 
 /**
  * Allows us to query for a character's friends.
  */
-export function getFriends(character) {
+export function getFriends(character: CharacterJS): Array<*> {
+// as you can see, with Flow
+// If you choose a schema structure like {friends?: Array<*>}
+// You must check it , or re-design your schema
```
leebyron: nit: please align comments to the other lines in this body. also no space before comma

src/__tests__/starWarsData.js
```diff
@@ -76,29 +79,39 @@ const artoo = {
...
-  2000: threepio,
-  2001: artoo,
+  '2000': threepio,
+  '2001': artoo,
 };
 
+type CharacterJS = {
+  id: string,
+  name?: string,
+  friends?: string[],
```
leebyron: Please use `Array<string>` formatting for consistency

src/__tests__/starWarsData.js
```diff
@@ -76,29 +79,39 @@ const artoo = {
 };
 
 const droidData = {
-  2000: threepio,
-  2001: artoo,
+  '2000': threepio,
+  '2001': artoo,
 };
 
+type CharacterJS = {
```
leebyron: just `Character` would be easier to understand. No need for the JS suffix

src/__tests__/starWarsSchema.js
```diff
@@ -105,25 +110,31 @@ const episodeEnum = new GraphQLEnumType({
  *     secretBackstory: String
  *   }
  */
+type CharacterJS = {
```
leebyron: this was already defined in the previous file - import it here rather than duplicating

src/__tests__/starWarsSchema.js
```diff
@@ -145,29 +156,37 @@ const characterInterface = new GraphQLInterfaceType({
...
+  appearsIn?: number[],
+  homePlanet?: string,
+};
 const humanType = new GraphQLObjectType({
   name: 'Human',
   description: 'A humanoid creature in the Star Wars universe.',
   fields: () => ({
     id: {
-      type: new GraphQLNonNull(GraphQLString),
+      type: nonNull(GraphQLString),
```
leebyron: please leave these as they were - this additional change distracts from the purpose of the PR

src/__tests__/starWarsSchema.js
```diff
@@ -145,29 +156,37 @@ const characterInterface = new GraphQLInterfaceType({
...
       type: GraphQLString,
       description: 'The name of the human.',
     },
     friends: {
-      type: new GraphQLList(characterInterface),
+      type: list(characterInterface),
       description: 'The friends of the human, or an empty list if they ' +
                    'have none.',
-      resolve: human => getFriends(human),
+      resolve: (human: HumanJS): Promise<CharacterJS>[] => getFriends(human),
```
leebyron: Please use Array<> instead of []

src/type/definition.js
```diff
@@ -1024,3 +1024,10 @@ export class GraphQLNonNull<T: GraphQLNullableType> {
     return this.ofType.toString() + '!';
   }
 }
+
+export type {
```
leebyron: Please write export at the definition of the type rather than exporting at the bottom of the file

leebyron:
Thanks for your hard work on this! I've left some comments inline, but please reduce the scope of this to only the changes necessary for this change - otherwise it's difficult to follow what was necessary and what was not

src/__tests__/starWarsData.js
iamchenxin: to fix Flow error: `non-string literal property keys not supported`, like:
```
const droidData = {
  2000: threepio,
  2001: artoo,
};
```

src/type/definition.js
iamchenxin: Seems these comes from `facebook/npm` branch, i remove them.

src/__tests__/starWarsData.js
```diff
@@ -1,10 +1,12 @@
+/* @flow */
+/* eslint quote-props: ["error", "as-needed",
+{ "keywords": true, "unnecessary": false }]*/
```
iamchenxin: to fix Flow error: `non-string literal property keys not supported`, like:
```
const droidData = {
  2000: threepio,
  2001: artoo,
};
```

iamchenxin:
Separate .gitignore to #437 .

src/__tests__/starWarsData.js
leebyron: I think you actually want `{numbers: true}` rather than `keywords` to handle this case.

I just landed a change which applies this across the whole project so that we would not need this special case.

src/__tests__/starWarsSchema.js
```diff
@@ -164,7 +167,7 @@ const humanType = new GraphQLObjectType({
       type: new GraphQLList(characterInterface),
       description: 'The friends of the human, or an empty list if they ' +
                    'have none.',
-      resolve: human => getFriends(human),
+      resolve: (human: Human): Array<Promise<Character>> => getFriends(human),
```
leebyron: are these resolve function type hints necessary? I would expect that Flow would not require them since you already typed `getFriends`

src/type/definition.js
```diff
@@ -478,12 +478,12 @@ export type GraphQLIsTypeOfFn = (
...
 ) => boolean
 
-export type GraphQLFieldResolveFn<TSource> = (
+export type GraphQLFieldResolveFn<TSource, TResult> = (
   source: TSource,
   args: {[argName: string]: mixed},
   context: mixed,
   info: GraphQLResolveInfo
-) => mixed
+) => TResult
```
leebyron: I'm curious if this actually aids in Flow's type checking? What issue would Flow not catch that it does catch after this change? I'd love to see an example so I can understand better.

src/__tests__/starWarsSchema.js
```diff
@@ -19,6 +20,8 @@ import {
 
 import { getFriends, getHero, getHuman, getDroid } from './starWarsData.js';
 
+import type {Character, Human, Droid} from './starWarsData.js';
```
leebyron: stylistic issue: please use spaces inside the `{ }` in import to match the others in the project

src/__tests__/starWarsData.js
```diff
@@ -76,29 +80,63 @@ const artoo = {
...
+  name?: string,
+  friends?: Array<string>,
+  appearsIn?: Array<number>,
+  primaryFunction?: string
+};
+/**
  * Helper function to get a character by ID.
  */
-function getCharacter(id) {
+function getCharacter(id): Promise<*> {
```
leebyron: You don't need to add Flow types to this function since it is not exported. Flow should automatically infer that a Promise is being returned.

src/__tests__/starWarsData.js
```diff
@@ -76,29 +80,63 @@ const artoo = {
...
   // Returning a promise just to illustrate GraphQL.js's support.
   return Promise.resolve(humanData[id] || droidData[id]);
 }
 
 /**
  * Allows us to query for a character's friends.
  */
-export function getFriends(character) {
+export function getFriends(character: Character): Array<*> {
```
leebyron: This should return `Array<Character>` instead of `Array<*>`, yes?

src/__tests__/starWarsData.js
iamchenxin: Its cool, will remove them.

src/__tests__/starWarsSchema.js
iamchenxin: Maybe it is necessary, the input and output datatype should be a part of  schema's design. A explicit declare give the resolver a constraint.(Makes it clear to other people, what data should be in and what data should be out).

src/__tests__/starWarsData.js
iamchenxin: Its a Promise<Character>,will amend this code style.

src/__tests__/starWarsData.js
iamchenxin: Typed all functions is some personal code style of me. Will remove this to fit the project's code style.

src/type/definition.js
iamchenxin: Im not sure what the sentence `What issue would Flow not catch that it does catch after this change?` means. My english is poor.

If it is means What will be happened if in the future `TSource -> any` be fixed. The explanation is:
The resovler's hierarchy is something like this:
```
ROOT -> A
  A -> B
    B -> BF
        BF -> KK
    B -> B2
  A -> C
    C -> E
    C -> G
```
Ideally the `TResult` can not be typed ,because the Lower level's input type is Upper level's output. The best solution is only needed to type the input for every resovler, but type the result is not bad.
Because at now ,in inner code `TSource` is cast to `any`.  So this is a alternative solution,to link the type system from ROOT to leaf.( A -> B -> BF -> KK).  
If in the future, a new PR make the `TSource` type rightly pass to its children. The explicit type for `result`  will be no longer needed. and in the user's code the explicit typed `result` will be a redundancy. But the redundancy will not make any problem.(Because B is B).

iamchenxin: If `What issue would Flow not catch that it does catch after this change?` means what is the usage of `TResult`.
With `TResult` the Type system can check the out type for a resovle, and  ensure it is corresponding to the lower level .There is a example:
[a demonstration for a Relay's Connection constraint](https://github.com/iamchenxin/relayql/blob/master/example/starWarsSchema.js#L168) With the `TResult`, the type system will check the returned data the users output,if the user typo, like 'id' -> 'idd'. flow will give errors.
untitled

PR #428, opened 2016-07-10
[WIP] Add Typescript typings

calebmer:
I was working on typings for `graphql-js` that would help me better write my GraphQL APIs and I was wondering if the GraphQL team would be willing to see this added in `graphql-js` core? If the team isn’t willing to maintain this, close the issue. I though (and I’m sure others) would love to see this merged, however.

I don’t expect this to get merged now, these are just basic types of everything exported from the library so far. The declaration is also written with Typescript 2.0 `strictNullChecks` in mind so as it stands this declaration won’t work in Typescript 1.0

Note also that this implementation defines the type for `GraphQLType` as `GraphQLType<T>` where `T` represents the GraphQL type’s internal type. I believe this could solve a lot of problems I’ve had in the past when writing GraphQL schemas.

Things that need to be done until I would be confident in merging:
- [ ] Remove type definition for `undefined` (this was temporary as most editors don’t have TS 2.0 support yet).
- [ ] Add tests that `TSource` carries through type definitions properly.
- [ ] Add tests to ensure `GraphQLContext` can be extended. Typescript types are open-ended, therefore a global interface should be sufficient for most cases.
- [ ] Manage recursive types better for `GraphQLList` and `GraphQLNonNull`.
- [ ] Add types for `print`, `visit`, `visitInParallel`, `visitWithTypeInfo`, and `Kind` (untyped as far as I can tell in the source code).
- [ ] Add types for validation rules.
- [ ] Add types for introspection query.
- [ ] Add types for `TypeInfo`.
- [ ] Add documentation? (optional)
- [ ] Add tests :+1:

If you’re interested in this addition, let’s talk about what it would take to get this merged. If you aren’t interested, close the PR. I’ll probably use these typings anyway.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,930 @@
+type undefined = void
+
+// TODO: Test that this is open-ended.
+export interface GraphQLContext {}
+
+export function graphql (
+  schema: GraphQLSchema,
+  requestString: string,
+  rootValue?: any,
+  contextValue?: GraphQLContext,
+  variableValues?: { [key: string]: any } | undefined,
+  operationName?: string
+): Promise<GraphQLResult>
+
+type GraphQLResult = {
+  data?: any,
+  errors?: GraphQLError[],
+}
+
+export class GraphQLSchema {
+  constructor (config: GraphQLSchemaConfig)
+  getQueryType (): GraphQLObjectType<any>
+  getMutationType (): GraphQLObjectType<any> | undefined
+  getSubscriptionType (): GraphQLObjectType<any> | undefined
+  getTypeMap (): TypeMap
+  getType (name: string): GraphQLNamedType<any> | undefined
+  getPossibleTypes <T>(abstractType: GraphQLAbstractType<T>): GraphQLObjectType<T>
+  isPossibleType <T>(abstractType: GraphQLAbstractType<T>, possibleType: GraphQLObjectType<T>): boolean
+  getDirectives (): GraphQLDirective[]
+  getDirective (name: string): GraphQLDirective | undefined
+}
+
+type TypeMap = {
+  [typeName: string]: GraphQLNamedType<any>,
+}
+
+type GraphQLSchemaConfig = {
+  query: GraphQLObjectType<any>,
+  mutation?: GraphQLObjectType<any>,
+  subscription?: GraphQLObjectType<any>,
+  types?: GraphQLNamedType<any>[],
+  directives?: GraphQLDirective[],
+}
+
+// TODO: GraphQLList and GraphQLNonNull should circularly reference
+// GraphQLType. See the flow definitions.
+type GraphQLType<T> =
+  GraphQLScalarType<T> |
+  GraphQLObjectType<T> |
+  GraphQLInterfaceType<T> |
+  GraphQLUnionType<T> |
+  GraphQLEnumType<T> |
+  GraphQLInputObjectType<T> |
+  GraphQLList<T, any> |
+  GraphQLNonNull<T, any>
+
+export function isType (type: any): type is GraphQLType<any>
+
+// TODO: GraphQLList and GraphQLNonNull should circularly reference
+// GraphQLInputType. See the flow definitions.
+type GraphQLInputType<T> =
+  GraphQLScalarType<T> |
+  GraphQLEnumType<T> |
+  GraphQLInputObjectType<T> |
+  GraphQLList<T, any> |
+  GraphQLNonNull<T, any>
+
+export function isInputType <T>(type: GraphQLType<T>): type is GraphQLInputType<T>
+
+// TODO: GraphQLList and GraphQLNonNull should circularly reference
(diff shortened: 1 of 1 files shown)


lacker:
It seems hard to maintain, given that the library itself isn't written in typescript. Basically, if there are errors over time in the typescript types it seems like it would be better if issues were filed against some repo that had people that know things about typescript. I notice that there are types for typescript for express-graphql in DefinitelyTyped - https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/express-graphql - maybe that would be a better place for this?

calebmer:
I think types are similar to documentation. The closer they live to the actual code the more in-sync they will be :blush:. I’m fine either way, but I wanted to give in-repo a shot first :+1:

helfer:
These typings seem to be reasonably complete: https://github.com/nitintutlani/typed-graphql

calebmer:
@helfer so it’s missing one major thing that I’d like to see in `graphql-js` types, and that’s typings for the raw value.

So if you look [here in your linked types](https://github.com/nitintutlani/typed-graphql/blob/master/graphql.d.ts#L615-L623) there is no generic parameter, but in the official types I’d like to see a [generic parameter, say `T`](https://github.com/calebmer/graphql-js/blob/adaa0d87522bb55f0d9755710cd89843c878d7b2/index.d.ts#L47-L55) that represents the raw value. This leads to great intelligence [for resolver functions](https://github.com/calebmer/graphql-js/blob/adaa0d87522bb55f0d9755710cd89843c878d7b2/index.d.ts#L202-L207) because we get types for both the source object and the return value that are validated by our type checker.

It looks like there was some expirementation of this for the inline GraphQL flow types if you look [here](https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L734). I’m not sure of the history though or the future, it was just nice to see I wasn’t the first to have this idea :blush:

A couple other things to note about those types though, it’s an [ambient module](https://github.com/nitintutlani/typed-graphql/blob/master/graphql.d.ts#L2) and some methods [aren’t type guards when they could be](https://github.com/nitintutlani/typed-graphql/blob/master/graphql.d.ts#L625) (see [my](https://github.com/calebmer/graphql-js/blob/adaa0d87522bb55f0d9755710cd89843c878d7b2/index.d.ts#L68) version).

The types in this PR are pretty much complete and will function well for most purposes (even giving you the extra intelligence provided by the generic type). I’ve been using them since I opened this issue in my own projects. The todo list I wrote above is the last 10%  to get these types from good to excellent.

stubailo:
> A couple other things to note about those types though, it’s an ambient module and some methods aren’t type guards when they could be (see my version).

This could be solved by contributing to that package, no?

helfer:
I agree with @lacker and @stubailo. Let's keep the typings separate and all contribute to the same repo.

@calebmer sounds like your typings are more complete and maybe more useful, but it seems they haven't really been tested. For example, here's a [typo](https://github.com/calebmer/graphql-js/blob/adaa0d87522bb55f0d9755710cd89843c878d7b2/index.d.ts#L891) I found. 

Let's just pick some repo for typings that we can all agree on, and then contribute to that. Maybe we can start a new one that we're all contributors to, or we convince @nitintutlani to give us commit access to his repo so we can help keep it up to date.

calebmer:
Ok, sounds good. Once we have something official, I’ll close this issue :+1:

nitintutlani:
I have no issues allowing commit access. Ambient wrapper was added by @jperl last week. We can remove it.

jperl:
Sorry Typescript newbie here. What are the drawbacks of it being an ambient module?

calebmer:
So I haven't heard any hard arguments or official recommendations, but here are my conclusions on the topic after experiencing TypeScript.

1. Global scope is bad and so ambient modules are bad. Ambient modules are inherently global scope as after you define one it's detected everywhere (I believe they are also open ended though).
2. ES modules are JavaScript's future. There's a large push in the JavaScript community to make everyone use ES modules for a number of good reasons. Ambient modules while they may proxy an ES module interface, their declaration is different.
3. Tooling seems to discriminate against ambient modules. Tools like `typings` split split `dependencies` up in two, `depemdencies` and `ambientDependencies` (or `globalDependencies` depending on your version). While not hard evidence, this to me appears to be the TypeScript community "picking a side" by making ES module style typings the happy path in their config file.

Further reading:

- https://github.com/typings/typings/blob/master/docs/external-modules.md
- https://github.com/typings/typings/blob/master/docs/examples.md

helfer:
@nitintutlani That's great!

So how shall we do this? I think it would be very useful to start if @calebmer could make a PR on @nitintutlani 's repo with the improved typings. If there are too many changes and a complete rewrite would be better, then I think we should open a discussion thread first, so we can all agree on which version to use as a base. Does that work for everyone?

nitintutlani:
Please update me if anyone wish to collaborate directly into the repo.

PR #429, opened 2016-07-11
Use only ASCII characters

kassens:
In some environments it's preferable to only use ASCII characters. This changes
all non-ASCII characters to ASCII alternatives.


dschafer:
Looks good, thanks!

leebyron:
Thanks for this!

PR #436, opened 2016-07-18
Allow UnionTypes to accept a types thunk

ekosz:
This should help with module loading order creating null types.

As this is my first PR for the project, please let me know if my tests/style are inline for `graphql-js`. More than happy to fix things up. Thanks!

### DIFF
--- a/src/type/__tests__/definition-test.js
+++ b/src/type/__tests__/definition-test.js
@@ -342,11 +342,21 @@ describe('Type System: Example', () => {
     ];
     badUnionTypes.forEach(x => {
       expect(() =>
-        new GraphQLUnionType({ name: 'BadUnion', types: [ x ] })
+        new GraphQLUnionType({ name: 'BadUnion', types: [ x ] }).getTypes()
       ).to.throw(
         `BadUnion may only contain Object types, it cannot contain: ${x}.`
       );
     });
+  });
+
+  it('allows a thunk for Union\'s types', () => {
+    const types = [ ObjectType ];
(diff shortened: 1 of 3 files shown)


ekosz:
@leebyron Do you know what the plan is for other places in `graphql-js` that don't take a thunk?

src/type/__tests__/definition-test.js
```diff
@@ -342,13 +342,23 @@ describe('Type System: Example', () => {
...
         `BadUnion may only contain Object types, it cannot contain: ${x}.`
       );
     });
   });
 
+  it('allows a thunk for Union\'s types', () => {
+    const types = [ ObjectType ];
+    const union = new GraphQLUnionType({
+      name: 'ThunkUnion',
+      types: () => types
```
leebyron: Isn't the whole point of using a thunk to enable a delay of referencing `ObjectType` until access time?

Shouldn't this be `types: () => [ ObjectType ]` to illustrate that usage?

leebyron:
Thanks - there's no reason this couldn't have also accepted a thunk

src/type/__tests__/definition-test.js
leebyron: Ah, I see - you're using the same reference below for equality. Makes sense!

leebyron:
In general the idea is that thunks should be acceptable wherever breaking a circular dependency between Types might require them.

Unions are not very often in that loop since the Object type itself is typically where the cycle is found and broken, but certainly it makes sense to include them here as well.

PR #437, opened 2016-07-19
ignore vscode's config files

iamchenxin:


leebyron:
Sure why not.

Though just FYI, you can globally ignore files for your user account. You may wish to add `.vscode` to that global ignore list so you do not need to include this in every project's .gitignore

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

Issue #391, opened 2016-05-19
Returning too large of a number from Int resolver results in hard to debug errors


stubailo:
If I have a field that is supposed to return `GraphQLInt`, then there are two hard-to-debug things that can happen if I return a number that's not within the allowed range, for example `1463692053071` (a timestamp):

1. If the field is optional, then I simply get `null` on the client with no errors at all
2. If the field is required, then I get the error: `Cannot return null for non-nullable field Entry.createdAt.` This is extra misleading since I didn't actually return `null` - I returned a Number that was simply too large.

Sounds like this can just be fixed by adding an error in the coerceInt function instead of silently returning `null`.

uhoh-itsmaciek:
I ran into this same problem and [commented on another related issue](https://github.com/graphql/graphql-js/issues/292#issuecomment-223720563). As I point out there, I think that behavior is in violation of [the spec](https://facebook.github.io/graphql/#sec-Int):

>GraphQL servers should coerce non‐int raw values to Int when possible **otherwise they must raise a field error**. Examples of this may include returning 1 for the floating‐point number 1.0, or 2 for the string "2".

(emphasis added)

Would you accept a patch fixing this behavior?

leebyron:
Hey, I really apologize for letting this issue sit. You're absolutely correct @uhoh-itsmaciek that this is a real bug relative to the correct spec language.

I've attached a PR to fix it.

uhoh-itsmaciek:
Awesome, thanks @leebyron!

PR #444, opened 2016-07-20
[FIX] Throw field errors when failing to coerce Int/Float.

leebyron:
As illustrated by #391, when a value is provided for a field of type Int which cannot be represented by Int (e.g. a 64bit value is provided to the 32bit Int type), then the spec claims a field-error should be raised however currently `null` is returned directly instead.

Spec: https://facebook.github.io/graphql/#sec-Int

This updates to throw meaningful error messages when invalid Int and Float values cannot be coerced without losing information.


stubailo:
Thanks!

PR #445, opened 2016-07-20
[FIX] Include locations for errors related to failed value completion.

leebyron:
Error locations were previously only being added due to async value resolution and not during value completion. This meant if an error was raised due to an incorrect response from a resolver function that it would not have gotten wrapped in a location-aware GraphQLError object.

This includes a new intermediate phase which catches and assigns these locations before further propogating the error.

This also assigns non-enumable property for `originalError` so it is not included during JSON.stringify.

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -17,10 +17,10 @@ export class GraphQLError extends Error {
   message: string;
   stack: string;
   nodes: ?Array<Node>;
-  source: Source;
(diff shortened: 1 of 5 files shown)


leebyron:
cc @stubailo - curious if you've come across this issue yet. It seems related to #391

stubailo:
Didn't come across it except for that int thing, when there were no errors reported at all. I think I have been using some special error handling wrappers from @helfer though.

Issue #450, opened 2016-07-21
Extending a schema with directives


josephsavona:
The `extendSchema` utility supports extending types with new fields, but it isn't clear how to extend a schema with a new directive. Given a `GraphQLSchema` instance, and a directive DSL such as `directive @foo(id: ID!) on FIELD`, is there a way to get a version of the schema with this directive added?

robzhu:
It looks like extendSchema copies over the directives from the input schema: https://github.com/graphql/graphql-js/blob/f06051e592362767e6531c1ceb653f1ef3873500/src/utilities/extendSchema.js#L223-L224

Can you share some details about your use case so that I might be able to suggest alternatives?

josephsavona:
@robzhu It only copies the directives from the input schema and ignores any directives defined in the [extension](https://github.com/graphql/graphql-js/blob/f06051e592362767e6531c1ceb653f1ef3873500/src/utilities/extendSchema.js#L99). It looks like [this switch statement](https://github.com/graphql/graphql-js/blob/f06051e592362767e6531c1ceb653f1ef3873500/src/utilities/extendSchema.js#L117) needs to be modified to account for directives (?).

josephsavona:
The use case is that Relay defines its own directives (`@relay`) that it strips out when processing fields. We'd like to use `extendSchema` to be able to augment the application's schema with the definitions of our custom directives.

asiandrummer:
Closing with #455 committed.

PR #455, opened 2016-07-27
Update extendSchema to support new Directives

robzhu:
https://github.com/graphql/graphql-js/issues/450

This changes the extendSchema utility to accept and process new
Directive definitions. Existing directives cannot be extended or
replaced (should throw an error in such cases).

See extendSchema-test for examples.

### DIFF
--- a/src/utilities/extendSchema.js
+++ b/src/utilities/extendSchema.js
@@ -29,6 +29,10 @@ import {
 } from '../type/definition';
 
 import {
+  GraphQLDirective,
+} from '../type/directives';
+
+import {
   __Schema,
   __Directive,
   __DirectiveLocation,
@@ -58,6 +62,7 @@ import {
   SCALAR_TYPE_DEFINITION,
   INPUT_OBJECT_TYPE_DEFINITION,
   TYPE_EXTENSION_DEFINITION,
+  DIRECTIVE_DEFINITION,
 } from '../language/kinds';
 
 import type {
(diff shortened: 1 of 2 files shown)


src/utilities/extendSchema.js
```diff
@@ -220,13 +247,26 @@ export function extendSchema(
...
   });
 
   // Below are functions used for producing this schema that have closed over
   // this scope and have access to the schema, cache, and newly defined types.
 
+  function getMergedDirectives(): Array<GraphQLDirective> {
+    const existingDirectives = schema.getDirectives();
+    invariant(existingDirectives, 'schema must have default directives');
+
+    const mergedDirectives = existingDirectives.slice(0);
```
leebyron: I think `concat` would be more clear than slice.

How about:

```js
const newDirectives = Object.keys(directiveDefinitionMap)
  .map(directiveName => getDirective(directiveDefinitionMap[dirName]));

return existingDirectives.concat(newDirectives);
```

leebyron:
This is awesome work! Only the one minor suggestion

leebyron:
Actually - one higher level point of feedback. Perhaps it would slightly simplify things to collect the directive ASTs into an Array rather than a named Object map. I assume you're following the existing patterns in that file, but the others have a need of looking up type references by name that directives don't have, and it would probably simplify some of your code to just map over an array rather than having to Object.keys first.

leebyron:
Awesome

josephsavona:
Thanks!!

PR #456, opened 2016-07-29
separateOperations - a utility function for splitting an AST

leebyron:
A typical task using GraphQL at Facebook looks something like:

1. Load and parse all .graphql files which may contain operations or fragments.
2. Use `concatAST` to produce one AST that contains all operations and fragments.
3. Separate this all-encompasing AST into individual ASTs that represent each operation which could be sent to the server in isolation.

`separateOperations` fulfills this third step.

### DIFF
--- a/src/utilities/__tests__/separateOperations-test.js
+++ b/src/utilities/__tests__/separateOperations-test.js
@@ -1,0 +1,124 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+import { separateOperations } from '../separateOperations';
+import { parse, print } from '../../language';
+
+
+describe('separateOperations', () => {
+
+  it('separates one AST into multiple, maintaining document order', () => {
+
+    const ast = parse(`
+      {
+        ...Y
+        ...X
+      }
+
+      query One {
+        foo
+        bar
+        ...A
+        ...X
+      }
+
+      fragment A on T {
+        field
+        ...B
+      }
+
+      fragment X on T {
+        fieldX
+      }
+
+      query Two {
+        ...A
+        ...Y
+        baz
+      }
+
+      fragment Y on T {
+        fieldY
+      }
+
+      fragment B on T {
+        something
+      }
+    `);
+
+    const separatedASTs = separateOperations(ast);
+
+    expect(Object.keys(separatedASTs)).to.deep.equal([ '', 'One', 'Two' ]);
+
+    expect(print(separatedASTs[''])).to.equal(
+`{
+  ...Y
+  ...X
+}
+
+fragment X on T {
+  fieldX
+}
+
+fragment Y on T {
+  fieldY
+}
+`
+    );
+
+    expect(print(separatedASTs.One)).to.equal(
+`query One {
+  foo
+  bar
+  ...A
+  ...X
+}
+
+fragment A on T {
+  field
+  ...B
+}
+
+fragment X on T {
+  fieldX
+}
+
+fragment B on T {
+  something
+}
+`
+    );
+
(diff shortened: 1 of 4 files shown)


src/utilities/__tests__/separateOperations-test.js
```diff
@@ -0,0 +1,124 @@
...
+}
+
+fragment B on T {
+  something
+}
+`
+    );
+
+    expect(print(separatedASTs.Two)).to.equal(
+`fragment A on T {
```
robzhu: Why does the fragment appear before the query in this case?

src/utilities/__tests__/separateOperations-test.js
```diff
@@ -0,0 +1,124 @@
...
+
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+import { separateOperations } from '../separateOperations';
+import { parse, print } from '../../language';
+
+
+describe('separateOperations', () => {
+
+  it('separates one AST into multiple, maintaining document order', () => {
```
robzhu: Should we have an error test case for a circular dependency?

src/utilities/separateOperations.js
```diff
@@ -0,0 +1,82 @@
...
+      const toName = node.name.value;
+      (depGraph[fromName] ||
+        (depGraph[fromName] = Object.create(null)))[toName] = true;
+    }
+  });
+
+  // For each operation, produce a new synthesized AST which includes only what
+  // is necessary for completing that operation.
+  const separatedASTs = Object.create(null);
+  for (let i = 0; i < operations.length; i++) {
```
robzhu: Why not use forEach() here?

src/utilities/separateOperations.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Provides the empty string for anonymous operations.
+function opName(operation): string {
+  return operation.name ? operation.name.value : '';
+}
+
+// From a dependency graph, collects a list of transitive dependencies by
+// recursing through a dependency graph.
+function collectTransitiveDependencies(collected, depGraph, fromName) {
```
robzhu: Can we add flow type annotations to these parameters? (and for opName above)

stubailo:
Where do you usually keep your graphql files? In JavaScript, people seem to like putting the queries together with the UI code; in your apps do you just have these query files scattered throughout the code base? are they next to the UI features they power? or is there some central directory of queries?

leebyron:
@stubailo - in our native codebases we encourage colocation of queries to the code they relate to, typically this means if you have something like `FBNewsFeedHeaderUI.m` that you would have `FBNewsFeedHeader.graphql` in the same directory.

In the case that queries appear in code, which happens in a few cases in our codebases, you just need one additional step to extract the GraphQL document from the host document before continuing. This just replaces `fs.readFile()` - but once you get to graphql-js, it really cares nothing about actual concrete files on disk.

src/utilities/__tests__/separateOperations-test.js
leebyron: Good idea. Circular dependency is a validation error, but we certainly don't want our tools to break when operating on invalid queries either.

src/utilities/separateOperations.js
leebyron: Habit to avoid creating a function closure when not necessary. Modern js vms probably unroll foreach into for loops anyhow.

stubailo:
Ah, I see now - so you basically treat all GraphQL code in the whole app as one huge `Document`, where chunks are in various places throughout.

leebyron:
Exactly. A "Document" is an abstraction - it only represents the top level of an AST and doesn't have to map directly to a file source. Parsing each file gives us a Document, then we use concatAST to produce the one-big-Document that ends up being the input to all our other code-gen steps and operations.

leebyron:
For some more flavor on how we use this particular step. 

We persist operations at build time so we can look them up later by a cache id, reducing upload bytes and also making some guarantees about what any particular client is capable of (also helps logging).

This `separateOperations` takes in the one-big-Document and returns a collection of Documents, each of which is ready to be persisted in isolation.

stubailo:
Sounds great - I would love to use a similar approach in Apollo Client. We've been thinking about suggesting that people use unique operation names anyway, which would fit really nicely with this idea.

PR #460, opened 2016-08-09
Relaxing `getNamedType` method for accepting ConnectionType

nodkz:
Relay's ConnectionType should have `ofType` property, like `GraphQLNonNull` and `GraphQLList`. ConnectoionType is some kind of wrapper under NamedType.

So for any other wrappers, I offer to relax `getNamedType` for accepting any object with `ofType` property.

Cross PR with Relay: https://github.com/graphql/graphql-relay-js/pull/98

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -179,10 +179,7 @@ export type GraphQLNamedType =
 
 export function getNamedType(type: ?GraphQLType): ?GraphQLNamedType {
   let unmodifiedType = type;
-  while (
-    unmodifiedType instanceof GraphQLList ||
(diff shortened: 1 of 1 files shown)


leebyron:
Unfortunately this change breaks Flow analysis

leebyron:
But also, I don't believe the comparison between List/NonNull and Connection/Edge is appropriate. List/Nonnull are modifier types whereas the types created by Connection/Edge are true named types. This change along with your proposed change to graphql-relay-js could actually result in broken tools that inadvertently misinterpret Connections and Edges as wrapping types and unwrap them, resulting in false query invalidations.

PR #463, opened 2016-08-14
Refactors tokenizer to emit a Token double-linked-list

leebyron:
This refactors the tokenizer, simplifying both tokenizer and parser code slightly (fewer utility functions) and returns tokens as part of the AST's `loc` information. 

Location info is now created as an Object with non-writable defined properties, only `start` and `end` are enumerable. That removes the need for the `noSource` option, which was really only added to not choke JSON.stringify/util.inspect on the full source. Location info now also includes the `startToken` and `endToken` associated with every node in the AST, providing easy access to the token stream from any point.

Tokens also have gained a few new properties: `line` for the 1-indexed line they appear on and `prev`/`next` to navigate the dll of Tokens.

Finally, this introduces three new Token types: `<SOF>` for the empty token at beginning of the file stream, `,` for ignored comma tokens, and `Comment` for ignored comment tokens.

These tokens are *included* in the Token list allowing for easier navigation of the parsed document in GraphQL tools and as a basis for parsing comment descriptions in the schema language.

### DIFF
--- a/src/language/parser.js
+++ b/src/language/parser.js
@@ -11,9 +11,16 @@
 import { Source } from './source';
 import { syntaxError } from '../error';
 import type { GraphQLError } from '../error';
-import { lex, TokenKind, getTokenKindDesc, getTokenDesc } from './lexer';
-import type { Token } from './lexer';
+import {
+  createLexer,
+  TokenKind,
+  getTokenDesc
+} from './lexer';
+import type { Lexer } from './lexer';
 import type {
+  Location,
+  Token,
+
   Name,
   Variable,
 
@@ -121,27 +128,20 @@ export type ParseOptions = {
    * in the source that they correspond to. This configuration flag
    * disables that behavior for performance or testing.
    */
-  noLocation?: boolean,
-
-  /**
-   * By default, the parser creates AST nodes that contain a reference
-   * to the source that they were created from. This configuration flag
-   * disables that behavior for performance or testing.
-   */
-  noSource?: boolean,
-}
+  noLocation?: boolean
+};
 
 /**
  * Given a GraphQL source, parses it into a Document.
  * Throws GraphQLError if a syntax error is encountered.
  */
 export function parse(
-  source: Source | string,
+  source: string | Source,
   options?: ParseOptions
 ): Document {
-  const sourceObj = source instanceof Source ? source : new Source(source);
-  const parser = makeParser(sourceObj, options || {});
-  return parseDocument(parser);
+  const sourceObj = typeof source === 'string' ? new Source(source) : source;
+  const lexer = createLexer(sourceObj, options || {});
+  return parseDocument(lexer);
 }
 
 /**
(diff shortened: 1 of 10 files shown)


src/language/parser.js
```diff
@@ -152,23 +152,26 @@ export function parse(
...
   options?: ParseOptions
 ): Value {
-  const sourceObj = source instanceof Source ? source : new Source(source);
-  const parser = makeParser(sourceObj, options || {});
-  return parseValueLiteral(parser, false);
+  const sourceObj = typeof source === 'string' ? new Source(source) : source;
+  const lexer = createLexer(sourceObj, options || {});
+  expect(lexer, TokenKind.SOF);
+  const value = parseValueLiteral(lexer, false);
+  expect(lexer, TokenKind.EOF);
```
leebyron: This was an existing bug! Previously `parseValue('123garbage')` would return the value `123`, now it rightfully creates a parse error.

leebyron:
Open to feedback on this. cc @robzhu @dschafer @alangenfeld

leebyron:
See also #464

leebyron:
One change I plan on making is to include a `column` field in Token, which is very cheap for the lexer to calculate now that it's tracking line, and will allow replacing the existing expensive utility that converts character offset to line/column.

leebyron:
column now included in tokens (will replace utilities in follow-up PRs)

stubailo:
Does this mean that now by default stringifying the AST won't include `source`? That sounds pretty great.

leebyron:
That's right @stubailo - the option to omit source from AST location nodes has been removed, but instead all but `start` and `end` have been made non-enumerable so they don't appear in JSON.stringify/util.inspect - not having source appear in stringified AST was really the only reason that parse flag was available.

leebyron:
Actually, update on that. I was profiling this change (parser needs to be fast) and discovered that `Object.defineProperties` is horribly slow, like 100x the cpu time of a simple object creation. Since the parser's runtime is partially bottlenecked by object creation, this translated to a ~20x slowdown of the parser.

I've got a new approach posted up now that instead defines a prototype with `toJSON` and `inspect` defined to essentially achieve the same goal. The only downside is mocha/chai's deepEqual comparator doesn't allow for a hook like this, so the tests had to be edited slightly to compensate.

stubailo:
> parser needs to be fast

Thanks for considering this! The overwhelming of our users are currently using the parser on the client so this perf is definitely very important.

By the way, how did you profile this change? Do you have any ideas about how to automate that kind of stuff?

leebyron:
I used benchmark and microtime npm modules and ran the parser on the kitchen sink file. On a modern MBP I saw ~32,000 parses per second typical, and ~1,700 per second using the non-enumerable properties.

leebyron:
That level of benchmarking is usually only worthwhile in looking for larger swings in performance since the runner isn't super accurate in my experience and isn't very fast to run, so it can be a challenge to set up in an automated way.

stubailo:
Yeah, could be a good sanity check still, by running the baseline and new version on the same machine and making sure there isn't like a 2x regression. Maybe I'll try that soon.

PR #466, opened 2016-08-15
[test] [minor] Make test name unique

jaymitchell:
There is already a `it('multiple variables not used'` earlier in the test for a plain graph. Clarifying that this test is about a graph with fragments.


facebook-github-bot-1:

leebyron:
Thanks!

PR #468, opened 2016-08-22
Make the default resolver use args and context

lacker:
Right now the default resolver is really handy in some cases. If you have a field that uses args and/or context, though, you can't use a default resolver because the default resolver just throws them away, even if there's a function hanging off the source object. There's no real benefit to that behavior, so this PR passes args and context along to the function hanging off the source object if there is one. This makes it easier to use es6 classes in conjunction with graphql.js.

This is backward compatible unless you happened to use a default resolver method which relied on not receiving any arguments despite the schema allowing arguments, which seems perverse enough that we can ignore that case.

### DIFF
--- a/src/execution/__tests__/resolve-test.js
+++ b/src/execution/__tests__/resolve-test.js
@@ -66,6 +66,34 @@ describe('Execute: resolve function', () => {
     });
   });
 
+  it('default function passes args and context', async () => {
+    const schema = testSchema({
+      type: GraphQLInt,
+      args: {
+        addend1: { type: GraphQLInt },
+      },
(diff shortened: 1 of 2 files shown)


leebyron:
Yeah, agreed this is more flexible behavior. This default behavior started conservatively, but I like this behavior better.

> This is backward compatible unless you happened to use a default resolver method which relied on not receiving any arguments despite the schema allowing arguments, which seems perverse enough that we can ignore that case.

I bet this is more common than we anticipate. It shouldn't matter though - the next cut of graphql-js will be a major version.

PR #469, opened 2016-08-24
[RFC] Allows use of generated schemas for execution.

leebyron:
Right now our generated schemas (built from introspection, built from AST) explicitly do not allow their use in execution. They throw as soon as any resolver function is called. This conservative behavior was appropriate for early versions where we wanted to limit the surface area under use. Today it's becoming more clear that there are immediately valuable use cases for using generated schema in execution.

The ability to execute will still be somewhat limited: generated schema do not contain enough information to resolve Interface and Union types, and coerce values for custom Scalars.

Case 1: Resolvers live in your domain models, AST schema is useful:

```js
const schema = buildASTSchema(parse(`type MyTypeSystem { ... }`));
class MyObject { ... }
class MyQueryRoot { ... }
const response = graphql(schema, query, new MyQueryRoot())
```

This is pretty compelling for simpler schema. You can define your schema entirely in AST and allow resolvers to completely exist outside the GraphQL type domain. This approach is now immediately useful thanks to @lacker's #468.

Case 2: Querying your local cache:

```js
const schema = buildClientSchema(introspection);
const response = graphql(schema, query, myGraphQLCache);
```

I admit that this use case is more experimental and less thought through, but I believe that contributors & community may take this idea further. Conceptually it's compelling: given a query I want to submit and my local cache of results, what partial results can I use right away before the server returns? I'm absolutely sure there are missing pieces to truly enable this, but this should be a first step.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -704,7 +704,7 @@ export type GraphQLUnionTypeConfig = {
    * the default implementation will call `isTypeOf` on each implementing
    * Object type.
    */
-  resolveType?: GraphQLTypeResolveFn;
+  resolveType?: ?GraphQLTypeResolveFn;
   description?: ?string;
 };
 
--- a/src/utilities/__tests__/buildASTSchema-test.js
+++ b/src/utilities/__tests__/buildASTSchema-test.js
@@ -13,10 +13,11 @@ import { parse } from '../../language';
 import { printSchema } from '../schemaPrinter';
 import { buildASTSchema } from '../buildASTSchema';
 import {
+  graphql,
   GraphQLSkipDirective,
   GraphQLIncludeDirective,
   GraphQLDeprecatedDirective,
-} from '../../type/directives';
+} from '../../';
 
 /**
  * This function does a full cycle of going from a
@@ -32,6 +33,23 @@ function cycleOutput(body) {
 }
 
 describe('Schema Builder', () => {
+
+  it('can use built schema for limited execution', async () => {
+    const schema = buildASTSchema(parse(`
+      schema { query: Query }
+      type Query {
+        str: String
+      }
+    `));
+
+    const result = await graphql(
+      schema,
+      '{ str }',
+      { str: 123 }
+    );
+    expect(result.data).to.deep.equal({ str: '123' });
+  });
+
   it('Simple type', () => {
     const body = `
 schema {
--- a/src/utilities/__tests__/buildClientSchema-test.js
+++ b/src/utilities/__tests__/buildClientSchema-test.js
@@ -644,7 +644,7 @@ describe('Type System: build schema from introspection', () => {
     await testSchema(schema);
   });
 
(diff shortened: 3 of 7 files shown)


leebyron:
cc @dschafer @stubailo @josephsavona

lacker:
Case 1 is really nice here - it lets people set up a graphql server without learning the API for the GraphQLSchema, GraphQLObjectType and so on. So +1 for making this change.

Just a style thing but what do you think about a helper function that does buildASTSchema(parse(x)) ? `buildSchema` ? If this is going to be in the hello world of express-graphql (which I think it should be) it's one notch nicer to have people import one function rather than two.

leebyron:
I like that idea!

josephsavona:
This is great. Case 1 seems like a really solid use-case to make prototyping and iteration on servers easier. 

ship it!

josephsavona:
> querying your local cache

I agree that this direction is compelling - we're  providing a way for select fields to be resolved at runtime frofor Relay2

stubailo:
Love case 1, should be useful.

As for case 2:

> I'm absolutely sure there are missing pieces to truly enable this, but this should be a first step.

There are two disadvantages using this in Apollo Client or a similar tool, for me:

1. This requires the schema, whereas it's actually possible to execute a graphql query against static data without a schema, as we have demonstrated in the AC implementation.
2. This package is not optimized for having a small footprint, because most of the modules have a lot of internal dependencies, in addition, a lot of the utilities assume schema knowledge because in some cases it's necessary to disambiguate certain edge cases.

I'd be happy to collaborate with anyone to unblock this use case - there might be some small changes that can be made to make it possible for anyone to build a nice client-side cache using this package, which I think would be a big win for the community overall, and reduce the amount of custom work we need to do in Apollo Client.

stubailo:
@josephsavona why close this PR?

josephsavona:
Fat fingers on mobile phone, my bad

leebyron:
@stubailo - yeah, I'm with you there. I think the first case is compelling and the second is only just interesting.

There's definitely some fruit for future exploration in what you're talking about, but of course if you're executing queries locally without a schema, then this is kind of moot since it's about the behavior of the schema itself :)

stubailo:
Sigh......... I got nerd sniped. Feast your eyes: https://github.com/stubailo/graphql-anywhere

Next up, refactoring Apollo Client to use this package to read from the cache!

Gregoor:
Heya, I'm trying to use the generated schema with Relay, but I'm getting the "Generated Schema cannot use Interface or Union types for execution." error. From what I can tell, for nodeDefinitions to work, we need interfaces.
If someone can highlight what's holding Interface support back, I'd get started on working it in.

stubailo:
Hey, give this package a try: https://github.com/apollostack/graphql-tools

It's like `buildSchema` but with support for resolvers, interfaces, etc.

PR #470, opened 2016-08-24
Advantage common convention for naming root types.

leebyron:
This is a result of a conversation I had with @lacker that we should provide a stronger opinion on what to name the root types, and then advantage that opinion. Here's the result of that:

If you use the type names `Query` and `Mutation` for your root types, then printing the schema or parsing a schema IDL can omit the `schema { query: Query, mutation: Mutation }` boilerplate. If your root type names are anything other than these, then you have to include it.

There doesn't seem to be a significant downside to this except for the relative complication of understanding why sometimes a `schema {}` description is included and sometimes it isn't.

**Note: It's okay to make these changes now since the IDL is still technically experimental (It's not in the spec) but we want to nail down the spec'd behavior of the IDL pretty soon, and this would be behavior the spec would need to describe.**

### DIFF
--- a/src/utilities/__tests__/buildASTSchema-test.js
+++ b/src/utilities/__tests__/buildASTSchema-test.js
@@ -457,10 +457,6 @@ type Subscription {
 
   it('Unreferenced type implementing referenced interface', () => {
     const body = `
-schema {
-  query: Query
-}
-
 type Concrete implements Iface {
   key: String
 }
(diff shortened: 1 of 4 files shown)


leebyron:
cc @dschafer who may also have opinions on this.

dschafer:
> There doesn't seem to be a significant downsite to this except for the relative complication of understanding why sometimes a schema {} description is included and sometimes it isn't.

Yeah, and that conceptually matches the omission of the `query` keyword, so seems reasonable to me.

stubailo:
Super into this!

PR #471, opened 2016-08-25
buildSchema helper function

lacker:
Pretty simple - this just adds a `buildSchema` function that composes `buildASTSchema` and `parse`, as mentioned in the discussion on https://github.com/graphql/graphql-js/pull/469 . This makes a "hello world" experience one notch simpler; product developers can use this mechanism to build their schemas while ignoring the intermediate ast format altogether.

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -13,6 +13,8 @@ import invariant from '../jsutils/invariant';
 import keyValMap from '../jsutils/keyValMap';
 import { valueFromAST } from './valueFromAST';
 import { TokenKind } from '../language/lexer';
+import { parse } from '../language/parser';
+import { Source } from '../language/source';
 import { getArgumentValues } from '../execution/values';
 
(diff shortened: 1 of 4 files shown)


src/utilities/buildASTSchema.js
```diff
@@ -13,6 +13,8 @@ import invariant from '../jsutils/invariant';
 import keyValMap from '../jsutils/keyValMap';
 import { valueFromAST } from './valueFromAST';
 import { TokenKind } from '../language/lexer';
+import { parse } from '../language/parser';
+import { Source } from '../language/source';
```
leebyron: `import type { Source }` since this is only being used for flow types

leebyron:
Sweet! Super simple

lacker:
Fixed your feedback.

leebyron:
Boom

nishp1:
🎉 🎉  This is huge. Thanks @lacker @leebyron.

Issue #476, opened 2016-09-02
empty string coerced to 0


itajaja:
empty string it's a valid GraphQLInt since `Number('')` returns 0 ([code](https://github.com/graphql/graphql-js/blob/master/src/type/scalars.js#L23-L24)). I believe this is a bug

robzhu:
Thanks for filing this!

PR #477, opened 2016-09-05
Throw when parsing empty string to numeric scalar

robzhu:
Addresses https://github.com/graphql/graphql-js/issues/476


leebyron:
Nice fix

PR #479, opened 2016-09-07
Update to Flow 0.32

iamchenxin:
Update to Flow 0.32, and update some type definitions with 0.32.(Basically most of those errors in 0.29 are caused by missing `@flow` in ` src/language/kinds.js `).


leebyron:
Awesome thanks

PR #482, opened 2016-09-08
Carry context up to parseValue scalar resolver

rricard:
The goal of this PR is to be able to get the resolution context into the Scalar's resolver (useful for instance to create "FileUpload" scalars). It's not quite done yet but if I can get an early review to check if I'm going in an overall right direction with this and know if this has any chance to get merged, that would actually be super cool 😉 !

**TODO**:
- [x] `parseValue` gets the context injected
- [ ] `parseLiteral` gets the context injected
- [ ] TESTS!

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -216,7 +216,8 @@ function buildExecutionContext(
   const variableValues = getVariableValues(
     schema,
     operation.variableDefinitions || [],
-    rawVariableValues || {}
+    rawVariableValues || {},
+    contextValue || {},
   );
 
   return {
--- a/src/execution/values.js
+++ b/src/execution/values.js
@@ -39,11 +39,17 @@ import type { Argument, VariableDefinition } from '../language/ast';
 export function getVariableValues(
   schema: GraphQLSchema,
   definitionASTs: Array<VariableDefinition>,
-  inputs: { [key: string]: mixed }
+  inputs: { [key: string]: mixed },
+  contextValue: mixed
 ): { [key: string]: mixed } {
   return definitionASTs.reduce((values, defAST) => {
     const varName = defAST.variable.name.value;
-    values[varName] = getVariableValue(schema, defAST, inputs[varName]);
+    values[varName] = getVariableValue(
+      schema,
+      defAST,
+      inputs[varName],
+      contextValue
+    );
     return values;
   }, {});
 }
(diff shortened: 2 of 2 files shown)


leebyron:
I'm not opposed to this, it's definitely interesting but can you provide some more info about how this should be used? It seems a little bit weird to me that the way a scalar value would be interpreted would change depending on context.

rricard:
Okay, right, that's what I wanted to talk about 😉 !

So for me all of this has a real use in those places, let's take the case of uploading files:

- **client side**: where you can just pass a `File` object from a form as a variable alongside the graphql mutation, here's what it would look like:

```js
remoteGraphqlCall(`
  mutation UploadFile($file: File) {
    uploadFile(file: $file) {
      size
      name
    }
  }
`, {
  file: document.querySelector("#some-input").files[0],
});
```

On this side the responsibility is **on the client** to serialise that in an actual HTTP request: for instance one could create a custom Relay or Apollo network layer to perform this operation of separating the `File`s in a multipart request and send it that way to the server.

```js
function remoteGraphqlCall(query, variables) {
  const formData = new FormData();

  let regularVariables = {};

  Object.keys(variables).forEach(varName => {
    const value = variables[varName];
    if (value instanceof File) {
      formData.set(varName, value);
      regularVariables[varName] = varName;
    } else {
      regularVariables[varName] = value;
    }
  });

  formData.set('graphql', JSON.stringify({
    query,
    variables: regularVariables,
  }));

  fetch(ENDPOINT_URL, {
    method: "POST",
    body: formData,
  );
}
```

_Theoretically, graphql-js has nothing to do with that!_

- **server side**: we're receiving a multipart `POST` query on the GraphQL server endpoint. In order to process that, we can customise the way we receive the query in express-graphql (or similar) with something like multer as a middleware:

```js
app.use('/graphql', graphqlHttp(request => {
  const {variables, query} = JSON.parse(request.body.graphql);
  return {
    query
    schema,
    context: {
      files: request.files,
    },
    variables,
  };
}));
```

Then once we have this, we can effectively use the context to rematch in our scalar resolver the actual file (in an actual buffer):

```js
return context.files[value].buffer;
```

Yes, it's deviating a bit the usage of scalars in GraphQL but I find a certain elegance in having complex types that are not serializable as strings right away...

I'm not sure though I see more usages than upload with the context associated but, who knows, it can be a great escape hatch in some cases!

However, I can understand if you are not completely on board with this: firstly because I see other ways to achieve equivalent functionality. But what I like about this is using the typesystem to coerce complex objects.

Thanks for considering it!

rricard:
Closing, outdated, repoen if still interested.

PR #487, opened 2016-09-15
Fix GraphQLError missing positions/locations when node.loc.start === 0

martijnwalraven:
Because the GraphQLError constructor filtered out falsy values for
node.loc.start, that would also filter out 0.

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -101,7 +101,8 @@ export function GraphQLError( // eslint-disable-line no-redeclare
 
   let _positions = positions;
   if (!_positions && nodes) {
-    _positions = nodes.map(node => node.loc && node.loc.start).filter(Boolean);
+    _positions = nodes.filter(node => node.loc !== null)
+      .map(node => node.loc.start);
   }
   if (_positions && _positions.length === 0) {
     _positions = undefined;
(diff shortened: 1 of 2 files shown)


facebook-github-bot-2:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-9:

src/error/GraphQLError.js
```diff
@@ -101,7 +101,8 @@ export function GraphQLError( // eslint-disable-line no-redeclare
 
   let _positions = positions;
   if (!_positions && nodes) {
-    _positions = nodes.map(node => node.loc && node.loc.start).filter(Boolean);
+    _positions = nodes.map(node => node.loc && node.loc.start)
```
leebyron: alternatively this could be `nodes.filter(node => node.loc != null).map(node => node.loc.start)` though Flow may complain about that.

martijnwalraven: You're right, that's much clearer. I made the changes and both the Flow checks and the tests still seem to pass. (I did change `!=` into `!==`, got a linting error).

leebyron:
Awesome! Thanks so much for the snappy test case as well

PR #494, opened 2016-09-17
Add internal names to input object fields and arguments

calebmer:
`GraphQLScalarType` gives us the ability to transform input data into an arbitrary form usable in our `resolve` functions. I’d like to see a similar functionality for `GraphQLInputObjectType`.

In my GraphQL schemas, I have a number of fairly complex input objects which must be in a different form by the time they reach my `resolve` function. One could manually parse the input objects in a `resolve` function, however doing so would require recursion and therefore type information making this case difficult to implement correctly. It would also require keeping this extra parsing information in a data structure parallel to both the value and the type system.

90% of the time, the transform I want to perform is renaming input object field names alone, having this feature baked into `graphql-js` would help ease this use case. Especially for users who may expect their input objects in `snake_case` but would like to expose a `camelCase` API.

This PR adds an `internalName` to input object fields and arguments which the executor will use as the key when building input objects or argument objects. This is the least powerful approach, but it addresses 90% of my use cases (maybe not others?). It’s also likely the least controversial option and easiest to implement.

### DIFF
--- a/src/execution/__tests__/variables-test.js
+++ b/src/execution/__tests__/variables-test.js
@@ -64,6 +64,24 @@ const TestNestedInputObject = new GraphQLInputObjectType({
   },
 });
 
+const TestInternalNamesInputObject = new GraphQLInputObjectType({
+  name: 'TestInternalNamesInputObject',
+  fields: {
+    insa: { type: new GraphQLNonNull(GraphQLString), internalName: 'a' },
+    insb: { type: new GraphQLNonNull(GraphQLString), internalName: 'b' },
+    insc: { type: new GraphQLNonNull(GraphQLString) },
+    insd: {
+      type: new GraphQLInputObjectType({
+        name: 'TestInternalNamesNestedInputObject',
+        fields: {
+          insa: { type: new GraphQLNonNull(GraphQLString), internalName: 'a' },
+          insb: { type: new GraphQLNonNull(GraphQLString) },
+        }
+      })
+    }
+  },
+});
+
 const TestType = new GraphQLObjectType({
   name: 'TestType',
   fields: {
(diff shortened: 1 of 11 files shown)


calebmer:
A method like `assign` which is analogous to `resolve` is a more functional/extensible solution to this problem, an `internalName` field is the declarative solution. The reason an `assign` can't be implemented is that `astFromValue` requires us to go the other way around (extracting the field vs assigning it). A declarative approach solves this.

How is `astFromValue` used? Does it accept the internal representation of values? If not the more general solution of `assign` may be better.

leebyron:
I'd love to learn more about what your use case is.

Ideally this sort of thing does not live in the GraphQL layer but instead lives in your code that uses GraphQL - that overall keeps the system simpler, though that's just a starting position - motivating examples should lead us from there.

calebmer:
My use case is that I’m generating GraphQL schemas from a database layer. As such all the names for objects in the database are snake_case, but I want all of the GraphQL fields to be camelCase. In addition some fields get renamed with more complex context specific rules then just snake_case to camelCase. Therefore, I need a step which has extra type information to convert the GraphQL input objects into database objects that can be inserted for example. Furthermore, I may have nested input objects, so things get recursive quickly.

But yeah, for now this should really be an implementation detail of my app. If GraphQL-js could provide a happy path for this use case in the future that would be excellent, but it’s definitely not a blocker.

However, it would be nice to add tests for passing through extra field arguments like this https://github.com/graphql/graphql-js/blob/a725499b155285c2e33647a93393c82689b20b0f/src/type/definition.js#L925 or this https://github.com/graphql/graphql-js/blob/a725499b155285c2e33647a93393c82689b20b0f/src/type/definition.js#L419. And also enabling the passage of extra properties here https://github.com/graphql/graphql-js/blob/a725499b155285c2e33647a93393c82689b20b0f/src/type/definition.js#L445-L450 along with other places.

I do currently depend on this functionality to implement similar logic as this PR, it would be nice to see it tested :+1:

PR #495, opened 2016-09-18
fix: inherit descriptor properties from the original error

gajus:
The the default values for `{writable,configurable}` are `false`.
Therefore, when you defineProperty without explicitly setting
`{configurable}` to `true`, it defaults to `false`. As a result,
the stack cannot be overwritten by a latter program, e.g.
Bluebird will attempt to add [longStackTraces](http://bluebirdjs.com/docs/api/promise.longstacktraces.html).

Instead of explicitly setting property descriptor settings,
inherit descriptor properties of the original error object.

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -80,15 +80,15 @@ export function GraphQLError( // eslint-disable-line no-redeclare
   // Include (non-enumerable) stack trace.
   if (originalError && originalError.stack) {
     Object.defineProperty(this, 'stack', {
-      value: originalError.stack,
-      writable: true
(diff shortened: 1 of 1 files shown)


leebyron:
How about instead we just add `configurable: true` to these? That also solves the problem but in clearer code

gajus:
> How about instead we just add `configurable: true` to these? That also solves the problem but in clearer code

My personal preference is to inherit settings as opposed to overwrite them.

Though, I agree that in 99% of the use cases, the end result is the same.

Let me know if I should update the PR.

leebyron:
Closed by #505

PR #496, opened 2016-09-19
Add isDeprecated value to field and enum value definitions

asiandrummer:
From [this GraphiQL issue](https://github.com/graphql/graphiql/pull/165), we found out `isDeprecated` is not available from the schema API via introspection. Provide it by checking if `deprecationReason` exists, and also the tests for this change.


leebyron:
Nice. You should also change the introspection field to use this

src/type/__tests__/definition-test.js
```diff
@@ -169,6 +169,41 @@ describe('Type System: Example', () => {
 
   });
 
+  it('defines an enum type with deprected value', () => {
```
intellix: deprected -> deprecated

Issue #501, opened 2016-09-23
Observables instead of Promises


DxCx:
Hi guys,

As more I'm thinking about it the more i think it should be done,
So before I'm submitting a PR i would like to hear your thoughts about it.

So basically i want to add support for Observables (RxJs? Any other preferred library?) so resolver function will be able to return Observables as well (Observable<any> | Promise<any> | any)

Then create another execute function that will return Observable<GraphQLResult> while chaining the original execute to call it with `Observable.toPromise()` function.

This way GraphQLResult can be reactive and this can be really useful for subscriptions and later in the future `@live` \ `@stream` \ `@defer` directives..

For the initial implementation I'm planing to always take(1).. later make it fully reactive..

So, what do you think?

davidyaha:
I think `@defer` and `@stream` could also benefit from observables. 

It will allow us to use the sweet power of observables in the server as well as the client.
Excellent proposition @DxCx!

Muffasa:
Amazing idea. I think it will open a door for far better architecture and maintainability.
Further more observables natural for reactivity or pub/sub and they are by far the best way I currently know to manipulate data.

DxCx:
Was just able to confirm that my idea works pretty well with subscription. (on unit test)
However the idea is simple:
  1. graphqlReactive should be used instead of graphql
  2. an hot Observable (Or subject) is being saved in context object.
  3. subscription resolver is being triggered when subscription doc is sent.
    - subscription resolver should return the hot observable from context mixed with any operator
      that is needed because of arguments.
  4. graphqlReactive returns an observable of GraphQLResult which is updated for each change.

leebyron:
Closing this issue. Observables are exciting but do not yet have clear behavior relative to the GraphQL spec. Since this repository stands to be a production server and reference implementation, I don't think it's the right place to experiment with this.

I'm looking forward to seeing this sort of experimentation happen in additional libraries that don't need to rely on spec compliance. See more comments on this in #502.

Also just a note that the versions of `@defer` and `@stream` I've prototyped in Facebook's server have not required Observable, so we should be able to still pursue those ideas without this approach.

DxCx:
for everyone that is interested in this,
I've just release `graphql-rxjs` package which adds Observable support to graphql engine:
https://www.npmjs.com/package/graphql-rxjs
Enjoy 👍

corysimmons:
lol thanks @DxCx 👏

shtse8:
[...]

PR #502, opened 2016-09-23
Intial refactoring to work over observables instead of promises

DxCx:
First Implementation for #501 .
NOTE: Promises still works, this change does not break any existing usage.

TODO:
- [ ] Get CR Notes.
- [ ] Update any documentation? Which?
- [ ] Replace Observables Library? set it to peer instead of dependency? pending a decision on issues.
- [x] Reactive Execution:
  - [x] remove TODO and enable reactive execution.
  - [x] write tests with reactive execution.
  - [x] make sure to `take(1)`  when non-reactive request executes and observable is resolved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "prepublish": ". ./resources/prepublish.sh"
   },
   "dependencies": {
-    "iterall": "1.0.2"
+    "iterall": "1.0.2",
+    "rxjs": "^5.0.0-beta.12"
   },
   "devDependencies": {
     "babel-cli": "6.18.0",
--- a/src/execution/__tests__/observables-test.js
+++ b/src/execution/__tests__/observables-test.js
@@ -1,0 +1,81 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+import { executeReactive } from '../execute';
+import { parse } from '../../language';
+import {
+  GraphQLSchema,
+  GraphQLObjectType,
+  GraphQLString,
+} from '../../type';
+import {
+  Observable
+} from 'rxjs';
+
+describe('Execute: Handles Observables from resolvers', () => {
+  it('uses the named operation if operation name is provided', async () => {
+    const doc = 'query Example { first: a }';
+    const data = { a: Observable.of('b') };
+    const schema = new GraphQLSchema({
+      query: new GraphQLObjectType({
+        name: 'Type',
+        fields: {
+          a: { type: GraphQLString },
+        }
+      })
+    });
+
+    const result = await executeReactive(schema, parse(doc), data).toPromise();
+
+    expect(result).to.deep.equal({ data: { first: 'b' } });
+  });
+
+  it('does not query reactive', () => {
+    const doc = 'query Example { first: a }';
+    const data = { a: Observable.interval(5) };
+    const schema = new GraphQLSchema({
+      query: new GraphQLObjectType({
+        name: 'Type',
+        fields: {
+          a: { type: GraphQLString },
+        }
+      })
+    });
+
+    return executeReactive(schema, parse(doc), data).take(2).map(result => {
+      expect(result).to.deep.equal({ data: { first: '0' } });
+    }).toPromise();
+  });
+
+  it('does query reactive for subscriptions', () => {
+    const doc = 'subscription Example { first: a }';
+    const data = { a: Observable.interval(5) };
+    const schema = new GraphQLSchema({
+      query: new GraphQLObjectType({
+        name: 'QueryType',
+        fields: {
+          b: { type: GraphQLString },
+        }
+      }),
+
+      subscription: new GraphQLObjectType({
+        name: 'SubscriptionType',
+        fields: {
+          a: { type: GraphQLString },
+        }
+      })
+    });
+    let counter = 0;
+
+    return executeReactive(schema, parse(doc), data).take(5).do(result => {
+      expect(result).to.deep.equal({ data: { first: counter.toString() } });
+      counter++;
+    }).toPromise().then(fresult => {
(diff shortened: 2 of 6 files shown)


facebook-github-bot-7:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-9:

DxCx:
Hi @leebyron can you please review this?
I'm very eager to hear your thoughts about this.

leebyron:
While I'm excited to support fields that return Observable values, I don't support this approach.

First, we shouldn't take a dependency on a large library like rxjs, either directly or as a peer dependency. Second, we should minimize the set of changes to the core executor. Third, we still need to support Promises/Thenable as the most common use case.

leebyron:
I'm in favor of continued experimentation in this area though. This is closely aligned with some other earlier experiments to add @live to GraphQL.

DxCx:
Hi @leebyron thanks for the quick response :)
I agree about the rxjs dependency, that's why i've opened it under discuession, and pending to hear what do you think should be supported..
also, if you will note, i did the minimal change to run over observables rather then promises.
then, in the end of the pipe, execute converts the observable into a promise, so no interface is broken.
and promises are still supported.

leebyron:
Another question is what we actually want to see from a reactive GraphQL executor. We probably do not want to re-execute fields that are likely to have not changed and the output should be a series of diffs to send over the network rather than a series of whole values

DxCx:
hehe, seems that we are aligned, just released today [rxjs-diff-operator](https://www.npmjs.com/package/rxjs-diff-operator) which is exactly for that purpose.

the update logic is same as the old one, so basiclly, static values will not be re-executed,
promises are Observables.fromPromise(1) which is observable of 1 value, so they won't reexecute either.
the only re-execution that accurs from my testing was when the subscription is Observable,
then for each value emitted, the ObjectType resolvers will be updated with a new root value from the subscription, however, this is the expected behavior as i see it.

DxCx:
hi @leebyron, eventually i decided to create a seperated npm [(graphql-rxjs)](https://www.npmjs.com/package/graphql-rxjs) package for this for the moment being.
this way i can experiment it myself,
i'm planing on creating example app that uses it so you can experiment yourself.

stubailo:
Really excited to see where this goes. I think observables in the resolver return can be a neat way to do subscriptions as well.

DxCx:
hi @leebyron,

i have a server side example (getting graphiql with subscription over observables)
i thought i posted it a while ago and now i see that i didn't :(

so, if you want to take a look,
https://github.com/DxCx/webpack-apollo-server/tree/rxjs-subscriptions
specifically, on main.ts i define a channel (https://github.com/DxCx/webpack-apollo-server/blob/rxjs-subscriptions/src/main.ts#L15)
and then i use it inside a resolver (https://github.com/DxCx/webpack-apollo-server/blob/rxjs-subscriptions/src/schema/modules/subscription.ts#L10)

if you want to really play with it, just npm install & npm start.
then you will have a graphiql interface with the subscription on port 3000.

Other then that, i'm not sure what are you against in here,
are you against for the general concept of using Observables along with promises? (which feels much more intuitive versus the pub/sub solution found today)
or for the dependency of rxjs?

also, once this is resolved i would be happy to work on the @defer/@live/@stream directives as well as with observables it's very easy to implement.

DxCx:
@leebyron can you please reply?

leebyron:
Hey @DxCx, sorry for the delayed response.

Since this library primarily serves to be a production node server and reference implementation of the spec, I don't think it's the appropriate place for this kind of experimentation for fear that other implementations would see it as part of the direction of GraphQL, and this is not yet what we'd like to do. 

Notably - it's important that the execution code align to the spec to be a good reference implementation.

I do think it's a great idea to experiment with observable-based resolvers, it's an exciting idea that deserves to be explored - but I think that exploration deserves it's own repository which can move quickly and experiment with different ideas without worrying about breaking core users.

leebyron:
I'm going to close this PR since we don't want to pursue this in this repository - I'm looking forward to seeing your work as a separate package.

DxCx:
hey @leebyron 
after seen the work on subscription i was thinking,
if i'll rewrite this using AsyncIterator instead of Observables, is that something you will accept?

i mean, i want to introduce some execute function (`executeReactive`?) which is able to return more then one result potentially.

smkhalsa:
Hey @leebyron,

Now that AsyncIterator has [reached Stage 4](https://github.com/tc39/proposals/commit/a41dccc91675f41c65dee354c18c19946d6f2f45), I'd love to hear your thoughts on @DxCx 's proposal above to add AsyncIterator support to the execute function. The ability to return multiple results from the execute function would be a huge step forward for live queries. Thanks!

leebyron:
A live query would need a separate core executor. It might use some parts
of the algorithm which are the same, however the live executor will have
different concerns, so we should not convert the existing executor to be a
live executor.

I agree that for a live executor, AsyncIterators will be an important
implementation detail, however I’d recommend accepting either AIs or
Observables from resolver functions

smkhalsa:
@leebyron That makes sense. I believe that @DxCx 's [graphql-rxjs](https://github.com/DxCx/graphql-rxjs) takes this approach, with a separate `executeReactive` executor. 

Would you consider a PR for adding reactive executors to the library at this time? Thanks again!

leebyron:
At this point I think it makes sense for add on libraries to work out
different approaches. If one approach becomes robust to all use cases we
could consider merging it into this library

Issue #439, opened 2016-07-19
Make GraphQLError have a writeable `.stack` property


alexturek:
Bluebird checks to see if `stack` is a writeable property on a rejected object, and (incorrectly) warns that the object being rejected with is not an error.

[Offending bluebird line](https://github.com/petkaantonov/bluebird/blob/master/src/util.js#L257)

If the property `stack` was made writeable ([here](https://github.com/graphql/graphql-js/blob/master/src/error/GraphQLError.js#L37)), Bluebird would shut up.

Otherwise, when I throw an error out of an ApolloServer resolver, I see a lot of
```
[grunt-develop] > Warning: a promise was rejected with a non-error (OR WAS IT): [object Error]
    at /Users/aturek/project/node_modules/graphql/execution/execute.js:518:22
From previous event:
    at Promise.then (/Users/aturek/project/node_modules/sequelize/lib/promise.js:24:17)
    at completeValue (/Users/aturek/project/node_modules/graphql/execution/execute.js:510:19)
    at completeValueCatchingError (/Users/aturek/project/node_modules/graphql/execution/execute.js:466:21)
    at resolveField (/Users/aturek/project/node_modules/graphql/execution/execute.js:438:10)
    at /Users/aturek/project/node_modules/graphql/execution/execute.js:243:18
    at Array.reduce (native)
    at executeFields (/Users/aturek/project/node_modules/graphql/execution/execute.js:240:42)
    at executeOperation (/Users/aturek/project/node_modules/graphql/execution/execute.js:181:10)
    at /Users/aturek/project/node_modules/graphql/execution/execute.js:97:13
<truncated>
```

I can make this change if people sign off on it

alexturek:
See also: https://github.com/petkaantonov/bluebird/issues/1146#issuecomment-229076868

leebyron:
Wow, that's kind of annoying.

It makes my heart hurt a little bit to enable overwriting the stack information but let me see if I can't fix this.

alexturek:
I actually assume they're doing it with good intentions; namely so they can rewrite the stack to support more frames (i.e. trace the provenance of the Promise).

But yeah, I get it. Mutable data 👎

PR #505, opened 2016-09-26
Add 'configurable: true' to GraphQLError stack property

### LINKED ISSUE #488: resolve() hangs with rejected bluebird Promises
I ran into this issue with graphql-js 0.7.0. While invoking the Stripe API using Bluebird::promiseFromCallback(), Stripe rejected the invocations with their own error object, not an `Error` instance. This resulted in the graphQL function to hang, and not process the error correctly.

I was able to correct my code by manually converting their errors into `Error` instances, but it seems like graphql should handle non-error based Promise rejections more gracefully.

darthtrevino:
Bluebird will throw an error if an error used to reject a promise has `writable: true` but not `configurable: true`

See: 
https://github.com/graphql/graphql-js/issues/439
https://github.com/petkaantonov/bluebird/pull/1238

Fixes #488

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -81,14 +81,16 @@ export function GraphQLError( // eslint-disable-line no-redeclare
   if (originalError && originalError.stack) {
     Object.defineProperty(this, 'stack', {
       value: originalError.stack,
-      writable: true
+      writable: true,
+      configurable: true
     });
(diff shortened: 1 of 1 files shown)


darthtrevino:
@leebyron 🙏,,can you review this? #488 is a giant PITA

leebyron:
Thank you for this!

PR #508, opened 2016-10-03
Add utility functions to compare schemas

sam-swarr:
This PR introduces some checks that can be used to help verify if changes to a schema are "safe". The idea is that you generate a `GraphQLSchema` object that represents the schema before and after a schema change (for example by using `buildClientSchema`). You can then pass the `oldSchema` and `newSchema` into these utility functions. The functions each check for different types of breaking changes, i.e. changes in the `newSchema` that are not backwards compatible with `oldSchema`. They can be used to help prevent regressions caused by unsafe schema changes that would break older versions of clients.

### DIFF
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -63,6 +63,7 @@ export { assertValidName } from './assertValidName';
 
 // Helpful functions to compare two GraphQLSchemas and detect breaking changes.
 export {
+  findBreakingChanges,
   findRemovedTypes,
   findTypesThatChangedType,
   findBreakingFieldChanges,
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -61,12 +61,5 @@ export {
 // Asserts that a string is a valid GraphQL name
 export { assertValidName } from './assertValidName';
 
-// Helpful functions to compare two GraphQLSchemas and detect breaking changes.
-export {
-  findBreakingChanges,
-  findRemovedTypes,
-  findTypesThatChangedKind,
-  findFieldsThatChangedType,
-  findTypesRemovedFromUnions,
-  findValuesRemovedFromEnums,
-} from './schemaComparisons';
+// Compares two GraphQLSchemas and detects breaking changes.
+export { findBreakingChanges } from './schemaComparisons';
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -65,8 +65,8 @@ export { assertValidName } from './assertValidName';
 export {
   findBreakingChanges,
   findRemovedTypes,
-  findTypesThatChangedType,
-  findBreakingFieldChanges,
+  findTypesThatChangedKind,
+  findFieldsThatChangedType,
   findTypesRemovedFromUnions,
   findValuesRemovedFromEnums,
 } from './schemaComparisons';
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -60,3 +60,12 @@ export {
 
 // Asserts that a string is a valid GraphQL name
 export { assertValidName } from './assertValidName';
+
+// Helpful functions to compare two GraphQLSchemas and detect breaking changes.
+export {
+  findRemovedTypes,
+  findTypesThatChangedType,
+  findBreakingFieldChanges,
+  findTypesRemovedFromUnions,
+  findValuesRemovedFromEnums,
+} from './schemaComparisons';
--- a/src/utilities/schemaComparisons.js
+++ b/src/utilities/schemaComparisons.js
@@ -62,12 +62,10 @@ export function findRemovedTypes(
   const oldTypes = Object.keys(oldSchema.getTypeMap());
   const newTypes = new Set(Object.keys(newSchema.getTypeMap()));
   return oldTypes.filter(typeName => !newTypes.has(typeName)).map(
-    type => {
-      return {
-        type: BreakingChangeType.TYPE_REMOVED,
-        description: `${type} was removed`,
-      };
-    }
+    type => ({
+      type: BreakingChangeType.TYPE_REMOVED,
+      description: `${type} was removed`,
+    })
   );
 }
 
@@ -82,7 +80,7 @@ export function findTypesThatChangedKind(
   const oldTypeMap = oldSchema.getTypeMap();
   const newTypeMap = newSchema.getTypeMap();
 
-  const typesThatChangedType = [];
+  const breakingChanges = [];
   Object.keys(oldTypeMap).forEach(typeName => {
     if (!newTypeMap[typeName]) {
       return;
@@ -90,16 +88,23 @@ export function findTypesThatChangedKind(
     const oldType = oldTypeMap[typeName];
     const newType = newTypeMap[typeName];
     if (!(oldType instanceof newType.constructor)) {
-      typesThatChangedType.push(
+      breakingChanges.push(
         {
           type: BreakingChangeType.TYPE_CHANGED_KIND,
-          description: `${typeName} changed from a ` +
-            `${oldType.constructor.name} to a ${newType.constructor.name}`,
+          description: `${typeName} changed from ` +
+            `${prettifyTypeKind(oldType.constructor.name)} type to ` +
+            `${prettifyTypeKind(newType.constructor.name)} type`,
         }
       );
     }
   });
-  return typesThatChangedType;
+  return breakingChanges;
+}
+
+function prettifyTypeKind(typeKind: string): string {
+  const kind = typeKind.replace('GraphQL', '').replace('Type', '');
+  const article = [ 'A','E','I','O' ].includes(kind.charAt(0)) ? 'an' : 'a';
+  return article + ' ' + kind;
 }
 
 /**
--- a/src/utilities/schemaComparisons.js
+++ b/src/utilities/schemaComparisons.js
@@ -59,9 +59,9 @@ export function findRemovedTypes(
   oldSchema: GraphQLSchema,
   newSchema: GraphQLSchema
 ): Array<BreakingChange> {
-  const oldTypes = Object.keys(oldSchema.getTypeMap());
-  const newTypes = new Set(Object.keys(newSchema.getTypeMap()));
-  return oldTypes.filter(typeName => !newTypes.has(typeName)).map(
+  const oldTypeNames = Object.keys(oldSchema.getTypeMap());
+  const newTypes = newSchema.getTypeMap();
+  return oldTypeNames.filter(typeName => !newTypes[typeName]).map(
     type => ({
       type: BreakingChangeType.TYPE_REMOVED,
       description: `${type} was removed`,
@@ -92,8 +92,8 @@ export function findTypesThatChangedKind(
         {
           type: BreakingChangeType.TYPE_CHANGED_KIND,
           description: `${typeName} changed from ` +
-            `${prettifyTypeKind(oldType.constructor.name)} type to ` +
-            `${prettifyTypeKind(newType.constructor.name)} type`,
+            `${addArticle(oldType.constructor.prettyName)} type to ` +
+            `${addArticle(newType.constructor.prettyName)} type`,
         }
       );
     }
@@ -101,10 +101,9 @@ export function findTypesThatChangedKind(
   return breakingChanges;
 }
 
-function prettifyTypeKind(typeKind: string): string {
-  const kind = typeKind.replace('GraphQL', '').replace('Type', '');
-  const article = [ 'A','E','I','O' ].includes(kind.charAt(0)) ? 'an' : 'a';
-  return article + ' ' + kind;
+function addArticle(typeKind: string): string {
+  const article = [ 'A','E','I','O' ].includes(typeKind.charAt(0)) ? 'an' : 'a';
+  return article + ' ' + typeKind;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


leebyron:
This is awesome!

My high level feedback is that we should probably have a single function like `findBreakingChanges` which is capable of returning a full list of described breakages.

Otherwise it's up to you to call all of these functions in succession which could require some boilerplate.

sam-swarr:
> My high level feedback is that we should probably have a single function like findBreakingChanges which is capable of returning a full list of described breakages.

Good call. I added a top level `findBreakingChanges` check and added a test case.

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,220 @@
...
+): Set<string> {
+  const oldTypeMap: TypeMap = oldSchema.getTypeMap();
+  const newTypeMap: TypeMap = newSchema.getTypeMap();
+
+  const breakingFieldChanges: Array<string> = [];
+  for (const typeName: string in oldTypeMap) {
+    if (!{}.hasOwnProperty.call(oldTypeMap,typeName)) {
+      continue;
+    }
+    const oldType:GraphQLType = oldTypeMap[typeName];
```
leebyron: nit: space after `:`

But also, you don't need to add types to variables you create - Flow infers all of that for you. Only need to annotate function args & returns. You can remove all of them and Flow will keep on chugging.

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * Given two schemas, returns a Set containing descriptions of all the types of
+ * breaking changes covered by the other functions down below.
+ */
+export function findBreakingChanges(
+  oldSchema: GraphQLSchema,
+  newSchema: GraphQLSchema
+): Set<string> {
+  return new Set([
```
leebyron: What's the reasoning behind using Sets throughout instead of just Arrays?

It seems like there ends up being a lot of work converting between Sets and Arrays and Object maps and we should ensure it's worth doing all that.

Also, I bet users of this function will expect an array of errors rather than a Set.

It's cool to use Set as an implementation detail in deriving these breakages though!

e.g. this could become:

```js
return [
  ...findRemovedTypes(oldSchema, newSchema),
  ...findTypesThatChangedType(oldSchema, newSchema),
  ...findBreakingFieldChanges(oldSchema, newSchema),
  ...findTypesRemovedFromUnions(oldSchema, newSchema),
  ...findValuesRemovedFromEnums(oldSchema, newSchema)
];
```

src/utilities/index.js
```diff
@@ -60,3 +60,13 @@ export {
 
 // Asserts that a string is a valid GraphQL name
 export { assertValidName } from './assertValidName';
+
+// Helpful functions to compare two GraphQLSchemas and detect breaking changes.
+export {
+  findBreakingChanges,
```
leebyron: You probably only need to export this function (same below). All the other functions are really just implementation details of this one.

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,220 @@
...
+    )
+  );
+  return removedTypes;
+}
+
+/**
+* Given two schemas, returns a Set containing descriptions of any breaking
+* changes in the newSchema related to changing the type of a type.
+*/
+export function findTypesThatChangedType(
```
leebyron: A better name for this might be `findTypesThatChangedKind`

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,220 @@
...
+  }
+  return new Set(typesThatChangedType);
+}
+
+/**
+* Given two schemas, returns a Set containing descriptions of any breaking
+* changes in the newSchema related to the fields on a type. This includes if
+* a field has been removed from a type or if a field has changed type.
+*/
+export function findBreakingFieldChanges(
```
leebyron: `findFieldsThatChangedType` might be a better name relative to the names of the other functions

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,220 @@
...
+*/
+export function findTypesRemovedFromUnions(
+  oldSchema: GraphQLSchema,
+  newSchema: GraphQLSchema
+): Set<string> {
+  const oldTypeMap: TypeMap = oldSchema.getTypeMap();
+  const newTypeMap: TypeMap = newSchema.getTypeMap();
+
+  const typesRemovedFromUnion: Array<string> = [];
+  for (const typeName: string in oldTypeMap) {
```
leebyron: Use `Object.keys(oldTypeMap).forEach(typeName =>` instead.

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,220 @@
...
+    if (!{}.hasOwnProperty.call(oldTypeMap,typeName)) {
```
leebyron: this will be removed by the suggestion above, but in the future, this sort of check is better written:

```js
oldTypeMap.hasOwnProperty(typeName)
```

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,220 @@
...
+*/
+export function findValuesRemovedFromEnums(
+  oldSchema: GraphQLSchema,
+  newSchema: GraphQLSchema
+): Set<string> {
+  const oldTypeMap: TypeMap = oldSchema.getTypeMap();
+  const newTypeMap: TypeMap = newSchema.getTypeMap();
+
+  const valuesRemovedFromEnums: Array<string> = [];
+  for (const typeName: string in oldTypeMap) {
```
leebyron: same

leebyron:
Should this, rather than returning a list of strings, return a list of some kind of object, where `description` is one field, but another field is some constant that tells a machine what kind of breakage it is?

src/utilities/schemaComparisons.js
sam-swarr: Yep, I like that better.

src/utilities/schemaComparisons.js
sam-swarr: Updated

src/utilities/schemaComparisons.js
sam-swarr: Ah good to know. I removed type declarations for all local variables.

src/utilities/schemaComparisons.js
sam-swarr: Converted all for in loops to forEach.

src/utilities/schemaComparisons.js
sam-swarr: I had that, but then I got `TypeError: oldTypeMap.hasOwnProperty is not a function`. I guess Flow-typed objects don't inherit `hasOwnProperty`?

Doesn't really matter though since I took your advice above and converted to forEach loops.

src/utilities/index.js
sam-swarr: I kept the other functions exported for a couple of reasons. One reason is so that the unit test can test them individually. The other is that other users might be interested in only a subset of the checks. This way they can pick and choose.

sam-swarr:
> Should this, rather than returning a list of strings, return a list of some kind of object, where description is one field, but another field is some constant that tells a machine what kind of breakage it is?
Good idea. I added a `BreakingChange` object that is now returned by these functions. That should allow the caller more flexibility in handling the results.

src/utilities/index.js
leebyron: Unit tests should be importing from that file directly - not from this package module.

I don't think there's a good reason to only partially check for breaking changes? Perhaps if you want to filter out certain breakages a better way would be to filter based on data in the BreakingChange objects instead of manually assembling these functions yourself.

If there real need in the future then we could export these then - otherwise I'd prefer to keep the surface area of the exported API smaller and preserve the option value to change the implementation of `findRemovedTypes` in the future.

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,239 @@
...
+export const BreakingChangeType = {
+  FIELD_CHANGED_KIND: 'FIELD_CHANGED_KIND',
+  FIELD_REMOVED: 'FIELD_REMOVED',
+  TYPE_CHANGED_KIND: 'TYPE_CHANGED_KIND',
+  TYPE_REMOVED: 'TYPE_REMOVED',
+  TYPE_REMOVED_FROM_UNION: 'TYPE_REMOVED_FROM_UNION',
+  VALUE_REMOVED_FROM_ENUM: 'VALUE_REMOVED_FROM_ENUM',
+};
+
+export type BreakingChange = {
```
leebyron: nice!

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,239 @@
...
+ * changes in the newSchema related to removing an entire type.
+ */
+export function findRemovedTypes(
+  oldSchema: GraphQLSchema,
+  newSchema: GraphQLSchema
+): Array<BreakingChange> {
+  const oldTypes = Object.keys(oldSchema.getTypeMap());
+  const newTypes = new Set(Object.keys(newSchema.getTypeMap()));
+  return oldTypes.filter(typeName => !newTypes.has(typeName)).map(
+    type => {
```
leebyron: this arrow only has one statement, a return statement, which means it can be re-written as:

```
map(type => ({
  type: BreakingChangeType.TYPE_REMOVED,
  description: `${type} was removed`,
}))
```

note the wrapping `({})` around the object literal to disambiguate from `{}` which arrows treat as a statement body :/

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,239 @@
...
+
+  const typesThatChangedType = [];
+  Object.keys(oldTypeMap).forEach(typeName => {
+    if (!newTypeMap[typeName]) {
+      return;
+    }
+    const oldType = oldTypeMap[typeName];
+    const newType = newTypeMap[typeName];
+    if (!(oldType instanceof newType.constructor)) {
+      typesThatChangedType.push(
```
leebyron: `typesThatChangedKind`

or perhaps just `breakingChanges`?

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,239 @@
...
+        {
+          type: BreakingChangeType.TYPE_CHANGED_KIND,
+          description: `${typeName} changed from a ` +
+            `${oldType.constructor.name} to a ${newType.constructor.name}`,
```
leebyron: These error names might be less clear to non JS users if this utility is used along with other servers.

This might print:

`Foo changed from a GraphQLObjectType to a GraphQLScalarType`

but might be easier to understand as:

`Foo changed from an Object type to a Scalar type`

src/utilities/index.js
sam-swarr: > Unit tests should be importing from that file directly - not from this package module.

Turns out the test was already directly importing from the file.

> If there real need in the future then we could export these then - otherwise I'd prefer to keep the surface area of the exported API smaller and preserve the option value to change the implementation of findRemovedTypes in the future.

That's a very good point. Updated so that only `findBreakingChanges` is exported.

src/utilities/schemaComparisons.js
sam-swarr: Updated so that the more generic Object/Scalar/Interface term is used.

igorcanadi:
Awesome to see this come together!

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,244 @@
...
+ * Given two schemas, returns an Array containing descriptions of any breaking
+ * changes in the newSchema related to removing an entire type.
+ */
+export function findRemovedTypes(
+  oldSchema: GraphQLSchema,
+  newSchema: GraphQLSchema
+): Array<BreakingChange> {
+  const oldTypes = Object.keys(oldSchema.getTypeMap());
+  const newTypes = new Set(Object.keys(newSchema.getTypeMap()));
+  return oldTypes.filter(typeName => !newTypes.has(typeName)).map(
```
leebyron: No need to convert object to set:

```js
const oldTypeNames = Object.keys(oldSchema.getTypeMap());
const newTypes = newSchema.getTypeMap();
return oldTypeNames.filter(typeName => !newTypes[typeName]).map(
```

src/utilities/schemaComparisons.js
```diff
@@ -0,0 +1,244 @@
...
+            `${prettifyTypeKind(oldType.constructor.name)} type to ` +
+            `${prettifyTypeKind(newType.constructor.name)} type`,
+        }
+      );
+    }
+  });
+  return breakingChanges;
+}
+
+function prettifyTypeKind(typeKind: string): string {
```
leebyron: lol this is kind of amazing - was expecting something more like `if (x instanceof GraphQLEnumType) { return "an Enum"; }` but I'm cool with this.

denvned: I think just adding a `prettyName` static property to each graphql type class would comply the open-closed principle even better. It could be accessed as `oldType.constructor.prettyName`.

src/utilities/schemaComparisons.js
sam-swarr: Removed Set.

src/utilities/schemaComparisons.js
sam-swarr: Updated to use the new static `prettyName` property.

leebyron:
Nice work

sam-swarr:
Thanks for the thorough review!

PR #509, opened 2016-10-06
Allow resolveType to return a type name string

tgriesser:
This change permits `resolveType` functions to return the "name" string of a type definition rather than the concrete `GraphQLObjectType` instance. 

This is particularly useful when the `resolveType` function doesn't have access to the type instances in the immediate closure scope. While it's possible to accomplish the same with `info.schema.getType` in the resolve function, being able to return a string and have the executor figure it out felt a bit nicer.

Example:
```js
const PetType = new GraphQLInterfaceType({
  name: 'Pet',
  resolveType(obj) {
    return obj instanceof Dog ? 'Dog' :
      obj instanceof Cat ? 'Cat' :
      null;
  },
  fields: {
    name: { type: GraphQLString }
  }
});
```

### DIFF
--- a/src/execution/__tests__/abstract-test.js
+++ b/src/execution/__tests__/abstract-test.js
@@ -353,4 +353,79 @@ describe('Execute: Handles execution of abstract types', () => {
     });
   });
 
+  it('resolveType allows resolving with type name', async () => {
+    const PetType = new GraphQLInterfaceType({
+      name: 'Pet',
+      resolveType(obj) {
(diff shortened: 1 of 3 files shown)


leebyron:
This is cool - nice idea.

istarkov:
@leebyron  having this isn't it will be good to make default `resolveType` as

```javascript
resolveType(obj) {  
   return obj.constructor.name;
}
```

So it will work with es6 classes, and will allow to define interfaces using graphql schema definitions?

PR #521, opened 2016-10-18
Fix minor js syntax error in README example

cflynn07:


stubailo:
That's not a syntax error, it is valid JavaScript syntax.

cflynn07:
Ah yup you're right, my bad

PR #544, opened 2016-10-27
language support for NullValue

langpavel:
Work in progress [GraphQL RFC: Null value](https://github.com/facebook/graphql/pull/83/files)
- [x] Support `null` literal in Parser
- [x] Support `null` literal in Printer
- [x] `astFromValue`
- [x] `valueFromAST`
- [x] `getArgumentValues`, `getVariableValue` in **execution/values.js**
- [x] `isValidLiteralValue` in **utilities/isValidLiteralValue.js**
- [x] introspection of `__InputValue defaultValue` field in **type/introspection.js**
- [x] **utilities/schemaPrinter.js**
- [x] _some more tests_


langpavel:
Help wanted! :-)

src/language/ast.js
```diff
@@ -284,6 +286,12 @@ export type BooleanValue = {
   value: boolean;
 }
 
+export type NullValue = {
+  kind: 'NullValue';
+  loc?: Location;
+  // value: null;
```
leebyron: no need for this line - I think it's more confusing than helpful since the type doesn't actually have a `value` field, and if you attempted to read one, you'd see `undefined` instead of `null`

src/utilities/__tests__/astFromValue-test.js
```diff
@@ -33,10 +33,14 @@ describe('astFromValue', () => {
       { kind: 'BooleanValue', value: false }
     );
 
-    expect(astFromValue(null, GraphQLBoolean)).to.deep.equal(
+    expect(astFromValue(undefined, GraphQLBoolean)).to.deep.equal(
       null
     );
 
+    expect(astFromValue(null, GraphQLBoolean)).to.deep.equal(
```
leebyron: Could you also add a test for `astFromValue(null, new GraphQLNonNull(GraphQLBoolean))` to expect `null`

src/utilities/astFromValue.js
```diff
@@ -73,7 +76,10 @@ export function astFromValue(
     return astFromValue(_value, type.ofType);
   }
 
-  if (isNullish(_value)) {
+  // only explicit null, not undefined, NaN
+  if (_value === null) {
+    return ({ kind: NULL }: NullValue);
+  } else if (isNullish(_value)) {
```
leebyron: no need for the else since you're returning in the if branch

src/language/ast.js
langpavel: I was in doubt with this. Can I add `value` property? Because other literal AST nodes have `value` - so I keep it commented. So you suggest remove `value` or actually set it to null?

I think remove it will be better, but thank you for your careful review :+1:

src/utilities/__tests__/astFromValue-test.js
langpavel: @leebyron I'm little confused by this.
What is correct result for `astFromValue(null, new GraphQLNonNull(*))`?

In `astFromValue` implementation there is this comment at line 74:
> Note: we're not checking that the result is non-null.
> This function is not responsible for validating the input value.
So, it is correct to break this logic?

langpavel: Ok, it makes sense. I added condition and test in 02d71a06b847d8f3e85bfb6836e2d75ab1425218

langpavel:
@leebyron Can you review please?

I think I'm done with code, if you want some more tests I can add, but it seems I shift coverage a little up and I'm happy with mine :1234: :smile:

src/utilities/valueFromAST.js
```diff
@@ -93,13 +98,14 @@ export function valueFromAST(
...
         valueFromAST(fieldAST && fieldAST.value, field.type, variables);
-      if (isNullish(fieldValue)) {
-        fieldValue = field.defaultValue;
-      }
-      if (!isNullish(fieldValue)) {
+      if (fieldValue === null) {
+        obj[fieldName] = fieldValue;
+      } else if (!isNullish(fieldValue)) {
         obj[fieldName] = fieldValue;
+      } else if (fieldValue !== null) {
```
leebyron: This could just be `else` since `fieldValue !== null` is guaranteed to be true here, otherwise the first branch would have been followed.

Also, the two conditions above are the same.

I think you can further shorten to a trinary.

```js
// If no valid field value was provided, use the default value
obj[fieldName] = fieldValue === undefined || fieldValue !== fieldValue ? 
  field.defaultValue :
  fieldValue;
```

I think you can probably skip the `isNullish` in favor of inlining the check here. I think `isNullish(x) && x !== null` would be confusing to read.

src/utilities/valueFromAST.js
```diff
@@ -111,6 +117,7 @@ export function valueFromAST(
   );
 
   const parsed = type.parseLiteral(valueAST);
+  // TODO: Should be this condition ommited?
```
leebyron: This is a little bit tricky now, but I think the right decision is to leave this check alone.

Right now the definition for parseLiteral is to either return a value, or return null (or some nullish value) if the parse failed. I don't think it's ever appropriate for parseLiteral to return null and have that be intended as a null value to be used.

src/utilities/astFromValue.js
```diff
@@ -52,6 +54,7 @@ import { GraphQLID } from '../type/scalars';
  *
  * | JSON Value    | GraphQL Value        |
  * | ------------- | -------------------- |
+ * | null          | NullValue            |
```
leebyron: nitpick, could you put this entry at the bottom of the table, that way it's in the same order as the comments above valueFromAST.

src/utilities/__tests__/astFromValue-test.js
leebyron: Yeah - I like the change you made.

langpavel:
Ahh, sorry, I rebased! But I hope your input was not lost!
I have 5:37 A.M. here so I probably should go sleep today :smile: :night_with_stars:

leebyron:
Definitely get some rest. Thanks for the progress!

src/utilities/valueFromAST.js
langpavel: Ternary it is, thanks

src/utilities/valueFromAST.js
langpavel: Comment removed

src/utilities/astFromValue.js
langpavel: Moved

langpavel:
Hi @leebyron. I'm kindly asking you for next review round, please :-)

I still should write some more tests mainly for validating null in `NonNull` but I'm not sure where is best place to settle this tests. Can you hint me?

Thanks!

langpavel:
I also discovered that it is not completed yet, because it is still possible to set default value to null for NonNull fields... This is definitely not right but I don't know how I can check this correctly..

src/language/__tests__/schema-kitchen-sink.graphql
```diff
@@ -17,6 +17,7 @@ type Foo implements Bar {
   four(argument: String = "string"): String
   five(argument: [String] = ["string", "string"]): String
   six(argument: InputType = {key: "value"}): Type
+  seven(argument: Int = null): Type
```
cjoudrey: This might have been discussed elsewhere, but I'm confused why we want to support `null` as a **default** value.

How is this different from:

```graphql
seven(argument: Int): Type
```

Except that one will have `args["argument"]` set to `null` and the other `args["argument"]` will be undefined.

cc @rmosolgo

cjoudrey: Right, so by setting the default value of an input argument to `null` you're making it impossible to know if the user passed in `null` vs the user omitted the input value entirely.

```graphql
six(argument: Int): Type
seven(argument: Int = null): Type
```

In `six`, the value of `argument` could be `null | Int | undefined`.

In `seven`, the value of `argument` could be `null | Int`.

Since both `null` and `undefined` are falsy, I can't think of a case where one would explicitly set the **default** value to `null` in order to omit the `undefined` case.

Maybe when using Flow/TypeScript removing this 3rd case makes it easier to understand the code?

langpavel: Default value will be present inside `resolve`

Nullable type without value (and default value) will be omitted === `undefined`

[RFC](https://github.com/facebook/graphql/pull/83/files#diff-fe406b08746616e2f5f00909488cce66R746)

langpavel: @cjoudrey

> Since both null and undefined are falsy, I can't think of a case where one would explicitly set the default value to null in order to omit the undefined case.

It is valid language construct. I doesn't care if it is semantically useful - but someone may want explicitly tell that there is no semantical difference between undefined and null

leebyron: Yeah this is probably confusing as an *example* because I agree that it's a little strange to make a default value of `null` to rule out the "not provided" (aka undefined) case, but it's certainly possible now, so it's worth testing as an *edge case*.

leebyron:
> I also discovered that it is not completed yet, because it is still possible to set default value to null for NonNull fields... This is definitely not right but I don't know how I can check this correctly..

The appropriate fix for this is the DefaultValuesOfCorrectType validation rule. 

Related to this is the ArgumentsOfCorrectType validation rule which looks for literal values being passed to arguments, which should reject `field(arg: null)` for `field(arg: Int!): Int!`

src/jsutils/isInvalid.js
```diff
@@ -0,0 +1,16 @@
...
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+/**
+ * Returns true if a value is null, undefined, or NaN.
```
leebyron: comment is incorrect

langpavel:
@leebyron DefaultValuesOfCorrectType fixed just now, tests for ArgumentsOfCorrectType I write very soon

src/utilities/astFromValue.js
```diff
@@ -68,11 +71,19 @@ export function astFromValue(
...
+    return astValue;
+  }
+
+  // only explicit null, not undefined, NaN
+  if (_value === null) {
+    return ({ kind: NULL }: NullValue);
   }
 
+  // other nullish values
   if (isNullish(_value)) {
```
leebyron: perhaps use your new `isInvalid` function here?

src/utilities/astFromValue.js
```diff
@@ -68,11 +71,19 @@ export function astFromValue(
   const _value = value;
 
   if (type instanceof GraphQLNonNull) {
-    // Note: we're not checking that the result is non-null.
-    // This function is not responsible for validating the input value.
-    return astFromValue(_value, type.ofType);
+    const astValue = astFromValue(_value, type.ofType);
+    if (astValue && astValue.kind === NULL) {
+      return null;
```
leebyron: perhaps to avoid confusion when using this function, rather than returning `AST | null` it should return `AST | undefined` - it would just help reduce the slim chance the output is misrepresented and `null` is treated improperly. I think the only necessary changes is just to replace the `return null;` statements with `return;`

src/utilities/valueFromAST.js
```diff
@@ -57,13 +59,17 @@ export function valueFromAST(
   }
 
   if (!valueAST) {
+    return undefined;
```
leebyron: This can be re-written as `return;`.

src/utilities/valueFromAST.js
```diff
@@ -57,13 +59,17 @@ export function valueFromAST(
...
+  }
+
+  if (valueAST.kind === Kind.NULL) {
```
leebyron: I think you should include a short comment explaining that this is intentionally returning the *value* `null`, which is different from simply `return;`

src/utilities/valueFromAST.js
```diff
@@ -93,14 +99,11 @@ export function valueFromAST(
...
         valueFromAST(fieldAST && fieldAST.value, field.type, variables);
-      if (isNullish(fieldValue)) {
-        fieldValue = field.defaultValue;
-      }
-      if (!isNullish(fieldValue)) {
-        obj[fieldName] = fieldValue;
-      }
+
+      // If no valid field value was provided, use the default value
+      obj[fieldName] = isInvalid(fieldValue) ? field.defaultValue : fieldValue;
```
leebyron: This is way nicer than the original logic.

src/validation/__tests__/ArgumentsOfCorrectType-test.js
```diff
@@ -117,6 +117,70 @@ describe('Validate: Argument values of correct type', () => {
 
   });
 
+  describe('`null` values', () => {
```
leebyron: It's unnecessary to have a test for every scalar type - really you only need one case that expects a pass for the literal `null`, I'd title it `null into nullable type`. However you should also include a test for passing the literal `null` to a non-null type.

src/validation/__tests__/DefaultValuesOfCorrectType-test.js
```diff
@@ -68,6 +68,44 @@ describe('Validate: Variable default values of correct type', () => {
     `);
   });
 
+  it('variables with valid default null values', () => {
```
leebyron: These two tests are great. Use this as a model for ArgumentsOfCorrectType-test

src/validation/__tests__/DefaultValuesOfCorrectType-test.js
```diff
@@ -68,6 +68,44 @@ describe('Validate: Variable default values of correct type', () => {
...
+      query WithDefaultValues(
+        $a: Int! = null,
+        $b: String! = null,
+        $c: ComplexInput = { requiredField: null, intField: null }
+      ) {
+        dog { name }
+      }
+    `, [
+      defaultForNonNullArg('a', 'Int!', 'Int', 3, 20),
+      badValue('a', 'Int!', 'null', 3, 20, [
```
leebyron: Interesting: I'm surprised to see two errors each for these cases, but I suppose it makes sense.

leebyron:
Yeah this is looking awesome so far, thanks for your continued work

src/utilities/astFromValue.js
langpavel: fixed

src/utilities/valueFromAST.js
langpavel: I learned from `eslint`s [consistent-return](http://eslint.org/docs/rules/consistent-return) rule that explicit is better than explicit.

Second reason is semantically big difference between `null` and `undefined`.

And last, every minifier will replace `return undefined;` with simple `return;`

src/utilities/valueFromAST.js
langpavel: Yes, exactly like I write before.. but I will add this comment here too! :-)

langpavel:
@leebyron Except of changing behavior of `astFromValue` (which I think is bad idea) we should got this! What do you think? :yum:

src/validation/__tests__/ArgumentsOfCorrectType-test.js
langpavel: Ok, I reworked this

leebyron:
Which astFromValue change are you referencing?

src/validation/__tests__/DefaultValuesOfCorrectType-test.js
langpavel: Actually I was thinking about creating one specific error, but it is really unnecessary. This two errors give more information. New error should be able carry exactly same information so I declined this idea.

langpavel:
> Which astFromValue change are you referencing?

`astFromValue` returns `null`. You proposed `undefined` but expected result is AST node so null is no AST node.

Result for `null` value is `({ kind: NULL }: NullValue)` and this function will never returns `undefined`

leebyron:
Yeah it's a minor detail since you're right that the function either returns an AST node or it does not do so. My concern is that returning `null` to represent no AST node has the potential to misunderstand as a representation of null, vs implicitly returning `undefined` with `return;` also represents no AST node but is far less likely to be misunderstood.

langpavel:
> Yeah it's a minor detail...

Exactly, but it is exported interface covered with tests so I'm for not changing it's behavior..

But if you wish I can change tests and `return;` from this function.. Then it should be mentioned in changelog too..

langpavel:
Ok,I will go sleep, deep night here :-) Please think about all this.. I know it is big thing so it probably needs more time for landing then I think just now.. But I'm very excited with this null support here..

I know, some people can hate me in future for introducing this null thing again to their lives, but I love possibility to say: Hey system, I don't know this value! :-) 

And in this point I must say: Hey @leebyron, thank you for null RFC and I'm really happy I can work on this thing with you! 

Good night

leebyron:
Thanks for all your hard work - let's land this and see how it goes

nodkz:
Can not wait for 0.8.0 anymore ;)
For me, a custom package build with `NULL` works perfectly.
Thanks @langpavel and @leebyron!

If somebody wants to test `NULL` in args and `flowtype` in their app, I built `graphql-js` package on commit https://github.com/graphql/graphql-js/commit/2b717f1405ce04eabd5a44df18b5b9b5cdd80fa3. Just run:
```
npm install https://www.dropbox.com/s/zrd3b00dmx3jhae/graphql-0.7.2.tgz\?dl\=1
```

PS. Yarn does not work with this link (((

langpavel:
Awesome! Thanks!

PR #553, opened 2016-11-01
Enforces input coercion rules.

leebyron:
Before this diff, bad input to arguments and variables was often ignored and replaced with `null` rather than rejected. Now that `null` has a semantic meaning, and thanks to some recent changes to the spec (https://github.com/facebook/graphql/pull/221) - changes are necessary in order to enforce the stricter coercion rules.

This diff does the following:

* Implements the CoerceArgumentValues as described in the spec.
* Implements the CoerceVariablesValues as described in the spec.
* Alters valueFromAST and coerceValue (dual functions) to strictly enforce coercion, returning `undefined` implicitly when they fail to do so. It also fixes issues where undefined returns were being ignored as items in a list or fields in an input object.

### DIFF
--- a/src/error/locatedError.js
+++ b/src/error/locatedError.js
@@ -32,9 +32,9 @@ export function locatedError(
     'An unknown error occurred.';
   return new GraphQLError(
     message,
-    originalError && originalError.nodes || nodes,
-    originalError && originalError.source,
-    originalError && originalError.positions,
+    originalError && (originalError: any).nodes || nodes,
+    originalError && (originalError: any).source,
+    originalError && (originalError: any).positions,
     path,
     originalError
   );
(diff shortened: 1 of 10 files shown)


leebyron:
Some of this ends up being follow-up to @langpavel's #544 as edges cases with support for null interrelates to the strict coercion rules.

langpavel:
Well done!

thebigredgeek:
@leebyron it appears that errors thrown from `parseValue` in scalar resolvers are now completely swallowed, with their messages being appended to the invalid input message.  This is a considerable breaking change, as many people are using scalars for form validation (myself included) and this makes it impossible to propagate back custom error text for presentation on the client.  Is there any way that the error could retain a reference to the `originalError` as is the case with most other errors?

leebyron:
Certainly retaining a reference to `originalError` is intended. I'd support a PR which fixed this issue

PR #561, opened 2016-11-09
Add path to formatError

leebyron:
While path was added to the serializable error, it was not yet added to the default formatError function.

Adding this is technically breaking since it begins to expose new information from your API.


stubailo:
How does this addition to graphql-js interact with the spec RFC?

Does that mean we're pretty comfortable with the RFC and I can start pointing people to this feature?

leebyron:
Yeah I like error path, I'm excited about enshrining it in spec :) - I'd say this change is mostly incremental for graphql-js since path was already a serializable field on the error object, but now it is preserved for formatError as well. This change likely should have been made a while ago.

I'd call this an experimental implementation of a near complete spec addition.

stubailo:
Sounds good! Really excited to have this, it will make handling errors much better by default.

PR #562, opened 2016-11-10
Use linked-list to represent response path.

leebyron:
This is a performance improvement to the execution path that removes the need for creating a new array to represent the response path for each resolved field in the response. Instead this creates a single-linked-list which is naturally persistently immutable.

This exposes a breaking change: if you relied on `resolve(_, _, _, { path }) {` then you'll find `path` in this linked list format instead of an array.

### DIFF
--- a/src/execution/__tests__/executor-test.js
+++ b/src/execution/__tests__/executor-test.js
@@ -225,7 +225,7 @@ describe('Execute: Handles basic execution tasks', () => {
     );
     expect(info.returnType).to.equal(GraphQLString);
     expect(info.parentType).to.equal(schema.getQueryType());
-    expect(info.path).to.deep.equal([ 'result' ]);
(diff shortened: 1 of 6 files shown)


stubailo:
Can we export the flatten path function?

leebyron:
Ya, good idea. Updated. Also renamed to something a little more descriptive.

dschafer:
> This is a performance improvement to the execution path that removes the need for creating a new array to represent the response path for each resolved field in the response. Instead this creates a single-linked-list which is naturally persistently immutable.

Do we have a benchmark to verify this ends up being a perf win? Seems like it should be, but it looks to add a bit of complexity as well, so it'd be good to quantify the tradeoff.

leebyron:
We don't have a perf benchmark set up yet - though having that would be valuable. In practice perf benchmarks are really hard to emulate outside of a production environment though. This particular change is related to one we did internally a long while ago to squeeze out a small win.

PR #569, opened 2016-11-15
differentiate between input values with no default and with default = null

jjergus:
- introspection should return null if there is no default value and the string "null" if there is a default value that is null
- buildClientSchema should translate defaultValue "null" to null (as it already does) and null to undefined (fixed here)


jjergus:
@leebyron @igorcanadi @langpavel

leebyron:
Sweeeet thank you

PR #570, opened 2016-11-15
Add assertXXXType helpers

josephsavona:
Adds helper functions for asserting that a type instance conforms to the various categories of types, e.g. `assertXXXType(type: ?GraphQLType): GraphQLXXXType`.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -36,6 +36,14 @@ export type GraphQLType =
   GraphQLList<any> |
   GraphQLNonNull<any>;
 
+export function assertType(type: mixed): GraphQLType {
+  invariant(
+    isType(type),
+    `GraphQL: Expected ${String(type)} to be a GraphQLType.`
+  );
+  return (type: any);
+}
+
 export function isType(type: mixed): boolean {
   return (
     type instanceof GraphQLScalarType ||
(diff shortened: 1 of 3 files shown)


src/type/definition.js
```diff
@@ -92,6 +110,15 @@ export type GraphQLOutputType =
     GraphQLList<GraphQLOutputType>
   >;
 
+export function assertOutputType(type: GraphQLType): GraphQLOutputType {
+  invariant(
+    isOutputType(type),
+    'GraphQL: Expected type `%s` to be a GraphQLOutputType',
+    type.name
```
leebyron: `String(type)` - not all types are named, like list and non-null.

graphql-js's `invariant` doesn't support printf format strings - so instead this should be:

```js
`Expected type "${String(type)}" to be an output type.`
```

src/type/definition.js
```diff
@@ -263,6 +317,10 @@ export class GraphQLScalarType {
     return parser ? parser(valueNode) : null;
   }
 
+  toJSON(): mixed {
```
leebyron: let's do toJSON in a separate PR. The right fix for that is probably to share the method rather than call through

josephsavona:
@leebyron addressed comments

src/type/definition.js
```diff
@@ -64,6 +72,14 @@ export type GraphQLInputType =
     GraphQLList<GraphQLInputType>
   >;
 
+export function assertInputType(type: ?GraphQLType): GraphQLInputType {
+  invariant(
+    type != null && isInputType(type),
```
leebyron: eqeqeq applies here.

These predicate functions accept `?Type` so you can fully lose the `type != null`

leebyron:
🎉

Issue #574, opened 2016-11-15
Better Flow type defs for field resolvers


leebyron:
When writing resolvers for GraphQL types in a flow-typed environment, the Flow types should aid as much as possible.

- [ ] Do not require casting through `(: any)` for typical behavior.
- [ ] Proper typing of `context`
- [ ] Proper typing of `source`
- [ ] Proper typing of return
- [ ] Proper typing of arguments

leebyron:
This was originally inspired by discussion in #554.

jamiehodge:
@leebyron is there any progress on this meta issue? It would really helpful to get end-to-end typing of the graph and the possibility of leveraging that in the client.

jamiehodge:
@leebyron are you still working on this issue? Is there something I or others could help with?

corydeppen:
Has there been any additional consideration for what it will take to move forward with the remainder of the typings for resolvers?

sibelius:
@gtkatakura is using conditional mapped types to improve types: https://github.com/gtkatakura/fullstack-challenge/commit/36cc1050768e29824f7c7114d22c131d43aefef9

sibelius:
typing for GraphQLScalarType https://github.com/graphql/graphql-js/pull/1522

saihaj:
we no longer support Flow https://github.com/graphql/graphql-js/issues/2860

PR #575, opened 2016-11-15
Add TContext to type resolvers

leebyron:
Part of #574

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -366,7 +366,7 @@ export type GraphQLScalarTypeConfig<TInternal, TExternal> = {
 export class GraphQLObjectType {
   name: string;
   description: ?string;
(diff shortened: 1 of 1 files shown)


src/type/definition.js
```diff
@@ -366,7 +366,7 @@ export type GraphQLScalarTypeConfig<TInternal, TExternal> = {
 export class GraphQLObjectType {
   name: string;
   description: ?string;
-  isTypeOf: ?GraphQLIsTypeOfFn;
+  isTypeOf: ?GraphQLIsTypeOfFn<*, *>;
```
intellix: Any reason this isn't `<TSource, TContext>`? I'm updating the DefinitelyTyped @types/graphql and couldn't find this documented anywhere in Flow. Would be a great help to point me in the right direction.

PR #576, opened 2016-11-15
RFC: Add type params to Scalars.

leebyron:
This incrementally adds to #574, however a broader solution is needed to create a mirroring between graphql type definitions and internal flow types.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -37,7 +37,7 @@ import type { GraphQLSchema } from './schema';
  * These are all of the possible kinds of types.
  */
 export type GraphQLType =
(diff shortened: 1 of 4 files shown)


jamiehodge:
@leebyron have you considered the approach taken by https://github.com/gcanti/flow-io or https://github.com/andreypopp/validated?

leebyron:
Rebased, but still a fairly nasty breaking change to land since Flow would require all references to `GraphQLScalarType` to change

sibelius:
it is related to this one https://github.com/graphql/graphql-js/pull/1487

PR #578, opened 2016-11-16
Implement toJSON and inspect on all types.

### LINKED ISSUE #568: Implement toJSON on types for debuggability
Inspired by a relay compiler trick, let's implement toJSON to be the same method as toString on type instances.

leebyron:
This adds toJSON() and inspect() methods on all types which shares the same function as toString(). This means that converting a GraphQL type to JSON or inspecting it with console.log (in Node) will produce something more meaningful than the implementation details of the type.

Closes #568

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -314,7 +314,15 @@ export class GraphQLScalarType {
   toString(): string {
     return this.name;
   }
+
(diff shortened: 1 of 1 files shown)


rjkmurray:
Wil, this benefit everyone

PR #589, opened 2016-11-17
Add isNamedType and assertNamedType helpers

jamesgorman2:
Completes the set of isXxxType/assertXxxType helpers. Useful for building type repositories and the like.


josephsavona:
Nice, was going to submit a PR to add these myself.

PR #614, opened 2016-12-04
Make executor throw if no document is provided

jmurzy:
Currently, `buildExecutionContext` throws a null reference error

`Cannot read property 'definitions' of null`

when `document` is undefined. Instead, we should throw a friendlier error message.

Also fixes apollostack/graphql-server#112

🍺


robzhu:
Sweet fix!

PR #615, opened 2016-12-06
Pass GraphQLResolveInfo object to resolve functions in defaultFieldResolver

jonbretman:
When using `buildSchema` and then passing a `root` object to `graphql` you don't get access to the `GraphQLResolveInfo` object which seems to be the only way to perform introspection on the query.


leebyron:
Sweet

PR #629, opened 2016-12-14
add public aliases for type configs

calebmer:
Having access to the type config is very useful for building type building utilities. Take for instance a function `implementNodeInterface(type: GraphQLObjectType)` which clones the `type._typeConfig` object and adds an interface, verifies an `isTypeOf` field exists, handles a special `nodeResolve` field (which can be used by a query level `node` field), and adds the global `id` field required by the interface.

As utilities around a GraphQL schema definition could benefit from the access to a type config I propose with this PR that we make the config field public by removing the underscore. From `_typeConfig` to `typeConfig`, `_scalarConfig` to `scalarConfig`, `_enumConfig` to `enumConfig`, and etc.

Currently this PR does not actually rename anything, instead it opts to add a getter alias which just returns the config with an underscore. This has the nice side effect of also disallowing assignment to the property as `type.typeConfig = null` would be illegal.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -328,6 +328,10 @@ export class GraphQLScalarType {
     const parser = this._scalarConfig.parseLiteral;
     return parser ? parser(valueNode) : null;
   }
+  
+  get scalarConfig(): GraphQLScalarTypeConfig<*, *> {
+    return this._scalarConfig;
+  }
 
   toString(): string {
     return this.name;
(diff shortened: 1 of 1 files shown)


leebyron:
Hey @calebmer can you be more specific about how you would like to use these? Right now they're intentionally private access since some of their specification could change in the future and I'd prefer to avoid any breaking changes around that, however if popular utilities are already using them - I'd take a practical approach to not breaking them.

Also some feedback - I'd suggest just renaming these all to `.config` just to simplify.

leebyron:
Closing this aging PR - ideally we will refactor these APIs in the future to enable this sort of utility.

Issue #398, opened 2016-06-02
Support asynchronous resolveType method for abstract types


taion:
This might be more of a "does this make sense" question than a feature request, but does it make sense for `resolveType` to optionally be async on abstract types, the way that field resolution can be optionally async?

Or would this be weird given that `resolveType` always runs, while any particular field may not be required for a given query?

leebyron:
Is there a use case you have in mind for an async resolveType? Given that it's called from within field execution, it's plausible that support for this could be added, but it would add some complication to the executor and we would definitely want to ensure there's some compelling use case.

taion:
This comes up for a sort of weird reason.

In a lot of places in my GraphQL server, instead of passing around objects, I pass around "object stubs", so e.g. just something like `{ id: '5' }` or something.

Then, in my `resolve` function, I use DataLoader to retrieve the object if needed. The reason I do this is that it allows me to efficiently handle nested queries that look like:

```
fragment on Widget {
  nestedObj1 {
    nestedObj2 {
      nestedObj3 {
        field
      }
    }
  }
}
```

Where these nested objects share the same local ID, but e.g. correspond to different endpoints on my backend, as it allows me to skip querying the endpoints for `nestedObj1`, `nestedObj2`, and `nestedObj3` (assuming that I expect them to exist).

I want to explore using GraphQL types to control data fetching – imagine something like having `ActiveTodo` and `CompletedTodo` types, where the concrete type is determined by the value of some fields for the backend. This ought to then allow me to do stuff like fetch a list of todos, but only pull down fields as relevant for displaying active or completed todos as needed.

However, for consistency with the "stub object" thing above, I may not in general have access to the status synchronously when attempting to resolve the type.

So anyway it's an intersection of two somewhat weird use cases, and comes from wanting to maintain consistency in doing this across a codebase.

taion:
To forestall the X-Y problem, I wrote up what I'm actually trying to accomplish in #405.

sonewman:
I definitely have a use case for this, and was going to create an issue if the wasn't one.

I work for one of the largest grocery retailers in Europe and am currently leading development of a new gateway API for our online grocery shopping.

This API is purely being driven by Graphql.

In our use case we have multiple different backend services and new ones being developed to replace them as well as different business units with separate API services.

To combat this we are heavily using interfaces to resolve different mapping types, as well as dynamically resolving resources to fetch our data.

[...]

PR #631, opened 2016-12-15
Allow `isTypeOf` and `resolveType` to return a Promise

dherault:
Hello, 

The idea behind this PR is to allow `GraphQLObjectType#isTypeOf`, `GraphQLInterfaceType#resolveType` and `GraphQLUnionType#resolveType` to return a Promise.

The PR consists of:
- Changes to `execute.js` to support the new promise feature.
- New tests: Promise.resolve duplicates of the existing ones about `isTypeOf` and `resolveType` + new ones to test the cases where the promise is rejected.

This relates to issue #398. Use case is the following: 
For some projects fetching data can be expensive. To prevent that, the 'source' passed to resolveFns can be thought as an ID string referencing remote data, so the fields resolve only what is needed based on that ID. It works fine, except when abstract types require a sync resolving of the type behind an ID string.
The use case at my company: We create our GraphQL schema based on our RDF ontology, query with SPARQL, etc... (see this almost open-sourced project: [semantic-graphql](https://www.npmjs.com/package/semantic-graphql))

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -949,47 +949,55 @@ function completeAbstractValue(
     const runtimeTypePromise: Promise<GraphQLObjectType | string> = 
       (runtimeType: any);
     return runtimeTypePromise.then(resolvedRuntimeType =>
-      validateRuntimeTypeAndCompleteObjectValue(
+      completeObjectValue(
         exeContext,
         returnType,
         fieldNodes,
         info,
         path,
-        ensureType(exeContext.schema, resolvedRuntimeType),
+        ensureValidRuntimeType(
+          resolvedRuntimeType,
+          exeContext, 
+          returnType, 
+          fieldNodes,
+          info,
+          result
+        ),
         result
       )
     );
   }
 
-  return validateRuntimeTypeAndCompleteObjectValue(
+  return completeObjectValue(
     exeContext,
     returnType,
     fieldNodes,
     info,
     path,
-    ensureType(exeContext.schema, runtimeType),
+    ensureValidRuntimeType(
+      resolvedRuntimeType,
+      exeContext,
+      returnType, 
+      fieldNodes,
+      info,
+      result
+    ),
     result
   );
 }
 
-function ensureType(
-  schema: GraphQLSchema, 
-  typeOrName: string | GraphQLObjectType
-): GraphQLObjectType {
-  return typeof typeOrName === 'string' ? 
-    schema.getType(typeOrName) : 
-    typeOrName;
-}
-
-function validateRuntimeTypeAndCompleteObjectValue(
+function ensureValidRuntimeType(
+  runtimeTypeOrName: string | GraphQLObjectType | void
   exeContext: ExecutionContext,
   returnType: GraphQLAbstractType,
   fieldNodes: Array<FieldNode>,
   info: GraphQLResolveInfo,
-  path: ResponsePath,
-  runtimeType: GraphQLObjectType,
   result: mixed
-): mixed {
+): GraphQLObjectType {
+  const runtimeType = typeof runtimeTypeOrName === 'string' ? 
+    exeContext.schema.getType(runtimeTypeOrName) : 
+    runtimeTypeOrName;
+
   if (!(runtimeType instanceof GraphQLObjectType)) {
     throw new GraphQLError(
       `Abstract type ${returnType.name} must resolve to an Object type at ` +
@@ -1007,14 +1015,7 @@ function validateRuntimeTypeAndCompleteObjectValue(
     );
   }
 
-  return completeObjectValue(
-    exeContext,
-    runtimeType,
-    fieldNodes,
-    info,
-    path,
-    result
-  );
+  return runtimeType;
 }
 
 /**
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -941,30 +941,24 @@ function completeAbstractValue(
   result: mixed
 ): mixed {
   const runtimeType = returnType.resolveType ?
-    returnType.resolveType(
-      result,
-      exeContext.contextValue,
-      info
-    ) :
-    defaultResolveTypeFn(
-      result,
-      exeContext.contextValue,
-      info,
-      returnType
-    );
+    returnType.resolveType(result, exeContext.contextValue, info) :
+    defaultResolveTypeFn(result, exeContext.contextValue, info, returnType);
 
   if (isThenable(runtimeType)) {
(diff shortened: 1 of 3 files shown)


src/execution/execute.js
```diff
@@ -940,40 +940,53 @@ function completeAbstractValue(
...
 ): mixed {
-  let runtimeType = returnType.resolveType ?
-    returnType.resolveType(result, exeContext.contextValue, info) :
-    defaultResolveTypeFn(result, exeContext.contextValue, info, returnType);
+  const runtimeTypePromise = returnType.resolveType ?
+    Promise.resolve(returnType.resolveType(
+      result,
+      exeContext.contextValue,
+      info
+    )) :
```
dherault: You're right, I'll see about it

dherault: It's done. It allows fully synchronous abstract type resolution like before.
PS: I've rebased my PR so this diff is outdated

leebyron:
Thanks so much for digging into this!

I'll do a more thorough review, thanks for your patience :)

src/execution/execute.js
```diff
@@ -990,51 +1076,37 @@ function completeObjectValue(
...
-  for (let i = 0; i < fieldNodes.length; i++) {
-    const selectionSet = fieldNodes[i].selectionSet;
-    if (selectionSet) {
-      subFieldNodes = collectFields(
-        exeContext,
-        returnType,
-        selectionSet,
-        subFieldNodes,
-        visitedFragmentNames
+  return Promise.resolve(returnType.isTypeOf ?
```
leebyron: Similarly here - can we avoid the per-field creation for a Promise for the more typical case of `returnType.isTypeOf` returning synchronously?

src/execution/execute.js
```diff
@@ -977,6 +1022,47 @@ function completeAbstractValue(
 }
 
 /**
+ * If a resolveType function is not given, then a default resolve behavior is
+ * used which tests each possible type for the abstract type by calling
+ * isTypeOf for the object being coerced, returning the first type that matches.
+ */
+function resolveFirstValidType(
```
leebyron: Suggestion: move this to the bottom of the file for less entropy in the diff.

Also, instead of having a default value for arguments, you could keep the `defaultResolveTypeFn` function, but just have it collect the possibleTypes and then call this function.

src/execution/execute.js
```diff
@@ -977,6 +1022,47 @@ function completeAbstractValue(
...
+  value: mixed,
+  context: mixed,
+  info: GraphQLResolveInfo,
+  possibleTypes: Array<GraphQLObjectType>,
+  i: number = 0,
```
leebyron: I'm a bit concerned about this function calling itself recursively. There are cases where a great many types can be iterated through here and the recursion could cause significant overhead. Also, if the matching type happens to be last in a long list of possible types, then it would have to wait for each promise to resolve sequentially before a match could be found.

I suggest calling the `isTypeOf` functions synchronously such that any promises generated run in parallel rather than in series. That way, if any synchronously return `true`, then the function may return synchronously, even if some of the `isTypeOf` functions return Promises. That also preserves the existing behavior. However if some responses were `thenable` then you could use `Promise.all` to resolve them in parallel and then iterate through the results to determine a match.

dherault: Done. Here are some thoughts on the new implementation.
- The ordering of types is not preserved when mixing sync and async `isTypeOf`:
```js
const Dog = new GraphQLObjectType({
  isTypeOf: () => Promise.resolve(true),
  /* ... */
});

const Cat = new GraphQLObjectType({
  isTypeOf: () => true,
  /* ... */
});

const PetType = new GraphQLUnionType({
  types: [Dog, Cat]
  /* ... */
});
```
In a fully-sync schema, Dog would be picked as the correct type for any query on Pet, but in the proposed implementation, we resolve the sync ones first, so Cat is picked. In a fully async schema the ordering is preserved. (Keeping that ordering was what I was trying to achieve using the outdated recursive function)
- Error handling changed a bit. See [this diff](https://github.com/graphql/graphql-js/pull/631/commits/ae5295a2d3b3b19aa3e45129ac1a18b163ca9aea#diff-ad7767caa77907dedfe8013eaad8ff86):
```js
const DogType = new GraphQLObjectType({
  isTypeOf: source => Promise.resolve(source instanceof Dog),
  /* ... */
});

const CatType = new GraphQLObjectType({
  isTypeOf: () => Promise.reject('Oh no!'),
  /* ... */
});

const PetType = new GraphQLUnionType({
  types: [Dog, Cat]
  /* ... */
});
```
When querying for a Dog everything should be alright but since we run `Promise.all` on all async `isTypeOf` functions it throws the Cat error. Maybe that is not an issue since rejecting in `isTypeOf` would never be the developer's intent.

src/execution/execute.js
```diff
@@ -1026,15 +1122,34 @@ function defaultResolveTypeFn(
...
     }
   }
+
+  if (promisedIsTypeOfResults.length) {
+    return Promise.all(promisedIsTypeOfResults).then(isTypeOfResults => (
+      promisedIsTypeOfResultTypes[
+        isTypeOfResults.findIndex(isTypeOf => isTypeOf)
+      ]
+    ));
+  }
```
sonewman: I have a quick question, if one of these functions resolved true, but another one returns a rejected promise then the this will completely fail. Is this the expected behaviour?

While I think resolving all in parallel is a good idea, I was just thinking if the typeOf methods involve individual  database lookups or something similar then those kinds of implementations would be restricted to using the resolveType option to control synchronicity.

dherault: > I have a quick question, if one of these functions resolved true, but another one returns a rejected promise then the this will completely fail. Is this the expected behaviour?

I agree the new behavior is not optimal (see previous comments). Any thoughts on how to handle it are welcome.

> While I think resolving all in parallel is a good idea, I was just thinking if the typeOf methods involve individual database lookups or something similar then those kinds of implementations would be restricted to using the resolveType option to control synchronicity.

Yes, but in a way, resolving types in the order `info.schema.getPossibleTypes(abstractType)` gives us is also arbitrary. So a "first-to-resolve" strategy would not be that bad.

leebyron: Re: the order of possible types: At Facebook we actually have a version of this resolver that runs in development environments (due to performance concerns) that runs through all possible types and makes an assertion that exactly one of them is the correct type.

leebyron:
This is looking awesome, thanks! I'll do a bit more collaborative edits and then lets get this landed.

PR #632, opened 2016-12-16
Fix Regex for detecting hyphens in a word

Ashraf-Ali-aa:
When a string contains a hyphen the regex fail to match the string i.e "X-SitTEST-Proposition"

### DIFF
--- a/src/utilities/assertValidName.js
+++ b/src/utilities/assertValidName.js
@@ -8,7 +8,7 @@
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
 
-const NAME_RX = /^[_a-zA-Z][_a-zA-Z-0-9]*$/;
+const NAME_RX = /^[_a-zA-Z][_a-zA-Z0-9-]*$/;
 
 /**
  * Upholds the spec rules about naming.
@@ -30,7 +30,7 @@ export function assertValidName(
   }
   if (!NAME_RX.test(name)) {
     throw new Error(
-      `Names must match /^[_a-zA-Z][_a-zA-Z-0-9]*$/ but "${name}" does not.`
+      `Names must match /^[_a-zA-Z][_a-zA-Z0-9-]*$/ but "${name}" does not.`
     );
   }
 }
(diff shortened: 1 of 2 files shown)


src/utilities/assertValidName.js
```diff
@@ -8,7 +8,7 @@
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
 
-const NAME_RX = /^[_a-zA-Z][_a-zA-Z0-9]*$/;
+const NAME_RX = /^[_a-zA-Z][_a-zA-Z-0-9]*$/;
```
wincent: This is a pretty treacherous way to allow a hyphen, buried in the middle like that. Would be better to put the hyphen at the very beginning or end of the range; eg:

```
const NAME_RX = /^[_a-zA-Z][_a-zA-Z0-9-]*$/;
```

Ashraf-Ali-aa:
@wincent I using a Rest API that has custom HTTP headers which have hyphens in the word, am currently using https://github.com/yarax/swagger-to-graphql to autogenerate GraphQL but due to GraphQL.js is not validating API the tool fails to create the GraphQL schema. 

I did raise an issue on `swagger-to-graphql` https://github.com/yarax/swagger-to-graphql/issues/4

leebyron:
As @wincent mentioned, this change would break compliance with [the spec](http://facebook.github.io/graphql/#sec-Names). Dashes are intentionally not supported as they're not valid characters in identifiers in most modern languages, and GraphQL is designed to be useful alongside as many different environments as possible.

I would suggest creating an issue or PR against swagger-to-graphql in order to support swagger fields with characters outside of those supported by GraphQL field names.

jquense:
Is there some place to discuss this restriction, or propose something more flexible? The interoperability story makes alot of sense from a platform implementation standpoint. However, it's severely limiting when trying to do interoperability with any real life existing data schemas (especially json). There is no room to even replace invalid characters, as _ is the only none semantic char available.

wincent:
@jquense: best place to discuss changes to the spec would be the repo containing the spec: https://github.com/facebook/graphql

PR #644, opened 2016-12-21
chore(circ-dep): fixed circular dependancies

Urigo:
Fix circular dependency that fails the code after compiling the library with Rollup (because of https://github.com/rollup/rollup-plugin-commonjs/issues/105)


leebyron:
Sweet, thanks

Issue #647, opened 2016-12-25
Proposal: refactor for RxJS compatible


xareelee:
I'm starting to build another version of GraphQL library using [RxJS](https://github.com/ReactiveX/rxjs) instead of Promise for reactive programming. 

I won't add Rx to this library, just build another library and reuse most of code this library provides (with module dependency).

To minimize the efforts of maintenance, I need to refactor [this original graphql library](https://github.com/facebook/graphql) to extract Promise-associated code into separated files, and export some functions to make them importable for reuse. It won't change any behavior which this library provides. Just separate the concern of event handler (Promise, Rx, and so on) from the GraphQL design.

At a glance, I think I just need to refactor the file `graphql/execution/execute.js` mainly. The goal is make Promise-based functions become callback-based (Promise-free), and turn them into Promise-based in separated files (and also I can turn them into Rx-based easily).

DxCx:
Hi,

Already done by me and rejected by @leebyron.
See #502.

Ill probably release in the future an addon libraray which will hook graphql library and add observable support.

xareelee:
@DxCx and @leebyron,

It's different from #502 and my approach. I agree with @leebyron's comment: _"not taking a dependency on a large library"_.

I'm not planing to introduce Rx or any other dependency into graphql-js. I'll just refactor existing functions to use a specific callback design, and turn them into Promise. 

I just wrote an article to explain what it is and how I'll do. Please refer to my article: [A better async/reactive function design in JS (or any other languages) — the Universal Currying Callback (UCC) Design](https://gist.github.com/xareelee/e85c8b2134ff1805ab1ab2f1c8a037ce)

Simply put, I'll make the functions become callback-based and turn them back to Promise in a separated file for decoupling. The callback-based functions would be easily turned into Promise, Rx or any other async handler system.

```js
// A universal currying callback (UCC) function
function graphql_$ = 
  (schema, requestString, rootValue, contextValue, variableValues, operationName) => 
  (...callbacks /* please refer to the article */) => {
  ...
};
// You can use it directly
graphql_$(schema, query)(...callbacks);

// Or use it as Promise flovar (We'll still provide `graphql()` as Promise-based)
const graphql = promisify(graphql_$);
graphql(schema, query).then(...);

// Or Rx flavor
const graphqlRx = rxify(graphql_$);
graphqlRx(schema, query).subscribe(...);
```

It doesn't break any design or introduce dependency. Just refactor existing functions to UCC design and make UCC functions be public.

[...]

PR #649, opened 2016-12-26
refactor(graphql): extract graphql() logic into graphql$UCCF to suppo…

xareelee:
**This PR is for discussion only, not for real merge.**

This PR demonstrates how to convert a Promise-based function to a reactive-callback-based function to support all kinds of async handling system without any new dependency.

We use `promisify()` to turn it back to a Promise-based function for the default flavor. 

It currently is not **reactive** yet due to the upstream functions are not reactive (as Promise). We need to refactor all upstream functions to be reactive-callback-based.

More discussions on #647.

## TODO

- [x] make first refactor demo and pass the tests and lint.
- [ ] add strict Flow type and pass the type check.
- [ ] refactor the upstream functions of graphql to be reactive-callback-based (UCC functions) until it fully supports reactive programming.

For more info on UCC functions, please refer to [this article](https://gist.github.com/xareelee/e85c8b2134ff1805ab1ab2f1c8a037ce).

### DIFF
--- a/src/graphql.js
+++ b/src/graphql.js
@@ -13,8 +13,14 @@ import { parse } from './language/parser';
 import { validate } from './validation/validate';
 import { execute } from './execution/execute';
 import type { GraphQLSchema } from './type/schema';
-import type { ExecutionResult } from './execution/execute';
+// import type { ExecutionResult } from './execution/execute';
 
+type onNextBlock = (next: mixed) => void;
+type onErrorBlock = (error: Error, isContinued: ?boolean) => void;
+type onEndBlock = (info: ?mixed) => void;
+
+type UniversalCallbackFunction =
+  (onNext: onNextBlock, onError: onErrorBlock, onEnd: onEndBlock) => void;
 
 /**
  * This is the primary entry point function for fulfilling GraphQL operations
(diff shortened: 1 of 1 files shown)


leebyron:
Can you clarify the purpose of this PR? If you would like to use a callback interface with GraphQL, you can wrap the existing function that returns a Promise by proving your callback function to the "then" method. 

I see that this adds complexity, but I don't see the value

xareelee:
Hi @leebyron,

This is unfinished PR. Currently I just publish this PR for discussion.

Using callback-based functions to support reactive events without any 3rd-party dependency (e.g. Rx) is the value.

Promise-based functions can not be reactive due to they only can send one value for their lifetime. We can not use Promises as the reactive event sources to send the value continuously.

Converting the event handling of this library from Promise to reactive-callback system would be friendly to the FRP world. Users could then choose how to use this library — Promise-based or FRP-based (any flavor).

My goal is to make this library be reactive without any 3rd-party dependency, so we can send reactive events to **resolvers** and make graphql-js subscribable to receive data changes.

If adopting the reactive-callback design, we don't need to maintain another reactive version of graphql-js. When finishing this branch, we can make graphql-js return an `Rx.Observable` (or using any other FRP library):

```js
// Promise flavor
const graphqlPromise = promisify(graphql$UCCF);
graphqlPromise(schema, query).then(...);

// Rx flavor
const graphqlRx = rxify(graphql$UCCF);
grapqlRx(schema, query).subscribe(...);
```

I'm building a front-end data flow framework based on graphql-js. I need it to be reactive.

leebyron:
Thanks that helps me understand. 

Please refer to #502 where this was discussed at greater length. The goal of graphql-js is to be a production grade spec compliant reference implementation with readable and reusable source. As I mentioned in that issue, I'm very excited about the explorations in different ways to implement live systems in GraphQL, however I do not think it's appropriate to do that at the expense of making this particular library worse at its intended goals. I would encourage you to explore this kind of experimental off-spec behavior in a fork or additional library so that it can be fully explored without impact on this reference implementation

PR #656, opened 2017-01-05
support customType for buildSchema

uxnow:
let us can add custom type
```javascript
  const Url = new GraphQLScalarType({
    name: 'Url',
    serialize: String,
    parseValue: String,
    parseLiteral(ast) {
      return ast.kind === Kind.STRING || ast.kind === Kind.INT ?
        ast.value :
        null;
    }
  });

    const schema = buildASTSchema(parse(`
      schema { query: Query }
      type Query {
        str(url: Url): Url
      }
    `), {Url});

```

### DIFF
--- a/src/utilities/__tests__/buildASTSchema-test.js
+++ b/src/utilities/__tests__/buildASTSchema-test.js
@@ -17,6 +17,8 @@ import {
   GraphQLSkipDirective,
   GraphQLIncludeDirective,
   GraphQLDeprecatedDirective,
+  Kind,
+  GraphQLScalarType,
 } from '../../';
 
 /**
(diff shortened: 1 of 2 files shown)


DxCx:
it is already supported (not sure if documented and how much) using makeExecutableSchema and set Url as a resolver,
also note that you need to define the actual scalar type for it (scalar Url - on the typeDefs)

stubailo:
By the way, this exact functionality exists in `graphql-tools`: http://dev.apollodata.com/tools/graphql-tools/scalars.html

Like so:

```
import { buildExecutableSchema } from 'graphql-tools';

const Url = new GraphQLScalarType(...);

const schema = buildExecutableSchema({
  typeDefs: `... schema string here ...`,
  resolvers: { Url },
});
```

PR #669, opened 2017-01-17
Add resolver validation, check if value is a function

robzhu:
If a field's resolver is set to a constant value like "{}" we get a runtime error, "resolveFn is not a function". We can easily detect this case when we validate fields during schema creation. This PR adds a new validation rule to check that the resolve definition is a function.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -508,8 +508,8 @@ function defineFieldMap<TSource, TContext>(
     );
     invariant(
       isValidResolver(field.resolve),
-      `${type.name}.${fieldName} field resolver must be function or null/` +
-      `undefined, but got: ${String(field.resolve)}.`
+      `${type.name}.${fieldName} field resolver must be a function if ` +
+      `provided, but got: ${String(field.resolve)}.`
     );
     const argsConfig = fieldConfig.args;
     if (!argsConfig) {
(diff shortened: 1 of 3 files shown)


src/type/definition.js
```diff
@@ -506,6 +506,11 @@ function defineFieldMap<TSource, TContext>(
       `${type.name}.${fieldName} field type must be Output Type but ` +
       `got: ${String(field.type)}.`
     );
+    invariant(
+      isValidResolver(field.resolve),
```
leebyron: you could probably just inline this as `!field.resolve || typeof field.resolve === 'function'`

src/type/definition.js
```diff
@@ -506,6 +506,11 @@ function defineFieldMap<TSource, TContext>(
...
+      `${type.name}.${fieldName} field resolver must be function or null/` +
```
leebyron: how about `field resolve must be a function if provided`

src/type/definition.js
robzhu: "!field.resolve" means we will accept any "falsey" value such as 0 or the empty string. Is that what we want?

leebyron: You're right, that it would let in falsey values. We could instead use `field.resolve == null || typeof field.resolve === 'function'` to be more specific

robzhu: I tried that, but the linter treats == as an error. I can update the linter to allow it if you feel strongly that == is more expressive.

leebyron: We should probably enable `== null` in the linter. I think eslint now has the ability to whitelist that particular kind of usage. It's equivalent to the longer form.

PR #675, opened 2017-01-19
Validate Input Types do not define resolvers

robzhu:
GraphQLInputObjectTypes are confusing to new users of GraphQL. One general distinction between GraphQLObjectType and GraphQLInputObjectType is the use of resolvers: I can't think of a valid case for them to be present on GraphQLInputObjectType. Let's validate this during schema construction and throw a helpful error.


leebyron:
Sweet

PR #692, opened 2017-01-26
Convert error to warning for non-compliant use of __

### LINKED ISSUE #690: __ strictness breaks postgraphql introspection
With the publication of [0.9.0](https://github.com/graphql/graphql-js/releases/tag/v0.9.0)  which includes #600, `postgraphql` [broke](https://github.com/calebmer/postgraphql/pull/319). Postgraphql implemented `__id` in advance in order to support the customers use case, there is a pending request open [here](https://github.com/facebook/relay/issues/1061) that refers to the validity of it (even though it is not yet implemented). Might it be good to add an exception case if `__` is used for `__id: GraphQLID` in order to not-break that case?

leebyron:
This unbreaks various uses of graphql-js which were defining fields with __ while retaining some form of reporting warning for non-compliant use.

Fixes #690

### DIFF
--- a/src/type/__tests__/validation-test.js
+++ b/src/type/__tests__/validation-test.js
@@ -341,15 +341,26 @@ describe('Type System: Objects must have fields', () => {
     );
   });
 
-  it('rejects an Object type with reserved named fields', () => {
-    expect(
-      () => schemaWithFieldType(new GraphQLObjectType({
+  it('warns about an Object type with reserved named fields', () => {
+    /* eslint-disable no-console */
+    const realConsoleError = console.error;
+    const calls = [];
+    console.error = function () {
+      calls.push(Array.prototype.slice.call(arguments));
+    };
+    try {
+      schemaWithFieldType(new GraphQLObjectType({
         name: 'SomeObject',
         fields: { __notPartOfIntrospection: { type: GraphQLString } }
-      }))
-    ).to.throw(
-      'Name "__notPartOfIntrospection" must not begin with "__", which is reserved by GraphQL introspection.'
-    );
+      }));
+
+      expect(calls[0][0]).contains(
+        'Name "__notPartOfIntrospection" must not begin with "__", which is reserved by GraphQL introspection.'
+      );
+    } finally {
+      console.error = realConsoleError;
+    }
+    /* eslint-enable no-console */
   });
 
   it('rejects an Object type with incorrectly typed fields', () => {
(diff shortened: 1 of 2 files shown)


src/utilities/assertValidName.js
```diff
@@ -22,11 +25,20 @@ export function assertValidName(
...
   }
-  if (!isIntrospection && name.slice(0, 2) === '__') {
-    throw new Error(
-      `Name "${name}" must not begin with "__", which is reserved by ` +
-      'GraphQL introspection.'
-    );
+  if (!isIntrospection && name.slice(0, 2) === '__' && !hasWarnedAboutDunder) {
+    hasWarnedAboutDunder = true;
+    /* eslint-disable no-console */
+    if (console && console.error) {
```
wincent: When would `console` ever be defined but falsey? Just being paranoid?

src/utilities/assertValidName.js
```diff
@@ -22,11 +25,20 @@ export function assertValidName(
...
+      const stack = new Error().stack;
+      const stackOnly = stack ?
+        '\n' + stack.split('\n').slice(1).join('\n') :
+        '';
+      console.error(
+        `GraphQL: Name "${name}" must not begin with "__", which is ` +
+        'reserved by GraphQL introspection.' + stackOnly
```
wincent: "reserved for" better than "reserved by"?

src/type/__tests__/validation-test.js
```diff
@@ -341,15 +341,27 @@ describe('Type System: Objects must have fields', () => {
...
 
-  it('rejects an Object type with reserved named fields', () => {
-    expect(
-      () => schemaWithFieldType(new GraphQLObjectType({
+  it('warns when an Object type with reserved named fields', () => {
+    /* eslint-disable no-console */
+    const prevConsoleError = console.error;
+    const calls = [];
+    console.error = function () {
+      calls.push(Array.prototype.slice.call(arguments));
```
wincent:
```
calls.push(Array.from(arguments));
```

Probably works in this environment too.

src/utilities/assertValidName.js
leebyron: Just being paranoid. I'd rather not have a cascading failure because someone's running graphql-js in some environment that either has no `console` or has `console` but doesn't have `console.error`.

wincent: Would `typeof console !== 'undefined'` be even more robust then?

leebyron: I don't know what environment might have `console` set to `null` or `false` or something silly like that, but I think this check should robust against whatever value console might have.

Technically `typeof console === 'object'` would be more specific, but it wouldn't yield a different result

PR #695, opened 2017-01-27
Use flow %checks to reduce occurance of :any

leebyron:
This uses a pre-release flow feature to provide predicate functions with more type checking power. It also breaks apart some functions into multiple declare function statements for cases where more specific input types result in more specific output types.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -72,12 +72,13 @@ export type GraphQLInputType =
     GraphQLList<GraphQLInputType>
   >;
 
-export function isInputType(type: ?GraphQLType): boolean {
-  const namedType = getNamedType(type);
+export function isInputType(type: ?GraphQLType): boolean %checks {
   return (
-    namedType instanceof GraphQLScalarType ||
-    namedType instanceof GraphQLEnumType ||
-    namedType instanceof GraphQLInputObjectType
+    type instanceof GraphQLScalarType ||
+    type instanceof GraphQLEnumType ||
+    type instanceof GraphQLInputObjectType ||
+    type instanceof GraphQLNonNull && isInputType(type.ofType) ||
+    type instanceof GraphQLList && isInputType(type.ofType)
   );
 }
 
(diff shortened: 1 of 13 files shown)


src/type/definition.js
```diff
@@ -134,11 +136,10 @@ export type GraphQLLeafType =
   GraphQLScalarType |
   GraphQLEnumType;
 
-export function isLeafType(type: ?GraphQLType): boolean {
-  const namedType = getNamedType(type);
```
leebyron: Note: This was ambiguous before. Other isCompositeType / isAbstractType checks below do not include named type unwrapping, and it did not match the type definition above. Only one callsite actually relied on named type unwrapping, so moved this check there.

leebyron:
I'm not super pleased about the eslint and babel workarounds required to get this all to work. I just wanted to get this up to show how it's useful. Perhaps we go after these tools to improve compatibility first?

src/type/definition.js
leebyron: Pulled out into #697

src/validation/rules/FragmentsOnCompositeTypes.js
```diff
@@ -39,16 +40,18 @@ export function fragmentOnNonCompositeErrorMessage(
...
+          context.reportError(new GraphQLError(
+            inlineFragmentOnNonCompositeErrorMessage(print(node.typeCondition)),
+            [ node.typeCondition ]
+          ));
+        }
       }
     },
     FragmentDefinition(node) {
-      const type = context.getType();
+      const type = typeFromAST(context.getSchema(), node.typeCondition);
```
leebyron: Moved all the changes in this file to #698

leebyron:
Might wait for https://github.com/babel/babylon/pull/103 before landing this to avoid the flow comments.

kassens:
@leebyron this could be merged now. (btw, no need for comments around %checks)

PR #696, opened 2017-01-27
Quotes preserved `if` literal when used in properties

asiandrummer:
Related to 080e281f97470a309be70e5edbc9bf5107bc4e21, but for `if` case.


wincent:
Why is this needed? Isn't `transform-es3-property-literals` supposed to do this for us?

asiandrummer:
ah, true. Closing - sorry!

PR #699, opened 2017-01-27
RFC: More flow-friendly thenable checking.

leebyron:
This replaces the isThenable boolean check with a "getIfPromise" function. It's essentially the same, but rather than returning boolean, it's the identify function when it would have returned true, and returns void otherwise. That lets us give it a better type signature that Flow can do more with.

Open to feedback on this one, and open to bikeshedding on the name of this new function.

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -341,8 +341,9 @@ function executeFieldsSerially(
       if (result === undefined) {
         return results;
       }
-      if (isThenable(result)) {
-        return ((result: any): Promise<*>).then(resolvedResult => {
+      const promise = getPromise(result);
+      if (promise) {
+        return promise.then(resolvedResult => {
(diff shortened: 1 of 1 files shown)


wincent:
I like that this consolidates the `any` cast to a single location. The `getIfPromise` name seems all right. Other random ideas that cross my mind are `maybePromise`, or a `safePromise` wrapper which you would use like this:

```
safePromise(maybeAPromise).then(() => {
  // Will only be called if `maybeAPromise` had a `then` method...
});
```

(Costs: extra function creation, and attendant cost of capturing a closure.)

leebyron:
Yeah I'd like to avoid the costs of doing so, plus in a lot of those cases we're actually returning early if it's a promise, so that might not work.

I wish javascript supported if-let conditions, that would drop one LOC for each use.

PR #701, opened 2017-01-31
`findBreakingChanges` handles args

joelgriffith:
First pass at implementing `arg` handling in the `findBreakingChanges` utility.

Considered breaking:
- arg removal
- arg type change
- arg defaultValue change

Considered non-breaking:
- arg going from non-nullable to nullable (as long as the underlying type is the same).

Open to suggestions, feedback.


src/utilities/findBreakingChanges.js
```diff
@@ -105,6 +110,80 @@ export function findTypesThatChangedKind(
   return breakingChanges;
 }
 
+/**
+ * Given two schemas, returns an Array containing descriptions of any breaking
+ * changes in the newSchema related to removal of an arg.
```
wincent: Seems that "removal of an arg" is only a partial description of what this does. I'd suggest something like:

```
/**
 * Given two schemas, returns an Array containing descriptions of any breaking
 * changes in the newSchema related to arguments (such as removal or change
 * of type of an argument, or a change in an argument's default value).
 */
```

src/utilities/findBreakingChanges.js
```diff
@@ -105,6 +110,80 @@ export function findTypesThatChangedKind(
...
+              `${oldArgDef.name} was removed`,
+          });
+
+        // Arg changed type in a breaking way
+        } else if (
+          oldArgDef.type !== newArgDef.type &&
+          getNullableType(oldArgDef.type) !== newArgDef.type
+        ) {
+          breakingArgChanges.push({
+            type: BreakingChangeType.ARG_REMOVED,
```
wincent: Seems like this should be `ARG_CHANGED_KIND`, not `ARG_REMOVED`, and the tests will need to be updated too.

joelgriffith:
@wincent I've updated this PR with the changes you've requested. Would love to get this merged as a consumer of `findBreakingChanges`

wincent:
Changes look good, although I am still not sure whether a change in a default value should be considered a breaking change. Let me ask for a second opinion internally. Thanks for your patience!

joelgriffith:
Thanks! My concern is that consumers could be relying on a defaultValue that works one way, and then when it changes something could break (more specifically with InputTypes).

wincent:
Yep, I get that it is a behavior change, but it doesn't meet the definition of "breaking change" [from the spec](https://facebook.github.io/graphql/#sec-Validation), which says:

> Any change that can cause a previously valid request to become invalid is considered a `breaking change`.

wincent:
Ok, second opinion obtained! My reading of the spec is correct, so we should pull the value-change check from this PR. You're welcome to put that into a separate `findDangerousChanges` check though if you'd like.

joelgriffith:
Alright, I've changed a few implementation details of `findBreakingArgChanges` to just `findArgChanges` to try and re-use that funcitonality. There's now a top-level `findDangerousChanges` as well that consumers can use to find regressive-type changes even thought queries still work.

I'd be open to feedback on these as my sense is that only the `defaultValue` change in args is exposed in `findDangerousChanges`, though I can see things like removing `deprecated` fields being considered dangerous even though they're planned for.

joelgriffith:
@wincent, hopefully this will be the last pass for you! Thanks again for your time and feedback

joelgriffith:
Hate to pester, but I'm _craving_ this feature!

wincent:
Looking at it.

wincent:
Thanks for your work on this, @joelgriffith. I merged this to `master` with one tiny (comment) fix on top (578c76ce). Going to cut a release shortly that includes this change.

joelgriffith:
Thanks @wincent, been a pleasure!

jbreckel:
When is the next release planned?

wincent:
@jbreckel: no explicit plans, but it has been a couple months since we cut a release so we should probably do that again. I'll make an issue to remind us to get to this shortly.

Issue #587, opened 2016-11-17
Use parent in GraphQLUnionType's resolveType function


chentsulin:
Think of there are some actions with its type and descriptor:

```js
const action1 = {
  type: 'SEND_MAIL',
  descriptor: {
    email: 'test@t.com',
  }, 
};

const action2 = {
  type: 'SEND_SMS',
  descriptor: {
    phone: '(+886)09xxxxxxxx',
  }, 
};
```

Typically, we will use switch or if else condition to recognize action type then do something differently. Here is the `Action` Type:

```js
const Action = new GraphQLObjectType({
  name: 'Action',
  fields: () => ({
    type: {
        type: new GraphQLNonNull(ActionType), // GraphQLEnumType
    },
    descriptor: {
        type: new GraphQLNonNull(ActionDescriptor),
    },
  }),
});
```

But when I try out with `GraphQLUnionType`, I can't find a way to get type information inside the descriptor. 

```js
const ActionDescriptor = new GraphQLUnionType({
  name: 'ActionDescriptor',
  types: [
    SendMailDescriptor,
    SendSmsDescriptor,
  ],
  resolveType: (data, context, info) => {
     // data, context, info is not very helpful here
     // data will be something like { email: 'test@t.com' }  
     // but determine type with GraphQLEnumType is better
  },
});
```

Any ideas? How can I access parent (action) in ActionDescriptor's `resolveType` function?

leebyron:
It seems you could detect by the fields `email` or `phone` existing? Or perhaps I've missed the core issue?

chentsulin:
In my cases, descriptors' fields may change over the time and types' descriptor may have same fields, so it's not ideal to detect them by field-based checks. If we need to change fields' checking logic in `resolveType` when one of types changed, it will become more and more difficult to maintain after our types grew. Ideally, we should follow open-closed principle, and avoid change resolveType frequently.

We want to support over ten types in our union in the future, so i think that will be plainful.

khankuan:
Agree with @chentsulin. It seems to be risky to depend on field schema to determine object type. Did you find a resolution for this issue?

khankuan:
@leebyron I find abstract types a unique case where it might be useful to have the type resolved by the source. Compared to other types, resolution is static via `type: TypeX`. What do you think?

leebyron:
[...]

PR #708, opened 2017-02-06
Added source argument to resolveType function in siutations where typ…

khankuan:
…e can be determined by source instead of value object.

Related #587.

### DIFF
--- a/src/execution/__tests__/abstract-test.js
+++ b/src/execution/__tests__/abstract-test.js
@@ -430,4 +430,108 @@ describe('Execute: Handles execution of abstract types', () => {
     });
   });
 
+  it('resolveType with source value', async () => {
+    const DogType = new GraphQLObjectType({
+      name: 'Dog',
+      fields: {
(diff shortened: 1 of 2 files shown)


wincent:
This looks reasonably enough to me, and the implementation is clean. AFAIK this change is compatible with [the spec](https://facebook.github.io/graphql/#sec-Value-Completion), which says: 

> When completing a field with an abstract return type, that is an Interface or Union return type, first the abstract type must be resolved to a relevant Object type. This determination is made by the internal system using whatever means appropriate.

So I think extending `resolveType` to take the additional parameter is fine.

leebyron:
Thanks for your investigation into this, but in my opinion we should not make this change.

See my comments in the related issue #587 - this would expose a value that very few assumptions/guarantees can be made about, which makes it unsafe to access. That's a "foot-gun" for unsuspecting uses of it, and creates the potential for a change in one part of the schema causing breakage in a seemingly unrelated part of the schema.

PR #717, opened 2017-02-16
Soften name validation warnings to avoid CI issues

wincent:
A few things here:

- Downgrade `console.error` to `console.warn`.
- Strip "Error: " prefix from messages that can trip up CIs.
- Note in warning message that it will eventually become a hard error.

### DIFF
--- a/src/utilities/__tests__/assertValidName-test.js
+++ b/src/utilities/__tests__/assertValidName-test.js
@@ -1,0 +1,139 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+import { formatWarning } from '../assertValidName';
+
+/**
+ * Helper for dedenting indented template literals. This helps us to
+ * keep the tests pretty.
+ */
+function dedent(string) {
+  // Get lines, discarding empty leading and trailing lines.
+  const lines = string.replace(/^[ \t]*\n|\n[ \t]*$/g, '').split('\n');
+
+  // Find smallest indent.
+  const indent = lines.reduce((currentMinimum, line) => {
+    const whitespace = line.match(/^ +/);
+    return Math.min(
+      whitespace ? whitespace[0].length : 0,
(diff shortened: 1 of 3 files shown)


leebyron:
Thanks for doing the research and adding tests. This looks great. Feel free to merge.

I think beyond this we should follow-up with with a slightly more invasive change to separate the construction of a schema from the validation of a schema. 

When building a "source of truth" schema (e.g. you run a node.js server, and use GraphQL.js to present a GraphQL service) then we should encourage validation - strict validation - to ensure the schema is well formed. I'm not sure what the best way to accomplish this would be and am open to ideas. Perhaps `execute()` or other operations dependent on a valid schema would check some internally-set flag for "has been validated", and if it's not set, run validation? Just an idea.

When building a "reference" schema (e.g. you used introspection to learn about a schema, and have rebuilt it to power a tool like GraphiQL) then we have very different validation needs. Assumptions made by execution like an object adhering to its implemented interfaces, or (in this case) names being well formed may not break assumptions of the tools being used - we're both paying a computation cost to perform this validation and we're forcing tools to be more strict than they might need to be. Should GraphiQL stop working or report errors if it's used against a server that has an invalid schema? If it should, how do we define that line?

This diff is a great incremental step to easing the pain of this particular change, but it would be awesome to see a broader rethink of how schema validation fits into the whole story.

src/utilities/__tests__/assertValidName-test.js
```diff
@@ -0,0 +1,139 @@
+/**
```
asiandrummer: Love this

src/utilities/assertValidName.js
```diff
@@ -28,12 +29,13 @@ export function assertValidName(
   if (!isIntrospection && name.slice(0, 2) === '__' && !hasWarnedAboutDunder) {
     hasWarnedAboutDunder = true;
     /* eslint-disable no-console */
-    if (console && console.error) {
+    if (console && console.warn) {
       const error = new Error(
         `Name "${name}" must not begin with "__", which is reserved by ` +
-        'GraphQL introspection.'
+        'GraphQL introspection. In a future release of graphql this will ' +
```
asiandrummer: nit: graphql -> graphql-js?

wincent: I went with `graphql` because that is the NPM package name. `graphql-js` is only the name of the GitHub repo.

wincent:
Split off https://github.com/graphql/graphql-js/issues/719 to carry out discussion about the deeper changes to the validation model.

Issue #546, opened 2016-10-28
Better error message on "wrong type" found in schema


GreenAsJade:
Appended is the message you get if you make a mistake in a type declaration.

This is quite a likely mistake: it occurs when you copy a `type` from a query schema into an input schema and forget to change it from  `type` to `input`.

Ideally the message would include the text of the line that is the problem:

`type VariationInput {`

or at least the name of the thing that has the wrong type

`VariationInput`.

I had a look to see if I could quickly make a patch for this, but unfortunately there doesn't appear to be much context obviously available at the place this message is generated :(

```
/Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/jsutils/invariant.js:19
    throw new Error(message);
    ^

Error: Expected Input type.
    at invariant (/Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/jsutils/invariant.js:19:11)
    at produceInputType (/Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/utilities/buildASTSchema.js:238:29)
    at /Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/utilities/buildASTSchema.js:336:18
    at /Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/jsutils/keyValMap.js:36:31
    at Array.reduce (native)
    at keyValMap (/Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/jsutils/keyValMap.js:35:15)
    at makeInputValues (/Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/utilities/buildASTSchema.js:333:36)
    at fields (/Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/utilities/buildASTSchema.js:410:16)
    at resolveThunk (/Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/type/definition.js:115:40)
    at GraphQLInputObjectType._defineFieldMap (/Users/mgregory/NRN/src/agreeonit/apollo_server/node_modules/graphql/type/definition.js:569:20)
```

GreenAsJade:
Of course the next thing that happens is that you fix the primary declaration and forget to fix the internal fields:

Get `Error: Expected Input type.`

Change

```
type VariationInput {
   project: Project!
```
to
```
input VariationInput {
   project: Project!
```

and the exact same message appears!

[...]

PR #718, opened 2017-02-16
Adds message to stack trace of GraphQLError.

tcr:
With this sample code run in Node:

```js
import { graphql, buildSchema } from 'graphql';

const schema = buildSchema(`
  type Query error_here {
    value: String
  }
`);
```

You get an error that looks like it should have a message, but because of how we create GraphQLErrors, it does not:

```bash
$ babel-node test.js

/Users/trim/github/graphql-js/dist/language/parser.js:965
  throw (0, _error.syntaxError)(lexer.source, token.start, 'Expected ' + kind + ', found ' + (0, _lexer.getTokenDesc)(token));
  ^
GraphQLError
    at syntaxError (/Users/trim/github/graphql-js/dist/error/syntaxError.js:28:15)
    at expect (/Users/trim/github/graphql-js/dist/language/parser.js:966:9)
    at many (/Users/trim/github/graphql-js/dist/language/parser.js:1014:3)
    at parseObjectTypeDefinition (/Users/trim/github/graphql-js/dist/language/parser.js:683:16)
    at parseTypeSystemDefinition (/Users/trim/github/graphql-js/dist/language/parser.js:607:16)
    at parseDefinition (/Users/trim/github/graphql-js/dist/language/parser.js:148:16)
    at parseDocument (/Users/trim/github/graphql-js/dist/language/parser.js:106:22)
    at parse (/Users/trim/github/graphql-js/dist/language/parser.js:43:10)
    at buildSchema (/Users/trim/github/graphql-js/dist/utilities/buildASTSchema.js:461:25)
    at Object.<anonymous> (/Users/trim/Desktop/ok3/what.js:3:16
```

With this PR, we add the message before generating the stack trace. The result becomes:

```bash
$ babel-node test.js

/Users/trim/github/graphql-js/dist/language/parser.js:965
  throw (0, _error.syntaxError)(lexer.source, token.start, 'Expected ' + kind + ', found ' + (0, _lexer.getTokenDesc)(token));
  ^
GraphQLError: Syntax Error GraphQL (2:14) Expected {, found Name "error_here"

1:
2:   type Query error_here {
                ^
3:     value: String

    at syntaxError (/Users/trim/github/graphql-js/dist/error/syntaxError.js:28:15)
    at expect (/Users/trim/github/graphql-js/dist/language/parser.js:966:9)
    at many (/Users/trim/github/graphql-js/dist/language/parser.js:1014:3)
[...]

PR #722, opened 2017-02-17
Allow buildASTSchema to throw errors with source locations.

tcr:
See #546 for context. This patch is one possible implementation of having type errors in a schema list their source location. Given the following code:

```js
const { graphql, buildSchema } = require('graphql');

const schema = buildSchema(`
  type OutputType {
    value: String
  }

  input InputType {
    value: String
  }

  type Query {
    output: [OutputType]
  }

  type Mutation {
    signup (password: OutputType): OutputType
  }
`)
```

Instead of this error:

```sh
$ node schema.js
/Users/trim/github/graphql-js/dist/jsutils/invariant.js:19
    throw new Error(message);
    ^

Error: Expected Input type
    at invariant (/Users/trim/github/graphql-js/dist/jsutils/invariant.js:19:11)
    ...
```

It would throw this:

```sh
$ node schema.js
/Users/trim/github/graphql-js/dist/jsutils/invariant.js:19
    throw new Error(message);
    ^

Error: Expected Input type (15:23)

14:   type Mutation {
15:     signup (password: OutputType): OutputType
                          ^
16:   }

    at invariant (/Users/trim/github/graphql-js/dist/jsutils/invariant.js:19:11)
    ....
```

Caveats:

* We'd have to thread the Source object through buildASTSchema in order to have content for line numbers and the source error itself. In this implementation, I made this optional so buildASTSchema just uses it as context for the error if it's provided.
* In order to avoid recreating the `Source` object many times in `invariantError`, `buildASTSchema` now does so explicitly. But is this necessary?
* This makes our invariant calls more costly, because it constructs the error message and deduces the line numbers & syntax highlighting whether or not it's thrown.
* We have to give the `syntaxError` module knowledge of the AST, though with a bit of work this implementation could pass in a position instead.
* The implementation `invariantError` I chose here is pretty weak, there's probably a better pattern for wrapping `invariant(...)` we could use.
* Can we merge the formatting logic of `syntaxError` to use `invariantError`?
* Lastly, I didn't apply the invariantError to all functions, just a handful. That would have to be finished before merge.

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -86,6 +86,33 @@ export function GraphQLError( // eslint-disable-line no-redeclare
   path?: ?Array<string | number>,
   originalError?: ?Error
 ) {
+  // Define message so it can be captured in stack trace.
+  Object.defineProperty(this, 'message', {
+    value: message,
+    // By being enumerable, JSON.stringify will include `message` in the
+    // resulting output. This ensures that the simplist possible GraphQL
+    // service adheres to the spec.
+    enumerable: true,
+    writable: true
+  });
+
+  // Include (non-enumerable) stack trace.
+  if (originalError && originalError.stack) {
+    Object.defineProperty(this, 'stack', {
+      value: originalError.stack,
+      writable: true,
+      configurable: true
+    });
+  } else if (Error.captureStackTrace) {
+    Error.captureStackTrace(this, GraphQLError);
+  } else {
+    Object.defineProperty(this, 'stack', {
+      value: Error().stack,
+      writable: true,
+      configurable: true
+    });
+  }
+
   // Compute locations in the source for the given nodes/positions.
   let _source = source;
   if (!_source && nodes && nodes.length > 0) {
@@ -109,14 +136,6 @@ export function GraphQLError( // eslint-disable-line no-redeclare
   }
 
   Object.defineProperties(this, {
-    message: {
-      value: message,
-      // By being enumerable, JSON.stringify will include `message` in the
-      // resulting output. This ensures that the simplest possible GraphQL
-      // service adheres to the spec.
-      enumerable: true,
-      writable: true
-    },
     locations: {
       // Coercing falsey values to undefined ensures they will not be included
       // in JSON.stringify() when not provided.
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -90,7 +90,7 @@ export function GraphQLError( // eslint-disable-line no-redeclare
   Object.defineProperty(this, 'message', {
     value: message,
     // By being enumerable, JSON.stringify will include `message` in the
-    // resulting output. This ensures that the simplist possible GraphQL
+    // resulting output. This ensures that the simplest possible GraphQL
     // service adheres to the spec.
     enumerable: true,
     writable: true
--- a/src/error/syntaxError.js
+++ b/src/error/syntaxError.js
@@ -58,9 +58,9 @@ export function syntaxError(
  * returns an error containing the message, without context.
  */
 export function validationError(
+  source: ?Source,
+  node: ?ASTNode,
   message: string,
-  node?: ASTNode,
-  source?: Source
 ): GraphQLError {
   const position = node ? (node.loc ? node.loc.start : null) : null;
   if (position == null || source == null) {
--- a/src/error/syntaxError.js
+++ b/src/error/syntaxError.js
@@ -11,6 +11,29 @@
 import { getLocation } from '../language/location';
 import type { Source } from '../language/source';
 import { GraphQLError } from './GraphQLError';
+import type { ASTNode } from '../language/ast';
+
+/**
+ * Produces a string for formatting a syntax or validation error with an
+ * embedded location.
+ */
+export function printLocatedError(
+  source: Source,
+  position: number,
+  message: string,
+  description?: string,
+): GraphQLError {
+  const location = getLocation(source, position);
+  const body = `${message} (${location.line}:${location.column})` +
+    (description ? ' ' + description : '') +
+    '\n\n' + highlightSourceAtLocation(source, location);
+  return new GraphQLError(
+    body,
+    undefined,
+    source,
+    [ position ]
+  );
+}
 
 /**
  * Produces a GraphQLError representing a syntax error, containing useful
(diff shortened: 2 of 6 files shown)


src/error/syntaxError.js
```diff
@@ -33,6 +34,25 @@ export function syntaxError(
 }
 
 /**
+ * Produces a string for the invarant(...) function that renders the location
```
asiandrummer: Maybe we can even refactor these functions to be more generic:
```js
export function formatError(errorType: string) { ... }
export function syntaxError() { return formatError('syntax'); }
export function validationError() { return formatError('validation'); }
```
Not exactly like that, but you get the gist

src/error/syntaxError.js
```diff
@@ -33,6 +34,25 @@ export function syntaxError(
...
+  message: string,
+  node: ASTNode,
+  source: ?Source
+): string {
+  const position = node.loc ? node.loc.start : null;
+  if (position == null || source == null) {
+    return message;
+  }
+  const location = getLocation(source, position);
+  return `${message} (${location.line}:${location.column}) ` +
```
asiandrummer: I think this can actually return `new GraphQLError(...)`

src/utilities/buildASTSchema.js
```diff
@@ -518,7 +526,8 @@ export function getDescription(node: { loc?: Location }): ?string {
  * document.
  */
 export function buildSchema(source: string | Source): GraphQLSchema {
-  return buildASTSchema(parse(source));
+  const sourceObj = typeof source === 'string' ? new Source(source) : source;
```
asiandrummer: > In order to avoid recreating the Source object many times in invariantError, buildASTSchema now does so explicitly. But is this necessary?

Maybe we could construct a custom function globally to do this. I don't have a good answer for this as well, but yeah, passing an argument just for `invariantError` (or whatever we'll end up with) seems a bit unnecessary.

src/error/syntaxError.js
```diff
@@ -33,6 +34,25 @@ export function syntaxError(
 }
 
 /**
+ * Produces a string for the invarant(...) function that renders the location
+ * where an error occurred. If no source is passed in, it renders the message
+ * without context.
+ */
+export function invariantError(
```
asiandrummer: eh, maybe typeError/validationError?

src/error/syntaxError.js
tcr: @asiandrummer This prints strange output from the `invariant` function, which inlined would be:

```
if (!condition) {
    throw new Error(new GraphQLError(...));
}
```

But, I could change `invariant(...)` to check first if its message argument is a string, and if it's actually an error, just rethrow. Then I could make `validationError` return a `GraphQLError`. Sensible?

tcr:
Updated this PR:

* This now rolls up #718 to allow the GraphQLError test to function properly.
* Allows invariant(...) to rethrow errors if they are passed in. This change can be reverted if we want validationError() to simply return a string.
* Broke out formatError into its own method.

The caveat still exists that buildASTSchema requires both "Source" and "ast" to return validation errors with full source body.

src/error/GraphQLError.js
```diff
@@ -77,6 +77,16 @@ export function GraphQLError( // eslint-disable-line no-redeclare
   path?: ?Array<string | number>,
   originalError?: ?Error
 ) {
+  // Define message so it can be captured in stack trace.
```
asiandrummer: Sorry if I misunderstood, but any reason why this was moved to the top?

src/error/syntaxError.js
```diff
@@ -21,15 +44,29 @@ export function syntaxError(
...
-  return error;
+}
+
+/**
+ * Produces a GraphQLError for the invariant(...) function that renders the
+ * location where a validation error occurred. If no source is passed in, it
+ * returns an error containing the message, without context.
+ */
+export function validationError(
+  message: string,
```
asiandrummer: nit, but should we order the argument similar to the functions above? That probably means to move `source` to the top.

asiandrummer:
This looks good to me - maybe @leebyron should have another look at this

src/error/GraphQLError.js
tcr: @asiandrummer Ah, yeah this is the rebase over #718. I needed this to add the test case for the validation error—otherwise the message reports as just "GraphQLError" and we can't meaningfully test it.

src/jsutils/invariant.js
```diff
@@ -8,8 +8,11 @@
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
 
-export default function invariant(condition: mixed, message: string) {
+export default function invariant(condition: mixed, message: string | Error) {
   if (!condition) {
+    if (message instanceof Error) {
+      throw message;
+    }
```
leebyron: This bit is unnecessary - no need for this change, then you can change the `invariant`s below to just be `if (!condition) { throw validationError() }`

src/error/syntaxError.js
```diff
@@ -11,6 +11,29 @@
 import { getLocation } from '../language/location';
 import type { Source } from '../language/source';
 import { GraphQLError } from './GraphQLError';
+import type { ASTNode } from '../language/ast';
+
+/**
+ * Produces a string for formatting a syntax or validation error with an
+ * embedded location.
+ */
+export function formatError(
```
leebyron: We already have another function called `formatError` which does something slightly different.

How about `printLocatedError`?

src/error/syntaxError.js
```diff
@@ -21,15 +44,29 @@ export function syntaxError(
...
   );
-  return error;
+}
+
+/**
+ * Produces a GraphQLError for the invariant(...) function that renders the
+ * location where a validation error occurred. If no source is passed in, it
+ * returns an error containing the message, without context.
+ */
+export function validationError(
```
leebyron: This is only used from buildASTSchema, so perhaps it should live there? - Also perhaps this should be "Schema Error" instead?

leebyron:
Sorry for not seeing this sooner! This is great! There is also some discussion in #746 which might lead to even more coverage of schema validation improvements like this one.

Do you mind rebasing?

src/error/syntaxError.js
tcr: Hm, in the rebased diff it is now used in buildASTSchema.js and definition.js.

tcr:
Rebased—let me know if anything got lost in the transition.

cjoudrey:
@tcr was your intention to replace all other cases of `throw new Error` with an error class that includes location of the mistake?

https://github.com/graphql/graphql-js/blob/37717b8a6e2549ac1ce73acb936d98dba7b14d09/src/utilities/buildASTSchema.js#L153-L159

For example, in the above snippet it would be great that the error that is thrown includes the location of both schema definition AST nodes.

This would allow tools (like schema linters) to surface this information to the user. At the moment this isn't possible because a generic `Error` with only a `message` is thrown.

Happy to help if you don't have time to ship this.

src/type/definition.js
```diff
@@ -82,11 +85,16 @@ export function isInputType(type: ?GraphQLType): boolean %checks {
...
+  type: ?GraphQLType,
+  typeNode?: ASTNode,
+  source?: Source
+): GraphQLInputType {
+  invariant(isInputType(type),
+    validationError(
+      source,
+      typeNode,
+      `Expected ${String(type)} to be a GraphQL input type.`));
   return type;
```
leebyron: These methods are used throughout and not only in validation, so instead anywhere doing schema validation should be using `isInputType` and throwing the appropriate error instead of this general purpose function changing

src/utilities/buildASTSchema.js
leebyron: Also note that ast nodes include a reference to the source they come from, so you shouldn't need to pass the second argument around

tcr:
+1

PR #734, opened 2017-03-01
Add object level default resolve function

### LINKED ISSUE #733: Custom Default Resolver
We are finding ourselves writing a lot of field resolvers that have very similar logic. There are a number of ways to deal with this, but I would like to propose adding a `defaultResolve` option that will override the built in default resolver for any fields of an object.

### Before
```
  name: 'Post',
  fields: {
    title: {
      type: GraphQLString,
      resolve: (root, args, ctx) => ctx.db.load(root, 'title')
    },
    author: {
      type: GraphQLString,
      resolve: (root, args, ctx) => ctx.db.load(root, 'author')
    }
  }
})
```

### After
```
  name: 'Post',
  defaultResolve: (root, args, ctx, info) => ctx.db.load(root, info.fieldName)
  fields: {
    title: { type: GraphQLString },
    author: { type: GraphQLString }
  }
})
```

You can imagine how much this dries up code when you have a lot of fields. I am happy to supply a PR to implement this if the maintainers feel it would be a valuable addition.

southpolesteve:
Adds the ability to specify a resolve function for an entire object type. Closes #733

### DIFF
--- a/src/execution/__tests__/resolve-test.js
+++ b/src/execution/__tests__/resolve-test.js
@@ -20,10 +20,11 @@ import {
 
 describe('Execute: resolve function', () => {
 
-  function testSchema(testField) {
+  function testSchema(testField, resolve) {
     return new GraphQLSchema({
       query: new GraphQLObjectType({
         name: 'Query',
+        resolve,
         fields: {
           test: testField
         }
@@ -94,7 +95,7 @@ describe('Execute: resolve function', () => {
     });
   });
 
-  it('uses provided resolve function', async () => {
+  it('uses field provided resolve function', async () => {
     const schema = testSchema({
       type: GraphQLString,
       args: {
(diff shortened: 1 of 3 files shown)


brad-decker:
This looks pretty awesome!

southpolesteve:
I've updated this PR with documentation edits

southpolesteve:
export class NodeType extends GraphQLObjectType {
  constructor (config) {
    const oldFields = config.fields
    config.interfaces = config.interfaces || []
    config.interfaces.push(Node)
    config.fields = () => {
      const fields = typeof oldFields === 'function' ? oldFields() : oldFields
      Object.keys(fields).map((fieldName) => {
        if (!fields[fieldName].resolve) {
          fields[fieldName].resolve = customFieldResolver
        }
      })

      return {
        id: {
          type: new GraphQLNonNull(GraphQLID)
        },
        createdAt: {
          type: new GraphQLNonNull(Timestamp),
          resolve: redisFieldResolver
        },
        updatedAt: {
          type: new GraphQLNonNull(Timestamp),
          resolve: redisFieldResolver
        },
        ...fields
      }
    }
    super(config)
  }
}

leebyron:
Unfortunately I will close this PR after discussion in the related issue. There is too much confusion around how these functions would be called - other requests for similar features have looked for very different semantics (others often want a resolver for the *returnType* instead of *parentType*).

That said, I have no qualms with this pattern, however I would prefer the use of composition to provide this behavior - a change to the core GraphQLObject API is not necessary to enable this!

Issue #223, opened 2015-11-08
Merging schemas


xpepermint:
I work on a huge project. To ensue project's long-term maintainability we decided to split the application into 3 standalone pieces thus we now have 3 specialized teams. Each app exposes its own GraphQL schema. At the end we plan to create the 4th app which will merge all 3 pieces together into a `gateway` application using the `express-graphql` package.

**What is the best way to merge schemas?** What we would need is a mechanism similar to [Express Router](http://expressjs.com/guide/routing.html). Our current plan is to make all 3 apps available as private packages from which we can build a new schema for the main (4th) app. Well, we could make each app act as a standalone private graphql server and then the main application would just send requests to sub-apps but I believe it could potentially represent a huge overhead. Am I wrong? 

Are there any plans for supporting this? What do you suggest? How do you handle this at Facebook?

KyleAMathews:
Just have each project export its types and combine those types together
into a schema.

xpepermint:
@KyleAMathews Thanks, that's what we plan to do. I just want to make sure that our thinking is correct.

leebyron:
We maintain one schema at Facebook. Different teams are responsible for maintaining the types that are specific to their products, but there is only one final schema that combines all the types together into one GraphQL service available to our client applications.

xpepermint:
@leebyron Thanks!

worldsayshi:
@leebyron How is this done when using schema lang? If I run `buildSchema` it seems that I must first combine all the schema lang code into one string because I can't combine schemas once they are compiled as far as I can see..

leebyron:
You're correct. If you're using the schema language to build a schema than you need to first combine it into a single string before calling buildSchema. That typically is a nice developer experience for small to medium sized schema, though for more complex schema we recommend building the schema programmatically. Even in this case you still need to collect all the types to assemble the schema. A similar operation to collecting all the strings before calling buildSchema

worldsayshi:
Hmm, it seems that I don't strictly need to combine it to a string but a list of strings. There seems to be an util for combining schemas however, [`extendSchema`](https://github.com/graphql/graphql-js/blob/e83409712ef1493a7709a5754b0e72d1bd33c753/src/utilities/extendSchema.js#L108). But it takes a `GraphQLSchema` and a `Document`. Not sure how to produce a `Document`.

worldsayshi:
Oh! http://graphql.org/graphql-js/language/#parse

liamcurry:
[...]

PR #744, opened 2017-03-07
Forbid duplicate type definitions

turadg:
A GraphQL schema can't have more than one definition of the same type.

Facebook runs an integration test to prevent this:
https://github.com/graphql/graphql-js/issues/223#issuecomment-265366200

With this PR the `buildASTSchema` step throws, causing any tests that load the schema to fail.

### DIFF
--- a/src/utilities/__tests__/buildASTSchema-test.js
+++ b/src/utilities/__tests__/buildASTSchema-test.js
@@ -802,4 +802,22 @@ fragment Foo on Type { field }
       .to.throw('Specified query type "Foo" not found in document.');
   });
 
+  it('Forbids duplicate type definitions', () => {
+    const body = `
+schema {
+  query: Repeated
+}
+
+type Repeated {
+  id: Int
+}
+
+type Repeated {
+  id: String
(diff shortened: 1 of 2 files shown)


turadg:
@leebyron this is the bug we discussed at React Conf.

leebyron:
Constructing a GraphQLSchema should throw this error, correct?

Also, in the context of an AST document, it would be great to throw GraphQLError with reference to the problem AST nodes.

wincent:
Thanks @turadg. This looks reasonable to me and in line with the spec which [says](http://facebook.github.io/graphql/#sec-Type-System):

> All types within a GraphQL schema must have unique names. No two provided types may have the same name.

And schema construction is obviously the right time to throw the error too. So I'll merge this. Would you like to follow up and make @leebyron's suggested change of throwing a `GraphQLError` instead? Some tools may appreciate the metadata.

turadg:
Thanks @wincent .  I haven't encountered that case but if I do I bet I'll be back with another PR. :)

PR #753, opened 2017-03-14
Fix GitHub syntax highlighting

calebmer:
I often pull up the `src/language/ast.js` file as a reference when working with a GraphQL query. However, the syntax highlighting on GitHub is broken. This PR should maintain the same behavior while fixing the syntax highlighting on GitHub with the side benefit of exporting `TokenKind` separate of `Token`.

A before photo is attached below.

- Before: https://github.com/graphql/graphql-js/blob/2235563141a01f081cf22092fe2dcc401b7cdc83/src/language/ast.js#L76
- After: https://github.com/calebmer/graphql-js/blob/da09c4ef1e5dd5ae2d99a73296a3274a03b609c7/src/language/ast.js#L84

screen shot 2017-03-14 at 6 53 39 pm

### DIFF
--- a/src/language/ast.js
+++ b/src/language/ast.js
@@ -9,7 +9,6 @@
  */
 
 import type { Source } from './source';
-import type { Token } from './lexer';
 
 
 /**
@@ -42,6 +41,70 @@ export type Location = {
    * The Source document the AST represents.
    */
   source: Source;
+};
+
+/**
+ * Represents a range of characters represented by a lexical token
+ * within a Source.
+ */
+export type Token = {
+
+  /**
+   * The kind of Token.
+   */
+  kind: '<SOF>'
+      | '<EOF>'
+      | '!'
+      | '$'
+      | '('
+      | ')'
+      | '...'
+      | ':'
+      | '='
+      | '@'
+      | '['
+      | ']'
+      | '{'
+      | '|'
+      | '}'
+      | 'Name'
+      | 'Int'
+      | 'Float'
+      | 'String'
+      | 'Comment';
+
+  /**
+   * The character offset at which this Node begins.
+   */
+  start: number;
+
+  /**
+   * The character offset at which this Node ends.
+   */
+  end: number;
+
+  /**
+   * The 1-indexed line number on which this Token appears.
+   */
+  line: number;
+
+  /**
+   * The 1-indexed column number at which this Token begins.
+   */
+  column: number;
+
+  /**
+   * For non-punctuation tokens, represents the interpreted value of the token.
+   */
+  value: string | void;
+
+  /**
+   * Tokens exist as nodes in a double-linked-list amongst all tokens
+   * including ignored tokens. <SOF> is always the first node and <EOF>
+   * the last.
+   */
+  prev: Token | null;
+  next: Token | null;
 };
 
 /**
--- a/src/language/ast.js
+++ b/src/language/ast.js
@@ -45,27 +45,29 @@ export type Location = {
 
 /**
  * Represents the different kinds of tokens in a GraphQL document.
+ * This type is not inlined in `Token` to fix syntax highlighting on GitHub
+ * *only*.
  */
-export type TokenKind = '<SOF>'
-                      | '<EOF>'
-                      | '!'
-                      | '$'
-                      | '('
-                      | ')'
-                      | '...'
-                      | ':'
-                      | '='
-                      | '@'
-                      | '['
-                      | ']'
-                      | '{'
-                      | '|'
-                      | '}'
-                      | 'Name'
-                      | 'Int'
-                      | 'Float'
-                      | 'String'
-                      | 'Comment';
+type TokenKind = '<SOF>'
+               | '<EOF>'
+               | '!'
+               | '$'
+               | '('
+               | ')'
+               | '...'
+               | ':'
+               | '='
+               | '@'
+               | '['
+               | ']'
+               | '{'
+               | '|'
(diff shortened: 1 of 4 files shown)


src/language/ast.js
```diff
@@ -44,6 +44,30 @@ export type Location = {
 };
 
 /**
+ * Represents the different kinds of tokens in a GraphQL document.
+ */
+export type TokenKind = '<SOF>'
```
asiandrummer: https://github.com/graphql/graphql-js/blob/da09c4ef1e5dd5ae2d99a73296a3274a03b609c7/src/language/lexer.js#L84-L131 already defines this - maybe we should just reuse?

calebmer: Didn’t know that existed. At a minimum we shouldn’t export `TokenKind` then.

Does Flow have a way to get the type of all the values? I know that TypeScript has `keyof` which will union all the string literal key types. Something like [`$PropertyType<T, x>`](https://flowtype.org/docs/utility-types.html#propertytypet-x), but it returns a union of _all_ the property types and not just one?

Assuming the property types are all string literals we could then write:

```js
type Token = {
  kind: $PropertyTypes<TokenKind>,
  ...
};
```

This would also be a reverse [`$Keys<T>`](https://flowtype.org/docs/utility-types.html#keyst).

asiandrummer: I'm not sure if there's a support for that yet - we probably need to do in a verbose way for the time being.
Another thing I noticed is that `ast.js` uses `Token` object solely to typecheck `Location` object - we could potentially move `export type Token` to `lexer.js` and import the type from there.

calebmer: I like that better. I moved the `Token` type as it is on master into the `lexer.js` file.

wincent:
Thanks for this @calebmer. I've also found the broken highlighting in GitHub to be bothersome. Travis is complaining about a failure under Node 7 though, with a legit looking error message, although I am not sure why it would only happen in version 7.

calebmer:
@wincent fixed. It was indeed a legitimate Flow error. It looks like Flow only runs on one of the Travis jobs: https://github.com/graphql/graphql-js/blob/1eb41dda699138c0a56de2fa937e7b9e5499949a/.travis.yml#L29, so it would make sense that it only fails on Node.js 7. That’s smart 😊

wincent:
So this doesn't appear to fix the GitHub syntax highlighting, it merely moves the brokenness into a different file. I think this will work instead (at least, it works when I tested it in a Gist):

```diff
diff --git a/src/language/ast.js b/src/language/ast.js
index 3639ec9..001900d 100644
--- a/src/language/ast.js
+++ b/src/language/ast.js
@@ -44,15 +44,11 @@ export type Location = {
 };
 
 /**
- * Represents a range of characters represented by a lexical token
- * within a Source.
+ * Token kinds, defined at the top level to avoid broken syntax
+ * highlighting on GitHub.
  */
-export type Token = {
-
-  /**
-   * The kind of Token.
-   */
-  kind: '<SOF>'
+type TokenKind
+      = '<SOF>'
       | '<EOF>'
       | '!'
       | '$'
@@ -73,6 +69,17 @@ export type Token = {
       | 'String'
       | 'Comment';
 
+/**
+ * Represents a range of characters represented by a lexical token
+ * within a Source.
+ */
+export type Token = {
+
+  /**
+   * The kind of Token.
+   */
+  kind: Kind;
+
   /**
    * The character offset at which this Node begins.
    */
```

calebmer:
@wincent this approach was roughly my first commit: https://github.com/graphql/graphql-js/pull/753/commits/da09c4ef1e5dd5ae2d99a73296a3274a03b609c7

Moved away from that because @asiandrummer mentioned the existing naming conflict with `TokenKind` in `lexer.js` and recommend we move the `Token` type into `lexer.js`, but keep the union inline. I like the idea of `Token` being in `lexer.js` instead of `ast.js`, but that was not the original point of the PR.

If you want we can revert back to https://github.com/graphql/graphql-js/pull/753/commits/da09c4ef1e5dd5ae2d99a73296a3274a03b609c7 and then not export `TokenKind` for `ast.js` with a comment noting that the type alias is only to fix syntax highlighting.

wincent:
Yeah, the diff was just an example to highlight what kind of change would be need to be made to fix the highlighting problem rather than just moving it.

> If you want we can revert back to da09c4e and then not export `TokenKind` for `ast.js`

Yep, and you'll note that in my diff I didn't export it.

> with a comment noting that the type alias is only to fix syntax highlighting.

Yep, that's what I did in my diff too. I'd be happy with that *or* an alternative that moves stuff into `lexer.js` but *also* fixes the busted highlighting in that file. I'll leave it up to your judgment.

calebmer:
Ok, reverted back to the original changes in order to keep this PR minimal and focused. As I mentioned before, the best solution would probably be to run a `$PropertyTypes<T>` utility type which would be a reverse of [`$Keys<T>`](https://flow.org/en/docs/types/utilities/#keyst) (and a generalization of [`$PropertyType<T, x>`](https://flow.org/en/docs/types/utilities/#toc-propertytype)) in Flow on the `TokenKind` object exported in `lexer.js`. However, such a utility type is not available, so this should be the simplest solution that removes the syntax highlighting problem 😊

Other options for the type name are `TokenKindUnion` or `TokenKindLiteral` if having two `TokenKind`s in the codebase is too confusing.

wincent:
Thanks for this @calebmer. I think what you've done here strikes the right balance.

PR #755, opened 2017-03-15
allow deprecated fields/values to be extended

asiandrummer:
There isn't a way to extend a schema with types that contains deprecated fields/enum values as `extendSchema` util function ignores the `@deprecated` directive. This fixes it.

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -465,7 +465,13 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
   }
 }
 
-function getDeprecationReason(directives: ?Array<DirectiveNode>): ?string {
+/**
+ * Given a collection of directives, returns the string value for the
+ * deprecation reason.
+ */
+export function getDeprecationReason(
+  directives: ?Array<DirectiveNode>,
+): ?string {
   const deprecatedAST = directives && find(
     directives,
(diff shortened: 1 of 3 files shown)


src/utilities/buildASTSchema.js
```diff
@@ -465,7 +465,13 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
...
 }
 
-function getDeprecationReason(directives: ?Array<DirectiveNode>): ?string {
+/**
+ * Given a collection of directives, returns the string value for the
+ * deprecation reason.
+ */
+export function getDeprecationReason(
+  directives: ?Array<DirectiveNode>,
+): ?string {
```
leebyron: fits on one line?

src/utilities/extendSchema.js
```diff
@@ -506,7 +510,14 @@ export function extendSchema(
     return new GraphQLEnumType({
       name: typeNode.name.value,
       description: getDescription(typeNode),
-      values: keyValMap(typeNode.values, v => v.name.value, () => ({})),
+      values: keyValMap(
```
leebyron: much better!

src/utilities/buildASTSchema.js
asiandrummer: nah, this was producing a lint error

PR #767, opened 2017-03-22
Unify wording in error messages.

neelance:
Before this change there was:
- There can be only one argument named ...
- There can be only one input field named ...
- There can be only one variable named ...
- There can only be one fragment named ...
- There can only be one operation named ...

Now the last two match the others.


wincent:
Nice attention to detail. Thanks a lot!

Issue #719, opened 2017-02-16
Create more flexible context-specific validation model


wincent:
Extracting @leebyron's [comment](https://github.com/graphql/graphql-js/pull/717#issuecomment-280439861) from https://github.com/graphql/graphql-js/pull/717:

> I think beyond this we should follow-up with with a slightly more invasive change to separate the construction of a schema from the validation of a schema. 
>
> When building a "source of truth" schema (e.g. you run a node.js server, and use GraphQL.js to present a GraphQL service) then we should encourage validation - strict validation - to ensure the schema is well formed. I'm not sure what the best way to accomplish this would be and am open to ideas. Perhaps `execute()` or other operations dependent on a valid schema would check some internally-set flag for "has been validated", and if it's not set, run validation? Just an idea.
>
> When building a "reference" schema (e.g. you used introspection to learn about a schema, and have rebuilt it to power a tool like GraphiQL) then we have very different validation needs. Assumptions made by execution like an object adhering to its implemented interfaces, or (in this case) names being well formed may not break assumptions of the tools being used - we're both paying a computation cost to perform this validation and we're forcing tools to be more strict than they might need to be. Should GraphiQL stop working or report errors if it's used against a server that has an invalid schema? If it should, how do we define that line?
>
> This diff is a great incremental step to easing the pain of this particular change, but it would be awesome to see a broader rethink of how schema validation fits into the whole story.

Adding the "discussion" label.

asiandrummer:
If I understood the schema validation step correctly (that is, I'm assuming there isn't a validation rule that checks something other than the integrity of the schema), then I think an invalid schema should never exist. #717 and related touch on a specific situation where we try invalidating what was a valid schema before by enforcing a restriction to the schema generation - I feel like this should be treated as an edge case and probably something we'd need to consider in addition when we'd like to add more restrictions to the specifications/implementations later on.

GraphiQL is an interesting use case because it probably needs both variations of schema concepts - "source of truth" schema to execute a query and "reference" schema to power IDE-like features without the said computation cost. With that said, I feel like we'd be walking a dangerous zone if GraphiQL is allowed to interact with an invalid schema.

Another thought I had was the possible discrepancy between two variations of the schema. Is it possible to have a "reference" schema that is different (e.g. has invalid new/modified field/types) from the "source-of-truth" schema?

OlegIlyenko:
I think it would a very good addition to the library 👍 

[...]

PR #772, opened 2017-03-29
Do not warn about legacy __configs__ name

kassens:
We currently don't have a good way to silence warnings for legacy field names staring with `__`. This silences the two offenders used internally at Facebook until we find a way to clean them up from the schema or ignore them when parsing an existing schema.

### DIFF
--- a/src/utilities/assertValidName.js
+++ b/src/utilities/assertValidName.js
@@ -26,7 +26,16 @@ export function assertValidName(
       `Must be named. Unexpected name: ${name}.`
     );
   }
-  if (!isIntrospection && name.slice(0, 2) === '__' && !hasWarnedAboutDunder) {
+  // For legacy reasons, __type__ and __configs__ should not generate a warning
+  // until the usage of these fields is cleaned up from the Facebook internal
+  // schemas.
+  if (
+    !isIntrospection &&
+    name.slice(0, 2) === '__' &&
+    name !== '__configs__' &&
+    name !== '__type__' &&
+    !hasWarnedAboutDunder
+  ) {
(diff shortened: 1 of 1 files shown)


wincent:
Rather than put an FB-specific hack in the reference implementation, I'd much rather see us come up with a mechanism you could use to add exceptions to a whitelist. (Or failing that, a way of opting out of this validation.) It wouldn't need to be pretty, ergonomic or easy (we don't want to make opting out *too* easy), just possible. What do you think?

kassens:
Thoughts on how to do this? I'd like to avoid passing a flag through the call stack, which I'd argue pollutes this implementation even more then a small special case.

I could add a flag on `assertValidName`? Something like `require('graphql').assertValidName.WARN = false`, but that relies on calling that before loading the schema and mutating a module.

leebyron:
IMHO we should create warnings if you're creating a schema manually for runtime, but not create warnings if you're hydrating a schema from an introspection result. It's pretty useless for tools like graphiql or codegen systems to issue these warnings that are really in place for first-time schema construction.

Threading a variable through everything feels ugly, but so setting a global value :/

leebyron:
For example, maybe in `buildClientSchema` or `buildASTSchema` we can set some global context flag to ignore warnings?

leebyron:
I think most ideally validating a schema and constructing a schema can be separated into two separate functions. That would also improve schema construction performance by not running this validation inline every time. Perhaps schema validation could them be run as part of the all-inclusive `graphql` function.

wincent:
For reference, #719 is the issue tracking the separation of warnings according to runtime construction vs introspection.

PR #780, opened 2017-04-05
[validation] Fix infinite loop on invalid queries in OverlappingFields

dschafer:
`OverlappingFieldsCanBeMerged` would infinite loop when passed something like

```graphql
fragment A on User {
  name
  ...A
}
```

It's not `OverlappingFieldsCanBeMerged`'s responsibility to detect that validation error, but we still would ideally avoid infinite looping.

This detects that case, and pretends that the infinite spread wasn't there for the purposes of this validation step.

### DIFF
--- a/src/validation/__tests__/OverlappingFieldsCanBeMerged-test.js
+++ b/src/validation/__tests__/OverlappingFieldsCanBeMerged-test.js
@@ -938,4 +938,55 @@ describe('Validate: Overlapping fields can be merged', () => {
 
   });
 
+  it('does not infinite loop on recursive fragment', () => {
+    expectPassesRule(OverlappingFieldsCanBeMerged, `
+      fragment fragA on Human {name relatives { name ...fragA } },
+    `);
+  });
+
+  it('does not infinite loop on immediately spread fragment', () => {
+    expectPassesRule(OverlappingFieldsCanBeMerged, `
+      fragment fragA on Human {name ...fragA },
+    `);
+  });
+
+  it('finds invalid case even with immediately spread fragment', () => {
+    // You would not expect this to have three error messages, and you would be
+    // right. The trickiness here is that we don't detect the immediate spread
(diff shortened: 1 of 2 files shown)


src/validation/__tests__/OverlappingFieldsCanBeMerged-test.js
```diff
@@ -938,4 +938,55 @@ describe('Validate: Overlapping fields can be merged', () => {
...
 
+  it('does not infinite loop on recursive fragment', () => {
+    expectPassesRule(OverlappingFieldsCanBeMerged, `
+      fragment fragA on Human {name relatives { name ...fragA } },
+    `);
+  });
+
+  it('does not infinite loop on immediately spread fragment', () => {
+    expectPassesRule(OverlappingFieldsCanBeMerged, `
+      fragment fragA on Human {name ...fragA },
```
leebyron: add a test for immediate spreads but in a larger cycle

fragment fragA on Human {name ...fragB }
fragment fragB on Human {name ...fragA }

OlegIlyenko:
I can [confirm](https://github.com/sangria-graphql/sangria/issues/238#issuecomment-309791134) this issue :) This PR was open for quite a while now. Is there something else that prevents this PR from being merged?

dschafer:
Nah, I think this is mergeable; it's not perfect, but it's certainly an upgrade over what exists before. @leebyron, any objections?

sorenbs:
Thanks for moving this forward Oleg!

Just want to add that we have servers being restarted on a daily basis because of this. Looking forward to this fix landing.

leebyron:
I'd love to see the fix handle larger cycles, since those are also likely and this PR as is will only partially improve the situation.

If anyone is willing to help with this, that would be welcomed!

leebyron:
I took the tests from this PR and wrote a complete solution in #1111

Issue #783, opened 2017-04-07
breaking changes finds wrong errors


jbreckel:
~~on the `npm` branch~~ in the 0.9.2 release, `breaking-changes` finds a lot of errors, that shouldn't be errors:

```
{
  "description": "Query.type arg id has changed type from ID! to ID!",
  "type": "ARG_CHANGED_KIND",
},
```

wincent:
Thanks for the report @jbreckel.

Related to https://github.com/graphql/graphql-js/pull/701

What are you doing to repro this error?

jbreckel:
I created a M(n)WE: https://github.com/jbreckel/graphql-breaking-changes-error.
* `yarn run build` to generate a new `schema.json`
* `yarn run test` to test new `schema.js` against `schema.json`

*expected*: should be fine, because I just generated the schema

*actual:* 
```
 FAIL  ./breaking-changes.spec.js
  ● Breaking Changes › there should be no breaking changes

    expect(received).toEqual(expected)

    Expected value to equal:
      Array []
    Received:
      [{"description": "Query.test arg id has changed type from ID! to ID!", "type": "ARG_CHANGED_KIND"}]

    Difference:

    - Expected
    + Received

    -Array []
    +Array [
    +  Object {
    +    "description": "Query.test arg id has changed type from ID! to ID!",
    +    "type": "ARG_CHANGED_KIND",
    +  },
    +]
```

jbreckel:
after some more testing this is what it can be boiled down to:
```
const {
  findBreakingChanges,
  buildClientSchema,
  graphql,
} = require('graphql')

const { introspectionQuery } = require('graphql/utilities')
const { makeExecutableSchema } = require('graphql-tools')

describe('Breaking Changes', () => {
  it('test with graphql-tools', async () => {
    const schemaString = `
    schema {
      query: Query
    }

    type Query {
      test(id: ID!): String
    }
    `

    const Schema = makeExecutableSchema({
      typeDefs: schemaString,
    })

    const { data, errors } = await graphql(Schema, introspectionQuery)
    expect(findBreakingChanges(buildClientSchema(data), Schema)).toEqual([])
  })
})
```

This test works with `graphql@0.9.1` but fails in `graphql@0.9.2` due to the argument of `Query.test`.
It is not only `ID` but every argument is seen as changed and therefore breaking.:

`graphql@0.9.1`:
```
$ yarn upgrade graphql@0.9.1 && yarn test
yarn upgrade v0.21.3
[...]

PR #800, opened 2017-04-11
Moving from raw instance check to name checks in `findBreakingArgChanges`

joelgriffith:
Addresses https://github.com/graphql/graphql-js/issues/783. When I initially wrote the `findBreakingArgChanges`, doing comparison checks on the arg instances themselves was the state-of-the-art. It looks like the `findFieldsThatChangedType` now references the `name` property of the types to check if changes happened in a breaking way, which is what I've done in this PR.

Not sure how I can assert this correction in unit-tests... would be happy for feedback

### DIFF
--- a/src/utilities/__tests__/findBreakingChanges-test.js
+++ b/src/utilities/__tests__/findBreakingChanges-test.js
@@ -392,6 +392,54 @@ describe('findBreakingChanges', () => {
     ]);
   });
 
+  it('should not flag args with the same type signature as breaking', () => {
+    const oldType = new GraphQLObjectType({
+      name: 'Type1',
+      fields: {
+        field1: {
+          type: GraphQLInt,
+          args: {
+            id: {
+              type: new GraphQLNonNull(GraphQLInt),
+            },
+          },
+        },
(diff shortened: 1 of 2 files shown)


wincent:
Thanks for the quick fix! You're right that a unit test would be very useful here.

joelgriffith:
I'll stew over that and fix the flow issues as well...

wincent:
> I'll stew over that and fix the flow issues as well...

Awesome, thanks!

src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -392,6 +392,54 @@ describe('findBreakingChanges', () => {
     ]);
   });
```
joelgriffith: This test asserts were not using instance equality to validate that an argument has changed in a breaking way (this test will fail if so)

joelgriffith:
I've verified that this works in another project as expected.

wincent:
Looks good. Thanks for this. Will cut a release shortly to get this fix out.

PR #807, opened 2017-04-14
Export getVisitFn function

IvanGoncharov:
In need, this function in order to implement a custom wrapper for visitor function.
As a workaround, I copy-pasted this function into my lib which is not ideal :(


IvanGoncharov:
> can you make sure this is also exported at the module level?

@leebyron Done

PR #808, opened 2017-04-14
[RFC] Export all validation rules directly

leebyron:
It's become relatively standard to create a whitelist or blacklist of validation rules, and reaching into the package per rule can get messy. This exports all the rules at the higher levels of submodule and module as well to make this easier.

For explaination of whitelist of blacklist:

```js
// Whitelist
import { specifiedRules, ScalarLeafsRule } from 'graphql';

// A ruleset which only runs one rule.
const whitelistRules = [ ScalarLeafsRule ];
```

```js
// Blacklist
import { specifiedRules, ScalarLeafsRule } from 'graphql';

// A ruleset which only runs all but one rule.
const blacklistRules = specifiedRules.filter(rule => rule !== ScalarLeafsRule);

// Note: the wrong way to do a blacklist would be to import and specify all
// but one rule. That's a whitelist and would fail to pick up any added
// validation rules that may be amended to spec in the future.
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -247,7 +247,36 @@ export type {
 export {
   validate,
   ValidationContext,
+
+  // All validation rules in the GraphQL Specification.
   specifiedRules,
+
+  // Individual validation rules.
+  ArgumentsOfCorrectTypeRule,
+  DefaultValuesOfCorrectTypeRule,
+  FieldsOnCorrectTypeRule,
+  FragmentsOnCompositeTypesRule,
+  KnownArgumentNamesRule,
+  KnownDirectivesRule,
+  KnownFragmentNamesRule,
+  KnownTypeNamesRule,
+  LoneAnonymousOperationRule,
+  NoFragmentCyclesRule,
+  NoUndefinedVariablesRule,
+  NoUnusedFragmentsRule,
+  NoUnusedVariablesRule,
+  OverlappingFieldsCanBeMergedRule,
+  PossibleFragmentSpreadsRule,
+  ProvidedNonNullArgumentsRule,
+  ScalarLeafsRule,
+  UniqueArgumentNamesRule,
+  UniqueDirectivesPerLocationRule,
+  UniqueFragmentNamesRule,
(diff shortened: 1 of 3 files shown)


wincent:
Looks reasonable to me.

We're dumping a lot of things in a flat export namespace now, so I would probably lean towards more descriptive names, like, `defaultValidationRules` instead of `specifiedRules`, and maybe even `KnownFragmentNamesValidation` instead of `KnownFragmentNames` etc.

leebyron:
I'd like to keep the names consistent for now to avoid issues with migrating between imports, but I think that a renaming pass should be something we consider in the future - though that may come with a cost of helping everyone migrate to those names.

wincent:
Yeah, consistency is important, as is not breaking shit for people, but I'm pretty sure we'll want to do this at some point because the mess will eventually become unmanageable, and the longer we delay, the greater the pain (I'd probably trade the consistency today for reducing future pain TBH, but up to you).

leebyron:
Ok I agree - let's go for minor inconsistency now. Considering that reaching in to import the validation rules directly was not 100% supported before, this isn't really a big deal. Since `specifiedRules` was already exported I didn't rename that to avoid a breaking change, though agree a more specific `specifiedValidationRules` would have been more clear.

Once this has been out in a release for a while we could do a follow up that just achieves internal consistency with adding the `Rule` suffix to these functions.

src/index.js
```diff
@@ -247,7 +247,36 @@ export type {
...
+  KnownDirectivesRule,
+  UniqueDirectivesPerLocationRule,
+  KnownArgumentNamesRule,
+  UniqueArgumentNamesRule,
+  ArgumentsOfCorrectTypeRule,
+  ProvidedNonNullArgumentsRule,
+  DefaultValuesOfCorrectTypeRule,
+  VariablesInAllowedPositionRule,
+  OverlappingFieldsCanBeMergedRule,
+  UniqueInputFieldNamesRule,
```
wincent: I'd alphabetize these to make it obvious where to insert new rules in the future.

PR #812, opened 2017-04-17
Forbid 'true', 'false' and 'null' as names for Enum value

IvanGoncharov:
From [spec](https://github.com/facebook/graphql/blob/master/spec/Section%202%20--%20Language.md#enum-value):
```
EnumValue : Name but not true, false or null
```

\+ Move one of test cases into proper location

### DIFF
--- a/src/type/__tests__/validation-test.js
+++ b/src/type/__tests__/validation-test.js
@@ -1188,6 +1188,43 @@ describe('Type System: Enum types must be well defined', () => {
     );
   });
 
+  it('rejects an Enum type with incorrectly named values', () => {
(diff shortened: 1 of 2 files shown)


leebyron:
Awesome, thank you!

smolinari:
Um, there is a small grammar mistake in the type definition.

`Name "${valueName}" is can not be used for Enum value.`

"is" is too much. It might also be better this way

`Name "${valueName}" can not be used as an Enum value.`

😄 

Scott

IvanGoncharov:
@smolinari Thank you. I created separate PR for grammar fix.

PR #817, opened 2017-04-19
Add buildASTSchemaComponents function

IvanGoncharov:
At the moment you can use IDL only to describe entire Schema.
But in a few our projects we need to build only directives and associated types from IDL.
We work around this by appending following to the IDL:
```
 type Query {
    dummy: String
 }
```
This PR splits out the major part of `buildASTSchema` function code and allows to get types and directives before they are passed to `GraphQLSchema` constructor.

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -38,6 +38,9 @@ import type {
 } from '../language/ast';
 
 import { GraphQLSchema } from '../type/schema';
+import type {
+  GraphQLSchemaComponents,
+} from '../type/schema';
 
 import {
   GraphQLString,
@@ -127,6 +130,43 @@ function getNamedTypeNode(typeNode: TypeNode): NamedTypeNode {
  * has no resolve methods, so execution will use default resolvers.
  */
 export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
+  const config: any = buildASTSchemaComponents(ast);
+  const directives = config.directives;
+
+  if (config.query === null) {
+    throw new Error(
+      'Must provide schema definition with query type or a type named Query.'
+    );
+  }
+
+  // If specified directives were not explicitly declared, add them.
+  if (!directives.some(directive => directive.name === 'skip')) {
+    directives.push(GraphQLSkipDirective);
+  }
+
+  if (!directives.some(directive => directive.name === 'include')) {
+    directives.push(GraphQLIncludeDirective);
+  }
+
+  if (!directives.some(directive => directive.name === 'deprecated')) {
+    directives.push(GraphQLDeprecatedDirective);
+  }
+
+  return new GraphQLSchema(config);
+}
+
+/**
+ * This takes the ast of a schema document produced by the parse function in
+ * src/language/parser.js.
+ *
+ * If no schema definition is provided, then it will look for types named Query
+ * and Mutation.
+ *
+ * Given that AST it return object that can be used to construct a GraphQLSchema
+ * instance.
+ */
+export function buildASTSchemaComponents(ast: DocumentNode):
+GraphQLSchemaComponents {
   if (!ast || ast.kind !== Kind.DOCUMENT) {
     throw new Error('Must provide a document ast.');
   }
@@ -214,12 +254,6 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
     }
   }
 
-  if (!queryTypeName) {
-    throw new Error(
-      'Must provide schema definition with query type or a type named Query.'
-    );
-  }
-
   const innerTypeMap = {
     String: GraphQLString,
     Int: GraphQLInt,
@@ -237,24 +271,12 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
   };
 
   const types = typeDefs.map(def => typeDefNamed(def.name.value));
-
   const directives = directiveDefs.map(getDirective);
 
-  // If specified directives were not explicitly declared, add them.
-  if (!directives.some(directive => directive.name === 'skip')) {
-    directives.push(GraphQLSkipDirective);
-  }
-
-  if (!directives.some(directive => directive.name === 'include')) {
-    directives.push(GraphQLIncludeDirective);
-  }
-
-  if (!directives.some(directive => directive.name === 'deprecated')) {
-    directives.push(GraphQLDeprecatedDirective);
-  }
-
-  return new GraphQLSchema({
-    query: getObjectType(nodeMap[queryTypeName]),
+  return {
+    query: queryTypeName ?
+      getObjectType(nodeMap[queryTypeName]) :
+      null,
     mutation: mutationTypeName ?
       getObjectType(nodeMap[mutationTypeName]) :
       null,
@@ -264,7 +286,7 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
     types,
     directives,
     astNode: schemaDef,
-  });
+  };
 
   function getDirective(
     directiveNode: DirectiveDefinitionNode
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -224,13 +224,20 @@ export class GraphQLSchema {
 
 type TypeMap = { [typeName: string]: GraphQLNamedType };
 
-type GraphQLSchemaConfig = {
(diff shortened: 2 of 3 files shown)


src/utilities/buildASTSchema.js
```diff
@@ -132,10 +132,9 @@ function getNamedTypeNode(typeNode: TypeNode): NamedTypeNode {
  * If no schema definition is provided, then it will look for types named Query
  * and Mutation.
  *
- * Given that AST it constructs a GraphQLSchema. The resulting schema
- * has no resolve methods, so execution will use default resolvers.
+ * Given that AST it return object that can be used to construct GraphQLSchema.
  */
-export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
+export function buildASTSchemaConfig(ast: DocumentNode) {
```
wincent: I think we should define and specify a return type here (`GraphQLSchemaConfig`, perhaps?).

src/utilities/buildASTSchema.js
```diff
@@ -466,6 +447,42 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
...
+ * src/language/parser.js.
+ *
+ * If no schema definition is provided, then it will look for types named Query
+ * and Mutation.
+ *
+ * Given that AST it constructs a GraphQLSchema. The resulting schema
+ * has no resolve methods, so execution will use default resolvers.
+ */
+export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
+  const config:any = buildASTSchemaConfig(ast);
```
wincent: nit: if we were to keep this `any` (and I don't think we should), we'd want a space after the colon.

src/utilities/buildASTSchema.js
```diff
@@ -132,10 +132,9 @@ function getNamedTypeNode(typeNode: TypeNode): NamedTypeNode {
  * If no schema definition is provided, then it will look for types named Query
  * and Mutation.
  *
- * Given that AST it constructs a GraphQLSchema. The resulting schema
- * has no resolve methods, so execution will use default resolvers.
+ * Given that AST it return object that can be used to construct GraphQLSchema.
```
wincent: spelling: return → returns

suggest: "construct GraphQLSchema" → "construct a GraphQLSchema instance"

IvanGoncharov: 👍 Fixed

src/utilities/buildASTSchema.js
IvanGoncharov: @wincent I can't return `GraphQLSchemaConfig` since it require `query` property to be not null:
```
type GraphQLSchemaConfig = {
  query: GraphQLObjectType;
  mutation?: ?GraphQLObjectType;
```
So it should some type identical to `GraphQLSchemaConfig` but with `query` defined as:
```
query: ?GraphQLObjectType;
```
Can you please suggest best way how to do this in Flow?

src/utilities/buildASTSchema.js
IvanGoncharov: @wincent I added space. As for using proper type it's the same problem with typing that I described in my previous comment :(

src/utilities/buildASTSchema.js
wincent: Yeah, sorry: I didn't mean literally `GraphQLSchemaConfig`; I meant some name like it.

In the worst case you define two types with different names. Better though, would be to use an intersection type:

```
type Base = {
  mutation?: ?GraphQLObjectType;
  subscription?: ?GraphQLObjectType;
  types?: ?Array<GraphQLNamedType>;
  directives?: ?Array<GraphQLDirective>;
};

type A = Base & {
  query: GraphQLObjectType;
};

type B = Base & {
  query?: GraphQLObjectType;
};
```

(Leaving the actual names of `A` and `B` out just for illustration.)

IvanGoncharov: @wincent Thank you. I used "GraphQLSchemaComponents" as the name for the new type and also rename function to `buildASTSchemaComponents`.

src/utilities/buildASTSchema.js
IvanGoncharov: @wincent I tried to make `config: GraphQLSchemaComponents` but it sims that Flow don't recognize runtime checks like this one: https://github.com/graphql/graphql-js/pull/817/files#diff-2691b670eb7e51187edc1f918c624a70R468
So when I call `new GraphQLSchema(config)` flow produce an error saying `query` is missing. Can you please advice on how to make this possible in Flow?

IvanGoncharov: @wincent I think it last change suggested in your review. Can you please advice how it can be solved in Flow?

src/utilities/buildASTSchema.js
```diff
@@ -466,6 +452,42 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
...
+ * This takes the ast of a schema document produced by the parse function in
+ * src/language/parser.js.
+ *
+ * If no schema definition is provided, then it will look for types named Query
+ * and Mutation.
+ *
+ * Given that AST it constructs a GraphQLSchema. The resulting schema
+ * has no resolve methods, so execution will use default resolvers.
+ */
+export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
```
leebyron: This should be the first function in the file since its eponymous.

src/utilities/buildASTSchema.js
```diff
@@ -466,6 +452,42 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
...
+  const config: any = buildASTSchemaComponents(ast);
```
leebyron: Shouldn't need an `:any` cast here, at least not without clear comment as to why. Seems unsound

IvanGoncharov:
> I'm curious - if you only expect to parse types and directives and explicitly working around not wanting to provide a schema definition, why also parse the schema definition in this new function?

> Is the intent that you would use the response of it to call new GraphQLSchema()?
It doesn't sound like it from the PR description.

In PR description I only described the main problem we try to address with this PR. At the same time, sometimes we need to inject type or directive into schema build from user-supplied IDL. Currently, we use string concatenation and `concatAST` but it addresses only simple scenarios where we injecting static definition for which we already have IDL. In our new library, we want to analyze existing type and have the ability to inject types generated based on this analyzes.

I tried to address both problems in a single function but I'm happy to split it into `buildASTDefinitions` and `buildASTSchemaComponents` if needed.

> Also, this is also including the standard built-in types in the return, is that what you need from this function? It's a bit strange that built-in types are added but built-in directives are not.

No, it doesn't include standard built-in types. Please see this test-case: https://github.com/graphql/graphql-js/pull/817/files#diff-59b5c9e589eeac323b0666278abb95e9R47
Build types are used only to fill `type` properties of fields and arguments.

src/utilities/buildASTSchema.js
IvanGoncharov: Thanks, done.

src/utilities/buildASTSchema.js
IvanGoncharov: @leebyron  I tried to make config: GraphQLSchemaComponents but it sims that Flow don't recognize runtime checks like this one: https://github.com/graphql/graphql-js/pull/817/files#diff-2691b670eb7e51187edc1f918c624a70R145
So when I call new GraphQLSchema(config) flow produce an error saying `query` is missing. Can you please advice on how to make this possible in Flow?

src/utilities/buildASTSchema.js
```diff
@@ -127,6 +130,43 @@ function getNamedTypeNode(typeNode: TypeNode): NamedTypeNode {
  * has no resolve methods, so execution will use default resolvers.
  */
 export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
+  const config: any = buildASTSchemaComponents(ast);
```
mjmahone: I believe you can fix the any check with:
```
const config = buildASTSchemaComponents(ast);
if (!config.query) {
  throw new GraphQLError(
    'Must provide schema definition with query type or a type named Query.',
    ast,
  );
}
...
return new GraphQLSchema(config);
```

I believe the issue for flow is that `config.query` can be `undefined` in addition to `null`.

IvanGoncharov:
Closing since now we can create schemas without root type.

PR #837, opened 2017-04-29
Spec compliance: coercion of Int values

### LINKED ISSUE #827: `Int` input value coercion from floating point JSON literal value
According to the spec (https://facebook.github.io/graphql/#sec-Scalars):

> Since some common serializations (ex. JSON) do not discriminate between integer and floating‐point values, they are interpreted as an integer input value if they have an empty fractional part (ex. 1.0) and otherwise as floating‐point input value.

Based on this description, when I define a value in JSON for a variable of an `Int` type:

* `123.0` must be considered a valid value and coerce to `123`
* `123.456` must be considered an invalid value and should result in error based on `Int` input value coercion rules (https://facebook.github.io/graphql/#sec-Int):

    > When expected as an input type, only integer input values are accepted. All other input values, including strings with numeric content, must raise a query error indicating an incorrect type.

Do I understand it correctly?

I discovered that reference implementation allows `123.456` value for a variable of `Int` type which does not comply to the mentioned parts of the spec. Here is the line that successfully drops fraction digits:

https://github.com/graphql/graphql-js/blob/master/src/type/scalars.js#L30

leebyron:
This fixes an issue with coercing Int values when provided a floating point value which disagrees with the spec definition.

Updates test cases which were along the same incorrect line.

Note: this is technically a breaking change

Fixes #827

### DIFF
--- a/src/type/__tests__/serialization-test.js
+++ b/src/type/__tests__/serialization-test.js
@@ -33,17 +33,30 @@ describe('Type System: Scalar coercion', () => {
       GraphQLInt.serialize(-1)
     ).to.equal(-1);
     expect(
-      GraphQLInt.serialize(0.1)
-    ).to.equal(0);
-    expect(
-      GraphQLInt.serialize(1.1)
-    ).to.equal(1);
-    expect(
-      GraphQLInt.serialize(-1.1)
-    ).to.equal(-1);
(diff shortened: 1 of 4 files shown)


leebyron:
Merging for next major version release

PR #838, opened 2017-04-30
Make 'errors' top property in the response object.

IvanGoncharov:
It’s hard to notice `errors` property if it is hidden under multiple screens of JSON.
For example, if you run [this query](http://graphql.org/swapi-graphql/?query=%7B%0A%20%20allPeople%20%7B%0A%20%20%20%20people%20%7B%0A%20%20%20%20%20%20height%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A) you will see the following:
image
And only if you scroll a couple pages down you will see an error what is bad DX.
To fix this I just changed the order of `errors` and `data` fields so result of the same query look like this:
image

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -96,12 +96,12 @@ type ExecutionContext = {
 /**
  * The result of GraphQL execution.
  *
- *   - `data` is the result of a successful execution of the query.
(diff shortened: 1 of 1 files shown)


leebyron:
I'm into it. Good idea

PR #839, opened 2017-05-01
[RFC] Environment var to silence warning about invalid names.

leebyron:
This provides an opt-out of a spec compliance naming style for legacy schema


wincent:
Thanks @leebyron. I'll just add a quick test before merging this.

PR #848, opened 2017-05-13
Fix input coercion to allow null-valued Enums as arguments

### LINKED ISSUE #840: 0.9.5 broke Enum null values provided as arguments
A [GraphBrainz](https://github.com/exogen/graphbrainz) tests started failing with v0.9.5 and above. I have an enum like:

```js
export const CoverArtImageSize = new GraphQLEnumType({
  name: 'CoverArtImageSize',
  description: `The image sizes that may be requested at the [Cover Art
Archive](https://musicbrainz.org/doc/Cover_Art_Archive).`,
  values: {
    SMALL: {
      name: 'Small',
      description: 'A maximum dimension of 250px.',
      value: 250
    },
    LARGE: {
      name: 'Large',
      description: 'A maximum dimension of 500px.',
      value: 500
    },
    FULL: {
      name: 'Full',
      description: 'The image’s original dimensions, with no maximum.',
      value: null
    }
  }
})
```

(Yes, I had this `null` valued Enum even before it was 'officially' supported, and simply worked around it by checking for `'FULL'` in the resolver.)

As of v0.9.5, I'm no longer able to supply `FULL` as an argument, as in this query:

```graphql
  {
    lookup {
      release(mbid: "b84ee12a-09ef-421b-82de-0441a926375b") {
        coverArt {
          front(size: LARGE)
          back(size: SMALL)
          fullFront: front(size: FULL)
        }
      }
    }
  }
```

This now results in an error from the server:

```
GraphQLError: Argument "size" has invalid value FULL.
Expected type "CoverArtImageSize", found FULL.
```

But clearly `FULL` is a valid value. It seems the new `null` support from #836 didn't extend support to arguments and resulted in breaking what was at least workaround-able.

(I can see how this could indeed pose problems for people when determining "no argument provided" vs. "null/undefined enum value provided", but that seems solvable by checking for `'arg' in args` and making sure `graphql-js` only populates passed-in args.)

exogen:
This fixes my issue from #840 where null-valued Enums (supported since 0.9.5) were rejected when provided as arguments.

This reveals some minor code duplication in `isValidLiteralValue` and `valueFromAST`.

There are likely other places in the code where null-valued Enums cause problems; for example, in `isValidJSValue`: if a type is `GraphQLNonNull(EnumWithNullValue)`, then `null` should probably be accepted, since it should end up resolving to a name in the Enum and not actually `null`. But I didn't attempt to chase down all those edge cases; these changes are enough to get my test suite to pass.

Fixes #840.

### DIFF
--- a/src/utilities/valueFromAST.js
+++ b/src/utilities/valueFromAST.js
@@ -152,9 +152,11 @@ export function valueFromAST(
   );
 
   const parsed = type.parseLiteral(valueNode);
-  if (isNullish(parsed)) {
-    // null or invalid values represent a failure to parse correctly,
-    // in which case no value is returned.
+  if (type instanceof GraphQLEnumType ?
+      !type.serialize(parsed) : isNullish(parsed)) {
+    // null or invalid values represent a failure to parse correctly (unless
+    // we have a legitimately null-valued Enum), in which case no value is
+    // returned.
     return;
   }
 
--- a/src/utilities/__tests__/valueFromAST-test.js
+++ b/src/utilities/__tests__/valueFromAST-test.js
@@ -66,7 +66,13 @@ describe('valueFromAST', () => {
 
   const testEnum = new GraphQLEnumType({
     name: 'TestColor',
-    values: { RED: { value: 1 }, GREEN: { value: 2 }, BLUE: { value: 3 } }
+    values: {
+      RED: { value: 1 },
+      GREEN: { value: 2 },
+      BLUE: { value: 3 },
+      NULL: { value: null },
+      UNDEFINED: { value: undefined }
+    }
   });
 
   it('converts enum values according to input coercion rules', () => {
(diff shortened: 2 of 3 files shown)


exogen:
Add higher level tests, for `ArgumentsOfCorrectType` specifically.

src/utilities/valueFromAST.js
```diff
@@ -152,9 +152,11 @@ export function valueFromAST(
   );
 
   const parsed = type.parseLiteral(valueNode);
-  if (isNullish(parsed)) {
-    // null or invalid values represent a failure to parse correctly,
-    // in which case no value is returned.
+  if (type instanceof GraphQLEnumType ?
+      !type.serialize(parsed) : isNullish(parsed)) {
```
leebyron: Hmm this condition is a little gross now. I wonder if we should instead have something like `type.isValidValue()` or something along those lines to simplify this?

PR #867, opened 2017-05-18
Allow providing an object of arguments to graphql().

### LINKED ISSUE #356: Accept an object as the argument to graphql()

leebyron:
As well as execute() and subscribe()

Closes #356

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -202,7 +202,7 @@
     "quote-props": [2, "as-needed", {"numbers": true}],
     "quotes": [2, "single"],
     "radix": 2,
-    "require-yield": 2,
+    "require-yield": 0,
     "semi": [2, "always"],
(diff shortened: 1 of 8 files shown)


leebyron:
Planning changes: needs tests

taion:
This is basically the best thing ever. Thank you!

Issue #519, opened 2016-10-16
`findBreakingChanges` function does not find all possible breaking chnges


OlegIlyenko:
`findBreakingChanges` function was introduced in https://github.com/graphql/graphql-js/pull/508. Unfortunately it does not cover all of the cases. 

For example:
- it does not consider field arguments at all
- adding a not-null input field is a breaking change (the same is true for arguments)

In some cases it may even produce false positives. For instance, changing the type of output field `T → T!` (making it a not-null field) is a backwards compatible change. The opposite is true for input fields: `T! → T` (making type nullable) is also backwards compatible change.

I ported this utility function to sangria and improved it a bit, including fixes for mentioned issues (I collect non-breaking changes as well). Maybe these test cases would be helpful (though I still not 100% sure that I have considered all possible breaking changes too):

https://github.com/sangria-graphql/sangria/blob/master/src/test/scala/sangria/schema/SchemaComparatorSpec.scala

leebyron:
cc @sam-swarr to look into this.

sam-swarr:
Thanks @OlegIlyenko! I'll try to get a PR out to fix the issues you found, else I'm happy to review if someone beats me to it.

joelgriffith:
Fixed in https://github.com/graphql/graphql-js/pull/701

cc @sam-swarr

joelgriffith:
Once `0.9.3` is out, we can probably consider closing this?

OlegIlyenko:
Thanks a lot @joelgriffith for your work! 🏆 

There are still some things missing (like directives, scalar values, etc.), but as far as this particular issue is concerned, it's implemented. I guess it would be to create new issues as new types of diffs are needed.

wincent:
> I guess it would be to create new issues as new types of diffs are needed.

Sounds like a good idea to me.

PR #874, opened 2017-05-22
Add more functionality to findBreakingChanges

sam-swarr:
This PR shores up more of the holes in findBreakingChanges that were pointed out by @OlegIlyenko in https://github.com/graphql/graphql-js/issues/519.

Here's a list of things that were fixed or added:
  - breaking changes to to a field or field arg were not properly detected if the change involved a type modifer like List or NonNull (for example, changing a field from String to [String])
    - certain changes are still considered safe, such as moving from NonNull to nullable of the same type for input fields and field args as well as moving from nullable to NonNull for fields on Object and Interface types
  - adding a NonNull field arg is now detected as a breaking change
  - adding a NonNull field to an input type is now detected as a breaking change
  - breaking changes to field args on Interface types are now considered
  - removing an Interface type from the list of interfaces an Object type implements is now detected as a breaking change

### DIFF
--- a/src/utilities/findBreakingChanges.js
+++ b/src/utilities/findBreakingChanges.js
@@ -9,16 +9,22 @@
  */
 
 import {
-  getNamedType,
+  isNamedType,
   GraphQLScalarType,
   GraphQLEnumType,
   GraphQLInputObjectType,
   GraphQLInterfaceType,
+  GraphQLList,
+  GraphQLNonNull,
   GraphQLObjectType,
   GraphQLUnionType,
 } from '../type/definition';
 
-import type { GraphQLNamedType, GraphQLFieldMap } from '../type/definition';
+import type {
+  GraphQLNamedType,
+  GraphQLFieldMap,
+  GraphQLType,
+} from '../type/definition';
 
 import { GraphQLSchema } from '../type/schema';
 
@@ -31,6 +37,9 @@ export const BreakingChangeType = {
   VALUE_REMOVED_FROM_ENUM: 'VALUE_REMOVED_FROM_ENUM',
   ARG_REMOVED: 'ARG_REMOVED',
   ARG_CHANGED_KIND: 'ARG_CHANGED_KIND',
+  NON_NULL_ARG_ADDED: 'NON_NULL_ARG_ADDED',
+  NON_NULL_INPUT_FIELD_ADDED: 'NON_NULL_INPUT_FIELD_ADDED',
+  INTERFACE_REMOVED_FROM_OBJECT: 'INTERFACE_REMOVED_FROM_OBJECT',
 };
 
 export const DangerousChangeType = {
@@ -62,6 +71,7 @@ export function findBreakingChanges(
     ...findTypesRemovedFromUnions(oldSchema, newSchema),
     ...findValuesRemovedFromEnums(oldSchema, newSchema),
     ...findArgChanges(oldSchema, newSchema).breakingChanges,
+    ...findInterfacesRemovedFromObjectTypes(oldSchema, newSchema),
   ];
 }
 
@@ -153,7 +163,8 @@ export function findArgChanges(
     const oldType = oldTypeMap[typeName];
     const newType = newTypeMap[typeName];
     if (
-      !(oldType instanceof GraphQLObjectType) ||
+      !(oldType instanceof GraphQLObjectType ||
+        oldType instanceof GraphQLInterfaceType) ||
       !(newType instanceof oldType.constructor)
     ) {
       return;
@@ -169,46 +180,50 @@ export function findArgChanges(
 
       oldTypeFields[fieldName].args.forEach(oldArgDef => {
         const newArgs = newTypeFields[fieldName].args;
-        const newTypeArgIndex = newArgs.findIndex(
+        const newArgDef = newArgs.find(
           arg => arg.name === oldArgDef.name
         );
-        const newArgDef = newArgs[newTypeArgIndex];
-
-        const oldArgTypeName = getNamedType(oldArgDef.type);
-        const newArgTypeName = newArgDef ?
-          getNamedType(newArgDef.type) :
-          null;
-
-        if (!oldArgTypeName) {
-          return;
-        }
 
         // Arg not present
-        if (!newArgTypeName) {
+        if (!newArgDef) {
           breakingChanges.push({
             type: BreakingChangeType.ARG_REMOVED,
             description: `${oldType.name}.${fieldName} arg ` +
               `${oldArgDef.name} was removed`,
           });
-
-        // Arg changed type in a breaking way
-        } else if (
-          oldArgTypeName.name !== newArgTypeName.name
-        ) {
-          breakingChanges.push({
-            type: BreakingChangeType.ARG_CHANGED_KIND,
-            description: `${oldType.name}.${fieldName} arg ` +
-              `${oldArgDef.name} has changed type from ` +
-              `${oldArgDef.type.toString()} to ${newArgDef.type.toString()}`,
-          });
-
-        // Arg default value has changed
-        } else if (oldArgDef.defaultValue !== undefined &&
+        } else {
+          const isSafe = isChangeSafeForInputObjectFieldOrFieldArg(
+            oldArgDef.type,
+            newArgDef.type,
+          );
+          if (!isSafe) {
+            breakingChanges.push({
+              type: BreakingChangeType.ARG_CHANGED_KIND,
+              description: `${oldType.name}.${fieldName} arg ` +
+                `${oldArgDef.name} has changed type from ` +
+                `${oldArgDef.type.toString()} to ${newArgDef.type.toString()}`,
+            });
+          } else if (oldArgDef.defaultValue !== undefined &&
           oldArgDef.defaultValue !== newArgDef.defaultValue) {
-          dangerousChanges.push({
-            type: DangerousChangeType.ARG_DEFAULT_VALUE_CHANGE,
-            description: `${oldType.name}.${fieldName} arg ${oldArgDef.name} ` +
-              'has changed defaultValue',
+            dangerousChanges.push({
+              type: DangerousChangeType.ARG_DEFAULT_VALUE_CHANGE,
+              description: `${oldType.name}.${fieldName} arg ` +
+                `${oldArgDef.name} has changed defaultValue`,
+            });
+          }
+        }
+      });
+      // Check if a non-null arg was added to the field
+      newTypeFields[fieldName].args.forEach(newArgDef => {
+        const oldArgs = oldTypeFields[fieldName].args;
+        const oldArgDef = oldArgs.find(
+          arg => arg.name === newArgDef.name
+        );
+        if (!oldArgDef && newArgDef.type instanceof GraphQLNonNull) {
+          breakingChanges.push({
+            type: BreakingChangeType.NON_NULL_ARG_ADDED,
+            description: `A non-null arg ${newArgDef.name} on ` +
+              `${newType.name}.${fieldName} was added`,
           });
         }
       });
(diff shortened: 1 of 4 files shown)


src/utilities/findBreakingChanges.js
```diff
@@ -153,7 +160,8 @@ export function findArgChanges(
     const oldType = oldTypeMap[typeName];
     const newType = newTypeMap[typeName];
     if (
-      !(oldType instanceof GraphQLObjectType) ||
+      !(oldType instanceof GraphQLObjectType ||
+        oldType instanceof GraphQLInterfaceType) ||
```
sam-swarr: We weren't considering field arg changes on Interface types at all.

src/utilities/findBreakingChanges.js
```diff
@@ -169,46 +177,69 @@ export function findArgChanges(
...
-          breakingChanges.push({
-            type: BreakingChangeType.ARG_CHANGED_KIND,
-            description: `${oldType.name}.${fieldName} arg ` +
-              `${oldArgDef.name} has changed type from ` +
-              `${oldArgDef.type.toString()} to ${newArgDef.type.toString()}`,
-          });
-
-        // Arg default value has changed
-        } else if (oldArgDef.defaultValue !== undefined &&
+        } else {
```
sam-swarr: The logic before was only comparing the results of `getNamedType` which meant we weren't detecting breaking changes when dealing with type modifiers like NonNull or List. The logic unfortunately becomes more complex, but more correct.

src/utilities/findBreakingChanges.js
```diff
@@ -278,18 +310,74 @@ export function findFieldsThatChangedType(
           description: `${typeName}.${fieldName} was removed.`,
         });
       } else {
-        // Check if the field's type has changed in the new schema.
-        const oldFieldType = getNamedType(oldTypeFieldsDef[fieldName].type);
```
sam-swarr: Same deal as the comment in `findArgChanges`. We were previously only comparing the results of `getNamedType` which didn't properly consider changes involving List/NonNull.

src/utilities/findBreakingChanges.js
```diff
@@ -169,46 +177,69 @@ export function findArgChanges(
...
+                newArgDef.type instanceof GraphQLNonNull &&
+                getNamedType(oldArgDef.type) === getNamedType(newArgDef.type)
+              ) ||
+              // moving from non-null to nullable of the same type is safe for
+              // field args
+              (
+                !(newArgDef.type instanceof GraphQLNonNull) &&
+                isNamedType(newArgDef.type) &&
+                getNamedType(oldArgDef.type).name ===
+                  getNamedType(newArgDef.type).name
```
wincent: Again, do we need `getNamedType(newArgDef.type)` here only because of Flow? (seeing as we already check `isNamedType` above).

src/utilities/findBreakingChanges.js
```diff
@@ -278,18 +310,74 @@ export function findFieldsThatChangedType(
...
+              isNamedType(newFieldType) &&
+              getNamedType(oldFieldType).name ===
+                getNamedType(newFieldType).name
+            ) ||
+            // moving from nullable to non-null of the same type for a field on
+            // an object or interface type is safe
+            (
+              !(oldType instanceof GraphQLInputObjectType) &&
+              newFieldType instanceof GraphQLNonNull &&
+              isNamedType(oldFieldType) &&
```
wincent: You already checked `isNamedType(oldFieldType)` on line 317 above.

src/utilities/findBreakingChanges.js
```diff
@@ -278,18 +310,74 @@ export function findFieldsThatChangedType(
...
+              getNamedType(oldFieldType).name ===
+                getNamedType(newFieldType).name
+            );
+        } else if (oldFieldType instanceof GraphQLList) {
+          isSafe = newFieldType instanceof GraphQLList &&
+            getNamedType(oldFieldType) === getNamedType(newFieldType);
```
wincent: Will `[Int]` to `[Int!]` sneak through here?

src/utilities/findBreakingChanges.js
```diff
@@ -278,18 +310,74 @@ export function findFieldsThatChangedType(
...
+                getNamedType(newFieldType).name
+            ) ||
+            // moving from nullable to non-null of the same type for a field on
+            // an object or interface type is safe
+            (
+              !(oldType instanceof GraphQLInputObjectType) &&
+              newFieldType instanceof GraphQLNonNull &&
+              isNamedType(oldFieldType) &&
+              getNamedType(oldFieldType).name ===
+                getNamedType(newFieldType).name
```
wincent: Will `Int` to `[Int]!` sneak through here?

sam-swarr:
Thanks for the feedback! I updated the PR to properly handle complex nestings of NonNull/List types. To do so, I broke apart `findFieldsThatChangedType` into two separate helpers `findFieldsThatChangedTypeOnObjectOrInterfaceTypes` and `findFieldsThatChangedTypeOnInputObjectTypes` and each of those delegates the tricky type comparison logic to `isChangeSafeForObjectOrInterfaceField` and `isChangeSafeForInputObjectFieldOrFieldArg` respectively.

The key to handling arbitrarily nested composite types was to make the `isChangeSafe...` functions recursive. Also by factoring them out I could reuse `isChangeSafeForInputObjectFieldOrFieldArg` in both the input type field and field arg checkers since logically they are analogous.

sam-swarr:
Also added a bunch more test cases to check for the cases Greg mentioned.

src/utilities/findBreakingChanges.js
```diff
@@ -278,22 +354,128 @@ export function findFieldsThatChangedType(
...
+      }
+    });
   });
   return breakingFieldChanges;
 }
 
+function isChangeSafeForObjectOrInterfaceField(
+  oldType: GraphQLType,
+  newType: GraphQLType,
+): boolean {
```
leebyron: These two functions are pretty hard to read. You don't need to track this mutable `isSafe` variable, you could just return after each condition with standard `if` statements.

src/utilities/findBreakingChanges.js
```diff
@@ -278,22 +354,128 @@ export function findFieldsThatChangedType(
...
+  let isSafe = false;
+  if (isNamedType(oldType)) {
+    isSafe =
+      // if they're both named types, see if their names are equivalent
+      (
+        isNamedType(newType) && oldType.name === newType.name
+      ) ||
```
leebyron: I'm not sure this condition is doing what the comment above it is suggesting. If this is a named type, I would expect differing names to immediately return false, instead what is happening is that if falls through such that `isChangeSafeForObjectOrInterfaceField` is called.

src/utilities/findBreakingChanges.js
```diff
@@ -363,3 +545,36 @@ export function findValuesRemovedFromEnums(
...
+  const oldTypeMap = oldSchema.getTypeMap();
+  const newTypeMap = newSchema.getTypeMap();
+  const breakingChanges = [];
+
+  Object.keys(oldTypeMap).forEach(typeName => {
+    const oldType = oldTypeMap[typeName];
+    const newType = newTypeMap[typeName];
+    if (
+      !(oldType instanceof GraphQLObjectType) ||
+      !(newType instanceof oldType.constructor)
```
leebyron: I see this condition or variants like it throughout this code and am confused as to what it's attempting to compare. IIUC, this condition applies if `oldType` is not an Object type, otherwise if it is an Object type it applies if `newType` is an instanceof the `oldType`'s constructor (which can only be `GraphQLObjectType` at that point). Could conditions like this be factored into descriptive functions?

src/utilities/findBreakingChanges.js
```diff
@@ -278,22 +354,128 @@ export function findFieldsThatChangedType(
...
+        !(newType instanceof GraphQLNonNull) &&
+        isChangeSafeForInputObjectFieldOrFieldArg(
+          oldType.ofType,
+          newType,
+        )
+      );
+  }
+  return isSafe;
+}
+
```
wincent: Great job on making this complicated code very readable. About the only way I can think of to make it more readable — maybe — would be to extract some smaller functions; eg:

```
function isChangeSafeForObjectOrInterfaceField(
  oldType: GraphQLType,
  newType: GraphQLType,
): boolean {
  const recurse = isChangeSafeForObjectOrInterfaceField;
  return (
    namedTypesMatch(oldType, newType, recurse) ||
    listTypesMatch(oldType, newType, recurse) ||
    nonNullTypesMatch(oldType, newType, recurse)
  );
}

function namedTypesMatch(
  oldType: GraphQLType,
  newType: GraphQLType,
  recurse: (oldType: GraphQLType, newType: GraphQLType) => boolean,
): boolean {
  return (
    isNamedType(oldType) && ((
      // If they're both named types, see if their names are equivalent.
      isNamedType(newType) &&
      oldType.name === newType.name
    ) || (
     // Moving from nullable to non-null of the same underlying type is safe.
     newType instanceof GraphQLNonNull &&
     recurse(oldType, newType.ofType)
    ))
  ); 
}

// etc
```

Subjective though.

src/utilities/findBreakingChanges.js
wincent: Funny. I just left this comment:

> Great job on making this complicated code very readable.

src/utilities/findBreakingChanges.js
wincent: It is though, because when the names don't match, it will check `newType instance of GraphQLNonNull` (and fail).

src/utilities/findBreakingChanges.js
wincent: My inclination was also to `return` immediately, but I know some people consider it cleaner if a function has exactly one exit point.

sam-swarr: The `isSafe` var is leftover from before these were factored out into separate functions. Updated so that we just have inline returns.

src/utilities/findBreakingChanges.js
sam-swarr: That comment only applies to the first part of the compound conditional. I moved the comment inside the parentheses to make it clearer.

But yeah the fall through to the second part of the conditional is intentional. If `newType` is not a named type, we also want to see if its a non-null type because the change will be safe if the type it wraps is compatible with the original type.

src/utilities/findBreakingChanges.js
sam-swarr: I would take your suggestion if it was possible to share the `namedTypesMatch`/`listTypesMatch`/`nonNullTypesMatch` functions. But since `Interface/Object Types` and `Input Types` have slightly different notions of what's safe regarding nullability changes, it's not possible to share those functions and that means we'd just have 6 more helper functions.

I think it's better off as it is.

src/utilities/findBreakingChanges.js
sam-swarr: Yeah this pattern makes more sense in some of the other functions like in `findFieldsThatChangedTypeOnObjectOrInterfaceTypes` where we have:
```
if (
  !(oldType instanceof GraphQLObjectType ||
  oldType instanceof GraphQLInterfaceType) ||
  !(newType instanceof oldType.constructor)
) {
  return;
}
...
```

We want to make sure that `oldType` is either a `GraphQLObjectType` or a `GraphQLInterfaceType` and that the `newType` is the same kind. (If they are different kinds than that is a more major breaking change that is caught by `findTypesThatChangedKind`.

In this case though, it does look odd when we're only checking for one particular kind of type. I changed it so that we also check that `!(newType instanceof GraphQLObjectType)`. (I also changed other places where there's only a single kind of type we want).

sam-swarr:
I'm kinda bad with git, so I apologize that I keep overwriting the commit with the updates rather than push them as separate commits so that you can see what's changed in each iteration. I keep having to force push the updates to my fork (`git push -f origin master`) cause I think I messed something up earlier...

sam-swarr:
Ok that sounds good to me. I'll let @leebyron sign off on it then one of you can merge.

PR #879, opened 2017-05-25
Update gitignore

baer:
This PR updates the `.gitignore` file with a version that is auto-generated by www.gitignore.io. 

https://www.gitignore.io/api/osx,node,linux,eclipse,windows,intellij

Hopefully this will help keep custom ignores at a minimum and keep the fuss over things to ignore to a minimum.


.gitignore
```diff
@@ -1,12 +1,259 @@
...
+# .idea/misc.xml
+# *.ipr
+
+# Sonarlint plugin
+.idea/sonarlint
+
+### Linux ###
 *~
-*.iml
-.*.haste_cache.*
```
baer: Note: This is the only line in this PR that is removed (and not just moved into another section). This appears to be a Jest specific cache file that does not apply to this project.

leebyron:
I would prefer not track all files relevant to editors and other software in this project. Ideally our .gitignore should get smaller, not larger.

Any custom ignores related to software you use on your own machine should go in your [global .gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

baer:
I was only trying to flesh out the things that had already been identified. Since you don't think editor specific things are appropriate here (I agree) I was going to suggest removing them but it looks like you did this yourself in master 👏  last night.

wincent:
Good resolution to this! Thanks @baer for triggering the discussion and @leebyron for cleaning up the file. 👍

PR #884, opened 2017-05-26
Add basic README for subscriptions

baer:
I was reading through the source and noticed that there was a README of this format in every directory but this one. For consistency, I thought I'd add it here too :). While I was at it, I made the doc references link to the individual docs..

### DIFF
--- a/src/subscription/README.md
+++ b/src/subscription/README.md
@@ -1,7 +1,7 @@
 GraphQL Subscription
 -----------------
 
-The `graphql/subscription` module is responsible subscribing to updates on specific data.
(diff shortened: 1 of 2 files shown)


src/subscription/README.md
```diff
@@ -0,0 +1,9 @@
+GraphQL Subscription
+-----------------
+
+The `graphql/subscription` module is responsible subscribing to updates on specific data.
```
wincent: "responsible for"

baer: 🤦‍♂️

leebyron:
Great thanks!

We probably want to discourage this kind of usage soon, but consistency now is great.

PR #885, opened 2017-05-26
fix flow errors from turning on const_params experimental flag

asiandrummer:
Fixed a bunch of flow errors observed after turning on `const_params` experimental flag, by delegating the correct arguments to another function (`Impl` functions).


leebyron:
Makes sense 👍

PR #886, opened 2017-05-26
Subscriptions: Report or Reject when encountering Errors.

leebyron:
Currently the `subscribe()` function throws Errors, however this is awkward when used along with async functions which would expect a rejected iteration to represent failure. Also GraphQLErrors should be reported back to the client since they represent client-provided errors.

Updates test cases to represent this new behavior.

Includes two new utilities `asyncIteratorReject` and `asyncIteratorResolve` to help implement this.

### DIFF
--- a/src/subscription/subscribe.js
+++ b/src/subscription/subscribe.js
@@ -10,19 +10,21 @@
  */
 
 import { isAsyncIterable } from 'iterall';
+import { GraphQLError } from '../error/GraphQLError';
 import {
   addPath,
   assertValidExecutionArguments,
   buildExecutionContext,
+  buildResolveInfo,
   collectFields,
   execute,
   getFieldDef,
   getOperationRootType,
-  buildResolveInfo,
   resolveFieldValueOrError,
 } from '../execution/execute';
 import { GraphQLSchema } from '../type/schema';
 import invariant from '../jsutils/invariant';
+import asyncIteratorReject from './asyncIteratorReject';
 import mapAsyncIterator from './mapAsyncIterator';
 
 import type { ExecutionResult } from '../execution/execute';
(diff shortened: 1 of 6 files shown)


leebyron:
Derived from #868

src/subscription/subscribe.js
```diff
@@ -106,16 +110,33 @@ function subscribeImpl(
...
+      contextValue,
+      variableValues,
+      operationName,
+      subscribeFieldResolver
+    );
+  } catch (error) {
+    // GraphQLError represent "field" or "query" errors, to be returned to
+    // the requesting client as a GraphQL Response. Other errors are "internal"
+    // errors, which cause the event stream to fail.
+    return error instanceof GraphQLError ?
```
robzhu: Can error be anything other than GraphQLError or Error here?

src/subscription/subscribe.js
```diff
@@ -106,16 +110,33 @@ function subscribeImpl(
   fieldResolver,
   subscribeFieldResolver
 ) {
-  const subscription = createSourceEventStream(
+  // If arguments are missing or incorrect, throw an early error.
+  assertValidExecutionArguments(
```
robzhu: This can throw. If we want to be consistent, should this be changed to also yield a rejected async iterator?

src/subscription/subscribe.js
leebyron: Error is a super-class of GraphQLError and typically covers everything else, but technically *anything* can be thrown, so what we want is to check if it is a GraphQLError or otherwise

src/subscription/subscribe.js
leebyron: I'll add more comments here - the idea is to throw an early error so you get an immediate error with stack since bad arguments would be programmer error, not a graphql query issue - and you want to know that ASAP, not waiting until you `.next().then()`

leebyron:
Yeah, it's a bit weird that `createSourceEventStream` throws instead of rejects, another option would be to extend the mapping function to map errors as well and do the detection and wrapping there. I'll investigate

leebyron:
Updated with an extension to `mapAsyncIterator`. I like this variant better, reminds me a lot of `.then(onVal, onReject)` and mirrors the behavior. It also keeps the `subscribe()` implementation very simple which was an original goal, and ensures the `createEventSource()` method behaves more like an Async Generator.

leebyron:

PR #887, opened 2017-05-26
Subscriptions: Respond with error when failing to create source

leebyron:
If a subscribe resolve function throws or returns an error, that typically indicates an issue to be returned to the requesting client. This coerces errors into located GraphQLErrors so they are correctly reported.

(Note: this is a stacked diff. It depends on #886 landing first)


leebyron:
Derived from #868

leebyron:

PR #888, opened 2017-05-27
Subscriptions: Test that source errors pass through.

leebyron:
After discussion in #868, decided that errors emitted from a source event stream should be considered "internal" errors and pass through.

However errors encountered during GraphQL execution on a source event should be cons
idered "field" or "query" errors and be represented within that Response.

(Note: this is a stacked diff. It depends on #887 landing first)


leebyron:

PR #889, opened 2017-05-27
Sort fields before printing to get a more stable diff

tberman:
Sort the fields of an object similar to how types are being sorted.

### DIFF
--- a/src/subscription/__tests__/subscribe-test.js
+++ b/src/subscription/__tests__/subscribe-test.js
@@ -585,7 +585,7 @@ describe('Subscribe', () => {
         invalidEmailSchema,
         ast
       );
-    }).to.throw('Subscription must return AsyncIterable. Received: test');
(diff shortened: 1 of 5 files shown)


src/subscription/__tests__/subscribe-test.js
```diff
@@ -585,7 +585,7 @@ describe('Subscribe', () => {
         invalidEmailSchema,
         ast
       );
-    }).to.throw('Subscription must return AsyncIterable. Received: test');
+    }).to.throw('Subscription must return Async Iterable. Received: test');
```
wincent: Thanks for fixing.

wincent:
Thnaks @tberman.

leebyron:
Unfortunately this PR broke how this utility is used for printing schema in tests and tools - the sorted behavior changed which fields appear first which some tools rely on.

I'm reverting this change, I suggest if stable diffs are required that fields are sorted while still in the schema rather than during printing of the schema.

PR #891, opened 2017-05-31
add test cases to check support of negative floats and ints

shnellpavel:
This tests are used to generate json with test cases for alternative realization of graphql (https://github.com/neelance/graphql-go). 
As for this repo pull request will add test case for addional variants of int and float values (such cases was not supported by alternative one).


wincent:
Thanks @shnellpavel. I just restarted the CI build as it was likely a spurious failure due to GitHub's downtime incident today.

PR #898, opened 2017-06-08
Expose 'getArgumentValues' method

IvanGoncharov:
Currently, it's the only method that allows parsing directives with Type Safety.
We are already using this method in multiple projects by importing from `graphql/execution/values.js` directly but it is not convenient (+ typings doesn't work).

### DIFF
--- a/src/execution/index.js
+++ b/src/execution/index.js
@@ -7,6 +7,7 @@
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
 
+export { getArgumentValues } from './values';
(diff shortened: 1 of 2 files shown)


leebyron:
Could you speak more to the use case? This is an implementation detail - I'd like to make sure we're exporting methods that are generically useful and best fit the problem to be solved.

IvanGoncharov:
@leebyron This PR preceded #904. As I understand #904 is accepted as an idea and it's the only question of finalizing function prototype. So I close this one in favor of #904.

PR #899, opened 2017-06-09
Add source filepath information to formatted errors

Slava:
Working on Relay Compiler, I noticed that the origin of the error messages is difficult to track.
This PR adds additional information about the source filepath if such is passed into the compiler.

I wasn't sure if it is better to add additional arguments or pass it as an option in the options object.

### DIFF
--- a/src/error/formatError.js
+++ b/src/error/formatError.js
@@ -18,10 +18,13 @@ import type { GraphQLError } from './GraphQLError';
  */
 export function formatError(error: GraphQLError): GraphQLFormattedError {
   invariant(error, 'Received null or undefined error.');
+  // Pick location associated with the first AST node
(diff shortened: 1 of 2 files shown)


src/language/lexer.js
```diff
@@ -22,11 +22,13 @@ import { syntaxError } from '../error';
  */
 export function createLexer<TOptions>(
   source: Source,
```
leebyron: `Source` here represents the file, including the filepath

Slava:
Updated exposing the information from the `Source` object.

Slava:
The tests fail because I changed the type of formatted error to return the `origin` field. I can mass patch it in tests, but first let's agree that the `origin` field is alright to introduce here.

PR #902, opened 2017-06-12
Unify how 'kinds' are imported

IvanGoncharov:
At the moment there are two ways 'kinds' are imported:
```js
import * as Kind from '../language/kinds';
```
and
```js
import {
  NULL,
  ENUM,
  // ...
} from '../language/kinds';
```

This PR unifies this by using the first approach. IMHO this PR not only shaves a few lines of code but also improves readability. For example `NULL` as the name is confusing (especially for someone who used C language) and `Kind.NULL` solves this problem.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -10,7 +10,7 @@
 
 import invariant from '../jsutils/invariant';
 import isNullish from '../jsutils/isNullish';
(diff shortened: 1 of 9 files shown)


leebyron:
Seems fine to me!

PR #903, opened 2017-06-12
Fix parsing of default values in `buildASTSchema`, `extendSchema` and `buildClientSchema`

IvanGoncharov:
In the current implementation `buildASTSchema`, `extendSchema` and `buildClientSchema` register following stubs as `parseLiteral` and `parseValue`:
```js
      // ...

      // Note: validation calls the parse functions to determine if a
      // literal value is correct. Returning null would cause use of custom
      // scalars to always fail validation. Returning false causes them to
      // always pass validation.
      parseValue: () => false,
      parseLiteral: () => false,
    });
```

Because of that, `graphql-js` can't correctly parse default values of custom scalar types:
```graphql
scalar CustomScalar

type Query {
  someField(someArg: CustomScalar = 5): String
}
```

So to clarify. You can override those functions after the schema is built, but the default value from IDL will be lost.

To illustrate this I've added a test case in the first commit of this PR and here is the result:
screen shot 2017-06-09 at 11 20 21 pm

This PR solves this problem by providing reasonable defaults for `parseLiteral` and `parseValue`.

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -12,6 +12,7 @@ import invariant from '../jsutils/invariant';
 import isNullish from '../jsutils/isNullish';
 import * as Kind from '../language/kinds';
 import { assertValidName } from '../utilities/assertValidName';
+import { scalarValueFromAST } from '../utilities/scalarValueFromAST';
 import type {
   ScalarTypeDefinitionNode,
   ObjectTypeDefinitionNode,
@@ -345,7 +346,10 @@ export class GraphQLScalarType {
   // Parses an externally provided value to use as an input.
   parseValue(value: mixed): mixed {
     const parser = this._scalarConfig.parseValue;
-    return parser && !isNullish(value) ? parser(value) : undefined;
+    if (isNullish(value)) {
+      return undefined;
+    }
+    return parser ? parser(value) : value;
   }
 
   // Determines if an internal value is valid for this type.
@@ -357,7 +361,7 @@ export class GraphQLScalarType {
   // Parses an externally provided literal value to use as an input.
   parseLiteral(valueNode: ValueNode): mixed {
     const parser = this._scalarConfig.parseLiteral;
-    return parser ? parser(valueNode) : undefined;
+    return (parser || scalarValueFromAST)(valueNode);
   }
 
   toString(): string {
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -58,6 +58,10 @@ export { valueFromAST } from './valueFromAST';
 // Create a GraphQL language AST from a JavaScript value.
 export { astFromValue } from './astFromValue';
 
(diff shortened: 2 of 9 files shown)


src/utilities/scalarValueFromAST.js
```diff
@@ -0,0 +1,43 @@
...
+
+import keyValMap from '../jsutils/keyValMap';
+import * as Kind from '../language/kinds';
+import type { ValueNode } from '../language/ast';
+
+/**
+ * Create a JavaScript value from a GraphQL languate AST representation
+ * of Scalar.
+ */
+export function scalarValueFromAST(astValue: ValueNode): mixed {
```
leebyron: This looks very similar to `valueFromAST` - can that be used instead?

src/type/definition.js
```diff
@@ -333,7 +334,7 @@ export class GraphQLScalarType {
   // Parses an externally provided value to use as an input.
   parseValue(value: mixed): mixed {
     const parser = this._scalarConfig.parseValue;
-    return parser && !isNullish(value) ? parser(value) : undefined;
+    return parser && !isNullish(value) ? parser(value) : value;
```
leebyron: If a value parses as nullish, shouldn't it return clearly as `undefined` to not be confused with a true `null` value?

src/utilities/index.js
```diff
@@ -58,6 +58,10 @@ export { valueFromAST } from './valueFromAST';
 // Create a GraphQL language AST from a JavaScript value.
 export { astFromValue } from './astFromValue';
 
+// Create a JavaScript value from a GraphQL languate AST representation
```
leebyron: language

src/utilities/scalarValueFromAST.js
leebyron: If this function is for scalars, then it should likely not support LIST or OBJECT, correct?

IvanGoncharov: > This looks very similar to valueFromAST - can that be used instead?

`valueFromAST` builds value based on the type definition it accepts as the second argument.
In a case of a scalar, there is no such type definition.

> If this function is for scalars, then it should likely not support LIST or OBJECT, correct?

AFAIK, the spec doesn't explicitly forbid complex scalars as LIST or OBJECT. Moreover, it's a widespread practice to create such scalars. More details in this PR:
https://github.com/facebook/graphql/pull/325

src/utilities/index.js
IvanGoncharov: Fixed

src/type/definition.js
IvanGoncharov: Fixed.

leebyron:
Thanks for making this happen, Ivan. Great improvement

PR #904, opened 2017-06-12
Add 'getAppliedDirectives' function

IvanGoncharov:
This PR adds utility function that provides a convenient way to get arguments of one or more applied directives.

### DIFF
--- a/src/execution/values.js
+++ b/src/execution/values.js
@@ -11,6 +11,7 @@
 import { createIterator, isCollection } from 'iterall';
 
 import { GraphQLError } from '../error';
+import find from '../jsutils/find';
 import invariant from '../jsutils/invariant';
 import isNullish from '../jsutils/isNullish';
 import isInvalid from '../jsutils/isInvalid';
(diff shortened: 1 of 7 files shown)


src/execution/values.js
```diff
@@ -165,6 +165,31 @@ export function getArgumentValues(
 }
 
 /**
+ * Prepares an object map of directive name to argument values given a list of
+ * directive definitions and AST node.
+ */
+export function getAppliedDirectives(
+  directiveDefs: Array<GraphQLDirective>,
```
leebyron: How about accepting a Schema instance instead? Those define all relevant directives

src/execution/values.js
```diff
@@ -165,6 +165,31 @@ export function getArgumentValues(
 }
 
 /**
+ * Prepares an object map of directive name to argument values given a list of
+ * directive definitions and AST node.
+ */
+export function getAppliedDirectives(
```
leebyron: Or - what do you think about having this be a singular function - `getAppliedDirective()` which accepts a `GraphQLDirective` and a node and returns `void | { [string]: mixed }`?

IvanGoncharov: > How about accepting a Schema instance instead? Those define all relevant directives

@leebyron  I needed this function to only get directives relevant for my tool/library so it doesn't make sense to return all of them, e.g. `buildASTSchema` only needs the value of `@deprecated` and doesn't care about the rest of directives.

So I changed it to accept a `GraphQLDirective` as you suggested.
However, I renamed it to `getDirectiveArgs` since I want to keep `getAppliedDirectives` name for the function that you suggested above (accepts `GraphQLSchema`), which in addition to returning arguments will report unknown directives but it's a topic for a separate PR.

src/execution/values.js
IvanGoncharov: Answered below ⬇️

PR #905, opened 2017-06-12
Added printIntrospectionSchema to exports

aeiz:
Docs at http://graphql.org/graphql-js/utilities/ say "You can import either from the graphql/utilities module, or from the root graphql module" and lists printIntrospectionSchema, yet it is not exported via the root module.


leebyron:
Thanks for spotting this!

Issue #897, opened 2017-06-06
Allow begining pipe in union types


alexandrebodin:
Hi 

Everything was said in the titile. I would be great to be able to write:

```graphql
union unionType = 
  | TypeA
  | Type B
  | Type C
```
instead of
```graphql
union unionType = 
    TypeA
  | Type B
  | Type C
```

This would be great for prettier / graphql to make cleaner formatting. And overall go the same wayas Flow and Elm for example :)

robzhu:
Hey @alexandrebodin, thanks for filing this issue. The proposal reasonable and we should be able to get it added soon.

alexandrebodin:
@robzhu If you can give me some pointers I'm down for doing a PR

robzhu:
@alexandrebodin awesome, thanks for volunteering =). Start with `schema-parser-test.js:460`. It has an example of parsing a union type from graphql schema format. If you trace through that, you'll probably find yourself at `language/parser.js: parseUnionTypeDefinition`. My general advice is to write the failing test first and work backward. I can follow up on your PR with the related spec text changes. Take a look and let me know if you still want to commit to doing this work.

alexandrebodin:
@robzhu 

PR done here: https://github.com/graphql/graphql-js/pull/907 
(also opened a PR on the spec)

robzhu:
Thank you for your contributions! Both PRs have been merged, closing the issue now.

PR #907, opened 2017-06-13
Add support for leading vertical bar in union types

alexandrebodin:
Implementation of the leading vertical bar in the language parser.

Issue #897

### DIFF
--- a/src/language/__tests__/schema-parser-test.js
+++ b/src/language/__tests__/schema-parser-test.js
@@ -489,6 +489,21 @@ type Hello {
     expect(() => parse(body)).to.throw();
   });
 
+  it('Union with types and double vertical bar at the beginning', () => {
+    const body = 'union Hello = || Wo | Rld';
+    expect(() => parse(body)).to.throw();
+  });
+
+  it('Union with types and double vertical bar in the middle', () => {
+    const body = 'union Hello = Wo || Rld';
+    expect(() => parse(body)).to.throw();
+  });
+
+  it('Union with types and double vertical bar at the end', () => {
+    const body = 'union Hello = | Wo | Rld ||';
+    expect(() => parse(body)).to.throw();
+  });
+
   it('Union with types , leanding and ending vertical bar', () => {
     const body = 'union Hello = | Wo | Rld |';
     expect(() => parse(body)).to.throw();
--- a/src/language/__tests__/schema-parser-test.js
+++ b/src/language/__tests__/schema-parser-test.js
@@ -457,6 +457,28 @@ type Hello {
     expect(printJson(doc)).to.equal(printJson(expected));
   });
 
+  it('Union with two types and begining pipe', () => {
+    const body = 'union Hello = | Wo | Rld';
+    const doc = parse(body);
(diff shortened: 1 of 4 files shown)


src/language/__tests__/schema-printer-test.js
```diff
@@ -83,6 +83,8 @@ union Feed = Story | Article | Advert
 
 union AnnotatedUnion @onUnion = A | B
 
+union AnnotatedUnionTwo @onUnion = A | B
```
robzhu: Did you intend to have `| A | B`?

src/language/__tests__/schema-parser-test.js
```diff
@@ -457,6 +457,28 @@ type Hello {
...
+            typeNode('Rld', { start: 21, end: 24 }),
+          ],
+          loc: { start: 0, end: 24 },
+        }
+      ],
+      loc: { start: 0, end: 24 },
+    };
+    expect(printJson(doc)).to.equal(printJson(expected));
+  });
+
```
robzhu: Can you please also add test cases for `| Wo | Rld |`, `union Hello = Wo | Rld |`, and `union Hello = |`?

src/language/__tests__/schema-printer-test.js
alexandrebodin: Nope the printer will only print without it or with it I think it's better to leave it out and add it with prettier for ex

src/language/__tests__/schema-parser-test.js
alexandrebodin: Okay

alexandrebodin:
@robzhu I added severals news tests

robzhu:
@alexandrebodin, thank you for the high quality PR and fast turnaround!

src/language/parser.js
```diff
@@ -943,6 +943,9 @@ function parseUnionTypeDefinition(lexer: Lexer<*>): UnionTypeDefinitionNode {
  */
 function parseUnionMembers(lexer: Lexer<*>): Array<NamedTypeNode> {
   const members = [];
+  if (peek(lexer, TokenKind.PIPE)) {
+    skip(lexer, TokenKind.PIPE);
+  }
```
leebyron: The wrapping `peek` is unnecessary.  `skip` is effectively a call to `peek`.

src/language/parser.js
```diff
@@ -943,6 +943,9 @@ function parseUnionTypeDefinition(lexer: Lexer<*>): UnionTypeDefinitionNode {
  */
 function parseUnionMembers(lexer: Lexer<*>): Array<NamedTypeNode> {
```
leebyron: Every method in the parser has parse language atop it. It would be nice to include this here as well

leebyron:
This is not the only place in the IDL where bar-separated tokens exists. Should we remain consistent and update the others as well? Or would that not be fitting?

alexandrebodin:
@leebyron I could update the other places if you can point them to me.

leebyron:
Let me take a quick look

PR #914, opened 2017-06-16
RFC: Define custom scalars in terms of built-in scalars.

leebyron:
This proposes an additive change which allows custom scalars to be defined in terms of the built-in scalars. The motivation is for client-side code generators to understand how to map between the GraphQL type system and a native type system. As an example, a `URL` custom type may be defined in terms of the built-in scalar `String`. It could define additional serialization and parsing logic, however client tools can know to treat `URL` values as `String`. Presently, we do this by defining these mappings manually on the client, which is difficult to scale, or by giving up and making no assumptions of how the custom types serialize.

Another real use case of giving client tools this information is GraphiQL: this change will allow GraphiQL to show more useful errors when a literal of an incorrect kind is provided to a custom scalar. Currently GraphiQL simply accepts all values.

To accomplish this, this proposes adding the following:

* A new property when defining `GraphQLScalarType` (`ofType`) which asserts that only built-in scalar types are provided.

* A second type coercion to guarantee to a client that the serialized values match the `ofType`.

* Delegating the `parseLiteral` and `parseValue` functions to those in `ofType` (this enables downstream validation / GraphiQL features)

* Exposing `ofType` in the introspection system, and consuming that introspection in `buildClientSchema`.

* Adding optional syntax to the SDL, and consuming that in `buildASTSchema` and `extendSchema` as well as in `schemaPrinter`.

* Adding a case to `findBreakingChanges` which looks for a scalar's ofType changing.

### DIFF
--- a/src/language/__tests__/schema-kitchen-sink.graphql
+++ b/src/language/__tests__/schema-kitchen-sink.graphql
@@ -41,6 +41,8 @@ union AnnotatedUnionTwo @onUnion = | A | B
 
 scalar CustomScalar
 
+scalar StringEncodedCustomScalar = String
+
 scalar AnnotatedScalar @onScalar
 
 enum Site {
--- a/src/language/__tests__/schema-kitchen-sink.graphql
+++ b/src/language/__tests__/schema-kitchen-sink.graphql
@@ -41,7 +41,7 @@ union AnnotatedUnionTwo @onUnion = | A | B
 
 scalar CustomScalar
 
-scalar StringEncodedCustomScalar = String
+scalar StringEncodedCustomScalar as String
 
 scalar AnnotatedScalar @onScalar
 
--- a/src/language/ast.js
+++ b/src/language/ast.js
@@ -397,6 +397,7 @@ export type ScalarTypeDefinitionNode = {
   kind: 'ScalarTypeDefinition';
   loc?: Location;
   name: NameNode;
+  type?: ?NamedTypeNode;
   directives?: ?Array<DirectiveNode>;
 };
 
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -294,6 +294,7 @@ function resolveThunk<T>(thunk: Thunk<T>): T {
 export class GraphQLScalarType {
   name: string;
   description: ?string;
+  ofType: ?GraphQLScalarType;
 
   _scalarConfig: GraphQLScalarTypeConfig<*, *>;
 
@@ -301,6 +302,19 @@ export class GraphQLScalarType {
     assertValidName(config.name);
     this.name = config.name;
     this.description = config.description;
+    this.ofType = config.ofType || null;
+    if (this.ofType) {
+      const ofTypeName = this.ofType.name;
+      invariant(
+        ofTypeName === 'String' ||
+        ofTypeName === 'Int' ||
+        ofTypeName === 'Float' ||
+        ofTypeName === 'Boolean' ||
+        ofTypeName === 'ID',
+        `${this.name} may only be described in terms of a built-in scalar ` +
+        `type. However ${ofTypeName} is not a built-in scalar type.`
+      );
+    }
     invariant(
       typeof config.serialize === 'function',
       `${this.name} must provide "serialize" function. If this custom Scalar ` +
@@ -321,7 +335,8 @@ export class GraphQLScalarType {
   // Serializes an internal value to include in a response.
   serialize(value: mixed): mixed {
     const serializer = this._scalarConfig.serialize;
-    return serializer(value);
+    const serialized = serializer(value);
+    return this.ofType ? this.ofType.serialize(serialized) : serialized;
   }
 
   // Determines if an internal value is valid for this type.
@@ -332,7 +347,8 @@ export class GraphQLScalarType {
 
   // Parses an externally provided value to use as an input.
   parseValue(value: mixed): mixed {
-    const parser = this._scalarConfig.parseValue;
+    const parser = this._scalarConfig.parseValue ||
+      this.ofType && this.ofType.parseValue;
     return parser && !isNullish(value) ? parser(value) : undefined;
   }
 
@@ -344,7 +360,8 @@ export class GraphQLScalarType {
 
   // Parses an externally provided literal value to use as an input.
   parseLiteral(valueNode: ValueNode): mixed {
-    const parser = this._scalarConfig.parseLiteral;
+    const parser = this._scalarConfig.parseLiteral ||
+      this.ofType && this.ofType.parseLiteral;
     return parser ? parser(valueNode) : undefined;
   }
 
@@ -364,9 +381,10 @@ GraphQLScalarType.prototype.toJSON =
 export type GraphQLScalarTypeConfig<TInternal, TExternal> = {
   name: string;
   description?: ?string;
+  ofType?: ?GraphQLScalarType;
   serialize: (value: mixed) => ?TExternal;
-  parseValue?: (value: mixed) => ?TInternal;
-  parseLiteral?: (valueNode: ValueNode) => ?TInternal;
+  parseValue?: ?(value: mixed) => ?TInternal;
+  parseLiteral?: ?(valueNode: ValueNode) => ?TInternal;
 };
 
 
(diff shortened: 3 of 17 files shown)


leebyron:
If there's general agreement, I'll be following up with more test coverage as well as the minor spec text change to cover the introspection addition.

stubailo:
Can I add another wish list item here? I don't know if it's specifically relevant to this PR.

It would be really nice to be able to override the serialization and deserialization of built-in types in GraphQL-JS. Specifically, I always run into this with `ID` and MongoDB. It's super convenient to just be able to return a Mongo ObjectID which is already kind of like a string, but it's not possible to use the scalar serialization feature and the `ID` type at the same time. So you end up with awkward wrapper code, or you have to define a custom `MongoID` scalar.

leebyron:
I think that's out of scope here but happy to discuss in another issue or pr

src/language/__tests__/schema-kitchen-sink.graphql
```diff
@@ -41,6 +41,8 @@ union AnnotatedUnionTwo @onUnion = | A | B
 
 scalar CustomScalar
 
+scalar StringEncodedCustomScalar = String
```
stubailo: Yeah, I agree. `implements` seems like the closest thing, `:` seems reasonable as well

leebyron: @dschafer also suggested `:` IIRC. My preference for `=` was based on matching the rest of the SDL forms, but I agree that the implication of equality isn't correct.

I'm a bit worried that `:` is also overloaded - we most use it for defining the types or values to arguments but also aliases.

Maybe a keyword is best to make this as literate as possible. I kind of like `as`. Thoughts from others?

erydo:
@leebyron — An issue that seems directly relevant to this RFC is validation and scalar compatibility. Sometimes we'll want to transition a scalar field from "String" to something more descriptive, like "ID" or custom "DateTime" scalar. For example:

```graphql
type Query {
  foos(ids: [String!]): [Foo]
}
```
to:
```graphql
type Query {
  foos(ids: [ID!]): [Foo]
}
```

Unfortunately, client queries which refer to that field as `String` fail to validate, so either we need to make a breaking change, or perform type transitions by designating new field names—even though the actual JSON input variables or response are not changing.

```graphql
# This old query would fail to validate after the otherwise-benign schema change,
# and there is no solution currently short of updating all clients in lockstep,
# or adding a new argument name.
query($myFooIds: [String!]) {
  foos(ids: $myFooIds) { bar }
}
```

Would this PR help with that kind of transition? For example, would it make sense to allow input fields whose scalars serialize to compatible JSON could emit a warning on type mismatch, rather than a validation error?

asaf-romano:
Will this change also help with "asymmetric" scalars? For example, in our system we've a Token type which is (roughly) ``btoa(JSON.stringify(value)`` in its serialized form and also in client inputs (literal or value), but a simple JS object on the server side. In other words, ``parseValue`` and ``parseLiteral`` work on strings, whereas ``serialize`` works with objects.

src/type/definition.js
```diff
@@ -294,13 +294,27 @@ function resolveThunk<T>(thunk: Thunk<T>): T {
...
     this.description = config.description;
+    this.ofType = config.ofType || null;
+    if (this.ofType) {
+      const ofTypeName = this.ofType.name;
+      invariant(
+        ofTypeName === 'String' ||
+        ofTypeName === 'Int' ||
+        ofTypeName === 'Float' ||
+        ofTypeName === 'Boolean' ||
+        ofTypeName === 'ID',
```
IvanGoncharov: @leebyron `graphql-js` already have the same list in multiple places + we have the same code in all of our GraphQL projects. It would be great if you review #924 which adds `isBuiltInScalar` predicate.

src/type/definition.js
```diff
@@ -478,12 +480,14 @@ export class GraphQLScalarType {
   // Serializes an internal value to include in a response.
   serialize(value: mixed): mixed {
     const serializer = this._scalarConfig.serialize;
-    return serializer(value);
+    const serialized = serializer(value);
+    return this.ofType ? this.ofType.serialize(serialized) : serialized;
```
leebyron: This effectively chains the serialize functions together if more than one are provided. Should we consider having the same chaining behavior for parsing input values as well?

src/type/validate.js
```diff
@@ -261,6 +267,20 @@ function validateTypes(context: SchemaValidationContext): void {
   });
 }
 
+function validateScalarSerialization(
+  context: SchemaValidationContext,
+  scalarType: GraphQLScalarType,
+): void {
+  if (scalarType.ofType && !isSpecifiedScalarType(scalarType.ofType)) {
```
leebyron: I wonder if this is over-constrained. Perhaps it would be okay to allow a scalar to serialize in terms of any other scalar type?

My thought is that people add custom scalars for two reasons. First is to represent some higher level concept like "Email" or "PhoneNumber" where "serializes as String" is useful. Second is to represent a new kind of transport primitive like "BigInt" where they definitionally serialize as something new. However that first kind of higher-level concept scalar might want to say it serializes in terms of another non-spec scalar!

In other words - what makes the specification of a scalar special to warrant this limitation?

I believe we originally settled on this limitation to avoid needing to follow a chain of "ofType" to determine the final actual serialization type, but I think that might not be such a big deal

src/language/ast.js
```diff
@@ -448,6 +448,7 @@ export type ScalarTypeDefinitionNode = {
   +loc?: Location,
   +description?: StringValueNode,
   +name: NameNode,
+  +type?: NamedTypeNode,
```
leebyron: I think this key should be more descriptive. A type having a field called "type" is confusing

leebyron:
@asaf-romano - that would still work fine - the internal representation as an object would first go through your custom scalar's serialize function before going through the serializing type's. That guarantees the result of the query will be a string, but still gives control over the internal representation

leebyron:
Additional tests should be added to ensure validation works as expected

src/type/validate.js
IvanGoncharov: > I believe we originally settled on this limitation to avoid needing to follow a chain of "ofType" to determine the final actual serialization type, but I think that might not be such a big deal

I think that decision about allowing recursive type should be done in the scope of the entire spec since it affects the entire type system. For example, you couldn't query all information about type using `__type` anymore:
```graphql
{
   __type(name: "LocalPath") {
      name
      # ...
      ofType {
         name
         # ...
         # <= you can't construct a query that gets arbitrary long chain of `ofType`
      }
   }
}
```
Or the fact that you can introduce loops inside types, so we need to make `buildASTSchema`, `buildClientSchema` and others loop-resistant. E.g.:
```graphql
scalar A as C
scalar B as A
scalar C as B
```

I think allowing to better describe scalars isn't worth added complexity. However, if the similar changes are done in other places of type-system (e.g. https://github.com/facebook/graphql/issues/295) then the price is already paid and we could definitely add recursiveness to scalar types.

> However that first kind of higher-level concept scalar might want to say it serializes in terms of another non-spec scalar!

We can allow specifying scalar serializable as any other scalar without `as` in the definition.
```graphql
scalar BigInt
scalar Counter as BigInt # Valid
scalar UserCounter as Counter # Invalid
```
But with current terminology and SDL syntax, it would be very confusing 😞 
That's why I think we need to separate `custom scalar type`s and `scalar type`s:
```graphql
custom scalar BigInt
scalar Counter as BigInt # Valid
scalar UserCounter as Counter # Invalid
```
So we could say that scalar is serializable only as a standard scalar or a custom scalar.

-------------------------------
I think this PR in its current form solves >80% of use-cases so we can release it as is and relax rules afterward.

src/type/definition.js
IvanGoncharov: Agreed 👍

src/language/ast.js
IvanGoncharov: Fully agree. How about `serializableAs`?

IvanGoncharov:
@leebyron I rebased this PR on top of #1175 and working on validation tests and other suggestions.
Should I create separate PR or commit on top of this one?

IvanGoncharov:
@leebyron I rebased this PR on top of `master` and squshed first 3 PRs (make it easier to rebase).
And I working on addressing review comments and adding more tests.
Should I push it into this PR or create a new one?

IvanGoncharov:
@leebyron I rebased this PR on top of `master`. I hope you don't mind?

src/language/ast.js
IvanGoncharov: @leebyron Given current syntax `scalar Url as String`.
How about `asType`?

PR #915, opened 2017-06-16
Throw error if argument is unknown in getArgumentValues

RomanGotsiy:
`getArgumentValues` function doesn't throw if there are unknown arguments in AST.

We experienced this issue while parsing directive arguments from IDL.To demonstrate the issue I have added a test for unknown argument on `@deprecated` directive as the [first commit](https://github.com/graphql/graphql-js/pull/915/commits/40e045be3b1615dd4d9ca1e9672f7e5fa148257c) of this PR.

Also, after implementing the check I've found unknown argument usage in Subscribe tests (e.g. argument `priority` [here](https://github.com/graphql/graphql-js/blob/master/src/subscription/__tests__/subscribe-test.js#L110)) so I fixed it too (the [last commit](https://github.com/graphql/graphql-js/pull/915/commits/24e4b6d888b9fd000a3d597dcbdfb4e85a259d71) of PR)

As part of the PR, I have also simplified `getArgumentValues` function because it was hard to read and understand.

### DIFF
--- a/src/execution/values.js
+++ b/src/execution/values.js
@@ -15,6 +15,7 @@ import invariant from '../jsutils/invariant';
 import isNullish from '../jsutils/isNullish';
 import isInvalid from '../jsutils/isInvalid';
 import keyMap from '../jsutils/keyMap';
+import find from '../jsutils/find';
 import { typeFromAST } from '../utilities/typeFromAST';
 import { valueFromAST } from '../utilities/valueFromAST';
 import { isValidJSValue } from '../utilities/isValidJSValue';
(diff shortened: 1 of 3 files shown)


leebyron:
The existing behavior is intentional. We would expect validation to find and report unknown arguments. This is important because there can be skew between when a query is validated vs when it is executed, and ignoring unknown arguments is the correct behavior to ensure successful execution when this occurs.

leebyron:
Closing - but feel free to continue discussion

RomanGotsiy:
@leebyron thanks for the explanation. This sounds reasonable.

RomanGotsiy:
@leebyron btw, can you suggest a place where to put this logic?
I need this to work with IDL only. Currently `buildASTSchema` doesn't check directive locations and arguments.
Maybe I can create a wrapper function for [`getDirectiveValues`](https://github.com/graphql/graphql-js/blob/master/src/execution/values.js#L175) in `utils` for this and use it inside `buildASTSchema`?

leebyron:
I think the answer is that the validation system should be expanded to also handle the query language. 

In fact, there is already a validation rule for directives with correct arguments, so that should Just Work. I think you just need to run validation before calling buildASTSchema

PR #916, opened 2017-06-16
Fix identation

RomanGotsiy:
Multiple places in tests use ES6 multiline strings. Becuase of that indentation is broken:

It's really hard to read this code. When I first time had to read this file it melted my brain. I think this can discourage people from writing tests.

To fix that I added a simple deindentation ES6 string tag `dedent` which I used across the tests. Resulting code looks much better:

The one place where I left strings ident intact is parse tests where the number of character matters and changing indentation there can make it harder to reason about.

### DIFF
--- a/src/jsutils/dedent.js
+++ b/src/jsutils/dedent.js
@@ -1,0 +1,50 @@
+/* @flow */
+/**
+ *  Copyright (c) 2017, Facebook, Inc.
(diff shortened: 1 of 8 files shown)


sibelius:
can't we use prettier on this?

leebyron:
This is pretty awesome, thanks @RomanGotsiy!

PR #918, opened 2017-06-16
Update subscribe function to return a Promise of AsyncIterator or ExecutionResult

robzhu:
This change updates the Subscribe() function to return a Promise<AsyncIterator |  ExecutionResult>.

The reason for this change is to allow callers to distinguish between initialization and publish errors. 

When calling `subscribe`:
If the operation is successful, returns an AsyncIterator representing the response stream.
If the operation failed with an informative error for the client, returns an ExecutionResult with an `errors` field but no `value` field.
If the operation failed due to a system related problem, such as an unavailable pubsub, system, an error will be thrown. 

Example of how to call `subscribe` after these changes:

```js
try {
  const result = await subscribe(...);
  if (result.errors) {
    // errors are informative and can be sent to the client
  } else {
    // result is guaranteed to be an AsyncIterable at this point
    return result;
  }
} catch (error) {
  // system error, log the problem and send a failure response to the client.
}
```

### DIFF
--- a/src/subscription/subscribe.js
+++ b/src/subscription/subscribe.js
@@ -256,7 +256,7 @@ export async function createSourceEventStream(
     info
   );
 
-  // Throw located GraphQLError if subscription source fails to resolve.   
+  // Throw located GraphQLError if subscription source fails to resolve.
   if (subscription instanceof Error) {
     throw locatedError(
       subscription,
--- a/src/subscription/subscribe.js
+++ b/src/subscription/subscribe.js
@@ -10,19 +10,21 @@
  */
 
 import { isAsyncIterable } from 'iterall';
+import { GraphQLError } from '../error/GraphQLError';
 import {
   addPath,
   assertValidExecutionArguments,
   buildExecutionContext,
+  buildResolveInfo,
   collectFields,
   execute,
   getFieldDef,
   getOperationRootType,
-  buildResolveInfo,
   resolveFieldValueOrError,
 } from '../execution/execute';
 import { GraphQLSchema } from '../type/schema';
 import invariant from '../jsutils/invariant';
+import asyncIteratorReject from './asyncIteratorReject';
 import mapAsyncIterator from './mapAsyncIterator';
 
 import type { ExecutionResult } from '../execution/execute';
@@ -34,8 +36,9 @@ import type { GraphQLFieldResolver } from '../type/definition';
  *
  * Returns an AsyncIterator
  *
- * If the arguments to this function do not result in a legal execution context,
- * a GraphQLError will be thrown immediately explaining the invalid input.
+ * If the client-provided arguments to this function do not result in a
+ * compliant subscription, a GraphQL Response with descriptive errors and no
+ * data will be yielded.
  *
  * Accepts either an object with named arguments, or individual arguments.
  */
@@ -132,7 +135,13 @@ function subscribeImpl(
       variableValues,
       operationName,
       fieldResolver
-    )
+    ),
+    error => {
+      if (error instanceof GraphQLError) {
+        return { errors: [ error ] };
+      }
+      throw error;
+    }
   );
 }
 
@@ -140,7 +149,11 @@ function subscribeImpl(
  * Implements the "CreateSourceEventStream" algorithm described in the
  * GraphQL specification, resolving the subscription source event stream.
  *
- * Returns an AsyncIterable, may through a GraphQLError.
+ * Returns an AsyncIterable.
+ *
+ * If the client-provided arguments to this function do not result in a
+ * compliant subscription, iteration will fail with a GraphQLError explaining
+ * the invalid input. These should be caught and yielded as a GraphQL Response.
  *
  * A Source Stream represents the sequence of events, each of which is
  * expected to be used to trigger a GraphQL execution for that event.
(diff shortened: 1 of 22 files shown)


src/subscription/subscribe.js
```diff
@@ -32,10 +35,17 @@ import type { GraphQLFieldResolver } from '../type/definition';
 /**
  * Implements the "Subscribe" algorithm described in the GraphQL specification.
  *
- * Returns an AsyncIterator
+ * Returns a Promise<AsyncIterator | Error | ExecutionResult>
```
leebyron: Promise<AsyncIterator | ExecutionResult>

leebyron: Or maybe it's easier to read if this spells it out. Something like:

Returns a Promise which resolves to either an ExecutionResult (indicating failure to subscribe) or an AsyncIterator of ExecutionResult values, triggered by the underlying subscribed event stream.

src/subscription/subscribe.js
```diff
@@ -32,10 +35,17 @@ import type { GraphQLFieldResolver } from '../type/definition';
...
+ *
+ * If the client-provided arguments to this function do not result in a
+ * compliant subscription, a GraphQL Response with descriptive errors and no
+ * data will be returned.
  *
- * If the arguments to this function do not result in a legal execution context,
- * a GraphQLError will be thrown immediately explaining the invalid input.
+ * If the the source stream could not be created due to faulty subscription
+ * resolver logic or underlying systems, an Error will be returned, not thrown.
```
leebyron: Promise rejected?

src/subscription/subscribe.js
```diff
@@ -59,7 +69,7 @@ declare function subscribe(
   operationName?: ?string,
   fieldResolver?: ?GraphQLFieldResolver<any, any>,
   subscribeFieldResolver?: ?GraphQLFieldResolver<any, any>
-): AsyncIterator<ExecutionResult>;
+): Promise<AsyncIterator<ExecutionResult> | Error | ExecutionResult>;
```
leebyron: This type should match the type above

src/subscription/subscribe.js
```diff
@@ -96,6 +107,16 @@ export function subscribe(
     );
 }
 
+// This function checks if the error is a GraphQLError. If it is, convert it to
```
leebyron: Nits:

Use /* for multi-line doc blocks.

Also, it would be nice to move this fn lower in the file - keeping `subscribe()` and `subscribeImpl()` directly next to each other in the file.

src/subscription/subscribe.js
```diff
@@ -106,44 +127,49 @@ function subscribeImpl(
...
-      schema,
-      document,
-      payload,
-      contextValue,
-      variableValues,
-      operationName,
-      fieldResolver
-    )
-  );
+  ).then(subscription => {
```
leebyron: remove the wrapping `{}` and `return` since this arrow function is of a single expression

src/subscription/subscribe.js
```diff
@@ -106,44 +127,49 @@ function subscribeImpl(
...
+        document,
+        payload,
+        contextValue,
+        variableValues,
+        operationName,
+        fieldResolver
+      ),
+      convertOrThrowError
+    );
+  }).catch(convertOrThrowError);
```
leebyron: You can provide `convertOrThrowError` as the 2nd argument to `then()` - which both reduces the number of wrapped promises, but also avoids calling `convertOrThrowError` two times in the case an error is thrown.

src/subscription/subscribe.js
```diff
@@ -220,16 +249,23 @@ export function createSourceEventStream(
     info
   );
 
-  if (subscription instanceof Error) {
-    throw subscription;
-  }
+  return Promise.resolve(subscription).then(resolvedSubscription => {
```
leebyron: `buildExecutionContext` above may throw a GraphQLError if client-provided values don't coerce properly - those should be properly formatted in the response via returning a rejected promise. You can test this by passing bad variable values along with a query.

The fix should probably be to wrap `return new Promise(resolve => {})`starting above line 196, ending in `resolve(subscription)` and followed by this `.then()`

robzhu: I added a test case that validates bad query variables are returned as an ExecutionResult. However, the `return new Promise` approach you suggested led to pretty gnarly logic for dealing with the nested promise coming back from resolveFieldValueOrError. The current version uses await/async to work around it, but happy to revise if you see don't think the tradeoff in code clarity is worth it.

cedrelek:
Thx for this. I hope the next release will include this PR :)

src/subscription/subscribe.js
robzhu: Here's what the code looks like using a wrapped promise: https://gist.github.com/robzhu/d2d9e3238b586e67569646a5944154b8

src/subscription/subscribe.js
```diff
@@ -122,44 +148,57 @@ function subscribeImpl(
...
+ * A Source Event Stream represents a sequence of events, each of which triggers
+ * a GraphQL execution for that event.
  *
  * This may be useful when hosting the stateful subscription service in a
  * different process or machine than the stateless GraphQL execution engine,
  * or otherwise separating these two steps. For more on this, see the
  * "Supporting Subscriptions at Scale" information in the GraphQL specification.
  */
-export function createSourceEventStream(
+export async function createSourceEventStream(
```
leebyron: My only mild concern is that by including an async function in non-test code that we may need to add a dependency to package.json for an async function runtime.

As a follow-up task, let's test this to see if we need to do so, ideally we can ensure no new dependencies.

leebyron:
No excuse for my long delay on this supremely excellent PR. Error behavior is so difficult, and getting all the corner cases correct and well tested deserves a serious shout out. Thank you for making GraphQL more awesome!

PR #924, opened 2017-06-21
Add predicates to for built-in types.

IvanGoncharov:
I find myself copy-pasting the same list of built-in scalar names in every new project we start.
In many cases, built-in types/directives should be handled differently so having such simple predicates save a lot of code lines.

Because of circular dependencies, I couldn't figure out file suitable for `builtInTypes` and `isBuiltInType`. That's why I created a new one: `src/type/builtins.js`.
Is it an appropriate filename or should I put somewhere else?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -59,6 +59,7 @@ export {
   DirectiveLocation,
 
   // Scalars
+  builtInScalars,
   GraphQLInt,
   GraphQLFloat,
   GraphQLString,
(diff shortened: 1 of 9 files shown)


IvanGoncharov:
rebased

leebyron:
I made some final edits - renaming a few things for consistency, but I really like this addition. Thanks @IvanGoncharov

leebyron:
See 134184e2817dbdd017ea2ee791fa94ce4de892e6 for the final edits

Issue #771, opened 2017-03-29
input validation too loose for string and boolean


itajaja:
this

```js
{ "foo": [1, 2, 3]}
```

is a correct payload for the muation

```js
export default mutationWithClientMutationId({
  name: 'Test',

  inputFields: {
    foo: { type: new GraphQLNonNull(GraphQLString) },
  },
...
```

happening beacuse of [this](https://github.com/graphql/graphql-js/blob/master/src/type/scalars.js#L91-L92).

Clearly `[1, 2, 3]` is not a string and this should probably be rejected. This kind of coercion makes sense for output types, but for mutation input this seems very odd, even more so considering that other things are validated (non-nullness, integers). I mean, if `[1, 2, 3]` is a string, then why isn't `"foo"` a valid integer? I am seeing a disparity here. Is there a reason for that?

leebyron:
This looks like a bug similar to #827 

https://facebook.github.io/graphql/#sec-String

vladar:
Coercion of booleans also seems invalid: 

```json
{"foo": "false"}
```

will be coerced to `true` by GraphQL Boolean Type without raising errors ([here](https://github.com/graphql/graphql-js/blob/master/src/type/scalars.js#L117)) even if specs [explicitly says](http://facebook.github.io/graphql/#sec-Boolean):

```
When expected as an input type, only boolean input values are accepted. 
All other input values must raise a query error indicating an incorrect type.
```

leebyron:
Now fixed in master and will be included in the next release

PR #925, opened 2017-06-21
Throw TypeError when coercing an array into a GraphQLString

robzhu:
Previously, we were using the default JS logic to coerce arrays to strings. This PR detects and disallows conversion of arrays into strings. 

https://github.com/graphql/graphql-js/issues/771

### DIFF
--- a/src/type/scalars.js
+++ b/src/type/scalars.js
@@ -88,6 +88,15 @@ export const GraphQLFloat = new GraphQLScalarType({
   }
 });
 
+function coerceString(value: mixed): ?string {
+  if (Array.isArray(value)) {
+    throw new TypeError(
+      `String cannot represent an array value: [${String(value)}]`
+    );
+  }
+  return String(value);
+}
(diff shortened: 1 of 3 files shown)


src/type/scalars.js
```diff
@@ -88,14 +88,23 @@ export const GraphQLFloat = new GraphQLScalarType({
...
+  return String(value);
+}
+
 export const GraphQLString = new GraphQLScalarType({
   name: 'String',
   description:
     'The `String` scalar type represents textual data, represented as UTF-8 ' +
     'character sequences. The String type is most often used by GraphQL to ' +
     'represent free-form human-readable text.',
   serialize: String,
```
robzhu: Note to reviewer: thoughts on updating the serialize behavior as well? See coerceFloat

leebyron: Yeah - I think this change is actually more compelling for serializing than parsing input values.

Issue #912, opened 2017-06-14
Define extra properties to errors


kbariotis:
From the [specs](http://facebook.github.io/graphql/#sec-Errors)

> GraphQL servers may provide additional entries to error as they choose to produce more helpful or machine‐readable errors, however future versions of the spec may describe additional entries to errors.

As far as I understand, we should be able to define custom properties to errors that are being exposed to the GraphQL response. Like this:

```
{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
+    "type": "NOT_FOUND",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ]
    }
  ],
  "data": {
    "hero": {
      "name": "R2-D2",
      "heroFriends": [
        {
          "id": "1000",
          "name": "Luke Skywalker"
        },
        null,
        {
          "id": "1003",
          "name": "Leia Organa"
        }
      ]
    }
  }
}
```

Is this currently possible? Should we consider adding it to this library?

IvanGoncharov:
@kbariotis Currently it's possible only by providing a custom function to format errors. See example for `express-graphql`: https://github.com/graphql/express-graphql#debugging-tips

I'm already working on PR that adds `extensions` arg  to constructor of `GraphQLError`, in your example argument value would be `{type: 'NOT_FOUND'}`. I still need to add tests, so I plan to submit my PR in the next couple of days.

kbariotis:
@IvanGoncharov thats great! Let me know if you want my review or any help.

Also, I am not using `express` but `graphql-js` directly, so the `formatError` is not really helpful.

IvanGoncharov:
@kbariotis I submitted my implementation for review as #928

PR #928, opened 2017-06-23
Allow to extend GraphQL errors with additional properties

IvanGoncharov:
Implementing suggestion from #912 
In some cases, it's required to provide additional details inside errors.
At the same time, we don't want to force users to override `formatError` method.

This PR adds `extensions` property to `GraphQLError` and extend error object in the response with fields from it.

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -26,7 +26,8 @@ declare class GraphQLError extends Error {
     source?: ?Source,
     positions?: ?Array<number>,
     path?: ?Array<string | number>,
-    originalError?: ?Error
+    originalError?: ?Error,
+    extensions?: ?{ [key: string]: mixed },
   ): void;
 
   /**
@@ -76,6 +77,11 @@ declare class GraphQLError extends Error {
    * The original error thrown from a field resolver during execution.
    */
   originalError: ?Error;
+
+  /**
+   * The original error thrown from a field resolver during execution.
+   */
+  extensions: ?{ [key: string]: mixed };
 }
 
 export function GraphQLError( // eslint-disable-line no-redeclare
@@ -84,7 +90,8 @@ export function GraphQLError( // eslint-disable-line no-redeclare
   source?: ?Source,
   positions?: ?Array<number>,
   path?: ?Array<string | number>,
-  originalError?: ?Error
+  originalError?: ?Error,
+  extensions?: ?{ [key: string]: mixed }
 ) {
   // Compute locations in the source for the given nodes/positions.
   let _source = source;
@@ -146,7 +153,10 @@ export function GraphQLError( // eslint-disable-line no-redeclare
     },
     originalError: {
       value: originalError
-    }
+    },
+    extensions: {
+      value: extensions
+    },
   });
 
   // Include (non-enumerable) stack trace.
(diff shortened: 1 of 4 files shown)


kbariotis:
This is awesome @IvanGoncharov !

So, I've included your branch in a little example and tried to throw an error inside a resolved like this:
```javascript
throw new GraphQL.GraphQLError(
  'Requested resource not found',
  null,
  null,
  null,
  null,
  null,
  { code: 'NOT_FOUND' }
);
```

The response that got out from GraphQL still didn't have the `code` attribute:
```
{ data: { createMessage: null },
  errors:
   [ { GraphQLError
         at Promise (/Users/Bariotis/Dev/develop/graphql-example/schema.js:61:17)
         at Promise (<anonymous>)
         at resolve (/Users/Bariotis/Dev/develop/graphql-example/schema.js:56:16)
         at resolveOrError (/Users/Bariotis/node_modules/graphql/execution/execute.js:474:12)
         at resolveField (/Users/Bariotis/node_modules/graphql/execution/execute.js:460:16)
         at /Users/Bariotis/node_modules/graphql/execution/execute.js:248:20
         at <anonymous>
         at process._tickCallback (internal/process/next_tick.js:169:7)
         at Function.Module.runMain (module.js:607:11)
         at startup (bootstrap_node.js:158:16)
       message: 'Requested resource not found',
       locations: [Array],
       path: [Array] } ] }
```

Did I miss anything?

IvanGoncharov:
@kbariotis  You need to call `formatError` for every error in `data.errors` array.
For example:
```js
if (result.errors) {
  result.errors = result.errors.map(formatError);
}
```

kbariotis:
@IvanGoncharov aw I get it now! May I ask what's the reason of that instead of directly returning the formated error?

IvanGoncharov:
@kbariotis Idea is to allow for more flexibility. For example, you can also serialize `stack` property to have stack trace tied to every error.

kbariotis:
@IvanGoncharov I wasn't able to make it work even with the `formatError` function. Would you please provide an example of PoC? Thank you

leebyron:
Nice, I like how simple this ended up

src/error/GraphQLError.js
```diff
@@ -76,6 +77,11 @@ declare class GraphQLError extends Error {
    * The original error thrown from a field resolver during execution.
    */
   originalError: ?Error;
+
+  /**
+   * The original error thrown from a field resolver during execution.
```
taion: This docstring should be fixed.

jsonfry:
For anyone else that doesn't immediately find this obvious, to use this just `throw new GraphQLError('Not available', null, null, null, null, null, { code: 'NOT_FOUND' });` where-ever, and then in whatever you get back from running `graphql` you `formatErrors`, e.g.

```
import { formatError } from 'graphql'; 

graphql(schema, requestString)
  .then(result => { 
    if (result.errors) {
      result.errors = result.errors.map(formatError);
    }
    ...
  });
```

IvanGoncharov:
> to use this just throw new GraphQLError('Not available', null, null, null, null, null, { code: 'NOT_FOUND' });

Since 2d08496720088dbe65ebea312c8526bd48fb8ee8 you can use your own custom errors:
```js
class MyError extends Error {
  constructor(message, code) {
    super(message);
    this.extensions = { code };
  }
}

throw new MyError('Not available', 'NOT_FOUND')
```

PR #935, opened 2017-06-29
Inline invariant transform

kassens:
This makes the format string of `invariant` lazily evaluated similar to what the
React transforms do. This makes loading a large schema noticeably faster.

Internal benchmark: https://phabricator.intern.facebook.com/P57579506


wincent:
Sweet. Thanks @kassens.

PR #939, opened 2017-07-04
Support ES modules for use in Webpack2/Rollup

baer:
This PR changes the build in the following ways:
  * A clean step is added to ensure predictable builds
  * Babel config has been divided by `env`
  * The build now outputs both a `dist/es5` and a `dist/modules` version of the code
  * `jsnext:main` was added

As of this PR, DCE from Rollup/Webpack 2+ should be supported!


baer:
This PR is up to date and should work as designed :D

.babelrc
```diff
@@ -1,26 +1,59 @@
...
-  "presets": [["env", {
-    "targets": {
-      "node": 4,
-      "browsers": [
-        "ie 9",
-        "ios 9",
-        "last 2 chrome versions",
-        "last 2 edge versions",
-        "last 2 firefox versions",
+  "env": {
```
baer: _Note_: The babel config merging strategy does not allow the common plugins to be defined at the top level. I'm investigating this babel bug and will file an issue on that project when I have a good repro.

package.json
```diff
@@ -8,7 +8,8 @@
     "Daniel Schafer <dschafer@fb.com>"
   ],
   "license": "BSD-3-Clause",
-  "main": "index.js",
+  "main": "es5/index.js",
+  "jsnext:main": "modules/index.js",
```
RomanGotsiy: @baer, `jsnext:main` [is deprecated](https://github.com/rollup/rollup/issues/969). Use `module` instead.

baer: Done! Thanks, I did not know that.

.babelrc
leebyron: Any updates on this?

baer: No, not at present. Is it blocking?

danez: I think the `env` option is broken conceptually in some ways, that's why there is still discussion to deprecate it in favor of the new `.babelrc.js` config coming in babel 7. That's why there is currently also no effort put into fixing all the broken cases with `env`.

A workaround for babel 6.x is to create a local preset and reference it in the babel config. That way you have full control about which plugin is activated in which environment by simply using javascript + environment variables.

https://github.com/babel/babel/issues/5276

baer: Wow! Thanks @danez, that's really exciting news!

PR #940, opened 2017-07-05
Fix Time

bondareko:


wincent:
Hi @bondareko. Would you mind rebasing this? It's hard to see what's going on in here, and I can't actually see any obviously new commits in the stack.

PR #946, opened 2017-07-12
Remove extraneous transformation

baer:
This PR removes the babel plugin [transform-es3-property-literals](https://babeljs.io/docs/plugins/transform-es3-property-literals/). It is extraneous compilation since the [quote-props](http://eslint.org/docs/rules/quote-props) ESLint rule defined [here](https://github.com/graphql/graphql-js/blob/master/.eslintrc#L202) already takes care of this.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -16,7 +16,6 @@
     "syntax-async-functions",
     "syntax-async-generators",
     "transform-class-properties",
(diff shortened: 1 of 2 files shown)


wincent:
Looks reasonable. Thanks!

PR #949, opened 2017-07-14
Account for query's offset in file for errors

dwwoelfel:
Adds support for providing the location in a file that the graphQL source was located. If the location is provided as part of the source, the line numbers in the formatted errors will be updated to account for the offset.

For example, in file Foo.js:

```
1  const x = 1
2 
3  graphql`
4  query {
5   ?
6  }
7  `
```

The lexer will throw an error with message:

```
Parse error: GraphQLError: Syntax Error Foo.js (5:2) Cannot parse the unexpected character "?".
4: query {
5:  ?
    ^
6: }
```

It's up to the caller to provide the correct offset in the source.

Not sure if `location` is the best name for the field in the `Source` object. Very open to suggestions!

### DIFF
--- a/src/error/syntaxError.js
+++ b/src/error/syntaxError.js
@@ -23,8 +23,9 @@ export function syntaxError(
 ): GraphQLError {
   const location = getLocation(source, position);
   const line = location.line + source.locationOffset.line - 1;
+  const column = location.column + source.locationOffset.column - 1;
   const error = new GraphQLError(
-    `Syntax Error ${source.name} (${line}:${location.column}) ${description}` +
+    `Syntax Error ${source.name} (${line}:${column}) ${description}` +
       '\n\n' +
       highlightSourceAtLocation(source, location),
     undefined,
@@ -41,22 +42,28 @@ export function syntaxError(
 function highlightSourceAtLocation(source, location) {
   const line = location.line;
   const lineOffset = source.locationOffset.line - 1;
+  const columnOffset = source.locationOffset.column - 1;
   const contextLine = line + lineOffset;
   const prevLineNum = (contextLine - 1).toString();
   const lineNum = contextLine.toString();
   const nextLineNum = (contextLine + 1).toString();
   const padLen = nextLineNum.length;
   const lines = source.body.split(/\r\n|[\n\r]/g);
+  lines[0] = whitespace(columnOffset) + lines[0];
   return (
     (line >= 2 ?
       lpad(padLen, prevLineNum) + ': ' + lines[line - 2] + '\n' : '') +
     lpad(padLen, lineNum) + ': ' + lines[line - 1] + '\n' +
-    Array(2 + padLen + location.column).join(' ') + '^\n' +
+    whitespace(2 + padLen + location.column + columnOffset - 1) + '^\n' +
     (line < lines.length ?
       lpad(padLen, nextLineNum) + ': ' + lines[line] + '\n' : '')
   );
 }
 
+function whitespace(len) {
+  return Array(len + 1).join(' ');
+}
+
 function lpad(len, str) {
-  return Array(len - str.length + 1).join(' ') + str;
+  return whitespace(len - str.length) + str;
 }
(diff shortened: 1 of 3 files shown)


src/language/__tests__/lexer-test.js
```diff
@@ -117,7 +116,26 @@ describe('Lexer', () => {
...
 
+  it('updates line numbers in error for file context', () => {
+    expect(() => {
+      const str = '' +
+      '\n' +
+      '\n' +
+      '    ?\n' +
+      '\n';
+      const source = new Source(str, 'foo.js', { line: 11, column: 0 });
+      return createLexer(source).advance()
```
wincent: Linter wants a semi here: https://travis-ci.org/graphql/graphql-js/jobs/253451244

src/language/source.js
```diff
@@ -8,18 +8,26 @@
...
 export class Source {
   body: string;
   name: string;
+  location: Location;
 
-  constructor(body: string, name?: string): void {
+  constructor(body: string, name?: string, location?: Location): void {
     this.body = body;
     this.name = name || 'GraphQL request';
+    this.location = location || { line: 1, column: 0 };
```
wincent: Should column be 1-indexed here as well?

src/error/syntaxError.js
```diff
@@ -22,9 +22,11 @@ export function syntaxError(
   description: string
 ): GraphQLError {
   const location = getLocation(source, position);
+  const line = location.line + source.location.line - 1;
```
wincent: I suspect we don't want to name this `source.location`, to avoid confusion with the existing `SourceLocation` type and `getLocation` function. Maybe we could call it something like `base` or `offset` or something else.

wincent: We should probably also handle the column offset too, but only on the first line. eg. if my source looks like this (ie. starting line 2, starting column 24):

```
  12345678901234567890123456789
1:
2:     graphql` ... X ...
3:  Y ...
4:`
```

If there is an error at "X", we should report that as being on line 2, column 29, and if there is an error at "Y" then we should report it as being on line 3, column 3. Does that make sense?

src/language/__tests__/lexer-test.js
```diff
@@ -101,13 +101,12 @@ describe('Lexer', () => {
...
-    ?
-
-
-`)
+    const str = '' +
+    '\n' +
+    '\n' +
+    '    ?\n' +
+    '\n';
+    expect(() => lexOne(str)
```
wincent: Might be worth mentioning in the commit message why you made this change.

src/language/source.js
dwwoelfel: For some reason lines are 1-indexed and columns are 0-indexed. I just looked at what Emacs prints in the modeline when I wrote this, but just confirmed with the Babylon docs: https://github.com/babel/babylon/blob/master/ast/spec.md#node-objects

src/error/syntaxError.js
dwwoelfel: How about `locationOffset`?

Oh right! I meant to mention that in the pull request body. I wrote code to adjust for the column offset, but deleted it because it didn't seem worth the hassle for a pretty rare (as far as I can tell) special case. I can add the code again if you think it's important.

src/language/__tests__/lexer-test.js
dwwoelfel: I'll just put it back :)

src/language/__tests__/lexer-test.js
dwwoelfel: 👍

src/language/source.js
wincent: Wow. That really makes no sense at all to me. In Vim, Atom, VSCode, TextMate, both line and column are 1-indexed. `gcc` also reports errors using 1-indexed lines and columns, and I suspect many other source-processing tools will as well.

src/error/syntaxError.js
wincent: `locationOffset` sounds pretty good.

PR #961, opened 2017-07-21
Fix directive description lost in extendSchema

RomanGotsiy:
[`getDirective`](https://github.com/graphql/graphql-js/blob/master/src/utilities/extendSchema.js#L537) from `extendSchema` misses `description` field which is present in `buildASTSchema` ([here](https://github.com/graphql/graphql-js/blob/master/src/utilities/buildASTSchema.js#L274))

Added it + created test to cover this


wincent:
Thanks a bunch for this @RomanGotsiy. Looks good. I applied one tiny wording tweak and will merge this now.

Issue #979, opened 2017-08-02
defaultFieldResolver should bind source


dimitri-koenig:
Hi there

The defaultFieldResolver should bind source so that within the function call I can use "this" to access instance properties and methods.

I made a PR for this.

IvanGoncharov:
@dimitri-koenig Based on your last comment in #980 can please close this issue?

PR #980, opened 2017-08-02
defaultFieldResolver should bind source

dimitri-koenig:
PR for this issue: defaultFieldResolver should bind source #979

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -1247,7 +1247,8 @@ function (source, args, context, info) {
   if (typeof source === 'object' || typeof source === 'function') {
     const property = source[info.fieldName];
     if (typeof property === 'function') {
-      return source[info.fieldName](args, context, info);
+      /* eslint-disable no-useless-call */
+      return source[info.fieldName].apply(source, [ args, context, info ]);
     }
     return property;
   }
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -1248,7 +1248,7 @@ function (source, args, context, info) {
     const property = source[info.fieldName];
     if (typeof property === 'function') {
       /* eslint-disable no-useless-call */
-      return source[info.fieldName].apply(source, [ args, context, info ]);
+      return source[info.fieldName].call(source, args, context, info);
     }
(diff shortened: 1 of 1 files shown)


src/execution/execute.js
```diff
@@ -1247,7 +1247,8 @@ function (source, args, context, info) {
   if (typeof source === 'object' || typeof source === 'function') {
     const property = source[info.fieldName];
     if (typeof property === 'function') {
-      return source[info.fieldName](args, context, info);
+      /* eslint-disable no-useless-call */
+      return source[info.fieldName].apply(source, [ args, context, info ]);
```
papandreou: Function#call is [usually faster](https://jsperf.com/function-calls-direct-vs-apply-vs-call-vs-bind/6) than Function#apply, and since you're not passing varargs here, it'd probably be better to go:

```js
return source[info.fieldName].call(source, args, context, info);
```

dimitri-koenig:
Fair enough, there we go with call instead of apply

dimitri-koenig:
@papandreou Any more changes needed or is this PR ready to be merged?

papandreou:
@dimitri-koenig, I'm not in charge here, so I don't know :)

leebyron:
Thanks for your patience - could you please add unit tests for this?

I think this was always the intent, and was assuming that the `source[method](arg)` form would use `source` as `this` in the same way as the `source.method(arg)` form. This assumption is also behind why [no-useless-call](http://eslint.org/docs/rules/no-useless-call) is being triggered. I'm curious to know if eslint is also wrong, if so we should make sure an issue is filed on that project.

Adding a test for this may help either confirm that the existing code does this, or if it failed with the existing code that this solution is the right fix. Either way, the test would be really valuable.

dimitri-koenig:
I found the real issue, though I don't have an explanation for it:

In graphql-tools the same code is being used though an old version, where the call of the class method is being made like this:
```javascript
return property(args, context);
```

Changing this to the code used right now works:
```javascript
return source[info.fieldName](args, context, info);
```

So no need to change anything.

Issue #968, opened 2017-07-25
Adding an enum value is breaking change?


excitement-engineer:
I was wondering if adding a new value to an existing enum is considered a breaking change. The method `findBreakingChanges` method currently does not consider this to be a breaking change. If any clients break because of this addition does it mean that the clients implemented the GraphQL schema incorrectly? I was wondering if this is a deliberate omission in the `findBreakingChanges` method.

rmosolgo:
I didn't quite understand the question at first so I thought I'd share my thoughts: 

Adding an enum value isn't a breaking change in purely GraphQL terms: by adding an enum value on the schema, all existing queries will still run. Enum values hard-coded in those queries are still valid, even if they're from the subset of previous values. 

I guess it _could_ break client code. For example, if you have a hard-coded set of values, and you receive something from the server which _is not_ in that set of values, it could crash the client app. 

So: 

- Should adding an enum value be a breaking change? OR 
- Should all GraphQL clients have the assumption that _current_ enum values are a subset of _possible_ enum values?

excitement-engineer:
Thanks a lot for your help! 

As you say it is technically not a breaking change because the query will still run. However, it is still quite scary to add an extra enum state because clients can break if they simply assumed that the enum set is fixed. Unlike adding a new field to a schema, it is not as clear what the effects will be on the clients that depend on the API. 

Have you run into a project where you have had to extend an enum with additional values? If yes, did this cause any major issues?

@dschafer @leebyron I am also curious to hear if enums have been extended over time in Facebook's schema and how this was managed. Have you ever run into these issues?

leebyron:
It's not a schema-breaking change for the reasons @rmosolgo mentioned above. All existing queries continue to behave correctly.

It is an example of expanding the possible values a query could return. Other examples of this are adding a type to a union or adding an interface to a type. When building clients, it's best to be defensive against possible future expansions and handle those cases. For enums, we typically include an `else` or `switch default` clause when branching on them to handle cases the client doesn't know about. If your clients aren't programming defensively like this, then it's true that expanding the possible response values could cause issues for those clients - it's not an entirely safe change. 

At Facebook, we've encountered some minor issues in the past with expanding enum values - but good testing has helped avoid any serious issues.

If you're worried about this, I'd suggest an audit of your own code or just communicating the upcoming change clearly first.

wincent:
I wonder if we should consider adding this to the (potentially) "dangerous changes" list?

[...]

PR #984, opened 2017-08-05
Adding a value to an enum is now a dangerous change.

excitement-engineer:
Following the discussion in issue #968, adding a new value to an enum is now considered a dangerous change.

### DIFF
--- a/src/utilities/__tests__/findBreakingChanges-test.js
+++ b/src/utilities/__tests__/findBreakingChanges-test.js
@@ -32,6 +32,7 @@ import {
   findTypesRemovedFromUnions,
   findTypesThatChangedKind,
   findValuesRemovedFromEnums,
+  findValuesAddedToEnums,
   findArgChanges,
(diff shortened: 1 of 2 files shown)


excitement-engineer:
Given @leebyron's [comment](https://github.com/graphql/graphql-js/issues/968#issuecomment-317834310). Should Interface and Union expansions be considered 'dangerous' as well? If so, then I will update the PR accordingly.

wincent:
Thanks for this, @excitement-engineer.

> Should Interface and Union expansions be considered 'dangerous' as well? If so, then I will update the PR accordingly.

I would think so, but I'd almost prefer them to be separate PRs so that we have the ability to revert individually if something goes wrong.

excitement-engineer:
@wincent Thanks for checking the code. 

I will create separate PRs for the dangerous changes related to interface and union extensions when I find some spare time:) Thanks for your help!

PR #991, opened 2017-08-12
Adding a type to a union is now a dangerous change.

excitement-engineer:
Following the discussion in issue #968, adding a new type to a union is now considered a dangerous change.


leebyron:
Nice, thanks for this addition

PR #993, opened 2017-08-12
Expanded findBreakingChanges test with missing case.

excitement-engineer:
I noticed the findBreakingChanges did not perform a check for the breaking change `INTERFACE_REMOVED_FROM_OBJECT` so I added it:)


leebyron:
Gotta love tests, thanks!

PR #999, opened 2017-08-22
Simplify operationTypes validation

IvanGoncharov:
Simplify code and more importantly unify some of the exceptions.

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -165,9 +165,9 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
   }
 
   const operationTypes = schemaDef ? getOperationTypes(schemaDef) : {
-    query: nodeMap.Query && 'Query',
-    mutation: nodeMap.Mutation && 'Mutation',
-    subscription: nodeMap.Subscription && 'Subscription',
+    query: nodeMap.Query ? 'Query' : null,
+    mutation: nodeMap.Mutation ? 'Mutation' : null,
+    subscription: nodeMap.Subscription ? 'Subscription' : null,
   };
 
   if (!operationTypes.query) {
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -164,57 +164,13 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
     }
   }
 
-  let queryTypeName;
-  let mutationTypeName;
-  let subscriptionTypeName;
-  if (schemaDef) {
-    schemaDef.operationTypes.forEach(operationType => {
-      const typeName = operationType.type.name.value;
-      if (operationType.operation === 'query') {
-        if (queryTypeName) {
-          throw new Error('Must provide only one query type in schema.');
-        }
-        if (!nodeMap[typeName]) {
-          throw new Error(
-            `Specified query type "${typeName}" not found in document.`
-          );
-        }
-        queryTypeName = typeName;
-      } else if (operationType.operation === 'mutation') {
-        if (mutationTypeName) {
-          throw new Error('Must provide only one mutation type in schema.');
-        }
-        if (!nodeMap[typeName]) {
-          throw new Error(
-            `Specified mutation type "${typeName}" not found in document.`
-          );
-        }
-        mutationTypeName = typeName;
-      } else if (operationType.operation === 'subscription') {
-        if (subscriptionTypeName) {
-          throw new Error('Must provide only one subscription type in schema.');
-        }
-        if (!nodeMap[typeName]) {
-          throw new Error(
-            `Specified subscription type "${typeName}" not found in document.`
-          );
-        }
-        subscriptionTypeName = typeName;
-      }
-    });
-  } else {
-    if (nodeMap.Query) {
-      queryTypeName = 'Query';
-    }
-    if (nodeMap.Mutation) {
-      mutationTypeName = 'Mutation';
-    }
-    if (nodeMap.Subscription) {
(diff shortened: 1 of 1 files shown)


wincent:
Thanks @IvanGoncharov. Will take a look.

src/utilities/buildASTSchema.js
```diff
@@ -131,6 +131,7 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
     throw new Error('Must provide a document ast.');
   }
 
+
```
asiandrummer: nit: remove whitespace

src/utilities/buildASTSchema.js
```diff
@@ -164,57 +165,31 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
     }
   }
 
-  let queryTypeName;
-  let mutationTypeName;
-  let subscriptionTypeName;
+  let operationTypes = {
```
asiandrummer: nit: Should be `const`

src/utilities/buildASTSchema.js
```diff
@@ -164,57 +165,31 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
...
-    if (nodeMap.Mutation) {
-      mutationTypeName = 'Mutation';
-    }
-    if (nodeMap.Subscription) {
-      subscriptionTypeName = 'Subscription';
-    }
   }
 
-  if (!queryTypeName) {
+  if (!operationTypes['query']) {
```
asiandrummer: `Query`?
Also, we should probably create constants to avoid this type of typos.

src/utilities/buildASTSchema.js
```diff
@@ -280,8 +255,8 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
     });
   }
 
-  function getObjectType(typeNode: TypeDefinitionNode): GraphQLObjectType {
-    const type = typeDefNamed(typeNode.name.value);
+  function getObjectType(name: string): GraphQLObjectType {
```
asiandrummer: Hmm, this is an API change. Have we done a sweep to see if we're breaking anything? is `getObjectType` only used in this file?

src/utilities/buildASTSchema.js
```diff
@@ -254,12 +229,12 @@ export function buildASTSchema(ast: DocumentNode): GraphQLSchema {
...
-    mutation: mutationTypeName ?
-      getObjectType(nodeMap[mutationTypeName]) :
+    query: getObjectType(operationTypes['query']),
+    mutation: operationTypes['mutation'] ?
+      getObjectType(operationTypes['mutation']) :
       null,
-    subscription: subscriptionTypeName ?
-      getObjectType(nodeMap[subscriptionTypeName]) :
+    subscription: operationTypes['subscription'] ?
+      getObjectType(operationTypes['subscription']) :
```
asiandrummer: Same as above. Am I misunderstanding something? It looks like we aren't being consistent with the operation names (query vs. Query and etc)

src/utilities/buildASTSchema.js
IvanGoncharov: @asiandrummer Yes, since it doesn't have `export`

src/utilities/buildASTSchema.js
IvanGoncharov: @asiandrummer Done 👍

src/utilities/buildASTSchema.js
IvanGoncharov: @asiandrummer It's intentional since on line 174:
```js
operationTypes = {};
```

This is because schema definition is optional but if it present you should use it instead on relaying on type names.

src/utilities/buildASTSchema.js
IvanGoncharov: @wincent Fully remove strings now it: `operationTypes.subscription`.

> consistent with the operation names (query vs. Query and etc)

It's how spec working, inside schema definition it's `query` operation but if schema definition is absent then you should look for `Query` type. So naming is following the specification.

IvanGoncharov:
> I agree with the purpose of this PR, but I'm scared this would break some things potentially. Maybe I'm wrong

@asiandrummer I try to break my changes into small PRs that can be fully reviewed.

Right now I'm working on improving validation of SDL/IDL but before submitting my changes I want to simplify this function. Since if keep adding complexity to `graphql-js` without simplifying it code will quickly become unreadable.

src/utilities/buildASTSchema.js
asiandrummer:
```
const operationTypes  = schemaDef ? { ... } : {};
```
Above seems better to me, but it's more of a personal pref to me - again it's a nit

IvanGoncharov: @asiandrummer Great idea 👍 
I refactor out `getOperationTypes` function so it become:
```js
const operationTypes  = schemaDef ? getOperationTypes(schemaDef) : {
   ... 
};
```

asiandrummer:
@IvanGoncharov oh sorry, I meant to approve this PR a while ago. The changes look harmless - just wanted to double check that there are tests covering this part of the code, right?

IvanGoncharov:
> The changes look harmless - just wanted to double check that there are tests covering this part of the code, right?

@asiandrummer Yes, it's fully covered. Here is [negative cases](https://github.com/graphql/graphql-js/blob/master/src/utilities/__tests__/buildASTSchema-test.js#L639), positive cases are spread through the entire file so I can provide a single link for them.

leebyron:
Nice work, @IvanGoncharov

PR #1000, opened 2017-08-22
Remove duplicated code from buildASTSchema and extendSchema

IvanGoncharov:
Duplicated code caused issues in the past, for example, #961.
Also, it makes PRs that I working right now significantly bigger.
I tried to leave code intact as much as possible to simplify review.

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -65,6 +65,7 @@ import type {
   GraphQLNamedType,
   GraphQLInputType,
   GraphQLOutputType,
+  GraphQLFieldConfig,
 } from '../type/definition';
 
 import {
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
I did a second pass and unify even more code now both function share type caching logic.
@wincent Would be great to hear your thoughts on overall design and on the naming of the new class and methods?

wincent:
@IvanGoncharov: I'm going to look at this one now. If your intention is that it should replace #999 please go ahead and close that one.

wincent:
This is great. Really good to see the redundancy eliminated here. The naming looks fine to me. I'd suggest doing a sweep through making sure that only methods that need to be public on the `ASTDefinitionBuilder` class are actually public (at least `buildWrappedType`, for instance, looks like it could be `_buildWrappedType`.

IvanGoncharov:
> I'd suggest doing a sweep through making sure that only methods that need to be public on the ASTDefinitionBuilder class are actually public (at least buildWrappedType, for instance, looks like it could be _buildWrappedType).

@wincent Done ✅ 

> I'm going to look at this one now. If your intention is that it should replace #999 please go ahead and close that one.

I try to keep separate PRs for separate changes. So #999 change the same function but intended to simplify different parts of it. After one of the PRs will be merged I will rebase the second one.

Issue #1006, opened 2017-08-25
Breaking regression in 0.11.0


brettgoulder:
```
bgoulder-ltm1% node server.js
/Users/bgoulder/code/users/node_modules/graphql/subscription/subscribe.js:27
  var _ref = _asyncToGenerator(regeneratorRuntime.mark(function _callee(schema, document, rootValue, contextValue, variableValues, operationName, fieldResolver) {
                               ^

ReferenceError: regeneratorRuntime is not defined
    at /Users/bgoulder/code/users/node_modules/graphql/subscription/subscribe.js:27:32
    at Object.<anonymous> (/Users/bgoulder/code/users/node_modules/graphql/subscription/subscribe.js:95:2)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/bgoulder/code/users/node_modules/graphql/subscription/index.js:7:18)
```

I came across this error while just following the [GraphQL tutorial](http://graphql.org/graphql-js/).  I also tested with 0.10.5 and it is working properly. It seems to be an issue just with 0.11.0.

wincent:
Thanks for the report, @brettgoulder. Looking for suspicious items in the commit listing:

```
d7a80b6 0.11.0
d9178f4 remove dead code from buildASTSchema
0eb070f Expanded findBreakingChanges test with missing case. (#993)
0779c36 Update subscribe function to return a Promise of AsyncIterator or ExecutionResult (#918)
24a5038 Throw TypeError when coercing an array into a GraphQLString (#925)
85bfd13 Export "ExecutionArgs" type (#988)
26023b3 Adding a type to a union is now a dangerous change. (#991)
c156180 chore(package): update eslint to version 4.4.1
ee57e28 chore(package): update eslint to version 4.4.0
bb98768 Adding a value to an enum is now a dangerous change.
cbaf79c chore(package): update chai to version 4.1.1
e067f52 chore(package): update flow-bin to version 0.52.0
07a293c chore(package): update mocha to version 3.5.0
9a63126 chore(package): update flow-bin to version 0.51.1
01033b1 chore(package): update eslint-plugin-babel to version 4.1.2
ab0d655 Add missing export for "findBreakingChanges" function
[...]

PR #1007, opened 2017-08-26
Rewrite createSourceEventStream to avoid use of `async`

wincent:
We should avoid a runtime dependency on `async` support because we want the graphql package to work in as many contexts as possible with minimal configuration.

As of 0779c36f9ba621d2d4a045, even a simple invocation usage of the package like the one in the tutorial (http://graphql.org/graphql-js/) will cause an error in the absence of the regenerator runtime:

    ReferenceError: regeneratorRuntime is not defined

Rather than depending on the runtime, let's just rewrite the function to not use `async`. In order to do this, we have to drop the use of the `await` keyword inside the function and make it explicitly return a promise.

Flow and tests pass.

There is a fair bit of whitespace-only change in this, so you may want to view the diff without it (with `git diff -b` or even `git diff -w` for example).

Closes: https://github.com/graphql/graphql-js/issues/1006

### DIFF
--- a/src/subscription/subscribe.js
+++ b/src/subscription/subscribe.js
@@ -188,7 +188,7 @@ function subscribeImpl(
  * or otherwise separating these two steps. For more on this, see the
  * "Supporting Subscriptions at Scale" information in the GraphQL specification.
  */
-export async function createSourceEventStream(
+export function createSourceEventStream(
   schema: GraphQLSchema,
   document: DocumentNode,
   rootValue?: mixed,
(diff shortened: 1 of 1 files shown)


wincent:
Also tested this with the simple `server.js`:

```
var { graphql, buildSchema } = require('graphql');

// Construct a schema, using GraphQL schema language
var schema = buildSchema(`
  type Query {
    hello: String
  }
`);

// The root provides a resolver function for each API endpoint
var root = {
  hello: () => {
    return 'Hello world!';
  },
};

// Run the GraphQL query '{ hello }' and print out the response
graphql(schema, '{ hello }', root).then((response) => {
  console.log(response);
});
```

And a `npm link`-style symlink in a project with nothing else in it:

```
{
  "name": "repro",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "graphql": "0.11.0"
  }
}
```

Before:

```
ReferenceError: regeneratorRuntime is not defined
```

After:

```
{ data: { hello: 'Hello world!' } }
```

wincent:
Would also like to set up a lint rule to disallow the use of `async`/`await` outside the test suite, but I don't think there is an eslint rule for that. :cry:

wincent:
Merging this to get a mitigation out quickly. Can follow-up later if there is any other feedback.

robzhu:
Thanks for the fix, this was totally my fault =(.

wincent:
> Thanks for the fix, this was totally my fault

A very easy "mistake" to make. I blame not having a linter, but now we have one. :smile:

Issue #1008, opened 2017-08-26
Lint for use of async in runtime code


wincent:
To avoid issues like #1006 cropping up again, we want to lint against accidental use of `async` in runtime code (using it in tests is fine).

Unfortunately, I don't think this is a built in eslint rule, so we may have to implement a temporary hack (as well as filing an issue or PR upstream).

PR #1010, opened 2017-08-26
Add linter error preventing use of `async` outside the test suite

wincent:
We want to keep the runtime dependency footprint small, so that means avoiding use of `async` at runtime (which creates a dependency on the Regenerator runtime).

There is no built-in eslint rule for this, so we make a custom one.

Test plan:

Add an `async` function to a file, run `yarn run lint` and see:

    graphql-js/src/subscription/subscribe.js
    288:1  error  async functions are not allowed outside of the test suite  no-async

    ✖ 1 problem (1 error, 0 warnings)

Note that no errors are issued for the `async` functions in the test suite.

Closes: https://github.com/graphql/graphql-js/issues/1008

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -219,6 +219,8 @@
     "vars-on-top": 0,
     "wrap-iife": 2,
     "wrap-regex": 0,
-    "yoda": [2, "never", {"exceptRange": true}]
+    "yoda": [2, "never", {"exceptRange": true}],
+
+    "no-async": 2
   }
 }
--- a/resources/lint/no-async.js
+++ b/resources/lint/no-async.js
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) 2017, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+module.exports = function(context) {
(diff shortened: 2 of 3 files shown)


.eslintrc
```diff
@@ -219,6 +219,8 @@
     "vars-on-top": 0,
     "wrap-iife": 2,
     "wrap-regex": 0,
-    "yoda": [2, "never", {"exceptRange": true}]
+    "yoda": [2, "never", {"exceptRange": true}],
+
```
robzhu: extra line break intended here?

resources/lint/no-async.js
```diff
@@ -0,0 +1,25 @@
...
+module.exports = function(context) {
+  if (context.getFilename().match(/\b__tests__\b/)) {
+    return {};
+  } else {
+    return {
+      FunctionDeclaration: function(node) {
+        if (node.async) {
+          context.report(
+            node,
+            'async functions are not allowed outside of the test suite'
```
robzhu: I suggest adding more detail here: "async functions are not allowed outside of the test suite because older versions of NodeJS do not support async functions without transpilation. Instead, use explicit Promises."

.eslintrc
wincent: Yeah, I wanted to separate the custom rules. Would have put in a comment to make it clear but this is faux-JSON so there are no comments... (We do use a blank line as a separator higher up in the file, at least.)

resources/lint/no-async.js
wincent: Good call.

Issue #1011, opened 2017-08-27
Flow error in subscription/mapAsyncIterator in 0.11.1


excitement-engineer:
I wanted to update to graphql@0.11.1. However, when I run the flow check in my codebase I get an error in `subscription/mapAsyncIterator` despite not using this function in my codebase. 

I am currently running flow@0.52.0. This is the output I am getting in my console:

<img width="857" alt="screen shot 2017-08-27 at 19 12 34" src="https://user-images.githubusercontent.com/19558321/29752324-2df2aa14-8b5c-11e7-8aad-0615acb05c95.png">

Anyone know if this is an issue with graphql-js?

IvanGoncharov:
@excitement-engineer Yes this is an issue `graphql`.
Test for  `graphql` itself is passing since [.flowconfig](https://github.com/graphql/graphql-js/blob/master/.flowconfig) has:
```
[options]
experimental.const_params=true
```
As a tempory workaround, you can add the same thing to you `.flowconfig`.


@wincent @robzhu Code in question was added in #918, see [this code](https://github.com/graphql/graphql-js/commit/0779c36f9ba621d2d4a04539fec257cbbc5fe9e4#diff-5da861070d89e445a77546c3eacc5946R40).

It's exactly the same problem that blocks  update in `express-graphql` https://github.com/graphql/express-graphql/pull/383

I'm not sure why it's only affecting Node 7+.

robzhu:
Thanks for filing this issue. This is definitely an issue with graphql-js and we'll work to get it fixed ASAP.

wincent:
> I'm not sure why it's only affecting Node 7+.

I would have thought it's because we only run Flow on one Node version in order to speed up Travis runs, but it turns out that [we don't](https://github.com/graphql/express-graphql/blob/5f4e9a0cf1548a67db5dcaa622347c05c26fac44/.travis.yml) (in other repos, [we do](https://github.com/facebook/relay/blob/64d14b6e0cb9b00b9ddd663991beddf325123b98/.travis.yml)).

¯\\_(ツ)_/¯

wincent:
Just cut v0.11.2 so going to close this.

PR #1012, opened 2017-08-27
Fix flow error in mapAsyncIterator

robzhu:
https://github.com/graphql/graphql-js/issues/1011

This fix attempts to address the flow errors found in libraries that depend on GraphQL-js but do not have `experimental.const_params=true` in their .flowconfig. 

With this explicit null check, I removed the `experimental.const_params=true` option from .flowconfig and saw all tests still pass.

### DIFF
--- a/src/subscription/mapAsyncIterator.js
+++ b/src/subscription/mapAsyncIterator.js
@@ -39,8 +39,9 @@ export default function mapAsyncIterator<T, U>(
 
   let mapReject;
   if (rejectCallback) {
-    mapReject = error =>
-      asyncMapValue(error, rejectCallback).then(iteratorResult, abruptClose);
+    mapReject = error => rejectCallback ?
+      asyncMapValue(error, rejectCallback).then(iteratorResult, abruptClose) :
+      error;
   }
 
   return {
--- a/src/subscription/mapAsyncIterator.js
+++ b/src/subscription/mapAsyncIterator.js
@@ -39,11 +39,10 @@ export default function mapAsyncIterator<T, U>(
 
   let mapReject;
   if (rejectCallback) {
-    // The redundant rejectCallback null check provides compatibility with
-    // different flow config options.
-    mapReject = error => rejectCallback ?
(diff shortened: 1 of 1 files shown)


src/subscription/mapAsyncIterator.js
```diff
@@ -39,8 +39,9 @@ export default function mapAsyncIterator<T, U>(
 
   let mapReject;
   if (rejectCallback) {
-    mapReject = error =>
-      asyncMapValue(error, rejectCallback).then(iteratorResult, abruptClose);
+    mapReject = error => rejectCallback ?
```
wincent: A comment here would be helpful. Without `git blame`-ing this to see why the check is there, it's not obvious that it's for Flow.

robzhu:
Added a comment and replaced the redundant null check with a const variable to capture rejectCallback. 

> You mentioned changing the .flowconfig. Should that change also be included in this commit?

No, the change to .flowconfig was only to simulate running flow from dependent projects.

wincent:
> No, the change to .flowconfig was only to simulate running flow from dependent projects.

I meant though, not having that config means we're more likely to catch problems in dependent projects, at least for a while.

IvanGoncharov:
@wincent Can you please release `0.11.2` and update dependency in `express-graphql`?

robzhu:
> I meant though, not having that config means we're more likely to catch problems in dependent projects, at least for a while.

I'm fine with removing it to detect possible breakage in this repo early. @asiandrummer, you added that line in https://github.com/graphql/graphql-js/pull/885, any opposition/context?. In my branch, if I remove it, tests/flow still pass.

wincent:
@IvanGoncharov: v0.11.2 is out. express-graphql update should be out in a few minutes.

asiandrummer:
@robzhu @wincent I think I added the flag to fix the flow errors from Nuclide, for the exactly same reasons with why this PR exists (Nuclide had this flag, and graphql-js didn't). I think we should still have this flag, as it allows for a more strict flow check.

PR #988, opened 2017-08-08
Export "ExecutionArgs" type

IvanGoncharov:
I need this type to apply some transformation on args before passing it to `execute`.
This PR just expose type without changing any run-time behavior.
@wincent Can you please review it?

PR #1027, opened 2017-09-07
Add missing ExecutionArgs export

IvanGoncharov:
In #988 I forget to export `ExecutionArgs` in all necessary places 🤦‍♂️ 
This one-line PR fix that. 

### DIFF
--- a/src/execution/index.js
+++ b/src/execution/index.js
@@ -1,3 +1,4 @@
+/* @flow */
 /**
  *  Copyright (c) 2015, Facebook, Inc.
  *  All rights reserved.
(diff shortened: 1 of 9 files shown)


IvanGoncharov:
@wincent I was wondering why Flow silently ignored missing export. So I discovered that a few files missing `@flow` directive.
And [lexer.js](https://github.com/graphql/graphql-js/blame/master/src/language/lexer.js#L1) had `/* @flow /` since the first commit, not sure how Flow is parsing such comments.
I aso unify `@flow` directives to always be in a first line, so you can use simple shell script to find all files without it.
After enabling flow checks on more files two new issue appeared which I fixed in separate commit.

wincent:
Thanks @IvanGoncharov.

FWIW I don't really like the hoisting up of the `@flow` annotation to the top; it smells of special-casing to me and `@flow` shouldn't be more special than any other annotation that we might use (and we do use quite a few at FB). Having said that, consistency is good, and there is a precedent for this style, so... ¯\\_(ツ)_/¯

PR #1029, opened 2017-09-09
Keep comments when printing a parsed query

astorije:
**WIP WIP WIP**

This follows attempt and discussion started at https://github.com/graphql/graphiql/pull/578.

Opening this early to validate: am I going into the wall here?

Because the printer takes a parsed AST as an input, what this is doing for now is trying to keep comments in said parsed AST.
But is it something actually reasonable? Should comments be kept in the AST? If not, I'm not sure how to achieve this without changing the printer so that it takes the result of `lexer` as an input instead of the result of `parse`.

I was recommended in the link above to take example from [prettier](https://github.com/prettier/prettier) which already supports printing comments (as well as other goodies such breaking arguments into multiple lines when a line gets too long), but being entirely new to both codebases, it's rather difficult for me to understand, at least for now.

I'm more than happy to spend more time on this as I do want to achieve this, but any help would be welcome here :)

### DIFF
--- a/src/language/__tests__/kitchen-sink.graphql
+++ b/src/language/__tests__/kitchen-sink.graphql
@@ -52,6 +52,7 @@ fragment frag on Friend {
 }
 
 {
+  # Here is a comment
   unnamed(truthy: true, falsey: false, nullish: null),
   query
 }
--- a/src/language/__tests__/printer-test.js
+++ b/src/language/__tests__/printer-test.js
@@ -36,6 +36,18 @@ describe('Printer', () => {
   });
 
   it('correctly prints non-query operations without name', () => {
+    const queryWithComment = parse(dedent`query {
+      # Test
+      id
+    }`);
+    console.log(JSON.stringify(queryWithComment, null, 2));
+    expect(print(queryWithComment)).to.equal(dedent`
+      {
+        # Test
+        id
+      }
+    `);
+
     const queryAstShorthanded = parse('query { id, name }');
     expect(print(queryAstShorthanded)).to.equal(dedent`
       {
--- a/src/language/ast.js
+++ b/src/language/ast.js
@@ -126,6 +126,7 @@ export type ASTNode =
   | SelectionSetNode
   | FieldNode
   | ArgumentNode
+  | CommentNode
   | FragmentSpreadNode
   | InlineFragmentNode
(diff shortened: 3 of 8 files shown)


jeff-zohrab:
Hi @leebyron - Some questions which may help move this forward.  I feel this may be harder than it appears on the surface, though.

- re "The lexer returns a linked list of nodes which include comments" - the lexer-test.js test `it('produces double linked list of tokens, including comments'` gives the linked list, but I can't see a simple way to map this D-L-list to an AST as is used by the printer visitor method, `printDocASTReducer`
- `parser.parse` eventually calls `parser.skip()`, which calls `lexer.advance()`.  `advance` skips all comments (the same double-linked-list test above checks that).  I tried a hack to prevent skipping comments (see https://github.com/graphql/graphql-js/compare/master...jeff-zohrab:try-leaving-comments-during-parsing?expand=1), but that didn't work out easily as the parser AST doesn't include comments.  If you or @astorije have suggestions on my branch I can take another shot at it.

Mixing ASTs with human things like comments is tough, it's a common point of discussion for parsers (at least based on a few hours of reading).  Maybe comments should be included as part of the top-level grammar for the language, but that would introduce other issues.

Cheers!

PixnBits:
>`lexer.advance()`...skips all comments (the same double-linked-list test above checks that).

I saw the same problem, but then found that descriptions have been changed from comments to string literals. There is a [`commentDescriptions` option (part of the `BuildSchemaOptions` TS type)](https://github.com/graphql/graphql-js/blob/8c96dc8276f2de27b8af9ffbd71a4597d483523f/src/utilities/buildASTSchema.js#L88-L95) that explains this a bit more.

Changing the descriptions in my schema from `# description` to `"description"` retained the descriptions through `print(parse())` using v14.2.1 of `graphql-js`.

This PR may not be needed anymore?

IvanGoncharov:
@PixnBits It became a less urgent issue after descriptions were moved to string literals.
However it still important if you want to prettify query/SDL without loosing comments like prettier does, see #1557

astorije:
I'm going to close this to reduce clutter on this repo, as I'm not going to be able to work on it anytime soon. Anyone who feels like it can help taking over :)
Thanks for the support, everyone!

PR #1033, opened 2017-09-11
Correct types for 'path' & 'parentType' of GraphQLResolveInfo

IvanGoncharov:
In this PR I corrected types for `path` and `parentType` fields of `GraphQLResolveInfo`.
Without this info, I forced to add additional checks to my resolvers:
```ts
if (info.parentType instanceof GraphQLObjectType) {
  // ...
}
if (info.path) {
 // ...
}
```
Both checks always true and this is obvious from  code of [buildResolveInfo](https://github.com/APIs-guru/graphql-js/blob/2d43dd65a1b784ac790e66d0cc64a4c7b0a5a901/src/execution/execute.js#L701-L702) function.
But I forced to add above checks in all my resolvers 😞 

@wincent This PR affects only typings and doesn't change any runtime behaviour. Can you please review it?

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -643,7 +643,7 @@ export type GraphQLResolveInfo = {
   variableValues: {[variable: string]: mixed};
 };
 
-export type ResponsePath = { prev: ?ResponsePath, key: string | number };
+export type ResponsePath = { prev: ResponsePath | void, key: string | number };
 
 export type GraphQLFieldConfig<TSource, TContext> = {
   type: GraphQLOutputType;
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
I rebased this PR.

leebyron:
Thanks for rebasing, I'm taking a look now

src/type/definition.js
```diff
@@ -634,7 +634,7 @@ export type GraphQLResolveInfo = {
   fieldName: string;
   fieldNodes: Array<FieldNode>;
   returnType: GraphQLOutputType;
-  parentType: GraphQLCompositeType;
+  parentType: GraphQLObjectType;
```
leebyron: Good catch fixing this.

PR #1035, opened 2017-09-12
Check isDeprecated instead of deprecationReason in printDeprecated

ty2:
`isDeprecated` is to determine deprecation instead of `deprecatedReason`.

currently `"isDeprecated"` is totally ignored, for example, if the input is like `{"isDeprecated": false, "deprecationReason": "-"}` it will be treated as deprecated, it is wrong.


leebyron:
Thanks!

Issue #1045, opened 2017-09-25
graphql-js 0.11.3 and Flow 0.55.0 incompatibilities


finneganh:
Hi pals. Looks like the latest Flow upgrade's breaking changes are incompatible with GraphQL 0.11.3, specifically "Fixed a bug that caused unsoundness with $ObjMap."

https://github.com/facebook/flow/releases/tag/v0.55.0

Seeing a lot of errors like:

Error: node_modules/graphql/execution/execute.js.flow:289
289:     Object.create(null);
         ^^^^^^^^^^^^^^^^^^^ Object.create. This type is incompatible with
288:   const fragments: {[name: string]: FragmentDefinitionNode} =
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type

Error: node_modules/graphql/execution/execute.js.flow:350
350:     Object.create(null),
         ^^^^^^^^^^^^^^^^^^^ Object.create. This type is incompatible with the expected param type of
514:   fields: {[key: string]: Array<FieldNode>},
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type

leebyron:
Thanks for the report. We're looking into this, and think it might be a bug with Flow. We'll hopefully have this resolved soon

asiandrummer:
Thanks for the report @finneganh!

PR #1048, opened 2017-09-29
fix flow errors by annotating null prototypes

asiandrummer:
From `flow@0.55.0`, flow started doing a more strict `null` prototype check. Fix them by annotating them with:
```
__proto__: null
```

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -12,4 +12,4 @@
 experimental.const_params=true
 
 [version]
-^0.54.0
+^0.56.0
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "eslint": "4.4.1",
     "eslint-plugin-babel": "4.1.2",
     "eslint-plugin-flowtype": "2.35.0",
-    "flow-bin": "^0.54.0",
+    "flow-bin": "0.56.0",
     "isparta": "4.0.0",
     "mocha": "3.5.0",
     "sane": "2.0.0"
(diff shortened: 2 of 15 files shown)


leebyron:
I think this would be easier to understand if you defined a utility type for this in `jsutils/`

```js
export type ObjMap<T> = { [key: string]: T, __proto__: null };
```

Then you use `ObjMap<SomeType>` everywhere instead of `{ [key: string]: SomeType }` which I think is easier to understand.

asiandrummer:
Fixes and closes #1045.

leebyron:
Looks great! Though I think you may have missed some manual `__proto__: null` candidates in there

cloudkite:
This has a bit of an unfortunate side effect of having to add `__proto__: null` everywhere that you have strongly typed resolve functions

```
type Args = {
  id: string,
  __proto__: null,
}

async function resolve(root: *, { id }: Args, ctx: Context): Promise<*> {
```

any work arounds for this?

asiandrummer:
@leebyron oops, I only fixed the ones that gave me flow errors at the moment. Do we need to change all of them?

@cloudkite I'm not sure to be honest - I was inspired to make these fixes as a hotfix to unblock other tools depending on `graphql-js`. Does anyone know how we could fix this more properly? I'm open to suggestions.

leebyron:
This isn't great - we should probably specifically make `Args` a prototyped Object to avoid making people type them in this way.

PR #1053, opened 2017-10-02
make buildObjectDef more robust

mohawk2:
The introspection schema says `__Type.interfaces` is nullable. This protects the function from throwing an exception when it is in fact `null`.

### DIFF
--- a/src/utilities/__tests__/buildClientSchema-test.js
+++ b/src/utilities/__tests__/buildClientSchema-test.js
@@ -741,6 +741,38 @@ describe('Type System: build schema from introspection', () => {
       );
     });
 
+    it('throws when null interfaces given', () => {
+      const nullInterfaceIntrospection = {
+        __schema: {
+          queryType: { name: 'QueryType' },
+          types: [
+            {
+              kind: 'OBJECT',
+              name: 'QueryType',
+              fields: [
+                {
+                  name: 'aString',
+                  args: [],
+                  type: { kind: 'SCALAR', name: 'String', ofType: null },
+                  isDeprecated: false,
+                }
+              ],
+              interfaces: null,
+            }
+          ]
+        }
+      };
+
+      expect(
+        () => buildClientSchema(nullInterfaceIntrospection)
+      ).to.throw(
+        'Introspection result missing interfaces: {"kind":"OBJECT",' +
+        '"name":"QueryType","fields":[{"name":"aString","args":[],' +
+        '"type":{"kind":"SCALAR","name":"String","ofType":null},' +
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@mohawk2 If `kind === 'OBJECT'` then `interfaces` can't be `null`, see here:
http://facebook.github.io/graphql/October2016/#sec-Object
All fields that allow for null value explicitly marked as such, e.g. `description`.

mohawk2:
It's a fact that the introspection schema defines `__Type.interfaces` as nullable. Therefore, a GraphQL implementation that is slightly misbehaving might return a `null` and GraphQL itself would not object from a nullability point of view. That introspected schema would (in fact, did) blow up GraphiQL with a pretty red, almost incomprehensible stack trace.

The doc you cite says:

> `interfaces`: The set of interfaces that an object implements.

Are you saying that a "set" is inherently not null? Are you saying that the change I am suggesting here is wrong from some theoretical point of view?

IvanGoncharov:
> It's a fact that the introspection schema defines __Type.interfaces as nullable. Therefore, a GraphQL implementation that is slightly misbehaving might return a null and GraphQL itself would not object from a nullability point of view. 

@mohawk2 Personally I'm against supporting spec deviations especially as part of the reference implementation. 

> That introspected schema would (in fact, did) blow up GraphiQL with a pretty red, almost incomprehensible stack trace.

On the other hand, it would be great if `graphql-js` provides better error checking something like:
```js
invariant(Array.isArray(objectIntrospection.interfaces), '...')
```

> Are you saying that a "set" is inherently not null? Are you saying that the change I am suggesting here is wrong from some theoretical point of view?

I'm suggesting that invalid input should be rejected with descriptive error message.

mohawk2:
@IvanGoncharov On reflection I think you're right. I have adjusted this to now do an `invariant` instead. I fear making the stack trace more comprehensible is beyond me for now!

IvanGoncharov:
@mohawk2 I believe similar check was implemented by @leebyron as part of #1063
See here: https://github.com/graphql/graphql-js/pull/1063/files#diff-768f1266aca135746881bd4d9ea9093eR246

mohawk2:
@IvanGoncharov I believe that too! That's why there were merge conflicts, now resolved. I'm still pull-requesting this test, but @leebyron 's code change does the same thing mine did.

mohawk2:
To be clear, I am still keeping this PR open because it tests something that #1063 implemented but did not test as far as I can see. If I am wrong, let me know!

leebyron:
Thanks for adding a test!

PR #1058, opened 2017-10-06
Check if `process` is defined (for usage in the browser)

aortbals:
When running this library in the browser, Node.js's `process` global is not necessarily defined. Usage in the browser is great for [certain use cases](https://medium.com/taller-team/graphql-today-using-apollo-for-applications-that-still-depend-on-rest-apis-839895ce20d0). This adds a simple check that `process` is defined before using it in `assertValidName.js`


wincent:
Thanks @aortbals. Seems quite reasonable.

PR #1061, opened 2017-10-12
docs: Fix possible typo

puradox:
The argument to the GraphQLList class should be of type `GraphQLType`, but
the docs of `GraphQLList` does not reflect this. In this case, `Person` is
`undefined`, when it *probably* should have been `PersonType`.

I wasn't quite sure if this was a typo or not, but I didn't think it would hurt to
send this PR anyway.


wincent:
Thanks @puradox. Looks good to me.

PR #1064, opened 2017-10-13
Upgrade to Flow v0.57.1

gabelevi:
So unfortunately the `type Thunk<T> = (() => T) | T` type is a little ambiguous, and Flow v0.57.0 doesn't like it much. It has trouble telling which member of the union to choose. Adding return types to the thunks helps disambiguate.

@calebmer and I were discussing maybe adding a "anything but a function type" so we could write

```js
type Thunk<T: $NotAFunction> = (() => T) | T
```

but that's future work.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -12,4 +12,4 @@
 experimental.const_params=true
 
 [version]
-^0.56.0
+^0.57.0
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "eslint": "4.4.1",
     "eslint-plugin-babel": "4.1.2",
     "eslint-plugin-flowtype": "2.35.0",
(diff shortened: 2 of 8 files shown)


gabelevi:
For the record, the errors looked like this before I fixed them: https://gist.github.com/gabelevi/8cf9afd5c2b33372c79b4f496a073fb6

leebyron:
> @calebmer and I were discussing maybe adding a "anything but a function type" so we could write `$NotAFunction`

There have been lots of cases where I've actually wanted `$Not<T>` for this kind of disambiguation!

leebyron:
Could we use exact types for the internal definitions in type/definition.js to help flow understand that `() => { ... }` is not one of the object types its looking for?

Issue #1057, opened 2017-10-06
returning null for a custom scalar's serialize


bradzacher:
I'm trying to build some custom types to create reusable type/value checking within my API.

The problem I've got is that if there's a problem with the serialization, there's no way to silently ignore the value.

For example, in mysql, a non-null `DATETIME` that's ignored at insert time is assigned the value `0000-00-00 00:00:00`.
The `mysql2` DB adapter automatically converts the db value to a JS date via the `Date` constructor, which for the zero'd value above gives an `Invalid Date`.

I was creating a custom date type with the `serialize` function as follows:
```TS
serialize: function(value : Date) {
    if (isNaN(value.getTime()) {
        return null
    }

    return customDateFormatter(value)     
},
```

The problem is that if you return null from the `serialize` function, graphql will automatically throw an error on your behalf (https://github.com/graphql/graphql-js/blob/e236ca23e26bff89de2e05cc95ba179d253662fd/src/execution/execute.js#L1015-L1021).

In `execute.js` `completeValue`: https://github.com/graphql/graphql-js/blob/e236ca23e26bff89de2e05cc95ba179d253662fd/src/execution/execute.js#L907-L910
The internals will happily silently return null (and not call `serialize`) if the value that comes into the function is null, but there is no way to stop it throwing if `serialize` returns `null`/`undefined`.

Considering that I can happily throw a handled exception from within the `serialize` function, how come you chose to also automatically throw one for me?

I can think of a lot of use cases where you might want to silently ignore the fact that `serialize` returns null.

Is it possible to get a way to make graphql-js allow a `serialize` function to return `null`?

bradzacher:
Reading through the graph-ql spec - in particular this section: https://facebook.github.io/graphql/October2016/#sec-Value-Completion

> 4. If *fieldType* is a Scalar or Enum type:
>     a. Return the result of “coercing” *result*, ensuring it is a legal value of *fieldType*, otherwise **null**.

It seems like the design of graphql-js goes against the spec in this instance.

bradzacher:
It's a bit fuzzy though, because this section: https://facebook.github.io/graphql/October2016/#sec-Scalars

> A GraphQL server, when preparing a field of a given scalar type, must uphold the contract the scalar type describes, either by coercing the value or producing an error.
> ...
> For example, a GraphQL server could be preparing a field with the scalar type `Int` ... If the server encounters some value that cannot be reasonably coerced to an `Int`, then it must raise a field error.

[...]

PR #1066, opened 2017-10-17
added config option to scalar type to allow the serialize to return null

bradzacher:
as per https://github.com/graphql/graphql-js/issues/1057.


leebyron:
Thanks for your work on this.

I'd like to avoid changing behavior via configuration flags in favor of improved behavior overall. I'll put up a PR to solve the original issue

PR #1081, opened 2017-11-03
Prettier!

kassens:
I was surprised to see graphql-js wasn't using prettier yet when my code wasn't auto-formatted on save.

This adds prettier via eslint plugin and `yarn prettier` command using the options that most closely reflect the current formatting decisions.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,8 @@
 
   "plugins": [
     "babel",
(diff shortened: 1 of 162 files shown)


trevorah:
This is cool, but wouldnt the diff be smaller if you got rid of `"trailingComma": "all"`?

PR #1088, opened 2017-11-15
Option to resolve conflicts when extending schema

brysgo:
I keep a copy of this `extendSchema` script so that I can add an argument to the field to choose which schema to go to (extended local schema, or original remote), not sure if there is general interest in doing similar things, but I will leave this here for discussion.


leebyron:
Thanks for putting this up, but I think this pattern is one I'd rather not encourage. While working on the SDL spec there is language explaining this case as invalid.

I think it's totally reasonable to have a script (or published package) which does this additional work though.

PR #1090, opened 2017-11-16
fix elasticsearch errors being mistaken for graphql errors

trevorah:
Elasticsearch [records its index path](https://github.com/elastic/elasticsearch-js/blob/6c4882681fe91c0c90901aa47f68b02dc9ed58f8/src/lib/transport.js#L298) and [attaches it to errors](https://github.com/elastic/elasticsearch-js/blob/14.x/src/lib/errors.js#L23). Its useful, but sadly that `path` property means that graphql-js assumes that an elasticsearch error is a GraphQLError, and so the index path gets leaked instead of the useful graphql query path.

The fix is to be a little bit stricter.

There is probably a better fix to recognise a GraphQLError across contexts, but this will at least improve the situation. Plus, the added tests should make any better solution easier to test.

### DIFF
--- a/src/error/__tests__/locatedError-test.js
+++ b/src/error/__tests__/locatedError-test.js
@@ -1,0 +1,47 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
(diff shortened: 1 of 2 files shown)


leebyron:
Nice, thanks for adding tests!

trevorah:
@leebyron no worries, thanks for the awesome library!

PR #1094, opened 2017-11-19
Uniform parsing of queries with short-hand syntax with regular queries

### LINKED ISSUE #729: Some AST consistency questions
I'm currently working on some AST stuff, and I'm running into a few situations where there is different behavior in a way that doesn't seem intentional. Two examples so far:

### Operation name

```js
parse(`query { x }`);
```

returns `"name": undefined` in the operation definition.

```js
parse(`{ x }`);
```

returns `"name": null` in the operation definition.

### Variable definitions

```js
parse(`query { x }`);
```

returns `"variableDefinitions": []` in the operation definition.

```js
parse(`{ x }`);
```

returns `"variableDefinitions": null` in the operation definition.

### Conclusion

Looks like these two in particular are because of this line and similar ones that hardcode the default values: https://github.com/graphql/graphql-js/blob/b17d5f566292bb92e1512ba4935b23ff75a6dc22/src/language/parser.js#L275

@leebyron - Would you be interested in a PR to make these different situations consistent, for example to always return `null` for the operation name and `[]` for variable definitions?

I'd be happy to submit a PR with all of the situations I find, but first wanted to check if that would be desirable - I can imagine some tools might depend on this output not changing.

angly-cat:
To date, parsing queries with short-hand syntax returns different results from parsing regular queries.

`parse('query{a{b}}')` returns
```
...
  name: undefined,
  variableDefinitions: [],
...
```
However, `parse('{a{b}}')` returns
```
...
  name: null,
  variableDefinitions: null,
...
```
image

This shouldn't be the case. This PR made queries with short-hand syntax to be parsed as regular.

This inconsistency indirectly causes [this bug](https://github.com/graphql/graphiql/issues/234) to happen in GraphiQL. Despite that issue being closed, using GraphiQL without `express-graphql` still makes that bug to happen (with `null` value, actually, but this doesn't change much).

Closes #729.

### DIFF
--- a/src/language/__tests__/parser-test.js
+++ b/src/language/__tests__/parser-test.js
@@ -218,7 +218,7 @@ describe('Parser', () => {
           loc: { start: 0, end: 40 },
           operation: 'query',
           name: null,
-          variableDefinitions: null,
+          variableDefinitions: [],
           directives: [],
           selectionSet:
           { kind: Kind.SELECTION_SET,
@@ -288,7 +288,7 @@ describe('Parser', () => {
           loc: { start: 0, end: 29 },
           operation: 'query',
           name: null,
-          variableDefinitions: null,
+          variableDefinitions: [],
           directives: [],
           selectionSet:
           { kind: Kind.SELECTION_SET,
--- a/src/language/__tests__/parser-test.js
+++ b/src/language/__tests__/parser-test.js
@@ -270,6 +270,56 @@ describe('Parser', () => {
     );
   });
 
+  it('creates ast from nameless query without variables', () => {
+
+    const source = new Source(`query {
+  node {
+    id
+  }
+}
+`);
+    const result = parse(source);
+
+    expect(result).to.containSubset(
+      { kind: Kind.DOCUMENT,
+        loc: { start: 0, end: 30 },
+        definitions:
+        [ { kind: Kind.OPERATION_DEFINITION,
+          loc: { start: 0, end: 29 },
+          operation: 'query',
+          name: null,
+          variableDefinitions: null,
+          directives: [],
+          selectionSet:
+          { kind: Kind.SELECTION_SET,
+            loc: { start: 6, end: 29 },
+            selections:
+            [ { kind: Kind.FIELD,
+              loc: { start: 10, end: 27 },
+              alias: null,
+              name:
+              { kind: Kind.NAME,
+                loc: { start: 10, end: 14 },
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@angly-cat 😮 Wow, you solved the mystery of `Unknown operation named \"undefined\".` message.
👏  

> This PR made queries with short-hand syntax to be parsed as regular.

I think `null` is the correct value for missing name since it consistent with rest of AST format, e.g. [alias](https://github.com/angly-cat/graphql-js/blob/d0c75b128594766e757cea2198ad0dc9d44d87a8/src/language/parser.js#L399), [defaultValue](https://github.com/angly-cat/graphql-js/blob/d0c75b128594766e757cea2198ad0dc9d44d87a8/src/language/parser.js#L341), etc.

You can change line [285](https://github.com/angly-cat/graphql-js/blob/d0c75b128594766e757cea2198ad0dc9d44d87a8/src/language/parser.js#L285) to:
```
let name = null;
```
As a bonus it should fix broken tests 😉

angly-cat:
> Wow, you solved the mystery of `Unknown operation named \"undefined\". message.`
👏

Thanks for your kind words.

> I think `null` is the correct value for missing name since it consistent with rest of AST format, e.g. alias, defaultValue, etc.

I agree.

However, all of my attempts to change output of short-hand syntax query parsing caused a test to fail. (OMG it was quite hard to make tests work on Windows) So, according to these `mocha` tests, short-hand syntax queries were being parsed correctly this whole time and, therefore, the regular queries were parsed incorrectly instead! So I made a couple of changes to unify the results of parsing short-hand syntax queries with regular. (And rebase my PR after that, hence the "duplicate" commit).

IvanGoncharov:
@angly-cat Great 👍
It would be great if you add test case to cover `parse('query{a{b}}')` to prevent future regressions.

angly-cat:
@IvanGoncharov the idea about adding a test was great, so, I added it.

IvanGoncharov:
@angly-cat Great 👍
Now we need to wait for someone with commit rights to merge it.

leebyron:
Nice work!

I think since other array-parsing operations default to empty arrays instead of null, that this should follow that convention, so I made some updates.

angly-cat:
> I think since other array-parsing operations default to empty arrays instead of null, that this should follow that convention, so I made some updates.

Agreed. Thanks, @leebyron.

PR #1096, opened 2017-11-20
Detect added nullable input field or added nullable argument

aolesky:
T23304543

findBreakingChanges now includes checks for adding a nullable field to an existing input, and adding a nullable argument. These changes won't necessarily break a schema, but they can potentially cause issues. I also added unit tests for these new detections.

### DIFF
--- a/src/utilities/__tests__/findBreakingChanges-test.js
+++ b/src/utilities/__tests__/findBreakingChanges-test.js
@@ -26,6 +26,7 @@ import {
   findBreakingChanges,
   findDangerousChanges,
   findFieldsThatChangedType,
+  findFieldsThatChangedTypeOnInputObjectTypes,
   findRemovedTypes,
   findTypesRemovedFromUnions,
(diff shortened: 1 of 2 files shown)


leebyron:
cc @sam-swarr @mjmahone - does this map to your understanding of potential issues?

sam-swarr:
@aolesky, thanks for writing this! The code looks great and ready to merge.

@leebyron, yep these are two good examples of cases that fall under potentially "dangerous" changes. Not technically breaking from the spec's perspective, but could cause issues if client code is not designed well. The diff looks good to me. I don't have the right privileges to accept it. Can you do that please?

mjmahone:
@leebyron yeah this is a good example. For instance, if your client code creates strict flow types for input objects, you may now need to update your client code when a nullable field is added to the input object.

leebyron:
@sam-swarr - you should be able to add yourself in the internal "oss" tool. I'll get this resolved and merged!

PR #1104, opened 2017-12-01
(Potentially Breaking) Allow serializing scalars as null.

### LINKED ISSUE #1057: returning null for a custom scalar's serialize

leebyron:
This changes the check for null/undefined to a check for undefined to determine if scalar serialization was successful or not, allowing `null` to be returned from serialize() without indicating error.

This is potentially breaking for any existing custom scalar which returned `null` from `serialize()` to indicate failure. To account for this change, it should either throw an error or return `undefined`.

Fixes #1057
Closes #1066

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -10,6 +10,7 @@
 import { forEach, isCollection } from 'iterall';
 import { GraphQLError, locatedError } from '../error';
 import invariant from '../jsutils/invariant';
+import isInvalid from '../jsutils/isInvalid';
 import isNullish from '../jsutils/isNullish';
 import type {ObjMap} from '../jsutils/ObjMap';
 
@@ -1013,7 +1014,7 @@ function completeLeafValue(
 ): mixed {
   invariant(returnType.serialize, 'Missing serialize method on type');
   const serializedResult = returnType.serialize(result);
-  if (isNullish(serializedResult)) {
+  if (isInvalid(serializedResult)) {
     throw new Error(
       `Expected a value of type "${String(returnType)}" but ` +
       `received: ${String(result)}`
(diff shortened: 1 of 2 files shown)


kerryChen95:
Hope for publishing this to npm as soon as possible 🙏

PR #1106, opened 2017-12-02
Test: Missing subscription field should not be an "internal error"

leebyron:
A remaining issue with subscriptions is attempting to subscribe to a missing field. The existing implementation creates an internal error which is an improper blame since there is no possibility to fix it internally - it is an issue with the query.

For consistency with the same scenario for query/mutation operations, this simply returns nothing. I interpret the equivalent of "undefined" for a subscribe operation as an empty iteration.

Note: this will require spec changes, however spec changes are necessary either way to resolve the ambiguity of this exact scenario.

### DIFF
--- a/src/subscription/__tests__/subscribe-test.js
+++ b/src/subscription/__tests__/subscribe-test.js
@@ -72,7 +72,6 @@ function emailSchemaWithResolvers(subscribeFn, resolveFn) {
           type: EmailEventType,
           resolve: resolveFn,
           subscribe: subscribeFn,
-          // TODO: remove
           args: {
             priority: { type: GraphQLInt },
           },
@@ -332,7 +331,7 @@ describe('Subscription Initialization Phase', () => {
     );
   });
 
-  it('throws an error for unknown root field', async () => {
+  it('resolves to an error for unknown subscription field', async () => {
     const ast = parse(`
       subscription {
         unknownField
@@ -341,10 +340,17 @@ describe('Subscription Initialization Phase', () => {
 
     const pubsub = new EventEmitter();
 
-    expectPromiseToThrow(
-      () => createSubscription(pubsub, emailSchema, ast),
-      'This subscription is not defined by the schema.',
-    );
+    const { subscription } = await createSubscription(pubsub, emailSchema, ast);
+
+    expect(subscription).to.deep.equal({
+      errors: [
+        {
+          message: 'The subscription field "unknownField" is not defined.',
+          locations: [{ line: 3, column: 9 }],
+          path: undefined,
+        },
+      ],
+    });
   });
 
   it('throws an error if subscribe does not return an iterator', async () => {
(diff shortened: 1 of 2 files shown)


src/subscription/__tests__/subscribe-test.js
```diff
@@ -347,7 +346,7 @@ describe('Subscription Initialization Phase', () => {
     );
   });
 
-  it('throws an error for unknown root field', async () => {
+  it('unknown field should result in closed subscription', async () => {
```
robzhu: I think asking for an unknown field is a 400-class error code (something about the request was malformed). In that case, I think the `await createSubscription` call should throw. 

My reasonings is that if we don't throw there, we'll have a subscription that is in the closed state, which does not communicate that the request was invalid or what was wrong with it.

src/subscription/subscribe.js
```diff
@@ -227,10 +227,10 @@ export function createSourceEventStream(
     const fieldNodes = fields[responseName];
     const fieldNode = fieldNodes[0];
     const fieldDef = getFieldDef(schema, type, fieldNode.name.value);
-    invariant(
-      fieldDef,
-      'This subscription is not defined by the schema.'
-    );
+
+    if (!fieldDef) {
+      return resolve(emptyAsyncIterator());
```
robzhu: This doesn't feel like the right return value for a missing field case.

leebyron: What would be a better one? Presuming an error becomes the domain of a validation step, we should produce something

leebyron:
Also remember that "subscribe" is the second half of a presumed "validate & subscribe" path. I think the missing piece is a validation rule that a subscription is well-formed.

There is symmetry if my proposed behavior to `execute` which skips over unknown fields such that `{ unknownField }` yields `{}`.

My suggestion here is that the "subscribe" operation is the wrong place for this error, it should instead be caught earlier during validation, and instead this operation should be durable to the malformed input

src/subscription/subscribe.js
robzhu: In https://github.com/facebook/graphql/blob/master/spec/Section%206%20--%20Execution.md#executing-requests we describe `Subscribe` as the async variant of `ExecuteQuery`. For `ExecuteQuery`, the spec says, "Return an unordered map containing {data} and {errors}."while `Subscribe` says `Return {responseStream}.`. Perhaps we should update these to be more similar, for example, Subscriptions resolving to an unordered map of {responseStream} and {errors}.

leebyron:
http://facebook.github.io/graphql/draft/#sec-Subscription-Operation-Definitions seems to handle this case, validation errors are typically reported as 400-level.

> Perhaps we should update these to be more similar, for example, 'Subscribe` could resolve to an unordered map of {responseStream} and {errors}.

I think the flow types we've defined more closely refer to a responseStream **or** a single error response payload

leebyron:
Updated description and commit to reflect "400" level error rather than current "500" level error after offline review with @robzhu

PR #1107, opened 2017-12-02
Subscriptions: report resolver errors during initialization

leebyron:
Expanding on the tests to cover all forms of subscription resolver errors.


leebyron:
Currently this PR just expands on the test to illustrate failure. 3 of the 4 cases are consistent, but the 4th case (rejected promise) results in different behavior, rejecting the containing promise for the subscription itself.

@robzhu do you remember if this behavior was intentional?

PR #1108, opened 2017-12-02
Fix typo in block string lexer tests

cjoudrey:
While looking over https://github.com/graphql/graphql-js/pull/926, I noticed what I believe is a bad copy/paste in the block string lexer tests.

cc @leebyron


leebyron:
Indeed! Thanks for the careful read and for the quick fix

PR #1115, opened 2017-12-05
RFC: Synchronous execution

leebyron:
**This is a breaking change. Existing uses of execute() outside of async functions which assume a promise response will need to wrap in Promise.resolve()**

This allows `execute()` to return synchronously if all fields it encounters have synchronous resolvers. Notable this is the case for client-side introspection, querying from a cache, and other useful cases.

Also introduces `graphqlSync()` as a top level function which asserts sync behavior and includes validation.

Note that the top level `graphql()` function remains a Promise to minimize the breaking change, and that `validate()` has always been synchronous.

### DIFF
--- a/src/execution/__tests__/sync-test.js
+++ b/src/execution/__tests__/sync-test.js
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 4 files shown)


IvanGoncharov:
>  @IvanGoncharov I'd like your opinion of if this will cover the more general cases you've encountered.

@leebyron Looks great 👍 and solves all the cases I encounter. One less promise to worry about 😄 
I've opened #1118  and #1120 for the small suggestions.

PR #1118, opened 2017-12-06
Export GraphQLArgs

IvanGoncharov:
It's useful for building wrappers around `graphql` since you can reuse `GraphQLArgs` type. 
BTW, [ExecutionArgs](https://github.com/graphql/graphql-js/blob/master/src/index.js#L225) is also exported.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -32,6 +32,7 @@
  */
 
 // The primary entry point into fulfilling a GraphQL request.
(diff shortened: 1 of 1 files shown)


leebyron:
Wow I totally forgot to do that in #1115! I factored out the arg type with the intention on doing this. Thanks for the review and follow up!

PR #1120, opened 2017-12-06
Test that introspection results are always sync.

IvanGoncharov:
Since being able to execute introspection queries synchronously was the primary motivation for #1115 I think it worth to test it 😃


leebyron:
Agreed. Thanks for adding tests!

Issue #1080, opened 2017-11-03
Split schema validation from instantiation


kassens:
Currently, the constructors do all the validation even when we already know that a  schema is valid. We should split up these steps to ensure we can do just the validation when we know the schema is valid.

Issue #1079, opened 2017-11-03
[meta] Speed up loading of schema


kassens:
This is a meta issue tracking progress on significantly speeding up loading a schema from disk. There are multiple incremental steps that should get us there.

- [x] Split validation from instantiation. If we have a schema persisted on disk and know that we already validated it before, there's no reason to validate it *again• every time we want to load it (#1080).
- [ ] Change from classes to simple data objects that are manipulated and queried using functions. This would allow the schema to be persisted and loaded as a plain JSON object.

Note: This is still just a rough outline and there's likely more steps to get there including making sure this is an incremental change and doesn't break existing users unexpectedly.

IvanGoncharov:
Closing since, schema loading performance increased dramatically in recent years.
Also, there are no other reports about schema loading performance.

PR #1124, opened 2017-12-07
Validate schema root types and directives

leebyron:
This moves validation out of GraphQLSchema's constructor (but not yet from other type constructors), which is responsible for root type validation and interface implementation checking.

Reduces time to construct GraphQLSchema significantly, shifting the time to validation.

This also allows for much looser rules within the schema builders, which implicitly validate while trying to adhere to flow types. Instead we use any casts to loosen the rules to defer that to validation where errors can be richer.

This also loosens the rule that a schema can only be constructed if it has a query type, moving that to validation as well. That makes flow typing slightly less nice, but allows for incremental schema building which is valuable

Improves #1079 
Partial #1080 
Closes #722

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -393,11 +393,18 @@ export function getOperationRootType(
 ): GraphQLObjectType {
   switch (operation.operation) {
     case 'query':
-      return schema.getQueryType();
+      const queryType = schema.getQueryType();
+      if (!queryType) {
+        throw new GraphQLError(
+          'Schema does not define the required query root type.',
+          [operation],
+        );
+      }
+      return queryType;
     case 'mutation':
       const mutationType = schema.getMutationType();
       if (!mutationType) {
-        throw new GraphQLError('Schema is not configured for mutations', [
+        throw new GraphQLError('Schema is not configured for mutations.', [
           operation,
         ]);
       }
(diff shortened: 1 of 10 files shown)


leebyron:
This builds atop @kassens's fed961040e4d53c46bc1d85b48beba2303a7929c

src/execution/execute.js
```diff
@@ -393,7 +393,13 @@ export function getOperationRootType(
...
   switch (operation.operation) {
     case 'query':
-      return schema.getQueryType();
+      const queryType = schema.getQueryType();
+      if (!queryType) {
+        throw new GraphQLError('Schema is not configured for queries', [
+          operation,
+        ]);
+      }
+      return queryType;
```
IvanGoncharov: The error message suggests that it's normal to have schema without query and it will work fine if you don't send `query` operations against it. I think error message should say that `query` type is mandatory.

leebyron: Yeah, I agree. I'll make that change

IvanGoncharov:
@leebyron Wow, you did a lot of work to search for correct error location both in `astNode` and `extensionASTNodes` 👍 

Having `queryType` optional makes #817 more feasible and I will rebase it after this PR will be merged.

kassens:
Really excited about this. The one thing that seemed not 100% ideal was the `any` casts, but I suppose the only way around that would to actually have different types for validated and unvalidated schema which would be a breaking change.

leebyron:
Yeah, casting to `any` makes me feel bad inside, but it's the right thing to do here. I'll add some comments there explaining what's happening.

Issue #821, opened 2017-04-21
0.8.x breaking change with errors thrown from scalar resolvers


thebigredgeek:
It appears that 0.8.x introduced a breaking change with errors thrown from scalar resolvers.  Upon upgrading, it appears that errors thrown from scalar resolvers (exclusively) have, by the time they have reached `formatErrors`, been transformed into a shape like:

```bash
[1] [1]   message: 'Variable "$input" got invalid value {"stripeCardName":"asdasdada","stripeCardNumber":"42424242424223113","stripeCardExp":"00/00","stripeCardCVC":"233"}.\nIn field "stripeCardNumber": Expected type "CreditCardNumber", found "42424242424223113": The credit card number you entered appears to be invalid.',
[1] [1]   locations: [ { line: 1, column: 28 } ],
[1] [1]   path: undefined }
```
where the original error message was "The credit card number you entered appears to be invalid", and the original error object contained a bunch of useful meta data as an error created with [apollo-errors](https://github.com/thebigredgeek/apollo-errors) that instructed the client about validation issues (in this case, a credit card number that isn't valid according to ccn validation algorithms).

Was this intended?  It seems strange to force an error format like this, especially as a breaking change.

It looks like [this](https://github.com/graphql/graphql-js/pull/561) might have been the change, according to the [v0.8.0 release changelog](https://github.com/graphql/graphql-js/releases/tag/v0.8.0)

helfer:
@thebigredgeek the PR in question is [this](https://github.com/graphql/graphql-js/pull/553) one, and I believe it was very much intentional. Just looking at the error message and the PR, I'm guessing that maybe the error isn't being thrown from a resolver, but from `parseValue`?

thebigredgeek:
Yes, you are correct.  Is `parseValue` not part of the scalar's resolver, though?

PR #1126, opened 2017-12-07
Fix unhandled error when parsing custom scalar literals.

### LINKED ISSUE #910: Throwing errors from parseLiteral
After 0.10.1 throwing `GraphQLError` from `parseLiteral` function in a custom scalar type leads to unhandled promise rejection. Tracked it down to this commit: https://github.com/graphql/graphql-js/pull/883/commits/62619dfe7d3274628ad0ad7f7d91e0a42d4c0517#diff-e896d5930c71857fc9566b22defba666
Parsing has a try-catch block around it, but validation does not. Doesn't seem like it was intended, considering it says "refactor".

leebyron:
Fixes #910

This factors out the enum value validation from scalar value validation and ensures the same try/catch is used in isValidLiteralValue as isValidJSValue and protecting from errors in valueFromAST.

### DIFF
--- a/src/execution/__tests__/variables-test.js
+++ b/src/execution/__tests__/variables-test.js
@@ -613,7 +613,7 @@ describe('Execute: Handles inputs', () => {
           {
             message:
               'Variable "$value" got invalid value [1,2,3].\nExpected type ' +
-              '"String", found [1,2,3]: String cannot represent an array value: [1,2,3]',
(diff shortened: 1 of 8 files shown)


leebyron:
Ideally this also allows for custom error messages as suggested in #821 when relevant.

leebyron:
Looks like improving the issue raised in #821 will require a different and non trivial approach

Issue #981, opened 2017-08-02
Possible desire for `findDescriptionChanges` utility


eapache:
We needed it internally so I wrote a quick-and-dirty version based on `findBreakingChanges`. Right now it just walks two schemas and records all the types, fields, args, etc. which have descriptions added or changed, but not removed (aka anything that might need to be reviewed by a docs team).

If there is sufficient interest I don't mind cleaning it up and PRing it here cc @swalkinshaw

leebyron:
Sounds great, send a PR when willing

PR #1127, opened 2017-12-07
Add a utility to detect description changes

eapache:
We use this internally for pinging docs team for PR reviews.

Discussed briefly at #981 cc @leebyron.

Flow is complaining about some of this code, but Flow is just wrong; the equivalence check between `newType` and `oldType` provides the necessary safety guarantees. Not sure what the preferred solution is here.

### DIFF
--- a/src/utilities/findDescriptionChanges.js
+++ b/src/utilities/findDescriptionChanges.js
@@ -1,0 +1,142 @@
+/* eslint-disable no-restricted-syntax */
+// @flow
+
+import {
+  GraphQLInterfaceType,
+  GraphQLObjectType,
+  GraphQLEnumType,
+  GraphQLInputObjectType,
+} from '../type/definition';
+import type { GraphQLFieldMap } from '../type/definition';
+import { GraphQLSchema } from '../type/schema';
+
+/**
+ * Given two schemas, returns an Array containing descriptions of any
+ * descriptions that are new or changed and need review.
+ */
+export function findDescriptionChanges(
+  oldSchema: GraphQLSchema,
+  newSchema: GraphQLSchema,
+): Array<string> {
+  const oldTypeMap = oldSchema.getTypeMap();
+  const newTypeMap = newSchema.getTypeMap();
+
+  const descriptionChanges: Array<string> = [];
+
+  Object.keys(newTypeMap).forEach(typeName => {
+    const oldType = oldTypeMap[typeName];
+    const newType = newTypeMap[typeName];
+
+    if (newType.description) {
+      if (!oldType) {
+        descriptionChanges.push(
+          `Description added on new type ${newType.name}.`,
+        );
+      } else if (!oldType.description) {
+        descriptionChanges.push(`Description added on type ${newType.name}.`);
+      } else if (oldType.description !== newType.description) {
+        descriptionChanges.push(`Description changed on type ${newType.name}.`);
+      }
+    }
+
+    if (oldType && !(newType instanceof oldType.constructor)) {
+      return;
+    }
+
+    if (
+      newType instanceof GraphQLObjectType ||
+      newType instanceof GraphQLInterfaceType ||
+      newType instanceof GraphQLInputObjectType
+    ) {
+      const oldTypeFields: ?GraphQLFieldMap<*, *> = oldType
+        ? oldType.getFields()
+        : null;
+      const newTypeFields: GraphQLFieldMap<*, *> = newType.getFields();
+
+      Object.keys(newTypeFields).forEach(fieldName => {
+        const oldField = oldTypeFields ? oldTypeFields[fieldName] : null;
+        const newField = newTypeFields[fieldName];
+
+        if (newField.description) {
+          if (!oldField) {
+            descriptionChanges.push(
+              `Description added on new field ${newType.name}.${newField.name}`,
+            );
+          } else if (!oldField.description) {
+            descriptionChanges.push(
+              `Description added on field ${newType.name}.${newField.name}.`,
+            );
+          } else if (oldField.description !== newField.description) {
+            descriptionChanges.push(
+              `Description changed on field ${newType.name}.${newField.name}.`,
+            );
+          }
+        }
+
+        if (!newField.args) {
+          return;
+        }
+
+        newField.args.forEach(newArg => {
+          const oldArg = oldField
+            ? oldField.args.find(arg => arg.name === newArg.name)
+            : null;
+
+          if (newArg.description) {
+            if (!oldArg) {
+              descriptionChanges.push(
+                `Description added on new arg ${newType.name}.${
+                  newField.name
+                }.${newArg.name}.`,
+              );
+            } else if (!oldArg.description) {
+              descriptionChanges.push(
+                `Description added on arg ${newType.name}.${newField.name}.${
+                  newArg.name
+                }.`,
+              );
+            } else if (oldArg.description !== newArg.description) {
+              descriptionChanges.push(
+                `Description changed on arg ${newType.name}.${newField.name}.${
+                  newArg.name
+                }.`,
+              );
+            }
+          }
+        });
+      });
+    } else if (newType instanceof GraphQLEnumType) {
+      const oldValues = oldType ? oldType.getValues() : null;
+      const newValues = newType.getValues();
+      newValues.forEach(newValue => {
+        const oldValue = oldValues
+          ? oldValues.find(value => value.name === newValue.name)
+          : null;
+
+        if (newValue.description) {
+          if (!oldValue) {
+            descriptionChanges.push(
+              `Description added on enum value ${newType.name}.${
+                newValue.name
+              }.`,
+            );
+          } else if (!oldValue.description) {
+            descriptionChanges.push(
+              `Description added on enum value ${newType.name}.${
+                newValue.name
+              }.`,
+            );
+          } else if (oldValue.description !== newValue.description) {
+            descriptionChanges.push(
+              `Description changed on enum value ${newType.name}.${
+                newValue.name
+              }.`,
+            );
+          }
+        }
+      });
+    }
+  });
+
+  return descriptionChanges;
+}
--- a/src/utilities/findDescriptionChanges.js
+++ b/src/utilities/findDescriptionChanges.js
@@ -9,6 +9,28 @@ import {
 } from '../type/definition';
 import type { GraphQLFieldMap } from '../type/definition';
 import { GraphQLSchema } from '../type/schema';
+import invariant from '../jsutils/invariant';
+
+export const DescribedObjectType = {
+  FIELD: 'FIELD',
+  TYPE: 'TYPE',
+  ARGUMENT: 'ARGUMENT',
+  ENUM_VALUE: 'ENUM_VALUE',
+};
+
+export const DescriptionChangeType = {
+  OBJECT_ADDED: 'OBJECT_ADDED',
+  DESCRIPTION_ADDED: 'DESCRIPTION_ADDED',
+  DESCRIPTION_CHANGED: 'DESCRIPTION_CHANGED',
+};
+
+export type DescriptionChange = {
+  object: $Keys<typeof DescribedObjectType>,
+  change: $Keys<typeof DescriptionChangeType>,
+  description: string,
+  oldThing: any,
+  newThing: any,
+};
 
 /**
  * Given two schemas, returns an Array containing descriptions of any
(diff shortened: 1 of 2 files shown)


src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,142 @@
...
+          `Description added on new type ${newType.name}.`,
+        );
+      } else if (!oldType.description) {
+        descriptionChanges.push(`Description added on type ${newType.name}.`);
+      } else if (oldType.description !== newType.description) {
+        descriptionChanges.push(`Description changed on type ${newType.name}.`);
+      }
+    }
+
+    if (oldType && !(newType instanceof oldType.constructor)) {
```
leebyron: Flow cannot see this as a cross-type refinement, plus it's a bit challenging to understand. I would instead make instanceof invariants for the `oldType` within each case below

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,142 @@
...
+      return;
+    }
+
+    if (
+      newType instanceof GraphQLObjectType ||
+      newType instanceof GraphQLInterfaceType ||
+      newType instanceof GraphQLInputObjectType
+    ) {
+      const oldTypeFields: ?GraphQLFieldMap<*, *> = oldType
```
leebyron: Should this differentiate between the old type not existing vs the old type existing but the field not existing?

It could be overwhelming to see a ton of "Description added on new field" for every field on a type that was newly added, that might not be helpful.

Instead perhaps "New type added with description" and "New field added with description" should be different from "Description added on field"

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,142 @@
...
+      const newTypeFields: GraphQLFieldMap<*, *> = newType.getFields();
+
+      Object.keys(newTypeFields).forEach(fieldName => {
+        const oldField = oldTypeFields ? oldTypeFields[fieldName] : null;
+        const newField = newTypeFields[fieldName];
+
+        if (newField.description) {
+          if (!oldField) {
+            descriptionChanges.push(
+              `Description added on new field ${newType.name}.${newField.name}`,
```
leebyron: This wording is a little confusing. How do you add to something that is new? Add implies additive to the existing state

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,142 @@
...
+import { GraphQLSchema } from '../type/schema';
+
+/**
+ * Given two schemas, returns an Array containing descriptions of any
+ * descriptions that are new or changed and need review.
+ */
+export function findDescriptionChanges(
+  oldSchema: GraphQLSchema,
+  newSchema: GraphQLSchema,
+): Array<string> {
```
leebyron: I would suggest mirroring the return value of the breaking change detection which offers what kind of detection was encountered. It looks like there are a handful of types here to differentiate: new types, new fields on existing types, added descriptions where there were none, descriptions changed, descriptions removed.

leebyron: Also, if these are complex objects, you could contain references to the old/new schema elements themselves, that would allow you to use the output of this function to build a review tool. Currently you just have english text which requires manually looking up the described types to see what changes actually occurred, which would be limited in usefulness.

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,142 @@
...
+            );
+          } else if (oldValue.description !== newValue.description) {
+            descriptionChanges.push(
+              `Description changed on enum value ${newType.name}.${
+                newValue.name
+              }.`,
+            );
+          }
+        }
+      });
```
leebyron: Much of the logic above looks repetitive, with the exception of the strings referencing each kind of type. Could you factor out the common logic?

src/utilities/findDescriptionChanges.js
eapache: I did it this way because spelling out the `instanceof`s actually becomes really messy in its own right to handle the cases where `oldType` is null/undefined, and where `oldType` is present but a different kind from `newType`. I'll play with it a bit to see if I can make it nicer, but any suggestions would be very welcome.

src/utilities/findDescriptionChanges.js
eapache: > It could be overwhelming to see a ton of "Description added on new field" for every field on a type that was newly added, that might not be helpful.

The goal was to provide our docs team with a comprehensive set of all the new strings that needed reviewing; I'll type these differently when I add stronger types per your earlier comment, and then the caller can filter as desired.

src/utilities/findDescriptionChanges.js
eapache: I'll take a look. Javascript isn't my forte so for a first pass I definitely stuck with a simpler more explicit style.

src/utilities/findDescriptionChanges.js
leebyron: In this case, below line 50, you could just mirror the if statement above in an invariant while including the falsey check:

```
invariant(
  !oldType ||
  oldType instanceof GraphQLObjectType ||
  oldType instanceof GraphQLInterfaceType ||
  oldType instanceof GraphQLInputObjectType,
  'Expected oldType to also have fields'
);
```

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,142 @@
...
+      const oldTypeFields: ?GraphQLFieldMap<*, *> = oldType
+        ? oldType.getFields()
+        : null;
+      const newTypeFields: GraphQLFieldMap<*, *> = newType.getFields();
+
+      Object.keys(newTypeFields).forEach(fieldName => {
+        const oldField = oldTypeFields ? oldTypeFields[fieldName] : null;
+        const newField = newTypeFields[fieldName];
+
+        if (newField.description) {
```
odykyi: ```if (newField && newField.description) {```

eapache:
Sorry for the delay on this; finally got flow passing and I think I addressed all the code review comments. Please let me know if there's anything else.

leebyron:
Thanks for the update and sorry for the review delay - it appears the test coverage is still below where we usually try to keep it. Here's a link from the coveralls report: https://coveralls.io/builds/14868261/source?filename=src%2Futilities%2FfindDescriptionChanges.js

Could you add more tests to ensure all aspects of the new functionality are well tested?

eapache:
Sorry for the delay again, I think we're both just busy :)

I've added some tests, so Coveralls now says overall coverage has increased. There's only one line (86) that's not covered at all and I'm not sure why - is `newField.args` expected to be truthy even when the field has no arguments? Should that check be adjusted, or removed? Otherwise (with the exception of "branch missed" warnings I don't know what to do with since it's not clear *which* branch was missed) everything should be covered.

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,163 @@
...
+      newType instanceof GraphQLObjectType ||
+      newType instanceof GraphQLInterfaceType ||
+      newType instanceof GraphQLInputObjectType
+    ) {
+      invariant(
+        !oldType ||
+          oldType instanceof GraphQLObjectType ||
+          oldType instanceof GraphQLInterfaceType ||
+          oldType instanceof GraphQLInputObjectType,
+        'Expected oldType to also have fields',
```
IvanGoncharov: It shouldn't be an error.
For example:
```graphql
scalar Date
```
```graphql
type Date {
  day: Int
  month: Int
  year: Int
}
```

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,163 @@
...
+              DescribedObjectType.ARGUMENT,
+            ),
+          );
+        });
+      });
+    } else if (newType instanceof GraphQLEnumType) {
+      invariant(
+        !oldType || oldType instanceof GraphQLEnumType,
+        'Expected oldType to also have values',
+      );
```
IvanGoncharov: Same here:
```graphql
scalar Month
```

```graphql
enum Month {
  Jan
  Feb
  # ...
}
```

src/utilities/findDescriptionChanges.js
eapache: This was added to satisfy Flow (https://github.com/graphql/graphql-js/pull/1127#discussion_r155619374).

This (and the enum case) are breaking schema changes, so an error seemed appropriate? Would you prefer they were silently ignored? If so, how do I do that in a way that Flow will recognize?

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,163 @@
...
+} from '../type/definition';
+import type { GraphQLFieldMap } from '../type/definition';
+import { GraphQLSchema } from '../type/schema';
+import invariant from '../jsutils/invariant';
+
+export const DescribedObjectType = {
+  FIELD: 'FIELD',
+  TYPE: 'TYPE',
+  ARGUMENT: 'ARGUMENT',
+  ENUM_VALUE: 'ENUM_VALUE',
```
IvanGoncharov: Your missing directives + you need to check arguments of directives in `findDescriptionChanges`

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,163 @@
...
+};
+
+export const DescriptionChangeType = {
+  OBJECT_ADDED: 'OBJECT_ADDED',
```
IvanGoncharov: As I understand it trigger when any new type is added not only `GraphQLObjectType`. So I think it should have a different name.

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,163 @@
+/* eslint-disable no-restricted-syntax */
+// @flow
```
IvanGoncharov: Please use the same header license + `@flow strict` as other files

src/utilities/findDescriptionChanges.js
```diff
@@ -0,0 +1,163 @@
+/* eslint-disable no-restricted-syntax */
```
IvanGoncharov: Please remove and fix all eslint errors if any.

IvanGoncharov:
@eapache Sorry for not reviewing it earlier.

My 2¢: Different developers and teams have different needs but we can't add `find*Changes` function for every case. So I think we need more generic solutions for finding difference beetween two schema.

From #981:
> aka anything that might need to be reviewed by a docs team

In that case changes in `deprecationReason` is also relevant and should be reviewed so we need to add `findDeprecationReasonChanges`.

Moreover some documentation teams would be interested in being notified on new types without any description.

I also see use case for detecting any non-documentation (not in `description` or `deprecationReason`) change.

So I think instead of adding new `find*Changes` function for every usecase we should provide single `diffSchemas` function. That returns something like this:
```flow
type SchemaDifference = {|
  added: {|
    queryType?: GraphQLObjectType,
    // ...
    types: Array<GraphQLTypes>,
    directives: Array<GraphQLDirectives>,
  |},
  changed: {|
    queryType?: ValueDifference<GraphQLType>,
    // ...
    types: Array<TypeDifference>,
    directives: Array<DirectivesDifference>,
  |},
  // ...
|};

type ValueDifference<T> = { old: T, new: T };

type TypeDifference = ScalarTypeDifference | ObjectTypeDifference /* ... */

type ObjectTypeDifference = {|
  name: string,
  added: {|
    description: string,
    fields: Array<GraphQLField>,
    // ...
  |},
  changed: {|
    description: ValueDifference<string>,
    fields: Array<FieldDifference>,
    // ...
  |},
  // ...
|};
```

It will allow to implement your particular check as:
```js
const diff = diffSchema(oldSchema, newSchema);

for (const newType of diff.added.types) {
  if (newType.description) {
    console.log('Description of a new type ...');
  }
  newType.getFields().forEach(reportNewField);
}

for (const typeDiff of diff.changed.types) {
  if (typeDiff.added.description) {
    console.log('Added type description ...');
  }
  typeDiff.added.fields.forEach(reportNewField);
  if (typeDiff.changed.description) {
    console.log('Changed type description ...');
  }
  for (const fieldDiff of typeDiff.changed.field) {
    if (fieldDiff.added.description) {
      console.log('Added field description ...');
      // ...
    }
  }
}

function reportNewField(field) {
  if (field.description) {
    // ...
  }
}
```

I think it would be more generic approach and allow other developers to easily implement their project specific needs.

@leebyron What do you think?

src/utilities/findDescriptionChanges.js
IvanGoncharov: So if I do breaking change in my schema I don't need to notify my documentation team about other non-breaking changes in my schema.
BTW Breaking changes is normal in many cases for example if you have your API in public demo.

eapache: Yes, you're right. I was thinking too much about our specific case (where breaking changes already ping a substantial set of people for review and the docs changes are less relevant) but in the general case this should absolutely work.

I'm still not sure how to satisfy Flow without an invariant, but that's purely me not being much of a javascript developer.

eapache:
> Different developers and teams have different needs but we can't add find*Changes function for every case. So I think we need more generic solutions for finding the difference between two schemas.

You're not wrong, however a full `diffSchema` is well outside of my available time here (and probably outside of my skillset to implement correctly in js). If you'd prefer to move in that direction instead I'll wait until somebody else tackles it and we can keep just using our copy of this method internally for the time being.

The only thing I can think might still belong outside of `diffSchema` is detecting breaking changes, since to a certain extent that check is not just descriptive but also prescriptive of what changes count as breaking. Though if that is already well-specified elsewhere we could just have a `breaking: true/false` flag on the various different types returned by `diffSchema`?

> Moreover, some documentation teams would be interested in being notified of new types without description.

Also a good point. We just have CI fail in this case, but I imagine that isn't universal.

IvanGoncharov:
Closing since it's outdated but extracted `diffSchema` idea into #3224

PR #1134, opened 2017-12-11
Use validation helper to check if GraphQLSchema instance

dwwoelfel:
Checks if schema is valid with the `assertValidSchema` helper instead of a simple instanceof check. This gives the developer actionable error messages when there are multiple versions of graphql in node_modules.

### DIFF
--- a/src/utilities/extendSchema.js
+++ b/src/utilities/extendSchema.js
@@ -12,6 +12,7 @@ import keyMap from '../jsutils/keyMap';
 import { ASTDefinitionBuilder } from './buildASTSchema';
 import { GraphQLError } from '../error/GraphQLError';
 import { GraphQLSchema } from '../type/schema';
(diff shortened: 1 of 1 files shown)


src/utilities/extendSchema.js
```diff
@@ -72,10 +73,7 @@ export function extendSchema(
   documentAST: DocumentNode,
   options?: Options,
 ): GraphQLSchema {
-  invariant(
-    schema instanceof GraphQLSchema,
-    'Must provide valid GraphQLSchema',
-  );
+  assertValidSchema(schema);
```
leebyron: Oh good call to check this here - but I'm not sure we want to assert the whole schema is valid before extending it, we want to preserve the use case where you can extend an incomplete schema to get to a complete one.

I have an idea to extract just this check out, one sec while I test it out

PR #1144, opened 2017-12-14
Validate literals in a single rule with finer precision

### LINKED ISSUE #821: 0.8.x breaking change with errors thrown from scalar resolvers

leebyron:
This generalizes the "arguments of correct type" and "default values of correct type" to a single rule "values of correct type" which has been re-written to rely on a traversal rather than the utility function `isValidLiteralValue`. To reduce breaking scope, this does not remove that utility even though it's no longer used directly within the library. Since the default values rule included another validation rule that rule was renamed to a more apt "variable default value allowed".

This also includes the original errors from custom scalars in the validation error output, solving the remainder of #821.

Closes #821


leebyron:
I plan to open a PR making non-normative edits to the spec which better align to this refactoring as well

PR #1146, opened 2017-12-14
Find Breaking Changes: removed directives

kenleezle:
Introduced a function to find removed directives, along with a simple unit test.


src/utilities/findBreakingChanges.js
```diff
@@ -676,3 +677,21 @@ export function findInterfacesAddedToObjectTypes(
...
 }
+
+export function findRemovedDirectives(
+  oldSchema: GraphQLSchema,
+  newSchema: GraphQLSchema,
+): Array<BreakingChange> {
+  const removedDirectives = [];
+
+  oldSchema.getDirectives().forEach(directive => {
+    if (!newSchema.getDirective(directive.name)) {
```
mjmahone: This I believe makes this O(n^2): https://github.com/graphql/graphql-js/blob/master/src/type/schema.js#L206

A more efficient version, O(nlogn):

```
const newSchemaDirectiveNames = new Set(oldSchema.getDirectives().map(directive => directive.name));
oldSchema.getDirectives().forEach(directive => {
  if (!newSchemaDirectiveNames.has(directive.name) {
    ...
  }
}
```

kenleezle: 👍  thanks @mjmahone !

kenleezle: update

sam-swarr:
Nice, this looks great! Thanks @kenleezle.

Can you go ahead and update this PR to also include:
  - if a directive argument is removed
  - if a non-nullable directive argument is added
  - if a directive location is removed

kenleezle:
@leebyron CI is not running on my commits, can you ptal?

src/type/directives.js
```diff
@@ -74,6 +75,12 @@ export class GraphQLDirective {
       });
     }
   }
+
+  getArgumentMap(): ObjMap<GraphQLArgument> {
```
sam-swarr: Not sure we want to add this to the `Directive` class. It should be fine for `findBreakingChanges` to just use the `this.args`, and just use `Array.find` when checking for the existence of an arg. I'm not too worried about performance given that there aren't usually many directives to begin with.

kenleezle: that was my initial instinct. @mjmahone is that ok with you since you originally raised the concern?

lol @sam-swarr you are going to love what i did with `Schema::getDirectiveMap()` har har

kenleezle:
@sam-swarr @mjmahone I think I addressed your comments to date. PTAL. Thanks!

kenleezle:
closed this and replaced it with #1152 in an attempt to debug the CI not running problem.

PR #1148, opened 2017-12-14
Remove eslint comments unnecessary after 40f73fd

IvanGoncharov:


mjmahone:
Thank you for catching all these!

PR #1149, opened 2017-12-15
Fix 'path' argument. Add more visitor tests

IvanGoncharov:
As discussed in #1145 I started to work on tests for `visit` function.

### DIFF
--- a/src/language/__tests__/visitor-test.js
+++ b/src/language/__tests__/visitor-test.js
@@ -15,8 +15,86 @@ import { join } from 'path';
 import { TypeInfo } from '../../utilities/TypeInfo';
 import { testSchema } from '../../validation/__tests__/harness';
 import { getNamedType, isCompositeType } from '../../type';
+import * as Kind from '../kinds';
+
+function getNodeByPath(ast, path) {
+  let result = ast;
+  for (const key of path) {
+    expect(result).to.have.property(key);
+    result = result[key];
+  }
+  return result;
+}
+
+function checkVisitorFnArgs(ast, args, isEdited) {
+  const [node, key, parent, path, ancestors] = args;
+
+  expect(node).to.be.an.instanceof(Object);
+  expect(node.kind).to.be.oneOf(Object.values(Kind));
+
+  const isRoot = key === undefined;
+  if (isRoot) {
+    if (!isEdited) {
+      expect(node).to.equal(ast);
+    }
+    expect(parent).to.equal(undefined);
+    expect(path).to.deep.equal([]);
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@leebyron I added the simple test for `path` parameter and discover that `path` value differs between `enter` and `leave` callbacks. On `leave` path is always one element shorter, for example if you was called on enter to `OperationDefinition` node `path` value would be `['definitions', 0]` but on leave you will get `['definitions']`.

Is it intentional behaviour or a bug?

src/language/__tests__/visitor-test.js
```diff
@@ -17,6 +17,35 @@ import { testSchema } from '../../validation/__tests__/harness';
...
+      enter(node, key, parent, path) {
+        visited.push(['enter', path.slice()]);
+      },
+
+      leave(node, key, parent, path) {
+        visited.push(['leave', path.slice()]);
+      },
+    });
+
+    console.log(JSON.stringify(visited, null, 2));
```
mjmahone: Probably don't want to keep this console.log in the test

src/language/__tests__/visitor-test.js
```diff
@@ -17,6 +17,35 @@ import { testSchema } from '../../validation/__tests__/harness';
 import { getNamedType, isCompositeType } from '../../type';
 
 describe('Visitor', () => {
+  it('check path argument', () => {
```
mjmahone: nit: 'validates path argument' or 'checks path argument'

IvanGoncharov: @mjmahone Thanks 👍 Fixed now.

src/language/__tests__/visitor-test.js
IvanGoncharov: @mjmahone Consequences of committing at 3 AM 😄 
Removed.

IvanGoncharov:
I fixed this with `path` by a very small fix that's easy to verify.

As for testing visitor callback arguments, I decided it's too much to add it hardcoded snapshots inside tests. So I added `checkVisitorFnArgs` utility function that checks types and interdependencies of all callback arguments and calls it from every callback inside `visitor-tests.js`.

One thing to note is the current implementation of `visit` doesn't maintain consistency of arguments after AST is edited. For example, if you edit some node on `enter` you would get updated node on `leave` but `parent` and all ancestors will remain not updated. To workaround this problem I added `isEdited` argument that disables some of the checks inside `checkVisitorFnArgs` function.

@mjmahone @leebyron Can you please review this PR?

leebyron:
Nice work, @IvanGoncharov

PR #1150, opened 2017-12-15
Make `npm run watch` linting work right

mohawk2:
Without this update, using `npm run watch` always gives this error:

```
  execution/execute.js ...
/home/osboxes/graphql-js/src/execution/execute.js
  1:1  error  Definition for rule 'no-async' was not found  no-async

✖ 1 problem (1 error, 0 warnings)

  ✘ execution/execute.js
```


mjmahone:
Thanks for this fix!

PR #1152, opened 2017-12-15
Find breaking directive changes

kenleezle:
The GraphQL schema definition is continually evolving as server engineers add, modify, and delete GraphQL types and fields. These changes to the schema can be dangerous, particularly breaking changes where the new version of the schema is not backwards compatible with the old schema.

For example, consider a breaking change where a www server diff removes the type `Foo` from the schema. This change could potentially break the client in fbsource if there is any client .graphql files that refer to the `Foo` type.

In order to prevent these types of issue, we run tests on diffs that affect the schema which compare the old schema to the new schema and detect any breaking changes. The logic that detects the breaking changes lives in our open source graphql-js library: https://github.com/graphql/graphql-js/blob/master/src/utilities/findBreakingChanges.js.

This task:

The findBreakingChanges utility detects most types of breaking changes (see https://github.com/graphql/graphql-js/blob/master/src/utilities/findBreakingChanges.js#L31-L41 for a list of which kinds are currently detected). However, the utility currently does not detect breaking changes related to Directives**.

This task is to add support in findBreakingChanges to detect:
  - if a directive is removed from the schema
  - if a directive argument is removed
  - if a non-nullable directive argument is added
  - if a directive location is removed

**For more info on GraphQL Directives, you can read http://graphql.org/learn/queries/#directives as well as the official GraphQL spec: http://facebook.github.io/graphql/October2016/#sec-Type-System.Directives.
Basically, directives are annotations that can be supplied in .graphql files that can alter runtime behavior. Directives has a list of supported "locations", where in a graphql file they can appear. (Some annotations are used on fields, some on queries, etc).

-- @sam-swarr


kenleezle:
@sam-swarr @mjmahone CI passed

leebyron:
Nice work, @kenleezle!

PR #1157, opened 2017-12-17
Works without experimental flow

mjmahone:
For packages that do not use `experimental.const_params` in their `.flowconfig`, using v0.12.0 will cause them to have flow errors, as they treat arguments to functions as non-const. This fixes up all of the known cases.

To test, comment-out `experimental.const_params=true` from `.flowconfig`, and run flow.

### DIFF
--- a/src/type/introspection.js
+++ b/src/type/introspection.js
@@ -22,7 +22,6 @@ import {
   isListType,
   isNonNullType,
   isAbstractType,
-  isNamedType,
 } from './definition';
 import { GraphQLList, GraphQLNonNull } from '../type/wrappers';
 import { GraphQLString, GraphQLBoolean } from './scalars';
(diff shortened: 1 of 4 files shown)


src/type/introspection.js
```diff
@@ -467,9 +466,23 @@ export const introspectionTypes: $ReadOnlyArray<*> = [
...
   return (
-    isNamedType(type) &&
-    introspectionTypes.some(
-      introspectionType => introspectionType.name === type.name,
-    )
+    // Cannot use function calls to %checks unless flow's
+    // experimental.const_params=true is set.
+
+    // isNamedType(type)
+    type !== null &&
```
mjmahone: `!!type &&` prettiers to `Boolean(type) &&` which is a function call which makes flow fail without `experimental.const_params=true`. And `type &&` fails because it's not returning a boolean.

leebyron: We can probably Boolean wrap the whole expression then. But also it would be great to avoid the additional `instanceof` checks - we've managed to corral them into their individual predicate functions so far

leebyron:
I just edited this slightly in #1160 to also remove the flow option so this won't regress, I think it can also avoid needing to inline the other checks

src/type/introspection.js
billydye1: Fvchbvgg

PR #1162, opened 2017-12-17
Add special case for legacy servers.

leebyron:
Much older versions of GraphQL use `__configs__` as part of a pseudo-introspection meta programming layer. While this is technically against the spec, this adds temporary support.


leebyron:
Agreed. Removing this case would not constitute a breaking change - I think Facebook's server is the only one with this legacy use since it dates back to ~2013, a couple years before spec-writing and open-sourcing. Fun times.

PR #1164, opened 2017-12-17
DRY with MaybePromise

mohawk2:
`MaybePromise<T>` type, to avoid constantly declaring `Promise<some stuff> | some stuff`. Most visible in `execute.js` but also in a few other places.


leebyron:
I'm into it.

PR #1165, opened 2017-12-18
Bring back `getDescription` export

stubailo:
We were using this in graphql-tools with graphql@0.11, would be nice to get it exported again!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -315,6 +315,8 @@ export {
   buildASTSchema,
   // Build a GraphQLSchema from a GraphQL schema language document.
   buildSchema,
+  // Get the description from a schema AST node.
+  getDescription,
   // Extends an existing GraphQLSchema from a parsed GraphQL Schema
   // language AST.
   extendSchema,
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -35,6 +35,8 @@ import type {
   EnumValueDefinitionNode,
   InputObjectTypeDefinitionNode,
   DirectiveDefinitionNode,
+  StringValueNode,
+  Location,
 } from '../language/ast';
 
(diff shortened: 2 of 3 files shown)


stubailo:
OK looks like there are some typing things to work through, I will come back to it tomorrow.

leebyron:
You should also have this exported from index.js if you expect to use it in a dependent library, otherwise it might be too easy to regress again! We assume exports used between files are used purely internally.

It might be helpful to audit for other deep require uses as well to avoid future breakage

stubailo:
@leebyron happy to make that modification for sure, we definitely have 3-4 others we are using. Didn't realize that was the intention with the internal exports!

Any idea what the argument type should be here? I would have thought `ASTNode` would have been the right thing, but I guess not all nodes have descriptions. Maybe I should just write some custom duck type thing.

leebyron:
Try out `{ +description?: StringValueNode }`?

stubailo:
OK looks like it's all good now, thanks for the suggestions!

leebyron:
Nice. I'll likely have a point release out today or tomorrow that will include this.

Also, apparently this is your first commit to GraphQL.js. Congrats! 🎉

stubailo:
> Also, apparently this is your first commit to GraphQL.js. Congrats! 

That is SO SURPRISING. I'll look into contributing more! I guess I spent more time on GraphQL.org that I misattributed.

nodkz:
I'm also using `getDescription` method in graphql-compose.
Glad that it already fixed. Thanks, Sashko! 👍

Lee, all other "breaking" changes did not hit my tests and all works like a charm. 👌
With Flow just change array on $ReadOnlyArray in several places. Typings became much stronger. Thanks for this!!! 💪

negezor:
@leebyron hi, when can you update package?

PR #1166, opened 2017-12-18
Add test case for failing parsing

freiksenet:
Extend without fields followed by extend makes second extend to be parsed as types to implement. Note that locations are faulty in test (cause it's hard to deduce them).

### DIFF
--- a/src/language/__tests__/schema-parser-test.js
+++ b/src/language/__tests__/schema-parser-test.js
@@ -203,6 +203,38 @@ extend type Hello {
     expect(printJson(doc)).to.equal(printJson(expected));
   });
 
+  it('Extension without fields followed by extension', () => {
+    const body = `
+      extend type Hello implements Greeting
+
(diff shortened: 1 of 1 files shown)


leebyron:
Great catch! Thanks for reporting this.

We may need to rethink field-less type extensions to solve for this ambiguity :/

freiksenet:
I think allowing empty fields' blocks at parse time could work.

leebyron:
Yeah, I think you're right. Requiring a fields block, but allowing the fields block to be empty might have to be the answer here, even though it's a bit inconsistent with other grammar rules in GraphQL :/

I'll do a bit of brainstorming with a few other folks and we'll resolve this quickly

leebyron:
After brainstorming came up with what I think is a more consistent solution, though it does constitute more of a breaking change from existing usage. See #1169 for a path forward along with suggested migration tools

PR #1169, opened 2017-12-20
RFC: SDL - Separate multiple inherited interfaces with `&`

leebyron:
This replaces:

```graphql
type Foo implements Bar, Baz { field: Type }
```

With:

```graphql
type Foo implements Bar & Baz { field: Type }
```

With no changes to the common case of implementing a single interface.

This is more consistent with other trailing lists of values which either have an explicit separator (union members) or are prefixed with a sigil (directives). This avoids parse ambiguity in the case of an omitted field set, illustrated by #1166

This is a breaking change for existing uses of multiple inheritance. To allow for an adaptive migration, this adds a parse option to continue to support the existing experimental SDL: `parse(source, {allowLegacySDLImplementsInterfaces: true})`


leebyron:
Optimistically applied this change to the pending spec proposal (https://github.com/facebook/graphql/pull/90) in https://github.com/facebook/graphql/pull/90/commits/05fc2a7fa6ce32b210a352300c29afa28666e455

leebyron:
@freiksenet, @stubailo, @dschafer - would love your opinion on this change. I think it makes sense in the context of the whole, solves the ambiguity Mikhail pointed out, and offers some tools for migration. I'd like a second opinion to make sure I'm not crazy.

If there's some agreement, I'd release this as v0.13 to allow some space for the subsequent SDL change.

dschafer:
Conceptually feels super reasonable to have a symbol used here. The `&` symbol choice didn't resonate with me at first, but it matches very closely with the concept at https://flow.org/en/docs/types/intersections/#toc-intersections-of-object-types so it seems like a reasonable choice to me.

I have some reservations about how many symbols we're introducing for the SDL (`&` and `"""` come to mind), but they all seem to be used in a relatively constrained manner (so we're not losing that much option value), so it doesn't bother me that much.

One other option that I just thought of is that instead of:

```graphql
type Foo implements Bar & Baz { field: Type }
```

we could do

```graphql
type Foo implements Bar implements Baz { field: Type }
```

that is to say, `implements` **is** the symbol. Not sure I like that better, but it is another option.

dschafer:
> To allow for an adaptive migration, this adds a parse option to continue to support the existing experimental SDL: parse(source, {legacySDL: true})

I'd prefer to see this gated under `sdlWhitespaceForMultipleInterfaceInheritence` or the like, since I doubt this is the last time we'll make a change to the SDL.

freiksenet:
This looks great! Agree with @dschafer on specific SDL flags instead of generic ones.

dschafer:
Ah, I may have rederived why you went with a general one - if this is a single flag to refer to the "pre-spec" SDL (e.g. this would also include comments as documentation, for example), then I think I'm fine with it being general.

leebyron:
Specific flags makes sense in the context of the existing specific flag for comment descriptions. I'm hoping to phase them out over major releases to ease the path to adoption

I'm hoping this is the last real change to the SDL before codifying it (though of course, we could expand on it in the future in new proposals)

stubailo:
This looks great, glad it eliminates the ambiguity!

PR #1171, opened 2017-12-20
Parse option: allowLegacySDLEmptyFields

leebyron:
This offers a slightly smoother adoption path for existing code that uses the legacy empty field sets.


vjeux:
Is there a way to do a new release with this option?

PR #1174, opened 2017-12-20
Only trigger "one instance of graphql" error in DEV environments.

leebyron:
This ensures the additional checking for multiple graphql instances only occurs in non-production builds. Not only does this improve the performance for production builds, it also solves a spurious error case for minified builds:

Since this additional check compares the `.name` of the constructors as a proxy for determining if two classes are representationally equal, minified builds may result in spurious false-positives if class names are minified to single-character names and those characters are likely to conflict.

This doesn't require `process` to exist (defaulting to the non-production mode), but bundlers like Webpack should consider using a stripping transform (see: https://webpack.js.org/guides/production/#specify-the-environment) in production builds if they also use a minifier.

### DIFF
--- a/src/jsutils/instanceOf.js
+++ b/src/jsutils/instanceOf.js
@@ -16,16 +16,18 @@ declare function instanceOf(
   constructor: mixed,
 ): boolean %checks(value instanceof constructor);
 
-// eslint-disable-next-line no-redeclare
-export default function instanceOf(value, constructor) {
-  if (value instanceof constructor) {
-    return true;
-  }
-  if (value) {
-    const valueConstructor = value.constructor;
-    if (valueConstructor && valueConstructor.name === constructor.name) {
-      throw new Error(
-        `Cannot use ${constructor.name} "${value}" from another module or realm.
+export default (process && process.env.NODE_ENV !== 'production'
+  ? // eslint-disable-next-line no-shadow
+    function instanceOf(value: any, constructor: any) {
+      if (value instanceof constructor) {
+        return true;
+      }
+      if (value) {
+        const valueClass = value.constructor;
+        const className = constructor.name;
+        if (valueClass && valueClass.name === className) {
+          throw new Error(
+            `Cannot use ${className} "${value}" from another module or realm.
 
 Ensure that there is only one instance of "graphql" in the node_modules
 directory. If different versions of "graphql" are the dependencies of other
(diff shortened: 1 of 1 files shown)


cyrilgandon:
That's a big blocker for using new features of graphql-js, any idea when the next release will be? Thank you! 👍

pleunv:
Would love to see this released too, for us this causes issues in prod that for some reason work in dev.

danez:
Also waiting for this.

nparse:
Struggling with this issue too.

flip-it:
When can we expect this to be released?

sheerun:
Same issue, please release :)

mabellLU:
Same issue

Dermah:
FYI for those that desperately need this, we've had success rolling back to `graphql@0.11` while waiting for the above fix to be released.

mabellLU:
Really just need the update GraphQLErrors object with extensions

stantoncbradley:
+1 need GraphQLErrors as well!  Thanks!!

nparse:
If you guys having issues with GraphiQL, you can override qraphql dependency in package.json using the most recent version, like so
```
...
   "dependencies": {
     "graphiql": "^0.11.10",
+    "graphql": "git://github.com/graphql/graphql-js.git#npm",
...
```
It solved the problem for me.

BerndWessels:
@nparse is this making it into the next release?

Another but maybe related problem is that `uglify` with `mangle=true` results in this error:

```
module initialization error: Error

Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate "graphql" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.
at n.a (/var/task/index.js:1:156141)
at f (/var/task/index.js:1:3835)
at d (/var/task/index.js:1:3756)
at _ (/var/task/index.js:1:4392)
at n.q (/var/task/index.js:1:2738)
at new o (/var/task/index.js:1:25040)
at o (/var/task/index.js:1:25008)
at Object.n.a.r (/var/task/index.js:1:36681)
at t (/var/task/index.js:1:447)
at Object.<anonymous> (/var/task/index.js:1:83640)
```

Cabezaa:
Same problem that @BerndWessels  here... Using MeteorJS (using `uglify` )

I try with graphql 0.12.3 and 0.12.0

pleunv:
This is fixed in 0.13/0.13.1

alistairholt:
I'm using 15.5.0 in a React Native project (Expo) and this is causing the built binaries to crash with the `Ensure that there is only one instance of "graphql"…` error. I'm assuming this is because of `NODE_ENV` environment variable check, which isn't set. I know there is only one one instance of `graphql` in node_modules. Frustrating.

PR #1175, opened 2017-12-20
Printer: create special function to add description

IvanGoncharov:
Makes code more readable, at least for me 😄


leebyron:
Works for me

PR #1178, opened 2017-12-24
DRY with lists tests

mohawk2:
With these changes, the "Execute: Handles list nullability" tests read almost like a DSL. I claim this makes the tests more understandable, and therefore more maintainable.


leebyron:
Also note that test coverage dropped as a result of this PR which is surprising for a test refactor.

mohawk2:
@leebyron I'm a bit surprised about the test coverage point. I assume it's because it's measuring the total code volume including tests, and if the tests get smaller, the amount of unexercised code will get proportionally larger?

I have added comments hoping this renders the changes more legible. What do you think?

PR #1179, opened 2017-12-25
Update printSchema for recent SDL change to `implements`.

mohawk2:
Added test to prevent future regressions. Updated existing also.

### DIFF
--- a/src/utilities/__tests__/schemaPrinter-test.js
+++ b/src/utilities/__tests__/schemaPrinter-test.js
@@ -9,6 +9,7 @@ import { describe, it } from 'mocha';
 import { expect } from 'chai';
 import dedent from '../../jsutils/dedent';
 import { printSchema, printIntrospectionSchema } from '../schemaPrinter';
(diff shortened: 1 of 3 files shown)


src/utilities/__tests__/schemaPrinter-test.js
```diff
@@ -27,7 +28,9 @@ import { GraphQLDirective } from '../../type/directives';
 import { DirectiveLocation } from '../../language/directiveLocation';
 
 function printForTest(schema) {
-  return printSchema(schema);
+  const schemaText = printSchema(schema);
+  buildSchema(schemaText); // keep printSchema and buildSchema in sync
+  return schemaText;
```
IvanGoncharov: @mohawk2 It would be great to also test output of `buildSchema`:
```flow
expect(printSchema(buildSchema(schemaText)).to.equal(schemaText)
```

mohawk2:
@IvanGoncharov Agreed, changed!

leebyron:
Great catch! Thanks for the added test coverage

PR #1180, opened 2017-12-26
Simplify 'BooleanValue' print

IvanGoncharov:


leebyron:
Much clearer, thanks!

PR #1181, opened 2017-12-26
Fix 'astFromValue' to correctly handle integers and strings

IvanGoncharov:
Currently `astFromValue` converts `-1` to float, `0000` to integer and also escapes strings when it shouldn't.

### DIFF
--- a/src/utilities/astFromValue.js
+++ b/src/utilities/astFromValue.js
@@ -11,17 +11,7 @@ import { forEach, isCollection } from 'iterall';
 
 import isNullish from '../jsutils/isNullish';
 import isInvalid from '../jsutils/isInvalid';
-import type {
-  ValueNode,
-  IntValueNode,
-  FloatValueNode,
-  StringValueNode,
-  BooleanValueNode,
-  NullValueNode,
-  EnumValueNode,
-  ListValueNode,
-  ObjectValueNode,
-} from '../language/ast';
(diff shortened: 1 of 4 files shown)


src/utilities/astFromValue.js
```diff
@@ -11,17 +11,7 @@ import { forEach, isCollection } from 'iterall';
...
-  IntValueNode,
-  FloatValueNode,
-  StringValueNode,
-  BooleanValueNode,
-  NullValueNode,
-  EnumValueNode,
-  ListValueNode,
-  ObjectValueNode,
-} from '../language/ast';
+import type { ValueNode } from '../language/ast';
```
leebyron: Thanks for cleaning up these flow types!

src/utilities/astFromValue.js
```diff
@@ -157,3 +145,10 @@ export function astFromValue(value: mixed, type: GraphQLInputType): ?ValueNode {
   /* istanbul ignore next */
   throw new Error(`Unknown type: ${(type: empty)}.`);
 }
+
+/**
+ * IntValue:
+ *   - NegativeSign? 0
+ *   - NegativeSign? NonZeroDigit ( Digit+ )?
+ */
+const integerStringRegExp = /^-?(0|[1-9][0-9]*)$/;
```
leebyron: Much better. Thanks for including the comment with the lexeme rule

PR #1183, opened 2017-12-27
DRY with NonNull of Object tests

mohawk2:
Similar to #1178 - zaps all the repeated code and data, rendering it more comprehensible (at least to me).


leebyron:
lgtm

Issue #941, opened 2017-07-06
Lexographically sorted printSchema for comparison/diff stability?


benjie:
`printSchema` already sorts the type names alphabetically:

https://github.com/graphql/graphql-js/blob/ff4338daadb7f86e0d0700faff086dd46cbfb7b2/src/utilities/schemaPrinter.js#L74

It would also be useful for diffing if fields, arguments, enum values and various other entities were also sorted alphabetically. I have performed some local modifications and it seems the changes required are quite minor. However, I believe it's possibly the case that a schema sorted in one way is not identical to a schema sorted in another because the order of fields/arguments may affect the order in which they are resolved?

I therefore have the following questions:

1. Would ordering the printSchema alphabetically by default be an acceptable change?
2. If not, would it be acceptable to add a second argument added to printSchema that supports an enumeration of sort modes (e.g. 'natural' for the current ordering, 'alphabetic', and even something more exotic like ordering by heirarchy as in #281)?
3. Would you accept a pull request to add these changes?

---

_Keywords (to help others find this issue, also detailing what I searched for before): alphabetic, lexographic, lexical, order, sort, printSchema, print, stable_

benjie:
It seems this has already been attempted in #889 and since reverted in #920 because the relation `s = parse(print(s))` didn't hold.

benjie:
For anyone who needs this functionality for whatever reason; here's an ugly hack to tide you over, currently it sorts fields, args and enums via duck-typing:

```js
const { parse, buildASTSchema } = require("graphql");
const { printSchema } = require("graphql/utilities");

const printSchemaOrdered = originalSchema => {
  const schema = buildASTSchema(parse(printSchema(originalSchema)));
  const typeMap = schema.getTypeMap();
  Object.keys(typeMap).forEach(name => {
    const Type = typeMap[name];
    if (Type.getFields) {
      const fields = Type.getFields();
      const keys = Object.keys(fields).sort();
      keys.forEach(key => {
        const value = fields[key];

        // Move the key to the end of the object
        delete fields[key];
        fields[key] = value;

        // Sort args
        if (value.args) {
          value.args.sort((a, b) => a.name.localeCompare(b.name));
        }
      });
    }
    if (Type.getValues) {
      Type.getValues().sort((a, b) => a.name.localeCompare(b.name));
    }
  });

  return printSchema(schema);
};
```

leebyron:
[...]

PR #1185, opened 2017-12-28
[WIP] Add 'sortIntrospectionQuery' utility function

IvanGoncharov:
Based on discussion in #941 and #889
Some of GraphQL servers doesn't preserve the order of types/fields and we need to sort them to provide consistent results. I think it is the primary use case for schema sorting so it makes sense to sort the result of introspection query and it makes implementation super simple. 

As for other use-cases (e.g. tests) you can query introspection with `graphqlSync` then sort it and build sorted schema using `buildClientSchema`.

### DIFF
--- a/src/utilities/__tests__/sortIntrospection-test.js
+++ b/src/utilities/__tests__/sortIntrospection-test.js
@@ -1,0 +1,321 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+import dedent from '../../jsutils/dedent';
+import { graphqlSync } from '../../graphql';
+import { printSchema } from '../schemaPrinter';
+import { buildSchema } from '../buildASTSchema';
+import { buildClientSchema } from '../buildClientSchema';
+import { getIntrospectionQuery } from '../introspectionQuery';
+import { sortIntrospectionQuery } from '../sortIntrospection';
+
+function sortSDL(sdl) {
+  const schema = buildSchema(sdl);
+  const result = graphqlSync(schema, getIntrospectionQuery());
+  expect(result.errors).to.equal(undefined);
+  const introspection = sortIntrospectionQuery(result.data);
+  const sortedSchema = buildClientSchema(introspection);
+  return printSchema(sortedSchema);
+}
+
+describe('sortIntrospectionSchema', () => {
+  it('sort fields', () => {
+    const sorted = sortSDL(dedent`
+      input Bar {
+        barB: String
+        barA: String
+        barC: String
+      }
+
+      interface FooInterface {
+        fooB: String
+        fooA: String
+        fooC: String
+      }
+
+      type FooType implements FooInterface {
+        fooC: String
+        fooA: String
+        fooB: String
+      }
+
+      type Query {
+        dummy(arg: Bar): FooType
+      }
+    `);
+
+    expect(sorted).to.equal(dedent`
+      input Bar {
+        barA: String
+        barB: String
+        barC: String
+      }
+
+      interface FooInterface {
+        fooA: String
+        fooB: String
+        fooC: String
+      }
+
+      type FooType implements FooInterface {
+        fooA: String
+        fooB: String
+        fooC: String
+      }
+
+      type Query {
+        dummy(arg: Bar): FooType
+      }
+    `);
+  });
+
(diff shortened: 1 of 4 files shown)


src/utilities/__tests__/sortIntrospection-test.js
```diff
@@ -0,0 +1,321 @@
...
+
+      interface FooB {
+        dummy: String
+      }
+
+      interface FooC {
+        dummy: String
+      }
+
+      type Query implements FooA, FooB, FooC {
```
IvanGoncharov: Should be `type Query implements FooA & FooB & FooC` fixed #1179

src/utilities/sortIntrospection.js
```diff
@@ -0,0 +1,123 @@
...
+  };
+}
+
+function sortDirective(
+  directive: IntrospectionDirective,
+): IntrospectionDirective {
+  return {
+    name: directive.name,
+    description: directive.description,
+    locations: sortBy(directive.locations, x => x),
```
mjmahone: there's no reason to use `sortBy` here: just `directive.locations.slice().sort((a, b) => a.compareLocale(b));`

This is useful because then we can make `sortBy` take a mutable array, and avoid one array-copy-per-`sortByName`

src/utilities/sortIntrospection.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+function sortBy<T>(
+  array: $ReadOnlyArray<T>,
+  mapToKey: T => string,
+): $ReadOnlyArray<T> {
+  if (!Array.isArray(array)) {
+    throw new Error('Expected array got: ' + (array: empty));
+  }
+  return array.slice().sort((obj1, obj2) => {
```
mjmahone: a bit of a nit, but we can remove this `slice` with a bit of trickery. We also will no longer need the special `sortBy<T>`.

```
function sortFields(
  fields: $ReadOnlyArray<IntrospectionField>,
): $ReadOnlyArray<IntrospectionField> {
  return sortByName(fields.map(field => ({
    name: field.name,
    description: field.description,
    args: sortByName(field.args.slice()),
    type: field.type,
    isDeprecated: field.isDeprecated,
    deprecationReason: field.deprecationReason,
  }));
 }

// This cannot be called on an original ReadOnly array, or it would mutate them
function sortByName<T: { +name: string }>(
  array: Array<T>,
): $ReadOnlyArray<T> {
  return array.sort((obj1, obj2) => obj1.name.localeCompare(obj2.name));
}
```

This ends up being a little more code (you now have to `slice()` whenever you want to `sortByName`), but now we save one array copy per field list.

src/utilities/sortIntrospection.js
```diff
@@ -0,0 +1,123 @@
...
+export function sortIntrospectionQuery(
+  introspection: IntrospectionQuery,
+): IntrospectionQuery {
+  const schema = introspection.__schema;
+  return {
+    __schema: {
+      queryType: schema.queryType,
+      mutationType: schema.mutationType,
+      subscriptionType: schema.subscriptionType,
+      types: sortByName(schema.types).map(sortType),
```
mjmahone: similar to below, given `sortByName` now takes in a mutable array, we could save one copy here vs. the current code if we flipped this:
```
types: sortByName(schema.types.map(sortType)),
```

src/utilities/sortIntrospection.js
```diff
@@ -0,0 +1,123 @@
...
+      directives: sortByName(schema.directives).map(sortDirective),
```
mjmahone: And the same thing here

IvanGoncharov:
> a bit of a nit, but we can remove this slice with a bit of trickery

@mjmahone Great suggestion 👍
I changed `sortBy`:
```flow
function sortBy<T>(
  array: $ReadOnlyArray<T>,
  mapToKey: T => string,
  sortValue?: T => T,
): $ReadOnlyArray<T> {
  const newArray = sortValue ? array.map(sortValue) : array.slice();
  //...
}
```
What do you think?

I also remembered that `locations` is a new field and could be missing if you work with old GraphQL API so I change `sortBy` to simply ignore all non-array values.

IvanGoncharov:
I changed my mind sorting introspection have one big disadvantage you lose a lot of data when you convert schema to introspection (resolve callback, directives, etc.). I started working on `sortSchema` function.

** Please don't merge this PR **

leebyron:
Going to close this PR at your suggestion, @IvanGoncharov 

I agree with you that the right place for sorting is the GraphQLSchema instance itself - introspection is an intermediate representation, and since we have utilities for converting to and from introspection and schema instances, support for sorting a schema instance can be easily derived to sort any other intermediate representation, including introspection

PR #1187, opened 2017-12-29
Add 'introspectionFromSchema' utility function

IvanGoncharov:
Based on #1114 by @mjmahone

>>(await graphql(schema, introspectionQuery)).data is a pretty strange API that doesn't, IMO, match.

> That is the specified definition of introspection. If we'd like a nominally easier to understand API, we can certainly add that, but it should be a one-liner implemented using exactly that line of code. I think #1115 could help make that easier by guaranteeing a sync result and at least avoiding async when it's not necessary

I think it's better to use `execute` instead of `graphql` since you are not spending time validation introspection query. But `execute` required parsed query so it's not an one-liner anymore.
Checking for `errors` is also a good practice even those introspection query should produce them.
 And finally, you should manually cast the result to `IntrospectionQuery` type.

That's why I think we should have this wrapper function inside lib.

### DIFF
--- a/src/utilities/introspectionFromSchema.js
+++ b/src/utilities/introspectionFromSchema.js
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import invariant from '../jsutils/invariant';
+import { getIntrospectionQuery } from './introspectionQuery';
+import { GraphQLSchema } from '../type/schema';
+import { execute } from '../execution/execute';
+import { parse } from '../language/parser';
+import type {
+  IntrospectionQuery,
+  IntrospectionOptions,
+} from './introspectionQuery';
+
+/**
+ * Build an IntrospectionQuery from a GraphQLSchema
+ *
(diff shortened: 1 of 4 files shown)


src/utilities/introspectionFromSchema.js
```diff
@@ -0,0 +1,37 @@
...
+ * IntrospectionQuery is useful for utilities that care about type and field
+ * relationships, but do not need to traverse through those relationships.
+ *
+ * This is the inverse of buildClientSchema. The primary use case is outside
+ * of the server context, for instance when doing schema comparisons.
+ */
+export function introspectionFromSchema(
+  schema: GraphQLSchema,
+  options: IntrospectionOptions,
+): IntrospectionQuery {
```
leebyron: Just a note, it seems like the response type of this should be named `IntrospectionResult`

leebyron:
Nice

src/utilities/introspectionFromSchema.js
IvanGoncharov: @leebyron It's pretty confusing name but it already defined here:
https://github.com/graphql/graphql-js/blob/863fe2b9bfac689904f3640e912730766cf8a809/src/utilities/introspectionQuery.js#L117
Would be great to rename it to `IntrospectionResult` but I'm not sure how much code it will brake 😞

PR #1189, opened 2018-01-02
getTypeMap() does not include input types used in directives

Yogu:
The result of `getTypeMap()` recursively includes all types referenced by root types, so I would expect it to include types referenced by directives, too.

If this is not intended, I think a test asserting the opposite would be a good idea.

### DIFF
--- a/src/type/__tests__/schema-test.js
+++ b/src/type/__tests__/schema-test.js
@@ -10,6 +10,8 @@ import {
   GraphQLInterfaceType,
   GraphQLObjectType,
   GraphQLString,
+  GraphQLInputObjectType,
+  GraphQLDirective,
 } from '../';
 
 import { describe, it } from 'mocha';
@@ -26,6 +28,25 @@ const ImplementingType = new GraphQLObjectType({
   fields: { fieldName: { type: GraphQLString, resolve: () => '' } },
 });
 
+const DirectiveInputType = new GraphQLInputObjectType({
+  name: 'DirInput',
+  fields: {
+    field: {
+      type: GraphQLString,
+    },
+  },
+});
+
+const Directive = new GraphQLDirective({
+  name: 'dir',
+  locations: ['OBJECT'],
+  args: {
+    arg: {
+      type: DirectiveInputType,
+    },
+  },
+});
+
 const Schema = new GraphQLSchema({
   query: new GraphQLObjectType({
     name: 'Query',
@@ -38,6 +59,7 @@ const Schema = new GraphQLSchema({
       },
     },
   }),
+  directives: [Directive],
 });
 
 describe('Type System: Schema', () => {
@@ -51,6 +73,12 @@ describe('Type System: Schema', () => {
           'Check that schema.types is defined and is an array of all possible ' +
           'types in the schema.',
       );
+    });
+  });
+
+  describe('Type Map', () => {
+    it('includes input types only used in directives', () => {
+      expect(Schema.getTypeMap()).to.include.key('DirInput');
     });
   });
 });
--- a/src/type/__tests__/schema-test.js
+++ b/src/type/__tests__/schema-test.js
@@ -16,6 +16,7 @@ import {
 
 import { describe, it } from 'mocha';
 import { expect } from 'chai';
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
> The result of getTypeMap() recursively includes all types referenced by root types, so I would expect it to include types referenced by directives, too.

@Yogu You absolutely right 👍 

Do you want to fix it yourself?
If so, you can add support for directives to [typeMapReducer](https://github.com/graphql/graphql-js/blob/master/src/type/schema.js#L222). And add all directives to [initialTypes](https://github.com/graphql/graphql-js/blob/master/src/type/schema.js#L112) (make sense to rename it).

Yogu:
Ok great, I'll push a fix tomorrow.

Yogu:
@IvanGoncharov Can you have a look on the fix?

src/type/schema.js
```diff
@@ -269,3 +278,10 @@ function typeMapReducer(map: TypeMap, type: ?GraphQLType): TypeMap {
 
   return reducedMap;
 }
+
+function getDirectiveArgTypes(directive: GraphQLDirective) {
+  if (!isDirective(directive)) {
```
IvanGoncharov: Why do you need this check?

Yogu: There is a test that `validateSchema` throws a specific error when `directives` is invalid (a string instead of a proper instance). `new GraphQLSchema()` does not perform this validation, so for the mentioned test to pass, the constructor needs to accept the invalid directive.

IvanGoncharov: @Yogu Make sense 👍

src/type/schema.js
```diff
@@ -269,3 +278,10 @@ function typeMapReducer(map: TypeMap, type: ?GraphQLType): TypeMap {
...
+    return [];
+  }
+  return directive.args.map(arg => getNamedType(arg.type));
```
IvanGoncharov: @Yogu You don't need `getNamedType` here since it's already handled here:
https://github.com/Yogu/graphql-js/blob/7321a186854af3a095fe7ee52a2a1622dbf288b8/src/type/schema.js#L235-L237

src/type/__tests__/schema-test.js
```diff
@@ -26,6 +28,25 @@ const ImplementingType = new GraphQLObjectType({
...
+    },
+  },
+});
+
+const Directive = new GraphQLDirective({
+  name: 'dir',
+  locations: ['OBJECT'],
+  args: {
+    arg: {
+      type: DirectiveInputType,
```
IvanGoncharov: Would be great to also test wrapped type maybe as one more argument

src/type/schema.js
```diff
@@ -13,6 +13,7 @@ import {
   isUnionType,
   isInputObjectType,
   isWrappingType,
+  getNamedType,
```
IvanGoncharov: @Yogu Looks great now!

One last issue from me: Can you please remove this line?
It cause Travis build to fail:
image

leebyron:
This looks like it also won't work if a directive argument type is a list or non-null type. I would expect flow to also report an error with this PR

Yogu:
> This looks like it also won't work if a directive argument type is a list or non-null type. I would expect flow to also report an error with this PR

@leebyron I initially had code to unwrap the named type, but @IvanGoncharov [pointed out](https://github.com/graphql/graphql-js/pull/1189#discussion_r159432927) this should not be necessary. I'll add a test tomorrow to see if it works.

leebyron:
Ah - I think you're right, it's just not clear from the code. `initialTypes` is expected to be an array of named types, but the type reducer unwraps types since when recursing it encounters wrapped types

IvanGoncharov:
@leebyron In [Voyager](https://github.com/APIs-guru/graphql-voyager) we support selecting any type as a root of a graph:
image
Also, it's very useful for schema stitching since you need to find a minimal set of dependencies for the every injected type.

It would be great to reuse `typeMapReducer` so what do you think about extracting it into `utilities` under `getTypeDependencies` name? Would you merge such PR?

Issue #1188, opened 2017-12-31
"""\nabc\n""" and """abc""" are not distinguishable from the ast


vjeux:
Both of then look exactly the same. I would expect \n to be part of the value.

```js
"""abc"""
type T {
  a: Int
}

"""
abc
"""
type T {
  a: Int
}

```

```js
{ kind: 'StringValue',
  value: 'abc',
  block: true,
  loc: { start: 0, end: 9 } }
{ kind: 'StringValue',
  value: 'abc',
  block: true,
  loc: { start: 31, end: 42 } }
```

IvanGoncharov:
@vjeux It's intentional:
> Note: Block string values are interpreted to exclude blank initial and trailing lines and uniform indentation with {BlockStringValue()}.

https://github.com/facebook/graphql/blob/master/spec/Appendix%20B%20--%20Grammar%20Summary.md#lexical-tokens

vjeux:
Thanks for the quick reply. This is unfortunate as it means I have to read from the raw input to print it as it was entered... I can live with it though!

lydell:
I’m not 100% sure this is an issue: https://github.com/prettier/prettier/pull/3605#discussion_r159255170

IvanGoncharov:
@vjeux I saw you [removed workaround](https://github.com/prettier/prettier/commit/5356db0e696b970bff91a42bc644cf90c267b685#diff-e6e272291fd4f200aa8617e6548d62b3L129) from `prettier`.
Does it mean that your problem is solved and this issue could be closed?

vjeux:
Yes, sorry!

PR #1190, opened 2018-01-02
Fix print of block string with leading space and quotation

IvanGoncharov:
When I looked at #1188 I noticed that current implementation doesn't handle strings like that:
```
"""    space-led value "quoted string"
"""
```
Currently it's printed as:
```
"""    space-led value "quoted string""""
                                     ^ breaks parser
```

### DIFF
--- a/src/language/printer.js
+++ b/src/language/printer.js
@@ -236,7 +236,7 @@ function join(maybeArray, separator) {
  */
 function block(array) {
   return array && array.length !== 0
-    ? indent('{\n' + join(array, '\n')) + '\n}'
+    ? '{\n' + indent(join(array, '\n')) + '\n}'
     : '';
 }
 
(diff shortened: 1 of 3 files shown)


src/language/printer.js
```diff
@@ -249,7 +249,7 @@ function wrap(start, maybeString, end) {
 }
 
 function indent(maybeString) {
-  return maybeString && maybeString.replace(/\n/g, '\n  ');
+  return maybeString && '  ' + maybeString.replace(/\n/g, '\n  ');
```
IvanGoncharov: Can be replaced with:
```js
return maybeString && maybeString.replace(/^|\n/g, '$&  ');
```
But I chose easy to read version even those it more verbose.

leebyron:
What do you think about just simplifying block string printing to *always* be:

```
"""
Content
"""
```

That would require updating some test cases, but would be much more predictable

IvanGoncharov:
> What do you think about just simplifying block string printing to always be:

@leebyron I'm confused 😕 
If you transform `"""   space-led"""` into:
```
""""
   space-led
"""
```
You would lose prefix spaces.
Or you want to change the specification to remove leading and trailing spaces from block string?

leebyron:
Great question about that edge-case. Let me look into that and return to a decision soon

leebyron:
Let's stick with this approach, nice work

PR #1194, opened 2018-01-08
Allow use of legacy names for schema validation

### LINKED ISSUE #1184: Consider turning validation of fields with dunderscore prefix back into warning.
Based on some old discussion (not sure where anymore so can’t provide links) we decided to use `__id` for global object IDs in our schema.

Recently, an initial change to the name validation [started throwing errors](https://github.com/graphql/graphql-js/commit/379a3084392179d2cae92c211a4559f9836299c6#diff-8cc58e6f9e7daaea6d65e16ce1d35b16) for fields starting with a dunderscore. This was then [turned into a warning](https://github.com/graphql/graphql-js/commit/fe346190671c4ad0d666e8eef08da55f85dc7565#diff-8cc58e6f9e7daaea6d65e16ce1d35b16) so legacy schemas wouldn’t immediately fail to load. But then [this recent change](https://github.com/graphql/graphql-js/commit/461392db1eb05b21dc8b048ac352ea2fb37bdbbd#diff-8cc58e6f9e7daaea6d65e16ce1d35b16) turned the warnings into hard errors again.

While I completely understand that this name is not in line with the spec, graphql-js is < v1 which according to semver means anything can change, and in principle throwing an error makes sense; the problem is that I’d rather not have to make very hard breaking changes to our schema (and break current iOS app versions) until either:

* we are going to make sweeping changes anyways (we’re currently experimenting with stitching schemas from backend services, which could lead to a great time to introduce a v2 of our schema)
* there are technical reasons for graphql-js to no longer being able to support fields with dunderscore prefixes

I think the previous logged warning was big enough that people wouldn’t adopt new fields in their schema that use a dunderscore prefix.

PS: I’m mostly making this case for others in the community, I have no problem simply [disabling the check in our apps](https://github.com/artsy/emission/pull/913/commits/f53512181757868785f6929cff8b30ff831c7f2f#r158707392).

leebyron:
This offers an escape hatch for schema that are defined with legacy field names that were valid in older versions of the spec and no longer are via an explicit white-list.

Fixes #1184

### DIFF
--- a/src/type/__tests__/validation-test.js
+++ b/src/type/__tests__/validation-test.js
@@ -407,6 +407,43 @@ describe('Type System: Objects must have fields', () => {
       },
     ]);
   });
+
+  it('accepts an Object type with explicitly allowed legacy named fields', () => {
+    const schemaBad = new GraphQLSchema({
+      query: new GraphQLObjectType({
+        name: 'Query',
+        fields: { __badName: { type: GraphQLString } },
+      }),
+    });
+    const schemaOk = new GraphQLSchema({
+      query: new GraphQLObjectType({
+        name: 'Query',
+        fields: { __badName: { type: GraphQLString } },
+      }),
+      allowedLegacyNames: ['__badName'],
+    });
+    expect(validateSchema(schemaBad)).to.containSubset([
+      {
+        message:
+          'Name "__badName" must not begin with "__", which is reserved by ' +
+          'GraphQL introspection.',
+      },
+    ]);
+    expect(validateSchema(schemaOk)).to.deep.equal([]);
+  });
+
+  it('throws with bad value for explicitly allowed legacy names', () => {
+    expect(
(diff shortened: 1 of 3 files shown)


leebyron:
Curious for your thoughts on this approach, @alloy

src/type/__tests__/validation-test.js
```diff
@@ -407,6 +407,43 @@ describe('Type System: Objects must have fields', () => {
...
+      {
+        message:
+          'Name "__badName" must not begin with "__", which is reserved by ' +
+          'GraphQL introspection.',
+      },
+    ]);
+    expect(validateSchema(schemaOk)).to.deep.equal([]);
+  });
+
+  it('throws with bad value for explicitly allowed legacy names', () => {
```
alloy: Thorough 👌

Issue #1095, opened 2017-11-20
Extend the extend functionality


kbrandwijk:
Currently, it's only possible to extend object types by spec. However, when using schema stitching, I'm running into use case where it would be nice to be able to extend other parts of my schema, like unions, interfaces, enums and input types.

I would like to be able to use the same syntax as for object types. For interfaces, enums and input types, I don't see any issues with that. For unions, the syntax is a bit different, so I'm open to suggestions for how that might look. As an alternative, the use case for unions might also be dealt with when merging schemas. If you agree with that, I will open an issue on graphql-tools for that. But for interfaces, enums, and input types, I think the change should be on the graphql-js side.

Examples:
```
extend enum MyEnum {
  NewValue
}

extend interface MyInterface {
  NewField: String
}

extend input MyInput {
  NewField: String
}
```

leebyron:
Syntax for this exists, however `extendSchema` needs additional logic for these cases. This would be a great task for someone to pick up who seeks to learn about this library

mgrenier:
I've began the work to add the logic to `extendSchema` to support enum, input and interface extension.

I'm preparing a pull request with my changes. Is there any guidelines on how to submit a PR or I just have to refer to the `contribution.md`?

IvanGoncharov:
> I've began the work to add the logic to extendSchema to support enum, input and interface extension.

@mgrenier Great 👍 You can submit WIP PR and get some early feedback.
Also, don't forget to test for tricky use cases like this:
```graphql
extend interface User {
  friends: [User]
}
```

IvanGoncharov:
Implemented by @mgrenier and merged as #1373 🎉 
Available in `14.0.0-rc.1` 📦

dandv:
Thanks for implementing this. Is passing the extended input fields supported?  
The following [example](https://runkit.com/dandv/graphql-extend-input-object---can-t-pass-extended-field) fails with:

> GraphQLError: "password" is not defined

```js
const { graphql, buildSchema } = require('graphql');

const schema = buildSchema(`
  input User {
    username: String!
  }
  extend input User {
    password: String!
  }
  type Query {
    hello: String
  }
  type Mutation {
    createUser(user: User): Boolean
  }
`);

var root = { createUser: user => console.log(user) };

graphql(schema, `mutation {
  createUser(user: {
    username: "john"
    password: "secret"
  })
}`, root).then((response) => {
  console.log(response);
});
```

csantos1113:
[...]

Issue #922, opened 2017-06-20
buildASTSchema() should throw on encountering `extend` definition


leebyron:
`buildASTSchema()` exists to parse a fully complete schema definition and produce a schema. However currently if it encounters an `extend` definition, or any other definition it does not use for that matter, it silently skips over it. This can lead to a confusing debugging situation.

Instead, `buildASTSchema()` should throw when it encounters a definition that it does not use, similar to how `execute()` throws if a the provided document contains any definition other than operations.

OlegIlyenko:
Out of curiosity: is there something that prevents supporting `extend` in `buildASTSchema`?

For example this definition:

```graphql
type Queries {
  querySubset: QuerySubset
}

type Query {
  queries: Queries
}

extend type Queries {
  hello: String
}
``` 

Look semantically correct to me. When used in this form in may look a bit unusual, but the last definition could be loaded from a different file. From what I can tell, supporting `extend` does not have any downsides and provides more flexibility.

leebyron:
That would be new behavior, but I'd also be happy to entertain PRs adding that as well. If so, the same change should probably be made in extendSchema if multiple extend statements are found, and care should be put into ensuring that the order in which the extensions are applied result in the same output.

dyst5422:
I think this might be a breaking change for graphql-tools. I believe the way it currently does schema extension concatenates a number of schema language strings together and then pulls the extensions out to add them in a second step.

Would it make sense to warn on this instead of throwing?
Is there a specific reason `buildASTSchema()` shouldn't parse extensions?
Should there be a method that DOES handle extensions as well?

kaushik94:
Hey I would like to chip in on this one. Can I first take up the issue of adding a warning? Then I will try to understand and help implement handling `extend` statement?

viztor:
Does extend work now? @leebyron 
or do we expect them to work maybe in the future?

The upside I'm seeing is as such allowing people building graphql application to modularize their application schema

IvanGoncharov:
> Does extend work now?

@viztor Not at the moment but I plan to implement it for `14.1.0` release.
Currently, `buildSchema`/`extendSchema` is under heavy refactoring with the goal to implement #1389.

langpavel:
For usecase where somebody needs multiple files merge up to one AST I write this helper: https://gist.github.com/langpavel/9653b99afe993167fc4bacafbfcc7909
[...]

PR #1199, opened 2018-01-10
[WIP] Transfrorm schema API

IvanGoncharov:
As discussed in #1185 I rewrote sort function to sort `GraphQLSchema`.
In a process, I notice that `sortSchema` duplicates > 50% of code from `extendSchema` so I decided to extract common part.
After this change, `extendSchema` became way simpler and #1095 trivial to implement.
To make this PR easier to review I split it into two commits one implementing `sortSchema` and another changing `extendSchema`.

@leebyron What do you think?

### DIFF
--- a/src/utilities/extendSchema.js
+++ b/src/utilities/extendSchema.js
@@ -8,33 +8,18 @@
  */
 
 import invariant from '../jsutils/invariant';
-import keyMap from '../jsutils/keyMap';
-import objectValues from '../jsutils/objectValues';
 import { ASTDefinitionBuilder } from './buildASTSchema';
+import { SchemaTransformer } from './transformSchema';
 import { GraphQLError } from '../error/GraphQLError';
 import { isSchema, GraphQLSchema } from '../type/schema';
-import { isIntrospectionType } from '../type/introspection';
-
 import type { GraphQLSchemaValidationOptions } from '../type/schema';
-
 import {
   isObjectType,
   isInterfaceType,
-  isUnionType,
-  isListType,
-  isNonNullType,
   GraphQLObjectType,
   GraphQLInterfaceType,
-  GraphQLUnionType,
 } from '../type/definition';
-import { GraphQLList, GraphQLNonNull } from '../type/wrappers';
-
-import { GraphQLDirective } from '../type/directives';
-
 import { Kind } from '../language/kinds';
-
-import type { GraphQLType, GraphQLNamedType } from '../type/definition';
-
 import type {
   DocumentNode,
   DirectiveDefinitionNode,
@@ -195,10 +180,12 @@ export function extendSchema(
     typeDefinitionMap,
     options,
     typeRef => {
+      invariant(schemaTransformer);
       const typeName = typeRef.name.value;
-      const existingType = schema.getType(typeName);
-      if (existingType) {
-        return getExtendedType(existingType);
+      const type = schemaTransformer.transformType(typeName);
+
+      if (type) {
+        return type;
       }
 
       throw new GraphQLError(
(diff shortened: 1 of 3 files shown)


langpavel:
Hi @IvanGoncharov Can you pick up use case for this?
I see it sorts even enum values. I hope that this cannot be dangerous even in any badly taken implementation..

IvanGoncharov:
> Can you pick up use case for this?

@langpavel See #941

> I hope that this cannot be dangerous even in any badly taken implementation..

We are living in dangerous world 😆 Seriously, the order of things shouldn't make any difference and even if for some weird reason you depend on it just don't use `sortSchema` on your schema.

IvanGoncharov:
I renamed this function to `lexographicSortSchema` based on https://github.com/graphql/graphql-js/issues/941#issuecomment-349882674

src/utilities/transformSchema.js
```diff
@@ -0,0 +1,298 @@
...
+  EnumType?: (config: ExactEnumTypeConfig) => GraphQLEnumType,
+  InputObjectType?: (
+    config: ExactInputObjectTypeConfig,
+  ) => GraphQLInputObjectType,
+};
+
+/**
+ * Note: This class is not a part of public API and is a subject to changes
+ * in the future.
+ */
```
IvanGoncharov: I think implementing #922 and #1095 will be the real test for this API and maybe require some changes.

IvanGoncharov:
@leebyron Can you please give a quick look at PR?
This PR has some non-trivial changes (especially [SchemaTransformer](https://github.com/graphql/graphql-js/pull/1199/files#diff-4bc4724f7c0a96d36d69eff0bf5e310aR78) class) so I'm not sure if I'm on a right track.

src/utilities/transformSchema.js
```diff
@@ -0,0 +1,298 @@
...
+export class SchemaTransformer {
```
leebyron: I'm not sure why this is a class since the only think you can do with it is `instance.transformSchema()`. I'd suggest refactoring this to just export a function `transformSchema(schema, transformer)` since that presents the same behavior in a simpler API

src/utilities/transformSchema.js
```diff
@@ -0,0 +1,298 @@
...
+  isScalarType,
+  isObjectType,
+  isInterfaceType,
+  isUnionType,
+  isEnumType,
+  isInputObjectType,
+} from '../type/definition';
+
+type ExactSchemaConfig = $Exact<GraphQLSchemaConfig> & {
+  types: Array<*>,
```
leebyron: We should avoid `*` as much as possible - Flow often cannot resolve them, so these are either effectively all `any` or they may be derived to some type which could produce an error that is very difficult to understand. They appear a lot in this file, I think almost every case should be able to be replaced with a true type reference

src/utilities/transformSchema.js
```diff
@@ -0,0 +1,298 @@
...
+  isNonNullType,
+  isScalarType,
+  isObjectType,
+  isInterfaceType,
+  isUnionType,
+  isEnumType,
+  isInputObjectType,
+} from '../type/definition';
+
+type ExactSchemaConfig = $Exact<GraphQLSchemaConfig> & {
```
leebyron: Why $Exact all of the config types?

src/utilities/extendSchema.js
```diff
@@ -171,204 +154,87 @@ export function extendSchema(
...
     options,
-    (typeName, node) => {
-      const existingType = schema.getType(typeName);
-      if (existingType) {
-        return extendType(existingType);
-      }
+    typeRef => {
+      invariant(schemaTransformer);
+      const typeName = typeRef.name.value;
+      const type = schemaTransformer.transformType(typeName);
```
leebyron: This API is a bit confusing - this isn't a transform function, it's something that takes a name and return a type, so I'd expect this to be more like the `getType()` API

IvanGoncharov:
> That should be at least 3-4 separate PRs, each of which can use a previous PR as a base to make them reviewable as a stack

@leebyron After #1208 being merged what is a next PR I should extract?
Maybe it worth to add `toConfig` method into every `GraphQL*Type` class since it useful on its own and will simplify implementation of `transformSchema`.

What do you think adding `toConfig` and convert both `lexographicSortSchema' and `extendSchema` to use it as a next PR?

leebyron:
That could definitely be a good next step - I'll leave it to you - it's also fine to build up multiple PRs, having each cite the previous as a way to illustrate a path of changes

yaacovCR:
Inspired by this prior art, see now fleshed out mapSchema in https://github.com/ardatan/graphql-tools/pull/1463

One implementation consideration I had that might be useful, right now FieldMappers get a fieldConfig and can return a fieldConfig, but TypeMappers get a type and have to return a type, so that a given object type could actually be converted to a scalar as desired and vice a versa.

This could be changed to eliminate that unlikely use case. Alternatively separate OBJECT_TYPE and OBJECT_TYPE_CONFIG mappers could be used. At this point, I am happy with retaining the discrepancy, as it maintains flexibility with decreased API surface, but it does require some repetitive

```
function mapperFunction(originalConfig) {
  const originalConfig = objectType.toConfig();
  ....<create a new config based on old>...
  return new GraphqlObjectType(newConfig)
}
```

instead of

```
function mapperFunction(originalConfig) {
  ....<create a new config based on old>...
  return newConfig
}
```

IvanGoncharov:
Extracted in #3226

PR #1201, opened 2018-01-11
Cleanup Enums

IvanGoncharov:
I converted `Kind` into an object and created `KindEnum` and `TokenKindEnum` types.
To do this I wrapped enums into `Object.freeze` as [required by Flow](https://github.com/facebook/flow/blame/master/Changelog.md#L107-L116).
It's safe to use `Object.freeze` since it supported by [all browsers](https://github.com/graphql/graphql-js/blob/master/.babelrc#L6-L12) supported by this lib:
image

If for some reason `Object.freeze` can't be used in this lib I can wrap it into code comment as suggested here: https://github.com/facebook/flow/issues/5465#issuecomment-350422478

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -17,7 +17,7 @@ import type { ObjMap } from '../jsutils/ObjMap';
 import type { MaybePromise } from '../jsutils/MaybePromise';
 
 import { typeFromAST } from '../utilities/typeFromAST';
-import * as Kind from '../language/kinds';
+import { Kind } from '../language/kinds';
 import {
   getVariableValues,
   getArgumentValues,
--- a/src/execution/values.js
+++ b/src/execution/values.js
@@ -14,7 +14,7 @@ import keyMap from '../jsutils/keyMap';
 import { coerceValue } from '../utilities/coerceValue';
 import { typeFromAST } from '../utilities/typeFromAST';
 import { valueFromAST } from '../utilities/valueFromAST';
-import * as Kind from '../language/kinds';
+import { Kind } from '../language/kinds';
 import { print } from '../language/printer';
 import { isInputType, isNonNullType } from '../type/definition';
 import type { ObjMap } from '../jsutils/ObjMap';
--- a/src/index.js
+++ b/src/index.js
@@ -247,6 +247,7 @@ export type {
   TypeExtensionNode,
   ObjectTypeExtensionNode,
   DirectiveDefinitionNode,
+  KindEnum,
   DirectiveLocationEnum,
 } from './language';
 
--- a/src/index.js
+++ b/src/index.js
@@ -248,6 +248,7 @@ export type {
   ObjectTypeExtensionNode,
   DirectiveDefinitionNode,
   KindEnum,
+  TokenKindEnum,
   DirectiveLocationEnum,
 } from './language';
 
--- a/src/language/__tests__/parser-test.js
+++ b/src/language/__tests__/parser-test.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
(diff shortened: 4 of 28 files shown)


leebyron:
A couple thoughts:

First, do you have any performance metrics for this change? This replaces a local const with a dictionary lookup for `kind` in the parser and lexer and replaces a const int with a string length calculation in the lexer. I think it's important we ensure this isn't a parse time regression.

Second, could you explain the impact of this change on the use of the library? Aside from some extra verbosity cost of replacing `XYZ` with `Kind.XYZ` everywhere, this seems reasonable - though I'm not sure I understand if this has non-internal implications

leebyron:
I've also heard reports that `Object.freeze()` can result in slower object reads - do you know about this and if those reports apply to this use case?

IvanGoncharov:
> Second, could you explain the impact of this change on the use of the library? Aside from some extra verbosity cost of replacing XYZ with Kind.XYZ everywhere, this seems reasonable - though I'm not sure I understand if this has non-internal implications

I look at different parts of `graphql-js` almost every day for two reasons:
* to understand certain aspects of GraphQL
* how to reuse code from `graphql-js` so I don't need to reimplement it in my tools

And if I see inconsistency I can't understand is intentional and I just don't get something or it looks like that for historical reasons. In this case, I noticed recently added [directiveLocation.js](https://github.com/graphql/graphql-js/blob/master/src/language/directiveLocation.js) and was wondering why other enums are defined differently.

You absolutely right that this PR doesn't change anything for library consumers but I think such changes improve quality of "reference implementation".

> I've also heard reports that Object.freeze() can result in slower object reads - do you know about this and if those reports apply to this use case?

The only thing I could find is reports that applying `Object.freeze` on `prototype` of built-in types (Array, Object,...) were preventing Chrome from using built-in implementations of their methods.
It's not affecting this PR and already fixed, here is a great article explaining the details:
http://benediktmeurer.de/2017/08/10/frozen-prototypes/

> First, do you have any performance metrics for this change?

Not initially 😞 
But I already added [lexer benchmark](https://github.com/graphql/graphql-js/pull/1167/files#diff-157dcc15fe75ec1a0f0a6ad0cd72f8e4R22) into #1167 

> his replaces a local const with a dictionary lookup for kind in the parser and lexer

I didn't measure any changes both with #1167 and node profiler.

> const int with a string length calculation in the lexer

It did cause performance drop ~4% but only because I defined `literalTok` inside the function.
And I didn't measure any difference between int and a string length.

But when I moved `literalTok` in the global scope (e9098c2d96d4425129c216e06f93bbf84fea1665) it actually resulted in ~7% performance increase against current master and I also was able to measure a few percent increase in the parser benchmark.
It surprised me a lot so I run profiler (master on the left and `literalTok` with no other changes is on the right):
profile

So it looks like V8 inlined both build-in for `new` and `Tok` constructor inside `literalTok`.
I think it happened because V8 was able to figure out 100% types of `literalTok`:
```
[marking  <JSFunction literalTok (sfi = )> for optimized recompilation, reason: small function, ICs with typeinfo: 3/3 (100%), generic ICs: 0/3 (0%)]
[marking  <JSFunction readToken (sfi = )> for optimized recompilation, reason: hot and stable, ICs with typeinfo: 126/140 (90%), generic ICs: 0/140 (0%)]
```
So my speculation is that `ICs with typeinfo: 3/3 (100%)` responcible for this boost.
Maybe in future versions of V8 they will be able to optimise entire `readToken`.

leebyron:
Thanks for the analysis, @IvanGoncharov - this looks great

IvanGoncharov:
> Enum values are best left as representative opaque values, so we should avoid `value.length`.

Agree 👍 
Thanks for merging.

PR #1203, opened 2018-01-11
Improve dedent (fixes #1202)

### LINKED ISSUE #1202: Issues with the dedent string tag used for testing
GraphQL.js has a ES6 string tag helper function under `jsutils/dedent` that removes indentation from strings so that the unit tests can look nicer.

One undocumented side effect of the `dedent` function however confused me when reading some of the tests: The tag also adds escape sequences to the dedented string. Take for example this test:

```javascript
  it('Prints String Field With String Arg With Default', () => {
    const output = printSingleFieldSchema({
      type: GraphQLString,
      args: { argOne: { type: GraphQLString, defaultValue: 'tes\t de\fault' } },
    });
    expect(output).to.equal(dedent`
      schema {
        query: Root
      }

      type Root {
        singleField(argOne: String = "tes\t de\fault"): String
      }
    `);
  });
```

The output actually contains escape sequences, like `"tes\\t de\\fault"`, but the test looks likes as if it expects the output to be a non-escaped string.

The reason for this is that the `dedent` tag accesses `string.raw` instead of `string`.

I'm not sure whether this is intended or not. If it *is*, then it should at least be documented in `jsutils/dedent`, because having such a side effect is not what you expect from a function called "dedent". It should also be documented that the function only looks at spaces, it does *not* dedent tab-based indentation.

If you me know whether you prefer fixing the documentation or fixing the behavior, I can create a PR.

Cito:
As discussed in #1202, this changes the behavior of the dedent function that is used in unit tests to not return raw strings any more. Also added tests for the function itself.

### DIFF
--- a/src/jsutils/__tests__/dedent-test.js
+++ b/src/jsutils/__tests__/dedent-test.js
@@ -94,4 +94,20 @@ describe('dedent', () => {
         me: User
       }`).to.equal('type Query {\n  me: User\n}');
   });
+
+  it('supports expression interpolation', () => {
+    const name = 'Luke Skywalker';
+    const age = 42;
+    expect(dedent`
+        {
+          "me": {
+            "name": "${name}"
+            "age": ${age}
+          }
+        }
+      `).to.equal(
+      '{\n  "me": {\n    "name": "Luke Skywalker"\n' +
+        '    "age": 42\n  }\n}\n',
+    );
+  });
 });
--- a/src/jsutils/__tests__/dedent-test.js
+++ b/src/jsutils/__tests__/dedent-test.js
@@ -11,7 +11,7 @@ import dedent from '../dedent';
 
 describe('dedent', () => {
   it('removes indentation in typical usage', () => {
-    expect(dedent`
+    const output = dedent`
       type Query {
         me: User
       }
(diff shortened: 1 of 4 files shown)


src/language/__tests__/printer-test.js
```diff
@@ -190,7 +190,7 @@ describe('Printer: Query document', () => {
 
       fragment frag on Friend {
         foo(size: $size, bar: $b, obj: {key: "value", block: """
-          block string uses \"""
+          block string uses \\"""
```
leebyron: Instead of changing this to `\\`, could you use `String.raw` as discussed in #1202? This test case intends to show the actual GraphQL string written, so avoiding double-encoding is intentional.

src/utilities/__tests__/schemaPrinter-test.js
```diff
@@ -212,7 +212,7 @@ describe('Type System Printer', () => {
       }
 
       type Root {
-        singleField(argOne: String = "tes\t de\fault"): String
+        singleField(argOne: String = "tes\\t de\\fault"): String
```
leebyron: same

Cito:
That's true, using String.raw is clearer for the printer tests. I changed it and added another unit test to make sure interpolation also works properly when dedent is used as a string tag.

src/jsutils/__tests__/dedent-test.js
```diff
@@ -0,0 +1,113 @@
...
+      }
+      
+      type User {
+        id: ID
+        name: String
+      }
+      `).to.equal(
+      'type Query {\n  me: User\n}\n\n' +
+        'type User {\n  id: ID\n  name: String\n}\n',
+    );
```
IvanGoncharov: In the current form, these tests are hard to read.
Would be great to have test string in a separate variable and expected string in [readable form](https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/blockStringValue-test.js#L14-L21), e.g.:
```js
const output = dedent`
  type Query {
    me: User
  }
`;

expect(output).to.equal([
  'type Query {',
  '  me: User',
  '}',
].join('\n'));

Cito: Yes, that looks a bit nicer, particularly if you're using that style already elsewhere. I refactored it already.

Cito:
Rebased onto current master now.

PR #1205, opened 2018-01-19
Fix printed schema descriptions that end with a double quote (")

mostcowbell:
Schemas printed with `printSchema` that contain descriptions ending with a double-quote character are not parseable and throw a `GraphQLError: Syntax Error: Unterminated string` because the SDL description ends with 4 contiguous double-quotes

I've included a failing test case:

```js
const Root = new GraphQLObjectType({
      name: 'Root',
      fields: {
        onlyField: {
          type: GraphQLString,
          description: 'This field is "awesome"',
        },
      },
    });
    const Schema = new GraphQLSchema({ query: Root });
    const output = printSchema(Schema);
    expect(() => parse(output)).not.to.throw(GraphQLError);
```

### DIFF
--- a/src/utilities/__tests__/schemaPrinter-test.js
+++ b/src/utilities/__tests__/schemaPrinter-test.js
@@ -26,6 +26,8 @@ import {
 } from '../../';
 import { GraphQLDirective } from '../../type/directives';
 import { DirectiveLocation } from '../../language/directiveLocation';
+import { parse } from '../../language/parser';
+import { GraphQLError } from '../../error/GraphQLError';
 
 function printForTest(schema) {
   const schemaText = printSchema(schema);
@@ -849,7 +851,20 @@ describe('Type System Printer', () => {
     `;
     expect(output).to.equal(introspectionSchema);
   });
-
+  it('Prints a parseable schema when a comment ends with a doublequote (")', () => {
+    const Root = new GraphQLObjectType({
+      name: 'Root',
+      fields: {
+        onlyField: {
+          type: GraphQLString,
+          description: 'This field is "awesome"',
+        },
+      },
+    });
+    const Schema = new GraphQLSchema({ query: Root });
+    const output = printSchema(Schema);
+    expect(() => parse(output)).not.to.throw(GraphQLError);
+  });
   it('Print Introspection Schema with comment descriptions', () => {
     const Query = new GraphQLObjectType({
       name: 'Query',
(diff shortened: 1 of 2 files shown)


leebyron:
cc @IvanGoncharov - I know you had a similar fix land earlier. Care to review this?

IvanGoncharov:
> cc @IvanGoncharov - I know you had a similar fix land earlier. Care to review this?

@leebyron  @mostcowbell Yes I fix similar issue with `print` function in #1190

Adding space changing description string and if we want to omit it than possible solution would:
```
"""
This field is "awesome"
"""
```
Or as:
```
"""This field is "awesome"
"""
```

Also, you need to handle the [same edge case](https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/printer-test.js#L111-L125) as `print` function:
```
"""    space-led value "quoted string"
"""
```

On the other hand [descriptionLines](https://github.com/graphql/graphql-js/blob/master/src/utilities/schemaPrinter.js#L371-L387) already changes description string.
Moreover, all description values are in Markdown format so trailing spaces are ignored.

Personally, I think `print` should try to produce an output as close as possible to original but `printSchema` output should be readable and less-surprising.  So I think @mostcowbell solution is better but it should be explicitly documented in the tests.

mostcowbell:
@IvanGoncharov so I should add another assertion to my test?
```js
expect(output).to.equal(`
  schema {
    query: Root
  }
  
  type Root {
    """This field is "awesome" """
    onlyField: String
  }
`);
```
or just a comment about the output of a description ending in a double-quote?
```js
// Will be printed as `"""This field is "awesome" """` to avoid parsing issues
```

IvanGoncharov:
@mostcowbell I would suggest being as explicit as possible: 
```js
    const description = 'This field is "awesome"';
    const output = printSingleFieldSchema({ type: GraphQLString, description });
    expect(output).to.equal(`
      ...
    `);

    const recreatedRoot = buildSchema(output).getTypeMap()['Root'];
    const recreatedField = recreatedRoot.getFields()['singleField'];

    /* Note: Additional space added to prevent parser confusing trailing double
     * quote inside description with the end of the block string. It should
     * affect result since descriptions are interpreted as Markdown so trailing
     * spaces are ignored.
     */
    expect(recreatedField.description).to.equal(description + ' ');
```
Disclaimer: I'm not a native speaker :)

>   Prints a parseable schema when a comment ends with a doublequote (")s

And change test description to something like: `Prints a field description that ends with a doublequote (")s`

mostcowbell:
@IvanGoncharov I've updated my test and removed the extra dependencies I was importing into that file

leebyron:
I'd like to see a solution that also handles the edge cases you referenced, @IvanGoncharov - I'll pull in these test cases and ensure the solution accounts for them.

dncrews:
Wow @leebyron way to give credit to your contributors...

PR #1207, opened 2018-01-21
More accurate check for cache-hit in memoize3

mohawk2:
`WeakMap.get` returns `undefined` on cache miss. Checking for false instead of `undefined` could give a false cache-miss and recalculate when a false value is returned by the memoised function with those params.


leebyron:
Thanks for adding this!

PR #1226, opened 2018-02-07
Maintain allowed legacy names when extending a schema.

alloy:
v0.13.0-rc.1 adds the `allowedLegacyNames` option from #1194. Unfortunately, while trying to make a PR for [Relay](https://github.com/facebook/relay) that adds support for this, it turns out that Relay internally uses `extendSchema` from `graphql-js` but that function does not maintain this option and thus the new schema still complains.

This PR addresses that, it would be nice if this could go into a v0.13.0-rc.2 release.

### DIFF
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -90,6 +90,7 @@ const testSchema = new GraphQLSchema({
     }),
   }),
   types: [FooType, BarType],
+  allowedLegacyNames: ['__badName'],
 });
 
 describe('extendSchema', () => {
@@ -992,6 +993,16 @@ describe('extendSchema', () => {
       'Cannot extend type "UnknownType" because it does not exist in the ' +
         'existing schema.',
     );
+  });
+
+  it('maintains configuration of the original schema object', () => {
+    const ast = parse(`
+      extend type Query {
+        __badName: String
+      }
+    `);
+    const schema = extendSchema(testSchema, ast);
+    expect(schema.__allowedLegacyNames).to.eql(['__badName']);
   });
 
   describe('does not allow extending a non-object type', () => {
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -90,7 +90,6 @@ const testSchema = new GraphQLSchema({
     }),
   }),
   types: [FooType, BarType],
-  allowedLegacyNames: ['__badName'],
 });
 
 describe('extendSchema', () => {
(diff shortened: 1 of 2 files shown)


src/utilities/__tests__/extendSchema-test.js
```diff
@@ -994,6 +995,16 @@ describe('extendSchema', () => {
...
   });
 
+  it('maintains configuration of the original schema object', () => {
+    const ast = parse(`
+      extend type Query {
+        __badName: String
+      }
+    `);
+    const schema = extendSchema(testSchema, ast);
+    expect(schema.__allowedLegacyNames).to.eql(['__badName']);
```
IvanGoncharov: All other tests are using`to.equal` so I think it should be consistent.

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -90,6 +90,7 @@ const testSchema = new GraphQLSchema({
     }),
   }),
   types: [FooType, BarType],
+  allowedLegacyNames: ['__badName'],
```
IvanGoncharov: I don't think it worth to add this option to common test setup
It's better to define separate minimal schema inside test-case itself.

src/utilities/__tests__/extendSchema-test.js
alloy: I had tried that, but it looks like chai’s `equal` does an identity comparison, whereas `eql` does equality comparison. Seeing as the `__allowedLegacyNames` array gets cloned before passing to the new schema instance, an identity comparison can’t be used here.

src/utilities/__tests__/extendSchema-test.js
alloy: I like that, I just figured that this file was adding all fixture data to this single schema and followed suit. Happy to revisit and create a tiny schema in the test, though 👍

src/utilities/extendSchema.js
```diff
@@ -230,6 +230,8 @@ export function extendSchema(
     types,
     directives: getMergedDirectives(),
     astNode: schema.astNode,
+    allowedLegacyNames:
+      schema.__allowedLegacyNames && schema.__allowedLegacyNames.slice(0),
```
IvanGoncharov: You don't need to use `slice` since it's defined as:
https://github.com/graphql/graphql-js/blob/b333b30b238d040d6a8cf959d54f037ee26ae2c6/src/type/schema.js#L86

So `allowedLegacyNames: schema.__allowedLegacyNames` should be enough.

src/utilities/__tests__/extendSchema-test.js
IvanGoncharov: Did you tried `to.deep.equal`?

src/utilities/extendSchema.js
alloy: This is why I needed the equality check, basically. Flow didn’t want to accept a read-only array for the config, which is typed to accept a mutable array: https://github.com/graphql/graphql-js/blob/b333b30b238d040d6a8cf959d54f037ee26ae2c6/src/type/schema.js#L248

I was surprised too, though, so not being very familiar with Flow I’ll do another sanity check.

src/utilities/__tests__/extendSchema-test.js
alloy: Ah no, I tried `deepEqual` assuming that Chai was using Node’s `assert` API. Thanks, will try that 👍

leebyron:
Thanks for catching this!

alloy:
@IvanGoncharov Thanks for the review!

@leebyron Thanks for making those changes! When this is released I’ll make a PR for Relay to use this and expose a CLI option to specify legacy names.

PR #1228, opened 2018-02-07
Add more 'buildASTSchema' tests + cleanup

IvanGoncharov:
Tested that 'buildASTSchema' can build invalid schema suggested here:
https://github.com/graphql/graphql-js/pull/1222/files#r166484638

### DIFF
--- a/src/utilities/__tests__/buildASTSchema-test.js
+++ b/src/utilities/__tests__/buildASTSchema-test.js
@@ -36,7 +36,7 @@ function cycleOutput(body, options) {
 describe('Schema Builder', () => {
   it('can use built schema for limited execution', () => {
     const schema = buildASTSchema(
-      parse(dedent`
+      parse(`
         type Query {
           str: String
         }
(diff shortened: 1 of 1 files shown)


src/utilities/__tests__/buildASTSchema-test.js
```diff
@@ -32,61 +34,54 @@ function cycleOutput(body, options) {
...
-  it('can use built schema for limited execution', async () => {
+  it('can use built schema for limited execution', () => {
     const schema = buildASTSchema(
-      parse(`
-      schema { query: Query }
-      type Query {
-        str: String
-      }
-    `),
+      parse(dedent`
```
leebyron: I don't think we need to use `dedent` for these literals since we're not testing token locations nor doing string comparisons

IvanGoncharov: @leebyron It certain test failing because of buildSchema throw Error. It's easier to use location inside error to figure out what's wrong but without `dedent` location is not intuitive and hard to map on SDL string.

PR #1229, opened 2018-02-08
Update flow + "@flow strict"

IvanGoncharov:
It's such a pain to not be able to function args inside arrow functions, but the new version of Flow introduced `@flow strict` that makes function arguments constant.
Since `@flow strict` don't rely on `.flowconfig` it shouldn't cause the problem described in #1157.

It's a huge DX improvement for me since I spend 5-10 minutes debugging every time Flow is not able to figure out the type of function argument.


leebyron:
I merged in the new flow version, but we don't need to use strict pragmas since we're not using the strict linter yet

leebyron:
I was wrong - it seems the strict mode adds additional checks in addition to applying the linter

IvanGoncharov:
@leebyron Thanks for merge.
Why did you revert this code:
https://github.com/graphql/graphql-js/blob/5fe39262a308df944a87cc85b225228e7556aaa4/src/type/introspection.js#L467-L476

It was introduced in #1160 because Flow you reverted `experimental.const_params=true`:
https://github.com/graphql/graphql-js/pull/1160/files#diff-87c5e74dd1f7d923143e0eee611f598eR150 
Or I'm missing something?

PR #1230, opened 2018-02-08
Make 'ASTDefinitionBuilder' responsible only for build types from AST

IvanGoncharov:
It's part of #1199 
Its primary purpose is to make  'ASTDefinitionBuilder' responsible only for build types from AST.
It's not only made possible to extract `transformSchema` into the separate function but also make the code more modular and simple.

### DIFF
--- a/src/utilities/extendSchema.js
+++ b/src/utilities/extendSchema.js
@@ -203,7 +203,7 @@ export function extendSchema(
       const typeName = typeRef.name.value;
       const existingType = schema.getType(typeName);
       if (existingType) {
-        return extendType(existingType);
+        return getExtendedType(existingType);
       }
 
       throw new GraphQLError(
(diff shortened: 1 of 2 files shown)


src/utilities/extendSchema.js
```diff
@@ -241,30 +237,29 @@ export function extendSchema(
...
+        extendedType = extendInterfaceType(type);
+      } else if (isUnionType(type)) {
+        extendedType = extendUnionType(type);
+      } else {
+        extendedType = type;
+      }
+      extendTypeCache[type.name] = extendedType;
     }
-    return type;
+    return (extendedType: any);
```
leebyron: I suggest factoring this into two functions, `getExtendedType` and `extendType`, the first one performing caching (with a comment that extendType need only be called once per type). That should allow you to remove or at least further isolate the `any` cast.

IvanGoncharov: @leebyron Great idea 👍 I will make the change in a few minutes.

leebyron: Cast through any internally while keeping the correct function definition, and we'll report it to the flow team

IvanGoncharov: @leebyron Done. Can you please review?

src/utilities/extendSchema.js
```diff
@@ -275,30 +273,32 @@ export function extendSchema(
...
+        extendedType = extendObjectType(type);
+      } else if (isInterfaceType(type)) {
+        extendedType = extendInterfaceType(type);
+      } else if (isUnionType(type)) {
+        extendedType = extendUnionType(type);
+      }
     }
-    return type;
+    // Workaround: Flow should figure out correct type, but it doesn't.
+    return (extendedType: any);
```
leebyron: Rather than mutating a variable in each branch, just return the value in each branch. Flow can't figure out what the type of `extendedType` is, since there are so many different possible local mutations

IvanGoncharov: @leebyron I did that initially but it forces me to do typecast in every branch. Plus explaining workaround is more complicated.

> Flow can't figure out what the type of extendedType is, since there are so many different possible local mutations

Flow has exactly the same problem figuring types both with or without mutations. You can see it in CI logs for your commit and in Flow playground:
image
image

IvanGoncharov:
@leebyron a8b9539 is failing flow check and I don't know how to fix it: https://github.com/graphql/graphql-js/pull/1230#discussion_r167345484
What should I do?

IvanGoncharov:
@leebyron I finally figure our how fix flow error, see 6f1867c
Can you please review this change?

leebyron:
Looks great! Thanks for iterating towards this better version

PR #1234, opened 2018-02-08
Export NAME_RX name validation

gdi2290:
Validations needed for dynamic query generation.

For [Tipe](https://tipe.io) we dynamically create the shape of data so we have to validate and generate these queries.


leebyron:
This RX is an internal implementation detail, and only part of name validation. You should use `assertValidName` or if you want an API which does not throw, `isValidNameError` should be exported from the top index.js instead.

PR #1235, opened 2018-02-08
More allowed legacy names

alloy:
Since yesterday I went back and did some more code reviewing and realised there’s another few situations where the allowed legacy names configuration isn’t being passed on, apologies for not catching that yesterday.

### DIFF
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -229,14 +229,16 @@ export class GraphQLSchema {
 
 type TypeMap = ObjMap<GraphQLNamedType>;
 
-export type GraphQLSchemaConfig = {
-  query?: ?GraphQLObjectType,
-  mutation?: ?GraphQLObjectType,
-  subscription?: ?GraphQLObjectType,
-  types?: ?Array<GraphQLNamedType>,
-  directives?: ?Array<GraphQLDirective>,
-  astNode?: ?SchemaDefinitionNode,
+export type GraphQLSchemaValidationOptions = {|
+  /**
+   * When building a schema from a GraphQL service's introspection result, it
+   * might be safe to assume the schema is valid. Set to true to assume the
+   * produced schema is valid.
+   *
+   * Default: false
+   */
   assumeValid?: boolean,
+
   /**
    * If provided, the schema will consider fields or types with names included
    * in this list valid, even if they do not adhere to the specification's
@@ -246,6 +248,16 @@ export type GraphQLSchemaConfig = {
    * major release.
    */
   allowedLegacyNames?: ?Array<string>,
+|};
+
+export type GraphQLSchemaConfig = {
+  query?: ?GraphQLObjectType,
+  mutation?: ?GraphQLObjectType,
+  subscription?: ?GraphQLObjectType,
+  types?: ?Array<GraphQLNamedType>,
+  directives?: ?Array<GraphQLDirective>,
+  astNode?: ?SchemaDefinitionNode,
+  ...GraphQLSchemaValidationOptions,
 };
 
 function typeMapReducer(map: TypeMap, type: ?GraphQLType): TypeMap {
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -111,9 +111,9 @@ export class GraphQLSchema {
         '"allowedLegacyNames" must be Array if provided but got: ' +
           `${String(config.allowedLegacyNames)}.`,
       );
-      this.__allowedLegacyNames = config.allowedLegacyNames;
     }
 
+    this.__allowedLegacyNames = config.allowedLegacyNames;
     this._queryType = config.query;
     this._mutationType = config.mutation;
     this._subscriptionType = config.subscription;
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -247,7 +247,7 @@ export type GraphQLSchemaValidationOptions = {|
    * This option is provided to ease adoption and may be removed in a future
    * major release.
    */
-  allowedLegacyNames?: ?Array<string>,
+  allowedLegacyNames?: ?$ReadOnlyArray<string>,
 |};
 
 export type GraphQLSchemaConfig = {
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -1225,6 +1225,25 @@ describe('extendSchema', () => {
       query: new GraphQLObjectType({
         name: 'Query',
         fields: () => ({
+          id: { type: GraphQLID },
+        }),
+      }),
+      allowedLegacyNames: ['__badName'],
+    });
+    const ast = parse(`
+      extend type Query {
+        __badName: String
+      }
+    `);
+    const schema = extendSchema(testSchemaWithLegacyNames, ast);
+    expect(schema.__allowedLegacyNames).to.deep.equal(['__badName']);
+  });
+
+  it('adds to the configuration of the original schema object', () => {
+    const testSchemaWithLegacyNames = new GraphQLSchema({
+      query: new GraphQLObjectType({
+        name: 'Query',
+        fields: () => ({
           __badName: { type: GraphQLString },
         }),
       }),
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -1225,18 +1225,23 @@ describe('extendSchema', () => {
       query: new GraphQLObjectType({
         name: 'Query',
         fields: () => ({
-          id: { type: GraphQLID },
+          __badName: { type: GraphQLString },
         }),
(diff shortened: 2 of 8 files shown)


src/type/schema.js
```diff
@@ -111,9 +111,9 @@ export class GraphQLSchema {
         '"allowedLegacyNames" must be Array if provided but got: ' +
           `${String(config.allowedLegacyNames)}.`,
       );
-      this.__allowedLegacyNames = config.allowedLegacyNames;
     }
 
+    this.__allowedLegacyNames = config.allowedLegacyNames;
```
alloy: The `assumeValid` configuration is about whether or not to check the configuration for common mistakes, so regardless of that the allowed legacy names still need to be stored on the schema instance.

There’s no tests for `assumeValid` and the below configuration being set or not, so wasn’t sure what you’d like. Should I add a test for any configuration being set, regardless of `assumeValid`?

leebyron: Adding a test would be great!

Let me know if I understand this correctly - you could create a GraphQLSchema with *both* `__allowedLegacyNames` *and* `assumeValid`, then extend that schema with a legacy name (which might no longer be assumed invalid), and then expect validation to pass?

src/utilities/extendSchema.js
```diff
@@ -55,6 +55,16 @@ type Options = {|
...
+
+  /**
+   * If provided, the schema will consider fields or types with names included
+   * in this list valid, even if they do not adhere to the specification's
+   * schema validation rules.
+   *
+   * This option is provided to ease adoption and may be removed in a future
+   * major release.
+   */
+  allowedLegacyNames?: ?Array<string>,
```
leebyron: Should this be added here? If so, should the result of `extendSchema` include both the allowed legacy names of the input schema in addition to these?

src/utilities/buildASTSchema.js
```diff
@@ -89,6 +89,16 @@ type Options = {|
...
+
+  /**
+   * If provided, the schema will consider fields or types with names included
+   * in this list valid, even if they do not adhere to the specification's
+   * schema validation rules.
+   *
+   * This option is provided to ease adoption and may be removed in a future
+   * major release.
+   */
+  allowedLegacyNames?: ?Array<string>,
```
leebyron: This makes sense as an alternate API for constructing a schema. 👍

src/type/schema.js
alloy: Hmm, I now see that `validate` depends on the private `__validationErrors` property of a schema instance.

I’m now starting to think that maybe what Relay is doing isn’t entirely correct. First it creates a schema [with `assumeValid` set](https://github.com/facebook/relay/blob/181455c68f6870df93a1f1c2be0dcc0101399038/packages/relay-compiler/bin/RelayCompilerBin.js#L243), then after extending that schema a few times it expects to still be able to [validate that schema](https://github.com/facebook/relay/blob/181455c68f6870df93a1f1c2be0dcc0101399038/packages/relay-compiler/graphql-compiler/core/GraphQLValidator.js#L47), which appears to only really be possible because `extendSchema` does not copy over the `assumeValid` configuration.

Maybe `extendSchema` should _also_ copy over the `assumeValid` setting, but `validate` could accept an optional parameter to force performing the validations regardless of `__validationErrors` already being set?

leebyron: Not quite. extendSchema produces a new schema - while the provided schema may have been "valid" (either assumed or validated), the returned schema is different and may be invalid, so it must not copy over assumeValid or __validationErrors from the input. We would expect the extended schema to be validated by validateSchema().

Now, we may want to provide a similar assumeValid option for extendSchema, but that would be a separate thing.

validate checks __validationErrors first as a form of memoization. Since schema are considered immutable, we do not want to wastefully validate the same schema multiple times.

leebyron: Relay first builds a schema with the assumeValid set, because it expects that schema has been provided from a valid GraphQL server - whereas any client side extensions are within Relay's responsibility to validate

leebyron: Also, the `GraphQLValidator` second link is validating the graphql operations and fragments found within relay product code, not the schema

alloy: Ok, I see, that makes sense. So to answer your initial question, when an original schema has `__allowedLegacyNames` then extending that schema should still allow those original legacy names too, regardless from whether or not they were validated on the original schema.

alloy: That `validate` call in `GraphQLValidator` is definitely still leading to an error about the legacy name in the schema . Maybe it’s just a side-effect, though, I’ll double-check that in the morrow.

IvanGoncharov:
@alloy @leebyron In theory buildClientSchema also should accept `allowedLegacyNames`.

Idea mode: Can we create a global array in `assertValidName.js` and just export `addAllowedLegacyName` function. We guarantee to have only one instance of `graphql` so you could write:
```js
import { addAllowedLegacyName } from 'graphql';

addAllowedLegacyName('__badname');
someRelayFunctionWorkingWithSDL(`
  type Query {
     __badname: String
  }
`);
```
That way we minimize changes in `graphql-js` to one file and allow to use the relay and other 3rd-party libs without any changes.
What do you think?

src/type/schema.js
leebyron: Probably a side effect - the document validator first asserts that a valid schema if being used to validate a document. If the schema wasn't validated before that point, it will throw

alloy:
@IvanGoncharov I’m personally not a big fan of singletons, but have no idea if that pattern is already being employed in this library or if that’s something that’s acceptable. Will defer to @leebyron on that.

What I did do is consolidate the schema validation options into a single type that’s shared and added the `allowedLegacyNames` option to `buildClientSchema` as well.

alloy:
Oh, just remembered that I still need to add tests for the existing `assumeValid` behaviour. Will do that later today.

src/utilities/buildClientSchema.js
```diff
@@ -60,16 +60,7 @@ import type {
...
-type Options = {|
-  /**
-   * When building a schema from a GraphQL service's introspection result, it
-   * might be safe to assume the schema is valid. Set to true to assume the
-   * produced schema is valid.
-   *
-   * Default: false
-   */
-  assumeValid?: boolean,
-|};
```
IvanGoncharov: I still think it's better to have `type Options = {| ...GraphQLSchemaValidationOptions |}` even though there are no other options at the moment. It shows that `options` arg could be used for some other options beyond validation.

alloy: Gotcha, makes sense 👍

alloy:
@leebyron Alright, I think I’ve addressed all feedback.

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -1225,18 +1225,23 @@ describe('extendSchema', () => {
       query: new GraphQLObjectType({
         name: 'Query',
         fields: () => ({
-          id: { type: GraphQLID },
+          __badName: { type: GraphQLString },
```
leebyron: Could you add a new test rather than changing this existing one?

src/utilities/extendSchema.js
```diff
@@ -245,6 +240,14 @@ export function extendSchema(
     types.push(definitionBuilder.buildType(typeName));
   });
 
+  let allowedLegacyNames = [
+    ...(schema.__allowedLegacyNames || []),
+    ...((options && options.allowedLegacyNames) || []),
+  ];
+  if (allowedLegacyNames.length === 0) {
+    allowedLegacyNames = null;
+  }
```
leebyron: This merge code could be much simpler and avoid the mutable var:

```js
const schemaAllowedLegacyNames = schema.__allowedLegacyNames
const extendAllowedLegacyNames = options && options.allowedLegacyNames
const allowedLegacyNames = 
  schemaAllowedLegacyNames && extendAllowedLegacyNames
      ? schemaAllowedLegacyNames.concat(extendAllowedLegacyNames)
      : schemaAllowedLegacyNames || extendAllowedLegacyNames
```

src/utilities/extendSchema.js
```diff
@@ -240,13 +240,13 @@ export function extendSchema(
...
-  if (allowedLegacyNames.length === 0) {
-    allowedLegacyNames = null;
-  }
+  // Support both original legacy names and extended legacy names.
+  const schemaAllowedLegacyNames = schema.__allowedLegacyNames;
+  const extendAllowedLegacyNames = options && options.allowedLegacyNames;
+  const allowedLegacyNames = 
+    schemaAllowedLegacyNames && extendAllowedLegacyNames
+      ? schemaAllowedLegacyNames.concat(extendAllowedLegacyNames)
+      : schemaAllowedLegacyNames || extendAllowedLegacyNames;
```
alloy: Nice 👍

PR #1237, opened 2018-02-10
export isValidNameError

gdi2290:
PR mentioned in https://github.com/graphql/graphql-js/pull/1234#issuecomment-364283068


mjmahone:
Thanks you! This seems like a useful export.

IvanGoncharov:
@mjmahone Thanks for merging this super useful function 👍 
One suggestion: If you use "Squash and merge" option it will result in cleaner commit history, 1 vs 4 commits in this particular case.

PR #1240, opened 2018-02-12
Remove duplicate test

Cito:
Same test has been added twice in buildASTSchema-test


IvanGoncharov:
@Cito Good catch 👍
I made classic ctrl-c ctrl-v 🐞

PR #1241, opened 2018-02-12
Export ExecutableDefinitionNode

mjmahone:
Makes it easier to use standard fragment/operation definitions without pulling in schema definitions.


leebyron:
Ah yes, thanks. This should have always been there

PR #1243, opened 2018-02-13
ISSUE_TEMPLATE: Fix 'facebook/graphql' link

IvanGoncharov:
Currently, GFM renders this link as:
`https://github.com/graphql/graphql-js/blob/master/.github/github.com/facebook/graphql`


leebyron:
Thank you!

PR #1244, opened 2018-02-14
Use .mjs for module code

### LINKED ISSUE #1217: Provide .mjs enty for native ESM
This will make it possible for packages dependant on `graphql` to in turn do the same for their consumers, without encountering https://github.com/babel/babel/issues/7294.

[Node.js with `--experimental-modules` can run ESM in .mjs files](https://nodejs.org/api/esm.html#esm_ecmascript_modules).

Essentially, package.json needs to go from this:

```json
{
  "main": "index.js",
  "module": "module/index.js"
}
```

To this:

```json
{
  "main": "index",
  "module": "index.mjs"
}
```

Sibling `index.js` (CJS) and `index.mjs` (ESM) files need to be published. Node.js will load the relevant file, looking up `.mjs` before `.js` for `main` if `--experimental-modules` is enabled.

leebyron:
* Writes .mjs instead of .js for es-module code
* Moves .mjs files out of ./module to top level since they no longer conflict with .js files

Fixes #1217


leebyron:
Looks like we'll also need iterall to support mjs

jaydenseric:
Is there a reason why the `.mjs` files are nested under a `module` directory, instead of being `.js` file siblings? Other than looking tidier, imports like this could then work the same whether or not `--experimental-modules` is enabled:

```js
import { GraphQLError } from 'graphql/error'
```

leebyron:
I've been trying to discourage keying deeply into the package contents, which should be implementation details, but you're right that it's no longer necessary if the files would have different extensions

leebyron:
Just released https://github.com/leebyron/iterall/releases/tag/v1.2.0 which helps unblock this

PR #1247, opened 2018-02-15
Fix orList to not use or and comma between last two elements

danez:
Fixes 

```js
"A", "B", "C", "D", or "E"
```

and remove the last comma which seems misplaced.


leebyron:
This is actually intentional to avoid ambiguity.

There's no hard and fast rule for this sort of thing (human language is flexible) but we've stuck to a Chicago Style Guide style thus far.

See more: https://en.wikipedia.org/wiki/Serial_comma#Recommendations_by_style_guides

danez:
Oh I see, sorry for the noise then.

PR #1250, opened 2018-02-16
Visitor: fix description of "ancestors" arg + test

IvanGoncharov:
This description is incorrect:
https://github.com/graphql/graphql-js/blob/86d33b49a92eda05ee9ca7b3a6fe3a449a21de4c/src/language/visitor.js#L43-L46

Since `ancestors` never includes parent, and it's already tested here:
https://github.com/graphql/graphql-js/blob/86d33b49a92eda05ee9ca7b3a6fe3a449a21de4c/src/language/__tests__/visitor-test.js#L54

But I added a separate test to make it more obvious.

### DIFF
--- a/src/language/__tests__/visitor-test.js
+++ b/src/language/__tests__/visitor-test.js
@@ -74,7 +74,6 @@ describe('Visitor', () => {
         checkVisitorFnArgs(ast, arguments);
         visited.push(['enter', path.slice()]);
       },
-
       leave(node, key, parent, path) {
         checkVisitorFnArgs(ast, arguments);
         visited.push(['leave', path.slice()]);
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@leebyron Can you please review this PR?

IvanGoncharov:
@leebyron Can you please take a look?

src/language/__tests__/visitor-test.js
```diff
@@ -95,6 +94,34 @@ describe('Visitor', () => {
...
 
+  it('validates ancestors argument', () => {
+    const ast = parse('{ a }', { noLocation: true });
+    const nodesInPath = [];
+
+    visit(ast, {
+      enter(node, key, parent, path, ancestors) {
+        const inArray = typeof key === 'number';
+        const expectedAncestors = nodesInPath.slice(0, path.length - 1);
+        expect(ancestors).to.deep.equal(expectedAncestors);
```
mjmahone: Why is this not `expect(ancestors).to.deep.equal(nodesInPath);`? i.e. why is the slicing needed?

IvanGoncharov: @mjmahone Because we need `nodesInPath` without last element and `slice(0, path.length - 1)` does exactly that it returns an array without last element.

Basically, it tests this statement:
> All nodes and Arrays visited before reaching parent of this node

IvanGoncharov: @mjmahone I rewrote this test to make it more explicit.

IvanGoncharov:
@mjmahone It was my pleasure.

I think now `visit` coverage is near 100% plus its interface is fully documented so next step will be to work on simplifying this function as suggested in https://github.com/graphql/graphql-js/pull/1145#pullrequestreview-83641503

PR #1251, opened 2018-02-16
Benchmark task

leebyron:
This adds a benchmark task inspired by #1167 and #1163. Invoke with `yarn benchmark`.

Allows filtering down which benchmark to run, ala jest, and supplying which revisions to run against.

Closes #1167


resources/benchmark.js
```diff
@@ -0,0 +1,148 @@
...
+const beautifyBenchmark = require('beautify-benchmark');
+const { execSync } = require('child_process');
+const os = require('os');
+const path = require('path');
+
+// Like build:cjs, but includes __tests__ and copies other files.
+const BUILD_CMD = 'babel src --optional runtime --copy-files --out-dir dist/';
+const LOCAL = 'local';
+const LOCAL_DIR = path.join(__dirname, '../');
+const TEMP_DIR = os.tmpdir();
```
IvanGoncharov: @leebyron Both #1167 and #1163 were intended to enable benchmarking during development and it really inconvenient if after every small change you need to rebuild `HEAD` 😞

PR #1252, opened 2018-02-17
memoise wrapper values

mohawk2:
In a similar spirit to #1130, and extracted from code written updating #1054: use `WeakMap` to cache `GraphQLList` and `GraphQLNonNull` wrappers for any given type, so that the wrapper function will always return the same instance. This will make type-comparisons able to be quicker, and avoid constructing unnecessary duplicates.

### DIFF
--- a/src/type/wrappers.js
+++ b/src/type/wrappers.js
@@ -34,12 +34,19 @@ declare class GraphQLList<+T: GraphQLType> {
   // Note: constructors cannot be used for covariant types. Drop the "new".
   constructor(ofType: any): void;
 }
+const listCache = new WeakMap();
 // eslint-disable-next-line no-redeclare
 export function GraphQLList(ofType) {
   if (this instanceof GraphQLList) {
     this.ofType = assertType(ofType);
   } else {
-    return new GraphQLList(ofType);
+    const cachedValue = listCache.get(ofType);
+    if (cachedValue !== undefined) {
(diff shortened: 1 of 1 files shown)


leebyron:
Any benchmarking effects from this change, @mohawk2? Conceptually it makes sense, but would love to see real world impact

mohawk2:
@leebyron Surprisingly, the benchmark showed a very very small slowdown. Suspicious, I ran `yarn run testonly` a number of times on each branch which seemed to show the memoised version was faster by a tiny amount.

I suspected that whichever went second on the benchmark got a speed advantage. I then ran `yarn run benchmark --revs local master`, and compared it to `yarn run benchmark --revs master local`. The first one showed roughly similar speed, while the second showed some speedup from memoising.

To summarise the findings:
- the benchmark favours whichever runs second, and should probably drop the first say 3 runs of each one
- memoising shows approximately these effects (+ = faster, - = slower):
  - parsing +6%
  - execute introspection well within margin of variation
  - build from introspection +1% but well within margin of variation
  - build from AST -5%

The slowdown in building a schema from AST is not very surprising since memoising trades startup time against runtime.

My compliments to you and @IvanGoncharov for turning my bodgy prototype benchmark into quite a powerful thing! Two questions:
- would you like me to tweak the benchmark to drop first few runs?
- do you think adding the "profile" option I put together would add value?

PR #1257, opened 2018-02-21
fix: process is undefined when used in browser

milosd92:
When this library is used in browser, global variable `process` is undefined. This issue is related to earlier #1058.

### DIFF
--- a/src/jsutils/instanceOf.js
+++ b/src/jsutils/instanceOf.js
@@ -16,7 +16,8 @@ declare function instanceOf(
   constructor: mixed,
 ): boolean %checks(value instanceof constructor);
 
-export default (process && process.env.NODE_ENV !== 'production'
+export default (typeof process !== 'undefined' &&
+process.env.NODE_ENV !== 'production'
   ? // eslint-disable-next-line no-shadow
     function instanceOf(value: any, constructor: any) {
       if (value instanceof constructor) {


leebyron:
Does this still result in the correct behavior for packagers like Webpack?

Could you share a bit more information about how you're using graphql in the browser? Typically these production environment checks are replaced with `true` or `false` by build pipelines

milosd92:
I am using `print` function from this lib, to convert from query AST.
Yes this is replaced with `true` or `false` but when this code is executed from browser `process` does not exist.

I can add a plunkr to simulate this, if its needed.

taion:
@leebyron Could we get a release with this version?

seanparmelee:
Hi @leebyron @IvanGoncharov, I unfortunately cannot upgrade to v14 yet. Any chance we could back-port this to a 0.13.3 release? Happy to open a PR, but I didn't see a 0.13.x branch to open it against.

IvanGoncharov:
Hi @seanparmelee 
Sorry for the late response.

> Any chance we could back-port this to a 0.13.3 release? 

Currently, we have a very limited resource, so I don't think supporting pre-stable releases make sense.
Since almost all packages use `graphql` 📦 as `peerDependency` you can create your fork (as `0.13.4`) and host it on GitHub.

PR #1261, opened 2018-02-22
Unify extending of type system

IvanGoncharov:
It's a preparation for the #1095

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -184,8 +184,7 @@ export function buildASTSchema(
     },
   );
 
-  const types = typeDefs.map(def => definitionBuilder.buildType(def));
-
+  const types = definitionBuilder.buildTypes(typeDefs);
   const directives = directiveDefs.map(def =>
     definitionBuilder.buildDirective(def),
   );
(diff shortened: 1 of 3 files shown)


src/utilities/extendSchema.js
```diff
@@ -212,9 +181,6 @@ export function extendSchema(
   const extendTypeCache = Object.create(null);
 
   // Get the root Query, Mutation, and Subscription object types.
-  // Note: While this could make early assertions to get the correctly
-  // typed values below, that would throw immediately while type system
-  // validation with validateSchema() will produce more actionable results.
```
IvanGoncharov: This comment is not relevant anymore since they're no `any` cast here.
Moreover, the result of extending object type will always be an object type.

leebyron:
After a cursory read I think this looks great. However I'm going to hold off merging it while I'm investigating #1265

src/utilities/buildASTSchema.js
```diff
@@ -334,11 +339,15 @@ export class ASTDefinitionBuilder {
...
+    const interfaces = def.interfaces;
     return new GraphQLObjectType({
       name: typeName,
       description: getDescription(def, this._options),
       fields: () => this._makeFieldDefMap(def),
-      interfaces: () => this._makeImplementedInterfaces(def),
+      // Note: While this could make early assertions to get the correctly
+      // typed values, that would throw immediately while type system
+      // validation with validateSchema() will produce more actionable results.
+      interfaces: interfaces ? () => (this.buildTypes(interfaces): any) : [],
```
leebyron: Nice to avoid the lambda allocation if the AST defines no interfaces

leebyron:
Nice work

PR #1268, opened 2018-03-04
Execute: Remove excessive arguments

IvanGoncharov:
Using benchmark, I didn't measure any performance difference.

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -731,7 +731,7 @@ export function buildResolveInfo(
   // The resolve function's optional fourth argument is a collection of
   // information about the current execution state.
   return {
-    fieldName: fieldNodes[0].name.value,
+    fieldName: fieldDef.name,
     fieldNodes,
     returnType: fieldDef.type,
     parentType,
@@ -1171,7 +1171,6 @@ function completeObjectValue(
           exeContext,
           returnType,
           fieldNodes,
-          info,
           path,
           result,
         );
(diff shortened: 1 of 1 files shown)


src/execution/execute.js
```diff
@@ -760,16 +757,14 @@ export function resolveFieldValueOrError<TSource>(
     // TODO: find a way to memoize, in case this field is within a List type.
     const args = getArgumentValues(
       fieldDef,
-      fieldNodes[0],
-      exeContext.variableValues,
+      info.fieldNodes[0],
+      info.variableValues,
```
leebyron: Info should be opaque to the executor internals - it's only used to pass into resolvers. These two should be either passed in directly as arguments or pulled from exeContext

src/execution/execute.js
```diff
@@ -840,26 +832,18 @@ function completeValueCatchingError(
...
-      result,
-    );
+    const completed = completeValue(exeContext, returnType, info, path, result);
     if (isPromise(completed)) {
       return completed.then(undefined, error =>
         Promise.reject(
           locatedError(
             asErrorInstance(error),
-            fieldNodes,
+            info.fieldNodes,
```
leebyron: Similarly here - the `fieldNodes` value is an important argument to the `completeValue` functions - it should be passed directly instead of extracted from the info passed to resolvers

IvanGoncharov:
@leebyron I removed all changes that used data from the `info`.
Can you please take a look?

leebyron:
Great!

PR #1269, opened 2018-03-04
Tests: Remove excessive dependencies and async/await

IvanGoncharov:
After seeing async/await on sync test I had an OCD episode 😄
I think I removed ~70% of excessive async/await keywords.


leebyron:
Nice

PR #1276, opened 2018-03-06
Merge wrappers into type/definition to fix circular imports (#1275)

### LINKED ISSUE #1275: Circular import breaks building with Google Closure Compiler
There is a circular dependency between `type/definition.js` and `type/wrappers.js`, which breaks building graphql-js with the Google Closure Compiler (which has no solution to breaking circular imports yet, see [this issue](https://github.com/google/closure-compiler/issues/1883)).

The affected code is located here:

https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L35

and

https://github.com/graphql/graphql-js/blob/master/src/type/wrappers.js#L10-L11

Any chance we could break this cycle? Among other things, it would allow to include graphql-js seamlessly in ClojureScript via its new `:npm-deps` feature.

Jannis:
This fixes #1275.

All existing references to `src/type/wrappers.js` were updated. I also verified that the tests still pass after this change.


leebyron:
Thanks for fixing this! Happy to hear this unblocks use of closure compiler

PR #1278, opened 2018-03-07
Remove redundant Flow type casts

IvanGoncharov:
Not sure I catch all places since I used RegExp:
```sh
grep ' (.*: [A-Za-z0-9]*)[^ ]' -R src/ | grep -v any
```
Would be great if flow itself will produce errors/warnings on redundant casts.


leebyron:
Great!

Most of these mitigated flow issues in older versions, but flow has definitely improved. Very happy to see these go away

PR #1280, opened 2018-03-08
fix #1277 ensure interface has at least 1 concrete type

### LINKED ISSUE #1277: Throw error if no possible concrete types found
While using relay modern v1.5.0, I got the following when running relay-compiler (I promise this is a graphql issue):

```sh
ERROR:
Cannot read property 'some' of undefined
error An unexpected error occurred: "Command failed.
Exit code: 100
```
Note that the error has no stack trace!

The `some` that it mentioned was:
```js
	function hasConcreteTypeThatImplements(schema, type, interfaceName) {
	  return isAbstractType(type) && getConcreteTypes(schema, type).some(function (concreteType) {
	    return implementsInterface(concreteType, interfaceName);
	  });
	}
```

Ultimately, this was caused because when graphql built my schema, it crawled & found the GraphQLInterfaceType, but it didn't find the concrete types that implemented the interface.
This didn't throw an error in GraphQL; however, calling `schema.getPossibleTypes` returned `undefined`, which borked it in the compiler.

While it'd be nice for relay-compiler to catch errors & include a stack trace, I think the real issue is that GraphQL allows an abstract type to exist without any concrete types!

I propose closing this hole by calling schema.getPossibleTypes for each abstract type when building the schema. if the result is `undefined` then throw a friendly error telling the user to add that concrete type to their `rootSchema.types`.

Happy to PR if this sounds like a reasonable plan!

mattkrick:
Changes:

- [ ] `getPossibleTypes` is now guaranteed to return an array, just like its flow typing says it does
- [ ] Validation ensures each interface has at least 1 object type
- [ ] Conforms to Spec, Type Validation, Rule 4: https://github.com/facebook/graphql/pull/424

### DIFF
--- a/src/type/__tests__/schema-test.js
+++ b/src/type/__tests__/schema-test.js
@@ -76,19 +76,6 @@ const Schema = new GraphQLSchema({
 });
 
 describe('Type System: Schema', () => {
-  describe('Getting possible types', () => {
-    it('throws human-reable error if schema.types is not defined', () => {
-      const checkPossible = () => {
-        return Schema.isPossibleType(InterfaceType, ImplementingType);
-      };
-      expect(checkPossible).to.throw(
-        'Could not find possible implementing types for Interface in schema. ' +
-          'Check that schema.types is defined and is an array of all possible ' +
-          'types in the schema.',
-      );
-    });
-  });
-
   describe('Type Map', () => {
     it('includes input types only used in directives', () => {
       expect(Schema.getTypeMap()).to.include.key('DirInput');
--- a/src/type/__tests__/schema-test.js
+++ b/src/type/__tests__/schema-test.js
@@ -23,12 +23,6 @@ const InterfaceType = new GraphQLInterfaceType({
   fields: { fieldName: { type: GraphQLString } },
 });
 
-const ImplementingType = new GraphQLObjectType({
-  name: 'Object',
-  interfaces: [InterfaceType],
-  fields: { fieldName: { type: GraphQLString, resolve: () => '' } },
-});
-
 const DirectiveInputType = new GraphQLInputObjectType({
   name: 'DirInput',
   fields: {
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -8,6 +8,7 @@
  */
 
 import {
+  isAbstractType,
   isObjectType,
   isInterfaceType,
   isUnionType,
@@ -159,6 +160,8 @@ export class GraphQLSchema {
             this._implementations[iface.name] = [type];
           }
         });
+      } else if (isAbstractType(type) && !this._implementations[type.name]) {
+        this._implementations[type.name] = [];
       }
     });
   }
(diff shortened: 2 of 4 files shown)


src/type/validate.js
```diff
@@ -355,6 +358,21 @@ function validateObjectInterfaces(
...
+function validateInterfaces(
+  context: SchemaValidationContext,
+  iface: GraphQLInterfaceType,
+): void {
+  const possibleTypes = context.schema.getPossibleTypes(iface);
+
+  if (possibleTypes.length === 0) {
+    context.reportError(
+      `No concrete types found for Interface type ${iface.name}. ` +
+        `If only referenced via abstraction, add concrete types to schema.types array`,
```
IvanGoncharov: The schema can be create using SDL syntax so `add concrete types to schema.types array` part would be confusing in such case.
`concrete types` - better to avoid inventing new terminology.

Something like this would be great:
`Interface ${iface.name} should be implemented by at least one object type.`
Disclaimer: I'm grammatically challenged 😄

mattkrick: good call, fixed!

mattkrick:
@IvanGoncharov ready for another review!

src/type/schema.js
```diff
@@ -204,7 +207,7 @@ export class GraphQLSchema {
     if (!possibleTypeMap[abstractType.name]) {
       const possibleTypes = this.getPossibleTypes(abstractType);
       invariant(
-        Array.isArray(possibleTypes),
+        Array.isArray(possibleTypes) && possibleTypes.length > 0,
```
leebyron: This is a late checking invariant, not part of the scheme validation. This change shouldn't be necessary

mattkrick: @leebyron this was only necessary for getting the schema tests working

mattkrick: Fully agree. The invariant can go away because `getPossibleTypes` is now guaranteed to return an array.
The test can go away because there's no way `isPossibleType` can produce an error.

src/type/__tests__/schema-test.js
```diff
@@ -76,19 +76,6 @@ const Schema = new GraphQLSchema({
 });
 
 describe('Type System: Schema', () => {
-  describe('Getting possible types', () => {
```
mattkrick: `isPossibleType` can no longer throw, so this test is no longer needed

src/type/schema.js
```diff
@@ -159,6 +160,8 @@ export class GraphQLSchema {
             this._implementations[iface.name] = [type];
           }
         });
+      } else if (isAbstractType(type) && !this._implementations[type.name]) {
+        this._implementations[type.name] = [];
```
mattkrick: guarantees that `getPossibleTypes` returns an array, as it should per its flow type.

mattkrick:
@leebyron that should fix it up

src/type/validate.js
```diff
@@ -355,6 +358,21 @@ function validateObjectInterfaces(
...
 
+function validateInterfaces(
+  context: SchemaValidationContext,
+  iface: GraphQLInterfaceType,
+): void {
+  const possibleTypes = context.schema.getPossibleTypes(iface);
+
+  if (possibleTypes.length === 0) {
+    context.reportError(
+      `Interface ${iface.name} must be implemented` +
```
leebyron: Could you please expand the tests further to include a test which asserts this error occurs?

leebyron:
Great work!

Issue #1279, opened 2018-03-07
GraphQLInterfaceType: TypeError: Cannot read property 'name' of undefined


nicky-lenaers:
I have two types that implement an interface:

**1. project.type.ts**
```ts
import {
  GraphQLObjectType
} from 'graphql';

import { ProjectFields } from './project.fields';
import { ProjectInterfaceType } from './project-interface.type';

export const ProjectType: GraphQLObjectType = new GraphQLObjectType({
  name: 'ProjectType',
  interfaces: [ProjectInterfaceType],
  description: 'GraphQL Type for Project',
  fields: ProjectFields
});
```

**2. project-private.type.ts**
```ts
import {
  GraphQLObjectType
} from 'graphql';

import { ProjectPrivateFields } from './project.fields';
import { ProjectInterfaceType } from './project-interface.type';

export const ProjectPrivateType: GraphQLObjectType = new GraphQLObjectType({
  name: 'ProjectPrivateType',
  interfaces: [ProjectInterfaceType],
  description: 'Project Private Type',
  fields: ProjectPrivateFields
});
```

**project-interface.type.ts**
```ts
import {
  GraphQLInterfaceType
} from 'graphql';

import { ProjectFields, ProjectPrivateFields } from './project.fields';
import { ProjectType } from './project.type';
import { ProjectPrivateType } from './project-private.type';

export const ProjectInterfaceType: GraphQLInterfaceType = new GraphQLInterfaceType({
  name: 'ProjectInterfaceType',
  fields: ProjectFields,
  resolveType: (data, context) => {
    return ProjectType; // <-- I need to conditionally return either ProjectType or ProjectPrivateType
  }
});
```

I also have the types declared in my schema:
```ts
...
export const ProjectSchema = new GraphQLSchema({
  types: [ProjectPrivateType, ProjectType, ProjectInterfaceType],
  ...
...
```

Whenever I change the return type from the `resolveType` from `ProjectType` to `ProjectPrivateType` I get the following error:

```
TypeError: Cannot read property 'name' of undefined
    at ...\node_modules\graphql\type\schema.js:136:52
    at Array.forEach (<anonymous>)
    at ...\node_modules\graphql\type\schema.js:135:30
    at Array.forEach (<anonymous>)
    at new GraphQLSchema (...\node_modules\graphql\type\schema.js:132:32)
    ...
```

## Update
I found out that it seems like some circular reference not being resolved. If I simply return the name of the type as a string, it works:

```ts
import {
[...]

PR #1282, opened 2018-03-08
Throws descriptive error when non-type used instead of interface

IvanGoncharov:
Based on #1279


leebyron:
Thanks for tightening this up

PR #1284, opened 2018-03-08
 formatError: put all extensions inside 'extensions' property

IvanGoncharov:
Mirrors https://github.com/facebook/graphql/pull/407

This is breaking change, however, `extensions` is the relatively new feature that was added in `0.12.0`. 
Also, it could be easily workaround by providing `formatError` property to a server:
https://github.com/graphql/express-graphql#options
https://github.com/apollographql/apollo-server#options

Unintended changes: I didn't want to add `extensions: undefined` to every error inside tests so I refactored them.

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -86,7 +86,7 @@ declare class GraphQLError extends Error {
   /**
    * Extension fields to add to the formatted error.
    */
-  +extensions: ?{ [key: string]: mixed };
+  +extensions: { [key: string]: mixed } | void;
 }
 
 export function GraphQLError( // eslint-disable-line no-redeclare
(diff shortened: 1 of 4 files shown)


IvanGoncharov:
@leebyron Thanks for review 🎉 

> This is great - just some small suggestions inline.

Fixed in 80ca852ebd113ed4c39218244a6e863332912ae8. I tried dozens of ways how to initialize optional non-undefined readonly property and it's the simplest one I found.

> We'll also need some kind of adoption strategy to help users with the breaking change

`JSON.stringify(error)` is not affected.
As for `formatError`, if we assume people read Release notes before updating to the new version than it's easy. You just need to write that in GraphQL server of your choice `express-graphql`, `apollo-server`, etc. you need to supply `formatError` as:
```js
import { formatError as baseFormatError, /* ... */ } from 'graphql';

{
  // other options
  formatError(error) {
    const { extensions, ...rest } = baseFormatError(error);
    return { ...extensions, ...rest };
  },
}
```

leebyron:
Nice - Let's get #1305 landed first and I'll let you rebase over that, otherwise I really like this change.

IvanGoncharov:
@leebyron I rebased this PR on top of master.

leebyron:
Great work on this!

Issue #1291, opened 2018-03-16
Flow 0.68.0 compatibility


stevehollaar:
Flow 0.68 added some changes that introduce some new errors:

```
$> yarn flow
yarn run v1.5.1
$ /Users/stevehollaar/graphql-js/node_modules/.bin/flow
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/error/locatedError.js:25:38

Property path is missing in Error [1].

 [1] 19│   originalError: Error,
     20│   nodes: $ReadOnlyArray<ASTNode>,
     21│   path: $ReadOnlyArray<string | number>,
     22│ ): GraphQLError {
     23│   // Note: this uses a brand-check to support GraphQL errors originating from
     24│   // other contexts.
     25│   if (originalError && Array.isArray(originalError.path)) {
     26│     return (originalError: any);
     27│   }
     28│


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/subscription/mapAsyncIterator.js:25:14

Property return is missing in $AsyncIterator [1].

     src/subscription/mapAsyncIterator.js
      22│   const iterator = getAsyncIterator(iterable);
      23│   let $return;
      24│   let abruptClose;
      25│   if (typeof iterator.return === 'function') {
      26│     $return = iterator.return;
      27│     abruptClose = error => {
      28│       const rethrow = () => Promise.reject(error);

     node_modules/iterall/index.js.flow
 [1] 506│   & (<+TValue>(asyncIterable: AsyncIterable<TValue>) => AsyncIterator<TValue>)


[...]

PR #1292, opened 2018-03-19
locatedError: correct definition to pass Flow 0.68 check

IvanGoncharov:
Not sure how to fix other `$FlowFixMe` introduced in #1291 but this was a low hanging fruit.

### DIFF
--- a/src/error/locatedError.js
+++ b/src/error/locatedError.js
@@ -16,13 +16,12 @@ import type { ASTNode } from '../language/ast';
  * document responsible for the original Error.
  */
 export function locatedError(
-  originalError: Error,
+  originalError: Error | GraphQLError,
   nodes: $ReadOnlyArray<ASTNode>,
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
> Seems weird. GraphQLError should be a subset of Error, so not required to explicitly declare. I'll take a look

@leebyron It's happening because of this change in Flow:
```
Likely to cause new Flow errors:

Previously, Flow would allow you to write if (foo.unknownProp) { ... }.

Now Flow disallows testing unknown properties in conditionals. If foo is a
union type like { x: string } | { y: number }, x and y are known
properties and z would be an unknown property
```

So after this change, you can't test if an instance of `Error` has `path` property.
But the same check works for instances of `GraphQLError`.

leebyron:
Ahhh, yes that makes sense. Thanks for the explanation!

PR #1137, opened 2017-12-12
Better Predicates

leebyron:
Introduces new predicate and assertion functions for each kind of type mirroring the existing ones for the higher order types. It also includes predicates for directives and schema. This should remove the need to use `instanceof` in any usage of GraphQL.js

This consolidates the checking of instances which generalizes the cross-realm/multiple-module check so that the clearer error message is provided in nearly all scenarios, reducing confusion.

This also replaces nearly all `instanceof` with new predicate functions internally.

Fixes #1134
Closes #1016

PR #1297, opened 2018-03-23
Add support for Symbol.toStringTag

nyteshade:
**Changes**
 * Add static and instance property getters for Symbol.toStringTag for
 each exported class.

**Purpose**
Being able to compare class and class instances via internal class type
as per the definition and usage of Symbol.toStringTag allows other
libraries to validate types during runtime in this manner. It also
prevents them from having to patch the live values in their own
codebases.

**Contrived Example**
With no modules and vanilla JavaScript we should be able to do something
like the following.

```javascript
let type = new GraphQLObjectType({name: 'Sample'});

if (({}).toString.call(type) === '[object GraphQLObjectType]') {
  // we have the right type of class
}
```

However, with libraries such as `type-detect` or `ne-types` the code can
look far cleaner.

```javascript
// type-detect
let type = require('type-detect')
let obj = new GraphQLObjectType({name:'Example'})
assert(type(obj) === GraphQLObjectType.name)

// ne-types
let { typeOf } = require('ne-types')
let obj = new GraphQLObjectType({name:'Example'})
assert(typeOf(obj) === GraphQLObjectType.name)
```

There are a lot of libraries out there, despite doing nearly the same
thing in all cases, that support the usage of `Symbol.toStringTag` and
by adding support for that in the base GraphQL classes, all of these
libraries can be used with GraphQL.

_Note: I would have used `get [Symbol.toStringTag]() { ... }` style programming instead, but the linters threw errors about computed types. So I used ES5 style `Object.defineProperty` calls instead._

### DIFF
--- a/src/language/source.js
+++ b/src/language/source.js
@@ -8,6 +8,7 @@
  */
 
 import invariant from '../jsutils/invariant';
+import applyToStringTag from '../jsutils/applyToStringTag';
 
 type Location = {
   line: number,
@@ -41,3 +42,6 @@ export class Source {
     );
   }
 }
+
+// Conditionally apply `[Symbol.toStringTag]` if `Symbol`s are supported
+applyToStringTag(Source);
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -7,6 +7,7 @@
  * @flow strict
  */
 
+import applyToStringTag from '../jsutils/applyToStringTag';
 import instanceOf from '../jsutils/instanceOf';
 import invariant from '../jsutils/invariant';
 import isInvalid from '../jsutils/isInvalid';
@@ -586,6 +587,9 @@ export class GraphQLScalarType {
   inspect: () => string;
 }
 
+// Conditionally apply `[Symbol.toStringTag]` if `Symbol`s are supported
+applyToStringTag(GraphQLScalarType);
+
 // Also provide toJSON and inspect aliases for toString.
 GraphQLScalarType.prototype.toJSON = GraphQLScalarType.prototype.inspect =
   GraphQLScalarType.prototype.toString;
@@ -687,6 +691,9 @@ export class GraphQLObjectType {
   toJSON: () => string;
   inspect: () => string;
 }
+
+// Conditionally apply `[Symbol.toStringTag]` if `Symbol`s are supported
+applyToStringTag(GraphQLObjectType);
 
 // Also provide toJSON and inspect aliases for toString.
 GraphQLObjectType.prototype.toJSON = GraphQLObjectType.prototype.inspect =
@@ -937,6 +944,9 @@ export class GraphQLInterfaceType {
   inspect: () => string;
 }
 
+// Conditionally apply `[Symbol.toStringTag]` if `Symbol`s are supported
+applyToStringTag(GraphQLInterfaceType);
+
 // Also provide toJSON and inspect aliases for toString.
 GraphQLInterfaceType.prototype.toJSON = GraphQLInterfaceType.prototype.inspect =
   GraphQLInterfaceType.prototype.toString;
@@ -1015,6 +1025,9 @@ export class GraphQLUnionType {
   toJSON: () => string;
   inspect: () => string;
 }
+
+// Conditionally apply `[Symbol.toStringTag]` if `Symbol`s are supported
+applyToStringTag(GraphQLUnionType);
 
 // Also provide toJSON and inspect aliases for toString.
 GraphQLUnionType.prototype.toJSON = GraphQLUnionType.prototype.inspect =
@@ -1130,6 +1143,9 @@ export class GraphQLEnumType /* <T> */ {
   toJSON: () => string;
   inspect: () => string;
 }
+
+// Conditionally apply `[Symbol.toStringTag]` if `Symbol`s are supported
+applyToStringTag(GraphQLEnumType);
 
 // Also provide toJSON and inspect aliases for toString.
 GraphQLEnumType.prototype.toJSON = GraphQLEnumType.prototype.inspect =
@@ -1263,6 +1279,9 @@ export class GraphQLInputObjectType {
   toJSON: () => string;
   inspect: () => string;
 }
+
+// Conditionally apply `[Symbol.toStringTag]` if `Symbol`s are supported
+applyToStringTag(GraphQLInputObjectType);
 
 // Also provide toJSON and inspect aliases for toString.
 GraphQLInputObjectType.prototype.toJSON =
--- a/src/jsutils/applyToStringTag.js
+++ b/src/jsutils/applyToStringTag.js
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+/**
+ * The `applyToStringTag()` function checks first to see if the runtime
+ * supports the `Symbol` class and then if the `Symbol.toStringTag` constant
+ * is defined as a `Symbol` instance. If both conditions are met, the
+ * Symbol.toStringTag property is defined as a getter that returns the
+ * supplied class constructor's name.
+ *
+ * @method applyToStringTag
+ *
+ * @param {Class<*>} classObject a class such as Object, String, Number but
+ * typically one of your own creation through the class keyword; `class A {}`,
+ * for example.
+ */
+export function applyToStringTag(classObject: Class<*>): void {
+  const symbolType: string = typeof Symbol;
+  const toStringTagType: string = typeof Symbol.toStringTag;
+
+  if (symbolType === 'function' && toStringTagType === 'symbol') {
+    Object.defineProperty(classObject.prototype, Symbol.toStringTag, {
+      get() {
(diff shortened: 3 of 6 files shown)


nyteshade:
@facebook-github-bot I've sent a few cla@fb.com emails but your message persists

IvanGoncharov:
@nyteshade Thanks for PR 👍
The similar issue was previously discussed here: https://github.com/graphql/graphql-js/issues/491#issuecomment-356103649
Problem is that `graphql-js` mostly used as one of `peerDependencies` so there is a very high risk that developer accidentally uses multiple instances of this lib and checks that depend on a class name is not reliable in such case.

Moreover, many popular tools mangle class names (e.g. UglifyJS) this will introduce even more confusing when `typeOf` will return two different strings for two enums created by different copies of this lib. We had this problem previously see #1174.

So the safest approach is to use predicate functions like `isDirective`, `isType`, etc. See #1137 for the complete list.

nyteshade:
@IvanGoncharov 

Thanks for your time and for responding. I understand that need to try to protect engineers from the community from stubbing their toes on common pitfalls; such as having multiple versions of `graphql` installed in their projects.

Whenever we write libraries for the public we often have to toe the line of being opinionated or not when it it comes to how they should use the code we provide. I would argue that the PR should still be allowed in for a few reasons.

- Those folks who use `Object.prototype.toString.call/apply` and/or `Symbol.toStringTag` are usually aware of issues that can crop up from having multiple instances of library installed and choose to continue regardless. 
- Recent versions of `yarn` and `npm` are good at flattening the node_modules directory in such a manner as to prevent multiple installations as much as possible.
- Finally, you are absolutely correct that certain tools such as Uglify Js do indeed mung the names of things but it only really matters if the compilation tool used in this library does so and that can be controlled.
- Providing the JavaScript community to use the tools to available to them and the coding styles their teams and companies prefer

Additionally, something like this could be added to the graphql-js page.

**GraphQL JS and Symbol.toStringTag**
We are aware that some engineers prefer to do type detection that relies on Symbol.toStringTag. We at Facebook want to provide ways to do that with the graphql-js library. In doing so, we would like to make you aware of a common pitfall in doing so. If you have multiple versions of GraphQL installed, checking for specific instances can cause problems. You are forewarned.

IvanGoncharov:
@nyteshade Nice set of arguments 👍
I think to support `Symbol.toStringTag` has it's pros and cons so it should be an architectural decision by @leebyron 
Also, it would be great if you could provide a code example that would benefit from `Symbol.toStringTag` vs library specific predicates (`isDirective`, `isType`, ...).

src/language/source.js
```diff
@@ -41,3 +41,15 @@ export class Source {
     );
   }
 }
+
+Object.defineProperty(Source, Symbol.toStringTag, {
```
leebyron: Why add this to the constructor itself? Shouldn't only instances be encountered? Also, wouldn't this produce confusing results? I would expect a constructor function to toString as a function

src/language/source.js
```diff
@@ -41,3 +41,15 @@ export class Source {
...
+  get() {
+    return this.name;
+  },
+});
+
+Object.defineProperty(Source.prototype, Symbol.toStringTag, {
```
leebyron: Should this only occur in a `if (typeof Symbol === 'function' && Symbol.toStringTag)` check? Otherwise if embedded in an older node or browser context, this could throw

src/type/definition.js
```diff
@@ -585,6 +585,18 @@ export class GraphQLScalarType {
...
+  get() {
+    return this.name;
+  },
+});
+
+Object.defineProperty(GraphQLScalarType.prototype, Symbol.toStringTag, {
+  get() {
+    return this.constructor.name;
+  },
+});
```
leebyron: I'm noticing this implementation is identical for all types. Perhaps a function in jsutils to reduce code duplication?

nyteshade:
Thanks @leebyron, I'll make the requested changes and update the PR.

IvanGoncharov:
Hi @nyteshade,

Do you need any help with this PR?
We are planning to release `14.0.0 and it would be great to include this PR.

IvanGoncharov:
> I'll clean it up and resubmit for review in the next 48 hours. When is 14 slated for release?

No need to hurry there is no fixed date at the moment, I just reviewing issues/PRs for low-hanging fruits.
Some context here: https://github.com/graphql/graphql-js/pull/1346

nyteshade:
Off to bed, but, @IvanGoncharov, @leebyron, I'm hoping these changes are more towards everyones' liking. I'll check back later and see if coveralls ever finished.

IvanGoncharov:
@nyteshade Can you please do `git rebase` on this PR?
Since it's very hard to review: https://github.com/graphql/graphql-js/pull/1297/files

> I'll check back later and see if coveralls ever finished.

It looks like I broke coveralls 😞  I will try to figure out whats wrong.

src/jsutils/__tests__/symbolSupport-test.js
```diff
@@ -0,0 +1,58 @@
...
+
+describe('symbolSupportTests', () => {
+  // NOTE Symbol appeared in nodejs in 0.12.18 but was largely unusable in
+  // that format. Symbol.toStringTag showed up in node 6.4.0, but was available
+  // behind a flag as early as 4.9.1.
+  const [, major, minor, patch] = /^v(\d+)\.?(\d+)?\.?(\d+)?/.exec(
+    process.version,
+  );
+
+  it('should have Symbol in scope if the version >= 4.9.1', () => {
```
IvanGoncharov: @nyteshade We drop support for Node 4 in master

package-lock.json
```diff
@@ -0,0 +1,8370 @@
+{
```
IvanGoncharov: Can you please remove this file?

src/jsutils/__tests__/symbolSupport-test.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    if (minor === 9) {
+      expect(patch >= 1).to.equal(true);
+    }
+
+    expect(hasSymbolSupport()).to.equal(true);
+  });
+
+  it('should have Symbol in scope if the version >= 6.4.0', () => {
```
IvanGoncharov: I checked and looks like it supported since `6.0.0`:
https://node.green/#ES2015-built-ins-well-known-symbols-Symbol-toStringTag
image

------------------------
Given that `package.json` requires `>=6.0.0`:
https://github.com/graphql/graphql-js/blob/0a30b62964efe16e3e87b0e8fea40bc9665f9393/package.json#L21
Please remove two above tests.

src/jsutils/symbolSupport.js
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @param {string}  specificSymbol an optional name of a property on the
+ * `Symbol` class itself that statically refers to a predefined `Symbol`. If
+ * properly specified and the value is set, then true will be returned.
+ * @return {bool} true if `Symbol` is both defined and a function. Optionally
+ * true if the `Symbol` class is true, a predefined symbol name such as
+ * `toStringTag` is set on the `Symbol` class and it maps to an instance of
+ * `Symbol`. False in all other cases
+ */
+export function hasSymbolSupport(specificSymbol?: string): boolean {
```
IvanGoncharov: Since `hasSymbolSupport` was used only by test that would be removed you don't need to export this function anymore.
This allows making `applyToStringTag` default export similar to all other `jsutils/*` files.
\+ it would make sense to rename this file into `applyToStringTag.js`

src/jsutils/symbolSupport.js
```diff
@@ -0,0 +1,80 @@
...
+    // necessary to appease the lint and flowtype gods.
+    //
+    // ((typeof Symbol[specificSymbol]): string) !== 'symbol' makes lint angry
+    // and typeof Symbol[specificSymbol] !== 'symbol' makes flow angry
+    //
+    // The former thinks everything is too verbose the later thinks I am
+    // comparing Symbol instance rather than the string resulting from a call
+    // to typeof.
+    //
+    // Le sigh....
```
IvanGoncharov: I think it's premature generalization to create `hasSymbolSupport` since inside `applyToStringTag` you can write:
```js
if (typeof Symbol === 'function' && typeof Symbol.toStringTag === 'symbol') {
  Object.defineProperty(...)
}
```
\+ it passes Flow

nyteshade:
@IvanGoncharov; your requested changes have been made.

src/jsutils/symbolSupport.js
nyteshade: This doesn't work exactly as you've typed but a variant in the PR does what you've asked for. I believe the issue here has some fundamental conflict with using `typeof` with `Symbol`.

IvanGoncharov:
@nyteshade Thanks 👍 Last issue from me, please remove `package-lock.json`:
image

nyteshade:
@IvanGoncharov Whoops! Forgot to remove it. Done ✅

nyteshade:
@leebyron ?

IvanGoncharov:
@mjmahone I think all @leebyron comments were address and this PR is ready to be merged. Can you please review it?

mjmahone:
Yup it looks good. Thank you @nyteshade !

nyteshade:
My pleasure @mjmahone

PR #1300, opened 2018-03-28
Add undefined export definitions to make rollup happy

### LINKED ISSUE #1293: Compile issue with Rollup
Looks like the way index.js handles it's exports rolllup can't find it 

{ Error: 'GraphQLObjectType' is not exported by node_modules/graphql/type/index.js
    at error (/Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:168:15)
    at Module.error (/Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:17602:9)
    at handleMissingExport (/Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:17863:12)
    at /Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:19512:56
    at Array.reduce (native)
    at Graph.handleMissingExport (/Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:19511:27)
    at Module.traceExport (/Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:17720:28)
    at Module.trace (/Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:17695:43)
    at ModuleScope.findVariable (/Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:14890:28)
    at Scope.findVariable (/Users/Corey/Documents/workspace/mkr/node_modules/rollup/dist/rollup.js:11840:68)
  code: 'MISSING_EXPORT',
  url: 'https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module',
  pos: 1383,
  loc:
   { file: '/Users/Corey/Documents/workspace/mkr/node_modules/graphql/index.mjs',
     line: 41,
     column: 19 },
  frame: '39: export { GraphQLSchema,\n40: // Definitions\n41: GraphQLScalarType, GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType, GraphQLEnumType, GraphQLInputObjectType, GraphQLList, GraphQLNonNull, GraphQLDirective,\n                       ^\n42: // "Enum" of Type Kinds\n43: TypeKind,' }

[ ERROR ]  bundling: node_modules/graphql/index.mjs, line: 41
           'GraphQLObjectType' is not exported by node_modules/graphql/type/index.js

     L40:  // Definitions
     L41:  GraphQLScalarType, GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType, GraphQLEnumType, GraphQLInputOb
     L42:  // "Enum" of Type Kinds

gaberudy:
To fix #1293


acoreyj:
Can confirm this fixes my issue with using graphql and rollup (as part of Stenciljs project)

gaberudy:
@acoreyj I'm curious if you also had this issue (which I can't tell whether its a problem with graphql, rollup or the rollup-commonjs-plugin):

https://github.com/rollup/rollup/issues/1859

acoreyj:
@gaberudy

Hmm no I haven't had this issue in my project

gaberudy:
@leebyron that may be the case, but the transpiling is incorrect in this case.

For some reason, the validation sub-module gets correct codegen (that's what I used as a template to fix the other sub-modules).

My only guess is that it pulls in a default import, which may anchor the transpiler to enumerate in that scope all the export variables instead of doing just the Object.defineProperty on them.

I suspect that:

https://github.com/graphql/graphql-js/blob/920fb874bb0da804529095e25c4a7f1e17efb269/src/validation/index.js#L13

Maybe be responsible for having the correct transpiled code. Not sure if there is a way to fake this in the other sub-modules..

leebyron:
I'm not sure I understand why the compiled output is incorrect - do you have a test case that fails when loading it in? In the `node` repl I can require the module and submodules and get access to all exported values.

I believe Babel adds additional code to the validation submodule due to require statements being interleaved in the result. The additional code just initializes the export as `undefined` before the call to `Object.defineProperty` in case any interleaved require statement happens to cause a circular require, then the property will be defined on the exported object. The submodules that don't have this pattern don't have that additional code since the property cannot be accessed before `Object.defineProperty`. If I recall, Babel v7 changed that behavior slightly - so we may expect the compiled output to change when adopting Babel v7, however the runtime behavior should be identical.

gaberudy:
Yea, I see it's a what you are saying ,and man, your knowledge of Babel internal transpiling behavior is impressive.

Anyway, this shouldn't be accepted, and hopefully the rollup upstream maintainers see the shiny new way of the `mjs` side-by-side modules definitions and then dealing with the transpiled files becomes mute.

PR #1305, opened 2018-04-05
Error: make 'path' and 'locations' enumerable only if present

IvanGoncharov:
Also cleanup tests from 'undefined' and 'containSubset'.

@leebyron I split up non-breaking changes from #1284 
I think they are useful by itself since they make tests more readable.

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -151,7 +151,7 @@ export function GraphQLError( // eslint-disable-line no-redeclare
       // By being enumerable, JSON.stringify will include `locations` in the
       // resulting output. This ensures that the simplest possible GraphQL
       // service adheres to the spec.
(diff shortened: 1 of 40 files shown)


leebyron:
This looks great - do you mind rebasing and looking at the tests? Looks like the build is failing with test failures

IvanGoncharov:
@leebyron Fixed. To make validation tests less fragile I made them expect a subset of errors.

leebyron:
That seems okay, but I'm not sure that I would agree that makes them less fragile - it just means they're more likely to pass when really there is an error.

Ideally these should detect the exact set of schema validation errors and we should write the test so that each test introduces only the validation error it expects to catch

leebyron:
Going to merge this, but we should still try to improve those tests

PR #1306, opened 2018-04-05
Replace containSubset with native chai checks

IvanGoncharov:
`containSubset` plugin used in many places where native `chai` checks would be enough.

### DIFF
--- a/src/language/__tests__/lexer-test.js
+++ b/src/language/__tests__/lexer-test.js
@@ -35,7 +35,7 @@ describe('Lexer', () => {
   });
 
   it('accepts BOM header', () => {
-    expect(lexOne('\uFEFF foo')).to.containSubset({
+    expect(lexOne('\uFEFF foo')).to.contain({
       kind: TokenKind.NAME,
       start: 2,
       end: 5,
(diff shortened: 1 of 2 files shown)


leebyron:
Very nice. Seems like soon we can get rid of the dev dependency altogether?

IvanGoncharov:
> Seems like soon we can get rid of the dev dependency altogether?

@leebyron Would be great plus it we can remove [mocha-bootload](https://github.com/graphql/graphql-js/blob/master/resources/mocha-bootload.js) and simplify test scripts inside `package.json`.

But not possible yet, it still required in the parser tests.
Since if you try to match AST you will have a problem with `loc` properties.

One of the possible solutions is to use the fact that all `Location` instances have `toJSON`. So I was thinking about adding `toSnapshot` utility function that recursively walks through entire object and calls `toJSON` on all objects that have it. 
What do you think?

leebyron:
I'm not sure we should do that if the only value is simplifying the test infra since it just moves complexity rather than eliminates it - but it's certainly worth trying to see if the final result is much simpler.

PR #1307, opened 2018-04-06
Match exact set of errors in schema validation tests

IvanGoncharov:
> Ideally these should detect the exact set of schema validation errors and we should write the test so that each test introduces only the validation error it expects to catch

@leebyron Done

### DIFF
--- a/src/type/__tests__/validation-test.js
+++ b/src/type/__tests__/validation-test.js
@@ -33,12 +33,12 @@ const SomeScalarType = new GraphQLScalarType({
 
 const SomeInterfaceType = new GraphQLInterfaceType({
   name: 'SomeInterface',
-  fields: { f: { type: GraphQLString } },
+  fields: () => ({ f: { type: SomeObjectType } }),
 });
 
 const SomeObjectType = new GraphQLObjectType({
   name: 'SomeObject',
-  fields: { f: { type: GraphQLString } },
+  fields: () => ({ f: { type: SomeObjectType } }),
   interfaces: [SomeInterfaceType],
 });
 
@@ -77,7 +77,7 @@ const outputTypes = withModifiers([
   SomeInterfaceType,
 ]);
 
-const notOutputTypes = withModifiers([SomeInputObjectType]).concat(Number);
+const notOutputTypes = withModifiers([SomeInputObjectType]);
 
 const inputTypes = withModifiers([
   GraphQLString,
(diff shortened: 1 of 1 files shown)


src/type/__tests__/validation-test.js
```diff
@@ -991,21 +1079,47 @@ describe('Type System: Interface fields must have output types', () => {
...
   });
 
+  it('rejects a non-type value as an Interface field type', () => {
+    const schema = schemaWithInterfaceFieldOfType(Number);
+    expect(validateSchema(schema)).to.deep.equal([
+      {
+        message: `The type of BadInterface.badField must be Output Type but got: ${Number}.`,
+      },
+      {
+        message: `Expected GraphQL named type but got: ${Number}.`,
```
leebyron: This one feels low value and duplicative of the error below it, don't you think?

src/type/__tests__/validation-test.js
```diff
@@ -1078,21 +1205,37 @@ describe('Type System: Field arguments must have input types', () => {
...
   });
 
+  it('rejects a non-type value as a field arg type', () => {
+    const schema = schemaWithArgOfType(Number);
+    expect(validateSchema(schema)).to.deep.equal([
+      {
+        message: `The type of BadObject.badField(badArg:) must be Input Type but got: ${Number}.`,
+      },
+      {
+        message: `Expected GraphQL named type but got: ${Number}.`,
```
leebyron: Another example. This is an opportunity to improve the quality of these errors in the future

src/type/__tests__/validation-test.js
```diff
@@ -1144,21 +1289,37 @@ describe('Type System: Input Object fields must have input types', () => {
...
   });
 
+  it('rejects a non-type value as an input field type', () => {
+    const schema = schemaWithInputFieldOfType(Number);
+    expect(validateSchema(schema)).to.deep.equal([
+      {
+        message: `The type of BadInputObject.badField must be Input Type but got: ${Number}.`,
+      },
+      {
+        message: `Expected GraphQL named type but got: ${Number}.`,
```
leebyron: Same

src/type/__tests__/validation-test.js
IvanGoncharov: @leebyron Agree 👍 
However I don't see how `validate` can distinguish beetween those two cases:
```js
validateSchema(new GraphQLSchema({
  queryType: Number,
}));
```
```js
validateSchema(new GraphQLSchema({
  queryType: new GraphQLObjectType({
    name: 'Foo',
    fields: { bar: { type: GraphQLString } },
  }),
  types: [Number],
}));
```

In the first case `Expected GraphQL named type but got: ${Number}.` is redundant but in second one it's the only error that would be reported.
Can you suggest how to solve this?

PR #1310, opened 2018-04-07
Test: Cleanup unneeded returns and async/await

IvanGoncharov:


leebyron:
Thanks for cleaning this up a bit!

PR #1311, opened 2018-04-09
Make "nonnull-test.js" more explicit

IvanGoncharov:
It was really hard to read those tests.
Moreover, explicit version is shorter and this PR increase line count only because I added `path` to every error.

### DIFF
--- a/src/execution/__tests__/nonnull-test.js
+++ b/src/execution/__tests__/nonnull-test.js
@@ -568,7 +568,7 @@ describe('Execute: handles non-nullable types', () => {
                 type: GraphQLNonNull(GraphQLString),
               },
             },
-            resolve: async (_, args) => {
+            resolve: (_, args) => {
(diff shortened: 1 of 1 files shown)


leebyron:
Looks like this has merge conflicts and test failures - want to take another look, @IvanGoncharov ?

IvanGoncharov:
@leebyron Rebased and fixed. The reason was that #1286 changed the order of errors in a promise-heavy test. Not a problem since the order of error is implementation dependent. Maybe it worth to sort them, somehow?

It went unnoticed on master since `containSubset` doesn't check the order of elements in arrays.

leebyron:
Aha, I'm glad we're migrating away from containSubset - the array order thing is not what I would have suspected

leebyron:
Thanks for cleaning this up! It's much easier to follow now and I'm happy to see `path` included in the test expectations

PR #1317, opened 2018-04-20
Cleanup contain subset

IvanGoncharov:
The only `containSubset` usage that can't be replaced with standard chai is the ability to compare ASTs. So I created `locationsToJSON` helper function which is calling `toJSON` on every instance of `Loc`.

Simplier alternative would be `JSON.parse(JSON.stringify(ast))` but it will remove all `undefined` from payload.

### DIFF
--- a/src/language/__tests__/parser-test.js
+++ b/src/language/__tests__/parser-test.js
@@ -5,23 +5,21 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import { inspect } from 'util';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
 import { Kind } from '../kinds';
 import { expect } from 'chai';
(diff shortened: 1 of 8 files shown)


src/language/__tests__/parser-test.js
```diff
@@ -127,31 +124,15 @@ describe('Parser', () => {
...
-        },
-      ],
-    });
+    const ast = parse(`
+      # This comment has a \u0A0A multi-byte character.
+      { field(arg: "Has a \u0A0A multi-byte character.") }
+    `);
+
+    expect(ast).to.have.nested.property(
+      'definitions[0].selectionSet.selections[0].arguments[0].value.value',
```
leebyron: nice, I didn't know about this!

PR #1318, opened 2018-04-21
Execute: use consistent name for execution context

IvanGoncharov:
It helps to easily distinguish between user-provided context and execution context.

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -381,7 +381,6 @@ function executeOperation(
   operation: OperationDefinitionNode,
   rootValue: mixed,
 ): MaybePromise<ObjMap<mixed> | null> {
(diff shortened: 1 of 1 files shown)


src/execution/execute.js
```diff
@@ -378,9 +378,9 @@ export function buildExecutionContext(
  */
 function executeOperation(
   exeContext: ExecutionContext,
-  operation: OperationDefinitionNode,
```
leebyron: The existing function signature was easier to understand - `executeOperation()` should accept an `operation`.

leebyron:
Similarly, user-provided context should always appear as `contextValue` - I don't think that consistency is there

leebyron:
I merged in the contextValue change as well :)

PR #1319, opened 2018-04-21
Update to flow(0.70), prettier(1.12) and other dependencies

IvanGoncharov:


leebyron:
Thanks for bumping these

PR #1322, opened 2018-04-22
[WIP] Extend the extend functionality

mgrenier:
This is a WIP implementation for extending enum, input and union. See discussion in issue #1095

### DIFF
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -847,10 +847,19 @@ describe('extendSchema', () => {
       extend enum SomeEnum {
         NEW_ENUM
       }
+
+      directive @test(arg: SomeEnum) on FIELD
     `);
     expect(extendedSchema).to.not.equal(testSchema);
-    expect(printSchema(extendedSchema)).to.contain('NEW_ENUM');
-    expect(printSchema(testSchema)).to.not.contain('NEW_ENUM');
+
+    const oldEnum = testSchema.getType('SomeEnum');
+    const newEnum = extendedSchema.getType('SomeEnum');
+    const testDirective = extendedSchema.getDirective('test');
+
+    expect(oldEnum._values.length).to.equal(2);
+    expect(newEnum._values.length).to.equal(3);
+    expect(newEnum._values[2].name).to.equal('NEW_ENUM');
+    expect(newEnum).to.equal(testDirective.args[0].type);
   });
 
   it('extend input', () => {
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -21,6 +21,7 @@ import {
   GraphQLID,
   GraphQLString,
   GraphQLEnumType,
+  GraphQLInputObjectType,
   GraphQLNonNull,
   GraphQLList,
   isScalarType,
@@ -77,6 +78,13 @@ const SomeEnumType = new GraphQLEnumType({
   },
 });
 
+const SomeInputType = new GraphQLInputObjectType({
+  name: 'SomeInput',
+  fields: () => ({
+    fooArg: { type: GraphQLString },
+  }),
+});
+
 const testSchema = new GraphQLSchema({
   query: new GraphQLObjectType({
     name: 'Query',
@@ -87,6 +95,10 @@ const testSchema = new GraphQLSchema({
       someInterface: {
         args: { id: { type: GraphQLNonNull(GraphQLID) } },
         type: SomeInterfaceType,
+      },
+      someInput: {
+        args: { input: { type: SomeInputType } },
+        type: GraphQLString,
       },
     }),
   }),
@@ -195,6 +207,52 @@ describe('extendSchema', () => {
         name: String
         some: SomeInterface
         tree: [Foo]!
+        newField: String
+      }
+    `);
+  });
+
+  it('extends enums by adding new values', () => {
+    const extendedSchema = extendTestSchema(`
+      extend enum SomeEnum {
+        NEW_ENUM
+      }
+    `);
+    expect(printTestSchemaChanges(extendedSchema)).to.equal(dedent`
+      enum SomeEnum {
+        ONE
+        TWO
+        NEW_ENUM
+      }
+    `);
+  });
+
+  it('extends unions by adding new types', () => {
+    const extendedSchema = extendTestSchema(`
+      extend union SomeUnion = TestNewType
+
+      type TestNewType {
+        foo: String
+      }
+    `);
+    expect(printTestSchemaChanges(extendedSchema)).to.equal(dedent`
+      union SomeUnion = Foo | Biz | TestNewType
+
+      type TestNewType {
+        foo: String
+      }
+    `);
+  });
+
+  it('extends inputs by adding new fields', () => {
+    const extendedSchema = extendTestSchema(`
+      extend input SomeInput {
+        newField: String
+      }
+    `);
+    expect(printTestSchemaChanges(extendedSchema)).to.equal(dedent`
+      input SomeInput {
+        fooArg: String
         newField: String
       }
     `);
@@ -314,11 +372,21 @@ describe('extendSchema', () => {
       extend type Foo {
         deprecatedField: String @deprecated(reason: "not used anymore")
       }
+
+      extend enum SomeEnum {
+        DEPRECATED @deprecated(reason: "do not use")
+      }
     `);
     const deprecatedFieldDef = extendedSchema.getType('Foo').getFields()
       .deprecatedField;
     expect(deprecatedFieldDef.isDeprecated).to.equal(true);
     expect(deprecatedFieldDef.deprecationReason).to.equal('not used anymore');
+
+    const deprecatedEnumDef = extendedSchema
+      .getType('SomeEnum')
+      .getValue('DEPRECATED');
+    expect(deprecatedEnumDef.isDeprecated).to.equal(true);
+    expect(deprecatedEnumDef.deprecationReason).to.equal('do not use');
   });
 
   it('extends objects by adding new unused types', () => {
@@ -326,12 +394,32 @@ describe('extendSchema', () => {
       type Unused {
         someField: String
       }
+
+      enum UnusedEnum {
+        SOME
+      }
+
+      input UnusedInput {
+        someInput: String
+      }
+
+      union UnusedUnion = Unused
     `);
     expect(extendedSchema).to.not.equal(testSchema);
     expect(printTestSchemaChanges(extendedSchema)).to.equal(dedent`
       type Unused {
         someField: String
       }
+
+      enum UnusedEnum {
+        SOME
+      }
+
+      input UnusedInput {
+        someInput: String
+      }
+
+      union UnusedUnion = Unused
     `);
   });
 
@@ -507,6 +595,34 @@ describe('extendSchema', () => {
       interface NewInterface {
         buzz: String
       }
+
+      extend enum SomeEnum {
+        THREE
+      }
+
+      extend enum SomeEnum {
+        FOUR
+      }
+
+      extend union SomeUnion = Boo
+
+      extend union SomeUnion = Joo
+
+      type Boo {
+        fieldA: String
+      }
+
+      type Joo {
+        fieldB: String
+      }
+
+      extend input SomeInput {
+        fieldA: String
+      }
+
+      extend input SomeInput {
+        fieldB: String
+      }
     `);
     expect(printTestSchemaChanges(extendedSchema)).to.equal(dedent`
       type Biz implements NewInterface & SomeInterface {
@@ -518,9 +634,32 @@ describe('extendSchema', () => {
         newFieldB: Float
       }
 
+      type Boo {
+        fieldA: String
+      }
+
+      type Joo {
+        fieldB: String
+      }
+
       interface NewInterface {
         buzz: String
       }
+
+      enum SomeEnum {
+        ONE
+        TWO
+        THREE
+        FOUR
+      }
+
+      input SomeInput {
+        fooArg: String
+        fieldA: String
+        fieldB: String
+      }
+
+      union SomeUnion = Foo | Biz | Boo | Joo
     `);
   });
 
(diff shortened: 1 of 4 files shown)


src/utilities/extendSchema.js
```diff
@@ -28,6 +30,10 @@ import {
   GraphQLObjectType,
   GraphQLInterfaceType,
   GraphQLUnionType,
+  isEnumType,
+  GraphQLEnumType,
+  isInputObjectType,
+  GraphQLInputObjectType,
```
IvanGoncharov: Please put function imports together with other function imports.

src/utilities/extendSchema.js
```diff
@@ -258,10 +264,103 @@ export function extendSchema(
...
   }
 
+  function extendInputObjectType(
+    type: GraphQLInputObjectType,
+  ): GraphQLInputObjectType {
+    return new GraphQLInputObjectType({
+      name: type.name,
+      description: type.description,
+      fields: () => extendInputFieldMap(type),
+      astNode: type.astNode,
```
IvanGoncharov: You need to add `extensionASTNodes` here. But first you need to add it into `GraphQLInputObjectType` class.

src/utilities/extendSchema.js
```diff
@@ -258,10 +264,103 @@ export function extendSchema(
...
+    }
+
+    return newFieldMap;
+  }
+
+  function extendEnumType(type: GraphQLEnumType): GraphQLEnumType {
+    return new GraphQLEnumType({
+      name: type.name,
+      values: extendValueMap(type),
+      astNode: type.astNode,
```
IvanGoncharov: Also missing `extensionASTNodes`.

src/utilities/extendSchema.js
```diff
@@ -300,10 +411,25 @@ export function extendSchema(
...
+      });
+    }
+
     return new GraphQLUnionType({
       name: type.name,
       description: type.description,
-      types: type.getTypes().map(getExtendedType),
+      types: types,
       astNode: type.astNode,
       resolveType: type.resolveType,
```
IvanGoncharov: Also missing `extensionASTNodes`.

src/utilities/extendSchema.js
```diff
@@ -280,6 +379,18 @@ export function extendSchema(
     });
   }
 
+  function extendArgsMap(
+    args: GraphQLArgument[],
```
IvanGoncharov: It accepts and returns an array so it probably shouldn't have `Map` in its name.

src/utilities/extendSchema.js
```diff
@@ -280,6 +379,18 @@ export function extendSchema(
...
+  ): GraphQLArgument[] {
+    return args.map(arg => ({
```
IvanGoncharov: I think it's better to use `keyValMap` here so the result of this function can be assigned to `GraphQLFieldConfig::args` without additional manipulations.

src/utilities/extendSchema.js
```diff
@@ -258,10 +264,103 @@ export function extendSchema(
...
+      values: extendValueMap(type),
+      astNode: type.astNode,
+    });
+  }
+
+  function extendValueMap(type: GraphQLEnumType) {
+    const oldValueMap = Object.create(null);
+    const newValueMap = Object.create(null);
+    type.getValues().forEach(value => {
+      newValueMap[value.name] = oldValueMap[value.name] = {
```
IvanGoncharov: I think code here should match `extendFieldMap`:
https://github.com/graphql/graphql-js/blob/3cb450059d58f45dff95867a0674d7588be64933/src/utilities/extendSchema.js#L335-L336
So you either change `extendFieldMap` to do the same code or remove assigment chain here.

src/utilities/extendSchema.js
```diff
@@ -258,10 +264,103 @@ export function extendSchema(
...
+        extension.fields.forEach(field => {
+          const fieldName = field.name.value;
+          if (oldFieldMap[fieldName]) {
+            throw new GraphQLError(
+              `Field "${type.name}.${fieldName}" already exists in the ` +
+                'schema. It cannot also be defined in this type extension.',
+              [field],
+            );
+          }
+          newFieldMap[fieldName] = astBuilder.buildField(field);
```
IvanGoncharov: You can't use `buildField` here since it return type/interface field. You should extract new function from `_makeInputValues`.

src/utilities/extendSchema.js
```diff
@@ -258,10 +264,103 @@ export function extendSchema(
...
+  }
+
+  function extendValueMap(type: GraphQLEnumType) {
+    const oldValueMap = Object.create(null);
+    const newValueMap = Object.create(null);
+    type.getValues().forEach(value => {
+      newValueMap[value.name] = oldValueMap[value.name] = {
+        name: value.name,
+        description: value.description,
+        value: value.value,
```
IvanGoncharov: You shouldn't do that, because of this check:
https://github.com/graphql/graphql-js/blob/3cb450059d58f45dff95867a0674d7588be64933/src/type/definition.js#L1164
In the current version, it will convert all enum values without explicit `value` to `undefined`.
This is the very important case so please add a separate test for it.

mgrenier:
@IvanGoncharov I still have to wrap my head around the directive argument case. Most tests cases I tried gave me errors.

`extendSchema.js`
```js
function getMergedDirectives(): Array<GraphQLDirective> {
  const existingDirectives = schema.getDirectives();
  invariant(existingDirectives, 'schema must have default directives');

  return existingDirectives.concat(
    directiveDefinitions.map(node => astBuilder.buildDirective(node)),
  ).map(directive => new GraphQLDirective({
    name: directive.name,
    description: directive.description,
    locations: directive.locations,
    args: extendArgs(directive.args),
    astNode: directive.astNode,
  }));
}
```

`extendSchema-test.js`
```js
it('extend input', () => {
  const extendedSchema = extendTestSchema(`
    extend type Query {
      newField(testArg: TestInput): String
    }

    input TestInput {
      testInputField: String
    }

    directive @test(arg: TestInput) on FIELD
  `);

```

Seems like just passing the directives args trough the newly `extendArgs`, function to extend args type, result in a mixup reference : `Error: Schema must contain unique named types but contains multiple types named "TestInput".`

I need some help on this one.

src/utilities/extendSchema.js
```diff
@@ -291,6 +302,7 @@ export function extendSchema(
     Object.keys(oldFieldMap).forEach(fieldName => {
       const field = oldFieldMap[fieldName];
       newFieldMap[fieldName] = {
+        // name: fieldName,
```
IvanGoncharov: You don't need it 😀

src/utilities/extendSchema.js
```diff
@@ -379,16 +399,18 @@ export function extendSchema(
...
-    args: GraphQLArgument[],
-  ): GraphQLArgument[] {
-    return args.map(arg => ({
-      name: arg.name,
-      type: extendFieldType(arg.type),
-      defaultValue: arg.defaultValue,
-      description: arg.description,
-      astNode: arg.astNode,
-    }));
+  function extendArgs(args: GraphQLArgument[]): GraphQLFieldConfigArgumentMap {
```
IvanGoncharov: Please use Array<GraphQLArgument> similar to other places in this file

src/utilities/extendSchema.js
```diff
@@ -379,16 +399,18 @@ export function extendSchema(
...
+    return keyValMap(
+      args,
+      arg => arg.name,
+      arg => ({
+        name: arg.name,
+        type: extendFieldType(arg.type),
```
IvanGoncharov: Not a critical issue but you could think of a better name for this function since now it also used for arguments. How about `extendType`?

src/utilities/extendSchema.js
```diff
@@ -379,16 +399,18 @@ export function extendSchema(
...
-      defaultValue: arg.defaultValue,
-      description: arg.description,
-      astNode: arg.astNode,
-    }));
+  function extendArgs(args: GraphQLArgument[]): GraphQLFieldConfigArgumentMap {
+    return keyValMap(
+      args,
+      arg => arg.name,
+      arg => ({
+        name: arg.name,
```
IvanGoncharov: You don't need `name`, see:
https://github.com/graphql/graphql-js/blob/ae5b163d2e6c124107fa0971f6d838c8a7d29f51/src/type/definition.js#L842-L847

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -847,10 +847,19 @@ describe('extendSchema', () => {
...
     `);
     expect(extendedSchema).to.not.equal(testSchema);
-    expect(printSchema(extendedSchema)).to.contain('NEW_ENUM');
-    expect(printSchema(testSchema)).to.not.contain('NEW_ENUM');
+
+    const oldEnum = testSchema.getType('SomeEnum');
+    const newEnum = extendedSchema.getType('SomeEnum');
+    const testDirective = extendedSchema.getDirective('test');
+
+    expect(oldEnum._values.length).to.equal(2);
```
IvanGoncharov: nit: `expect(oldEnum._values).to.have.lengthOf(2)`

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -847,10 +847,19 @@ describe('extendSchema', () => {
...
+    expect(newEnum._values.length).to.equal(3);
+    expect(newEnum._values[2].name).to.equal('NEW_ENUM');
+    expect(newEnum).to.equal(testDirective.args[0].type);
```
IvanGoncharov: nit: `expect(testDirective).to.have.nested.property('args[0].type', newEnum)`

IvanGoncharov:
> I still have to wrap my head around the directive argument case. Most tests cases I tried gave me errors.

@mgrenier You code is correct, except you shouldn't recreate new directives, only old ones:
```js
  function getMergedDirectives(): Array<GraphQLDirective> {
    const existingDirectives = schema.getDirectives()    );
    invariant(existingDirectives, 'schema must have default directives');

    return existingDirectives.concat(
      existingDirectives.map(
        directive =>
          new GraphQLDirective({
            name: directive.name,
            description: directive.description,
            locations: directive.locations,
            args: extendArgs(directive.args),
            astNode: directive.astNode,
          }),
      ),
      directiveDefinitions.map(node => astBuilder.buildDirective(node)),
    );
  }
```

If you move out directive recreation into separate function it will make this code more readable.

Also don't forget that all code should be tested e.g.: values of `extensionASTNodes`, extending schema multiple times, all errors (e.g. adding duplicate enum value or input field), values of enum values, etc.
When possible it's better to extend existing tests.

IvanGoncharov:
@mgrenier Also please rebase your PR on top of master, since we updated Flow & Prettier.

src/utilities/extendSchema.js
```diff
@@ -258,10 +268,119 @@ export function extendSchema(
...
+
+  function extendValueMap(type: GraphQLEnumType) {
+    const newValueMap = Object.create(null);
+    const oldValueMap = keyMap(type.getValues(), value => value.name);
+    Object.keys(oldValueMap).forEach(valueName => {
+      const value = oldValueMap[valueName];
+      newValueMap[valueName] = {
+        name: value.name,
+        description: value.description,
+        value: value.hasOwnProperty('value') ? value.value : valueName,
```
IvanGoncharov: My bad, I got confused with something else 😞
You don't need this code, it should be `value: value.value`.

src/utilities/extendSchema.js
mgrenier: `extendType` is already used in this scope. Should I move the logic of `extendFieldType` into this function or used something like `extendValueType` ?

IvanGoncharov: @mgrenier I think it should be `extendType` and `extendNamedType` but it will make this review hard to read so I created #1327.

IvanGoncharov:
@mgrenier I review you PR one more time and found that you also need to extend this function: https://github.com/graphql/graphql-js/blob/ae5b163d2e6c124107fa0971f6d838c8a7d29f51/src/utilities/extendSchema.js#L409

mgrenier:
All review and rebase have been done

jgcmarins:
@mgrenier it still has conflicts with `src/utilities/extendSchema.js`

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -314,24 +372,54 @@ describe('extendSchema', () => {
       extend type Foo {
         deprecatedField: String @deprecated(reason: "not used anymore")
       }
+
+      extend enum SomeEnum {
+        DEPRECATED @deprecated(reason: "do not use")
+      }
```
IvanGoncharov: Test name is "extends objects with deprecated fields"
So I think it's better to create a separate one for enum values.

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -314,24 +372,54 @@ describe('extendSchema', () => {
...
+
+      enum UnusedEnum {
+        SOME
+      }
+
+      input UnusedInput {
+        someInput: String
+      }
+
+      union UnusedUnion = Unused
```
IvanGoncharov: I think 'extends schema by adding new unused types' would be better name for this test.

> union UnusedUnion = Unused

`Unused` became used 😄 It's better to create separate `DummyUnionMember` type.

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -507,6 +595,34 @@ describe('extendSchema', () => {
       interface NewInterface {
         buzz: String
       }
+
```
IvanGoncharov: `extends objects multiple times` => `extends types multiple times`

IvanGoncharov:
@mgrenier I notice decreased coverage:
https://coveralls.io/builds/16751750/source?filename=src%2Futilities%2FextendSchema.js#L157
https://coveralls.io/builds/16751750/source?filename=src%2Futilities%2FextendSchema.js#L289
Can you please investigate and add/extend tests.

src/utilities/extendSchema.js
```diff
@@ -257,14 +267,24 @@ export function extendSchema(
...
-    const existingDirectives = schema.getDirectives();
+    const existingDirectives = schema.getDirectives().map(extendDirectiveType);
     invariant(existingDirectives, 'schema must have default directives');
 
     return existingDirectives.concat(
       directiveDefinitions.map(node => astBuilder.buildDirective(node)),
     );
   }
 
+  function extendDirectiveType(type: GraphQLDirective): GraphQLDirective {
```
IvanGoncharov: function name is confusing it should be either `extendDirective` or `extendDirectiveTypes` (I like it more).

src/utilities/extendSchema.js
```diff
@@ -257,14 +267,24 @@ export function extendSchema(
   // this scope and have access to the schema, cache, and newly defined types.
 
   function getMergedDirectives(): Array<GraphQLDirective> {
-    const existingDirectives = schema.getDirectives();
+    const existingDirectives = schema.getDirectives().map(extendDirectiveType);
```
IvanGoncharov: Maybe we shouldn't recreate standard directives like `@skip`, `@include` and `@deprecate`.
@leebyron What do you think?

src/utilities/extendSchema.js
```diff
@@ -287,10 +307,119 @@ export function extendSchema(
...
+
+    // If there are any extensions to the values, apply those here.
+    const extensions = typeExtensionsMap[type.name];
+    if (extensions) {
+      extensions.forEach(extension => {
+        extension.values.forEach(value => {
+          const valueName = value.name.value;
+          if (oldValueMap[valueName]) {
+            throw new GraphQLError(
+              `Enum "${type.name}.${valueName}" already exists in the ` +
```
IvanGoncharov: `Enum` => `Enum value` or something simular

IvanGoncharov:
@mgrenier Do you need any help with this PR?
We are planning `14.0.0` and it would be great to include this PR.
I think this PR is ready it just need to be rebased and address review comments in tests.

src/utilities/extendSchema.js
mjmahone: I'm going with `extendDirective`

src/utilities/extendSchema.js
mjmahone: This is really elegant, so we can start with this. It would be a breaking change to disallow extending `@skip`/etc. but I like being able to extend them just like any other directive.

mjmahone:
This was merged in #1373

PR #1323, opened 2018-04-23
Fix extend schema from #1314

IvanGoncharov:
I was rebasing WIP PR for 'toConfig' feature and spotted couple problems with #1314.

### DIFF
--- a/src/utilities/extendSchema.js
+++ b/src/utilities/extendSchema.js
@@ -222,6 +222,12 @@ export function extendSchema(
     }
   });
 
+  const schemaExtensionASTNodes = schemaExtensions
+    ? schema.extensionASTNodes
+      ? schema.extensionASTNodes.concat(schemaExtensions)
+      : schemaExtensions
+    : schema.extensionASTNodes;
+
   const types = [
     // Iterate through all types, getting the type definition for each, ensuring
     // that any type not directly referenced by a field will get created.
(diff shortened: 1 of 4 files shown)


src/utilities/extendSchema.js
```diff
@@ -220,11 +213,21 @@ export function extendSchema(
         if (operationTypes[operation]) {
           throw new Error(`Must provide only one ${operation} type in schema.`);
         }
-        operationTypes[operation] = astBuilder.buildType(operationType.type);
+        const typeRef = operationType.type;
+        // Note: While this could make early assertions to get the correctly
+        // typed values, that would throw immediately while type system
+        // validation with validateSchema() will produce more actionable results.
+        operationTypes[operation] = (astBuilder.buildType(typeRef): any);
```
IvanGoncharov: Not sure how it worked previously since `astBuilder.buildType` returns `GraphQLNamedType` but `new GraphQLSchema` expects `query`/`mutation`/`subscription` to be `GraphQLObjectType`.
Looks like Flow bug 😞

IvanGoncharov:
I also notice that `validateSchema` doesn't search for schema error locations inside `extensionASTNodes`.

src/utilities/extendSchema.js
leebyron: I'm guessing that the more declarative `operationTypes` definition has firmer types which flow can now report on where previously the error was hidden

leebyron:
Thanks, this is great. I'm glad you caught this bug before the release!

PR #1327, opened 2018-04-27
extendSchema: Use consistent naming + inline 'getExtendedType' func

IvanGoncharov:
Based on this discussion: https://github.com/graphql/graphql-js/pull/1322#discussion_r184336060
I think naming should be consistent with types and predicates:
https://github.com/graphql/graphql-js/blob/ae5b163d2e6c124107fa0971f6d838c8a7d29f51/src/type/definition.js#L52-L63
https://github.com/graphql/graphql-js/blob/ae5b163d2e6c124107fa0971f6d838c8a7d29f51/src/type/definition.js#L462-L471


leebyron:
Makes sense to me - do you mind rebasing?

IvanGoncharov:
@leebyron Rebased

PR #1328, opened 2018-04-28
printError: fix padding of line numbers

IvanGoncharov:
I noticed that if an error happens on the last line 9 lines string all line numbers are padded with extra space.
It's happening because the padding is calculated based on the size of next line number without checking if this line exists.


leebyron:
Nice catch!

PR #1329, opened 2018-04-28
Don't call global fieldResolver on introspection fields

### LINKED ISSUE #1015: Custom global fieldResolver called on introspection fields
Ran into this when adding a custom schema wide `fieldResolver`. This resolver will also get called for introspection fields which was surprising to me. Is this the expected behavior?

IvanGoncharov:
Fixes #1015 
I also measured performance and the new version is ~~7% faster~~ in this test:
Update: I corrected my measurements and it became only 2.5% faster
```js
const {
  GraphQLObjectType,
  GraphQLSchema,
  GraphQLString,
  getIntrospectionQuery,
  execute,
  parse,
} = require('../dist');

const QueryRoot = new GraphQLObjectType({
  name: 'QueryRoot',
  fields: {
    onlyField: { type: GraphQLString },
  },
});

const schema = new GraphQLSchema({ query: QueryRoot });
const document = parse(getIntrospectionQuery());

for (let i = 0; i < 10000; ++i) {
  execute({ schema, document });
}
```
Probably because `obj => obj.property` is faster than:
https://github.com/graphql/graphql-js/blob/ae5b163d2e6c124107fa0971f6d838c8a7d29f51/src/execution/execute.js#L1264

### DIFF
--- a/src/type/__tests__/introspection-test.js
+++ b/src/type/__tests__/introspection-test.js
@@ -1446,4 +1446,26 @@ describe('Introspection', () => {
       },
     });
   });
+
+  it('executes an introspection query without calling global fieldResolver', () => {
+    const QueryRoot = new GraphQLObjectType({
+      name: 'QueryRoot',
+      fields: {
+        onlyField: { type: GraphQLString },
+      },
+    });
+
+    const schema = new GraphQLSchema({ query: QueryRoot });
+    const source = getIntrospectionQuery();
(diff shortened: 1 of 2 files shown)


leebyron:
I'm surprised the `typeof` calls are enough to add up to 2.5% - but this makes sense to keep introspection predictable in the context of a custom resolver function

PR #1331, opened 2018-05-01
[RFC]Add 'toConfig' method

IvanGoncharov:
I created to get early feedback about this feature so it's missing proper tests.
This `toConfig` is intended to help with schema transformation by minimizing the amount of code. It such big problem that many tooling write they own versions e.g. [recreateType](https://github.com/apollographql/graphql-tools/blob/4e68230c10a9fd1c6cb601869aa3a1ff8f9f0b90/src/stitching/schemaRecreation.ts#L33) from `graphql-tools`.
When I worked on `graphql-voyager` I created my own [intermidiate format](https://github.com/APIs-guru/graphql-voyager/blob/master/src/introspection/introspection.ts#L52) to do schema transformations.

Moreover such functionality prevents bug that occur when new field is added. For example, `lexicographicSortSchema` doesn't respect newly added `assumeValid` and `allowedLegacyNames` fields.


src/type/definition.js
```diff
@@ -782,6 +853,16 @@ export type GraphQLObjectTypeConfig<TSource, TContext> = {
...
 
+export type GraphQLObjectTypeExactConfig<TSource, TContext> = {|
+  name: string,
+  interfaces: Array<GraphQLInterfaceType>,
+  fields: GraphQLFieldExactConfigMap<TSource, TContext>,
+  isTypeOf: ?GraphQLIsTypeOfFn<TSource, TContext>,
+  description: ?string,
+  astNode?: ?ObjectTypeDefinitionNode,
+  extensionASTNodes: $ReadOnlyArray<ObjectTypeExtensionNode>,
+|};
```
IvanGoncharov: I can't use `GraphQL*Config` types since they have `Thunk` fields and a lot of fields marked as optional so without this new type code that uses `toConfig` should have a lot of unnecessary runtime checks and also have dead code to handle `Thunk`.

Plus I think it easier to keep this types up to date than update `extendSchema`, `lexicographicSortSchema`, and all similar functions inside 3rd-party tools.

freiksenet:
I've checked it against `recreateType` in graphql-tools and it looks like it matches what we have there, so it should be fine.

src/type/definition.js
mjmahone: I wonder whether we still need the `Thunk` fields in the `GraphQL*Config` types at all. This feels like a much cleaner expression of a GraphQL type without needing surrounding schema. I'll admit, I don't fully understand how the thunks are used currently, so take that idea as a relative newcomer's curiosity instead of a strong suggestion.

src/type/definition.js
```diff
@@ -782,6 +853,16 @@ export type GraphQLObjectTypeConfig<TSource, TContext> = {
   extensionASTNodes?: ?$ReadOnlyArray<ObjectTypeExtensionNode>,
 };
 
+export type GraphQLObjectTypeExactConfig<TSource, TContext> = {|
+  name: string,
+  interfaces: Array<GraphQLInterfaceType>,
```
mjmahone: Should this instead be:
```
interfaces: Array<GraphQLInterfaceTypeExactConfig<TSource, TContext>>,
```
I'm not sure which makes sense: we definitely don't want to end up with a Schema where an object points to a different instance of the same interface vs. what the schema itself has. So we need to make sure, anywhere we hydrate a type out of this config, that it ends up pointing to the schema's actual interface type, rather than the interface type used at build time (I see you're already doing this everywhere it's used, but others might not be so prudent). I'd even be tempted to make this just `Array<string>` just to prevent any mismatches. At that point, we're basically be creating a stricter version of the Introspection schema (we seem to need a non-Introspection, but Introspection-like, representation regardless, so I'm still OK with this).

My worry may be a complete non-issue if we don't allow exporting these new types and limit how we use them within the repo.

src/type/definition.js
alloy: As far as I have always understood, the thunks are there to avoid circular references at module import time when your schema is made up of types each in their own module. We have definitely run into that situation a couple of times and switched to using thunks.

For `toConfig`, however, it’s indeed not necessary as by the time it can be invoked the import stage should already have passed.

alloy:
@IvanGoncharov I’d love to have access to this 🙏 Currently I’ve copy-pasted some code from the `extendSchema` module into my own code https://github.com/artsy/metaphysics/pull/1346/files#diff-2092cb52bba31c5ad29eb09c6aa4c475R57.

I agree with the idea of not exporting this through the top-level `index.js` file yet, but other than that is there anything still blocking this?

src/type/definition.js
```diff
@@ -674,6 +696,23 @@ export class GraphQLObjectType {
     return this._interfaces;
   }
 
+  toConfig(): {|
+    ...GraphQLObjectTypeConfig<*, *>,
```
mjmahone: There's definitely another project for someone of figuring out how to get rid of all these `*` flow types (which are about as bad as `any`). Not this PR though.

src/type/schema.js
```diff
@@ -238,6 +238,27 @@ export class GraphQLSchema {
   getDirective(name: string): ?GraphQLDirective {
     return find(this.getDirectives(), directive => directive.name === name);
   }
+
+  toConfig(): {|
+    ...GraphQLSchemaConfig,
```
mjmahone: huh this must be new Flow wizardry: this is a nicely concise way of defining types in terms of other types!

PR #1332, opened 2018-05-01
 Allow to override default type resolver

IvanGoncharov:
Having global default `typeResolver` is very useful for writing GraphQL proxies and for mocking missing types. For example it would be very easy to mock types with missing resolver like this:
```js
import {
  defaultFieldResolver,
  defaultTypeResolver,
} from 'graphql';

function mockFieldResolver(value, args, ctx, info) {
  const default = defaultFieldResolver(value, args, ctx, info);
  return default === undefined ? mockType(info.returnType) : default;
}

function mockType(type) {
  //...
  if (type === GraphQLString) {
    return 'mock string';
  }
}

function mockTypeResolver(value, ctx, info) {
  const defaultType = defaultTypeResolver(value, ctx, info);
  if (defaultType === undefined) {
    const types = info.schema.getPossibleTypes(info.valueType);
    const random = Math.floor(Math.random() * (types.length - 1))
    return types[random];
  }
  return defaultType;
}

const result = graphql({
  schema, 
  source,
  fieldResolver: mockFieldResolver,
  typeResolver: mockTypeResolver,
});
```

### DIFF
--- a/src/execution/__tests__/executor-test.js
+++ b/src/execution/__tests__/executor-test.js
@@ -15,6 +15,7 @@ import { execute } from '../execute';
 import { Kind, parse } from '../../language';
 import {
   GraphQLSchema,
+  GraphQLInterfaceType,
   GraphQLObjectType,
   GraphQLList,
   GraphQLBoolean,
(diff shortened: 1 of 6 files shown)


src/execution/execute.js
```diff
@@ -1027,9 +1038,10 @@ function completeAbstractValue(
   path: ResponsePath,
   result: mixed,
 ): MaybePromise<ObjMap<mixed>> {
-  const runtimeType = returnType.resolveType
-    ? returnType.resolveType(result, exeContext.contextValue, info)
-    : defaultResolveTypeFn(result, exeContext.contextValue, info, returnType);
+  const resolveTypeFn = returnType.resolveType || exeContext.typeResolver;
+  const contextValue = exeContext.contextValue;
+  const typeInfo = { ...info, valueType: returnType, valuePath: path };
+  const runtimeType = resolveTypeFn(result, contextValue, typeInfo);
```
leebyron: Can we avoid creating a new object for every field resolution?

Perhaps these lines should be:

```
const runtimeType = returnType.resolveType
  ? returnType.resolveType(result, exeContext.contextValue, info)
  : exeContext.typeResolver
    ? exeContext.typeResolver(returnType, result, exeContext.contextValue, info)
    : defaultResolveTypeFn(returnType, result, exeContext.contextValue, info)
```

leebyron: Open to this type signature - I'm not sure how valuable `returnType` will be for a default resolver - if it deserves to be the first argument or not. I imagine `result` would still be the most useful.

Another option might be to just provide `returnType` as another argument after info to all type resolvers - not just default ones?

IvanGoncharov:
> Another option might be to just provide returnType as another argument after info to all type resolvers - not just default ones?

@leebyron Done 👍 
Can you please review one more time?

src/execution/__tests__/executor-test.js
```diff
@@ -1029,4 +1030,42 @@ describe('Execute: Handles basic execution tasks', () => {
...
+        name: 'Query',
+        fields: {
+          foo: { type: fooInterface },
+        },
+      }),
+      types: [fooObject],
+    });
+
+    function typeResolver(source, context, info, abstractType) {
+      return schema.getPossibleTypes(abstractType)[0];
```
Neitsch: Can we change this line to something like `return schema.getPossibleTypes(abstractType).find(graphQLType => graphQLType.name === "FooObject");`. Makes it clearer what we are returning.

IvanGoncharov:
> Just to make sure that I get this change correctly: You allow to specify a custom type resolved, that if specified replaces the default resolver. Does that sound right?

@Neitsch Yes, exactly 👍

src/execution/__tests__/executor-test.js
IvanGoncharov: @Neitsch Great suggestion 👍
One thing is that it's enough to return string with type name from `typeResolve` so matching type by name looks unnecessary if you can just return name.

So I just change it to be more explicit.

PR #1336, opened 2018-05-04
Scalars: reject array serialization/coercion

IvanGoncharov:
Working on #1333 I noticed that you can pass arrays as scalars in a query variable, e.g. [SWAPI example](http://graphql.org/swapi-graphql/?query=query%20(%24id%3A%20ID)%20%7B%0A%20%20person(personID%3A%20%24id)%20%7B%0A%20%20%20%20name%0A%20%20%7D%0A%7D&variables=%7B%0A%20%20%22id%22%3A%20%5B1%5D%0A%7D)
This PR is based on #925 but fixing same issue for types other than `String` e.g.
```js
Number([1]) // 1
```

### DIFF
--- a/src/type/scalars.js
+++ b/src/type/scalars.js
@@ -127,8 +127,8 @@ export const GraphQLID = new GraphQLScalarType({
     'response as a String; however, it is not intended to be human-readable. ' +
     'When expected as an input type, any string (such as `"4"`) or integer ' +
     '(such as `4`) input value will be accepted as an ID.',
-  serialize: String,
-  parseValue: String,
+  serialize: coerceString,
+  parseValue: coerceString,
   parseLiteral(ast) {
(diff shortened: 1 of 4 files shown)


src/type/scalars.js
```diff
@@ -87,7 +97,7 @@ export const GraphQLFloat = new GraphQLScalarType({
   },
 });
 
-function coerceString(value: mixed): ?string {
+function coerceString(value: mixed): string {
```
mjmahone: Oh interesting: technically the spec lets us coerce arrays into strings and booleans: https://github.com/facebook/graphql/blob/master/spec/Section%203%20--%20Type%20System.md#string

However it does NOT allow us to coerce arrays into floats or ints.

IvanGoncharov:
@mjmahone Thanks for the review 👍 Merged 🎉

Issue #1005, opened 2017-08-25
[RFC] Bump version number to 1.0


wincent:
While it's true that the GraphQL Spec itself [is labelled](https://github.com/facebook/graphql/blame/3761de18924f4ee3e0a63931b7d954f3bb37ca65/spec/GraphQL.md#L4-L8) as a "Working Draft" and an "Draft RFC Specification", and "Significant enhancement [is expected to] continue" ([see](https://github.com/facebook/graphql/blame/3761de18924f4ee3e0a63931b7d954f3bb37ca65/spec/GraphQL.md#L11-L12)), the truth is that the `graphql` NPM package is relatively solid at this point and being depended on in production by numerous companies.

I think it's time to move to 1.0+ versioning so that we can actually use the version number to transmit meaningful semantic information (ie. that a "major" version bump is a compatibility-breaking change). Our current version numbers (of the form v0.x.y) [transmit basically nothing](http://semver.org/#spec-item-4), because:

> Major version zero (0.y.z) is for initial development. Anything may change at any time.

This entry from the [Semver FAQ](http://semver.org/#how-do-i-know-when-to-release-100) is also helpful:

> **How do I know when to release 1.0.0?**
>
> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

I think all of those apply to this package, so let's do it with the next release.

IvanGoncharov:
@wincent Great idea 👍 
One thing to note is that after #927 be merged description inside comments will be deprecated.
I think it's better to declare something deprecated before `1.0.0` so we should wait for merge of #927.

In general, a big part of `graphql-js` is dependent on IDL/SDL which is not yet merged.
And there is a non-zero risk that something besides description will be changed so I recommend waiting until https://github.com/facebook/graphql/pull/90 merged.

leebyron:
Agree with both of you. Let's release v1.0 once we cut the Fall version of the spec next month

felixfbecker:
This would also help a lot with https://github.com/graphql/graphql-js/issues/491#issuecomment-333334173

razor-x:
Just stumbling on this thread and wondering what ever happened with this as according to the above comment this should have already happened sometime in October / November. Thanks.

felixfbecker:
[...]

PR #1338, opened 2018-05-04
Run tests on node 10. Drop support for node 4.0

IvanGoncharov:
Based on the discussion in #1005 next release should be `14.0.0`.
I think it's important to drop node 4.0 before this release.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,7 @@
   "presets": [["env", {
     "modules": false,
     "targets": {
-      "node": 4,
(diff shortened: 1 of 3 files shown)


mjmahone:
We may need to cut a minor release prior to `14.0.0`: this seems like something we should merge right before cutting the major-version-bumping release, but not in any minor release.

IvanGoncharov:
@mjmahone Agree 👍

PR #1342, opened 2018-05-08
Coverage: Switch isparta to nyc

IvanGoncharov:
Isparta produces incorrect coverage reports:
image

Moreover there is zero chance it will be fixed since `isparta` is [officially unmantained](https://github.com/douglasduteil/isparta/commit/7e89d4e467f54558396533462122e73ea4e9dc31).
So I switched `isparta` to [nyc]() which is official CLI for istanbul and also replaced `babel-node` workaround with [offically recomended](https://babeljs.io/docs/setup/) `babel-registry`/`babel-polyfill`.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -10,6 +10,7 @@ yarn.lock
 package-lock.json
 
 .babelrc
+.nycrc
 CONTRIBUTING.md
 node_modules
 coverage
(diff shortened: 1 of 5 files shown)


leebyron:
Thank you for fixing this!

Issue #1324, opened 2018-04-25
Spec Compliance: Strict input coercion for scalars


spawnia:
The GraphQL specification defines strict rules for input coercion, especially of scalar types: http://facebook.github.io/graphql/October2016/#sec-Scalars

However, inputs that are passed in as variables are not validated accordingly, while literals are. `parseValue` mostly behaves the same as `serialize` - which allows for some loose coercion.

If i understand the specification correctly, such loose coercion is only allowed for results - outbound information that the server sends to the client. It makes sense in this case, the server should not break the contract it defines in its schema.

Inputs, regardless if they are provided as literals or as variables, are inbound information which come from the client and are sent to the server. In that case, strict validation makes absolute sense - if the client sends bad data, the server should not quietly coerce the value but rather throw an error.

While literals and variables have to be handled differently at first, they both fall into the category of Input Coercion. The spec is really clear on how inputs should be handled, see [Section 6.1.2](http://facebook.github.io/graphql/October2016/#sec-Coercing-Variable-Values):
> If the operation has defined any variables, then the values for those variables need to be coerced using the input coercion rules of variable’s declared type.

What is currently happening is that the result coercion rules, which are different from input coercion rules, are applied to input variables. On the other hand, input literals are treated different than input variables - while according to the specification, they should be handled the same in regards to the coercion.

It seems like there is a fundamental misunderstanding here. If it is on my side, maybe you can clear it up for me? If not, i think resolving this issue to ensure spec compliance requires breaking changes to the implementation.

IvanGoncharov:
@spawnia Can you please provide some examples?

spawnia:
Sure. Given the following schema:

````graphql
type Query {
	example(number: Int): Example
}

type Example {
    number: Int
}
````

Now, if i send the following Query i get a type error:

````
{
	example(number: "123"){
		number
	}
}
````

However, if i pass the same value as a variable, the value gets converted into a string:

````
query ExampleQuery($value: Int){
	example(number: $value){
		number
	}
}

// JSON
{
	"value": "123"
}
````

IvanGoncharov:
> It seems like there is a fundamental misunderstanding here. If it is on my side, maybe you can clear it up for me? 

@spawnia You completely right, specification clear state that input coercion should be strict. Mover some behavior is ridiculous for example you can pass `{}` and it will be coerced to `'[object Object]'`.

> If not, i think resolving this issue to ensure spec compliance requires breaking changes to the implementation.

[...]

PR #1346, opened 2018-05-11
0.13.3

mjmahone:


IvanGoncharov:
@mjmahone There are a few breaking changes that were merged into `master` since `0.13.2`. The biggest breaking change is #1284 but there is also #1274 and #1277.
Not sure that that bumping version to `0.13.3` would be enough.

mjmahone:
Yeah I think that's a good idea, especially if it is enforced on PRs themselves: we wouldn't pull in PRs that have semver-major changes until just before we'd want a new major version released.

:/ I unfortunately simultaneously really need to bump a version, don't want to break people currently on `13.2`, and don't have the knowledge yet to pull in many semver-major changes in preparation for the 14.0.0 release. I may have to disrupt @leebyron's plans a bit and bump this to 0.14.0 (treating it as a minor, yet still backwards-incompatible, bump), and let the major release start with 15.0.0.

IvanGoncharov:
>  I may have to disrupt @leebyron's plans a bit and bump this to 0.14.0 (treating it as a minor, yet still backwards-incompatible, bump), and let the major release start with 15.0.0.

I think it's best possible solution in the current situation 👍

IvanGoncharov:
@mjmahone I don't think there are any critical PRs that should be merged before release, except for #1353 which fixes CI for `node v9`.
Also, it would be nice to include `babel7` migration (#1338 and #1350) in this release since it was long planned and block other improvements to build system, e.g. #1348

leebyron:
Let's wait on this - we published that the next version would be v14 and we've merged some breaking changes.

leebyron:
In fact - why not just update this to be a v14 bump instead

leebyron:
0.13.0 -> 0.14.0 is semver equivalent - starting with 0. effectively treats the next number as a major breaking. I'm not sure it's necessary to release 0.14 instead of 14.0? Why bother with changing plans?

mjmahone:
Yeah I think we should move to 14.0 based on this discussion. Not sure though what is necessary to get there: were there other breaking changes we need to pull in?

IvanGoncharov:
@mjmahone @leebyron I think people wanted `14.0.0` to solve the issue with incompatible `peerDependencies`. But `14.0.0` will be the improvement in that sense only if we don't plan to release `15.0.0` in foreseen future.

So I think we should review all proposed breaking changes that we have at the moment and either reject or merge them.
For example #1324 propose to make `graphql-js` spec compliant but it's a breaking change at the same time.
I can review all issues/PRs and compile a list.

To minimize possible risks I think we need to make `14.0.0-rc.1` release similar to [0.13.0-rc.1](https://github.com/graphql/graphql-js/releases/tag/v0.13.0-rc.1).

I also think it would be great if release notes for `14.0.0` would contain deprecation notice and define sunset version (`15.0.0` or other) for:
  - legacy SDL syntax
  - allowedLegacyNames
  - [onOperation, onFragment, onField](https://github.com/graphql/graphql-js/blob/bd8aa362efc2dcafb54ed1b7286577a6ee7b4c59/src/type/introspection.js#L95-L117)
  - [isValidJSValue](https://github.com/graphql/graphql-js/blob/2536f144e30fb33aadec3916dfb728043fae573b/src/utilities/isValidJSValue.js#L14),[isValidLiteralValue](https://github.com/graphql/graphql-js/blob/17a0bfd5292f39cafe4eec5b3bd0e22514243b68/src/utilities/isValidLiteralValue.js#L23),[introspectionQuery](https://github.com/graphql/graphql-js/blob/b333b30b238d040d6a8cf959d54f037ee26ae2c6/src/utilities/index.js#L14)

PR #1352, opened 2018-05-16
Validate Input Objects do not contain non-nullable circu…

spawnia:
…lar references

The first two to tests are there to avoid false positives when checking for circular references.

Tests 3 and 4 are failing right now but show what should happen.

https://github.com/facebook/graphql/pull/445 contains the proposed additions to the spec.

### DIFF
--- a/src/language/__tests__/schema-parser-test.js
+++ b/src/language/__tests__/schema-parser-test.js
@@ -802,54 +802,6 @@ input Hello {
     );
   });
 
-  it('Allow simple input object with nullable circular reference', () => {
-    expect(() =>
-      parse(`
-        input Hello {
-          world: String
-          self: Hello
-        }`),
-    ).to.not.throw();
-  });
-
-  it('Allow input object with circular reference broken up by a list', () => {
-    expect(() =>
-      parse(`
-        input Hello {
-          world: String
-          self: [Hello!]!
-        }`),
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@spawnia Thanks for PR, your SDL examples look great 👍

However `parse` function doesn't do any validation beyound checking that correctness of GraphQL grammar.
This check should be performed by `validateSchema` function:
https://github.com/graphql/graphql-js/blob/4aca994d2b0447293ee700e687d333daf4113589/src/type/validate.js#L48-L57

Can you please move your tests here:
https://github.com/graphql/graphql-js/blob/master/src/type/__tests__/validation-test.js#L1397

IvanGoncharov:
@spawnia Preferably test cases should match examples from https://github.com/facebook/graphql/pull/445

spawnia:
Unfortunately, implementation of this is not trivial and i think i will not be able to do it soon, i might implement it in PHP since i know that better.

IvanGoncharov:
@spawnia Don't want this proposal to stuck so I created a draft implementation in #1359. I don't think it's ready to be merged but hopefully, it allows us to continue the conversation in https://github.com/facebook/graphql/pull/445

IvanGoncharov:
@spawnia Since #1359 superseded this PR can I close it?

PR #1353, opened 2018-05-17
CI: fix nyc doesn't proxy error status from mocha

IvanGoncharov:
This PR fixes issue introduced in #1342
It results in one node version always being green even if others will fail:
image
https://travis-ci.org/graphql/graphql-js/builds/379852096?utm_source=github_status&utm_medium=notification

Fix is based on this comment: https://github.com/istanbuljs/nyc/issues/433#issuecomment-260173503


IvanGoncharov:
@kassens Thanks 👍

PR #1355, opened 2018-05-24
extendSchema incorrectly throws "multiple types" error when input type used as a query argument.

jesstelford:
As per the New Issue template, this PR contains a failing unit test showing the issue (I don't have a fix).

Here's a live example you can play around with: https://codesandbox.io/s/9orv5vlovw?module=%2Fsrc%2Findex.js

The error received is:

```
Error: Schema must contain unique named types but contains multiple types named "FooInput"
```

I've had a stab at tracking it down, and it looks like it arises due a reference inequality between two objects. When investigating the two objects, their structures appear to be slightly different:

screen shot 2018-05-24 at 11 18 53 am

Stepping through the code, I noticed the first instance of `FooInput` to be put in the `map` (`map[type.name]` in the screenshot above) is the one encountered as part of recursing into `type Query { foo(where: FooInput ...`. Not sure if that helps at all?

Note also that manually building the schema (ie; not parsing from the SDL) doesn't appear to exhibit this bug (hence the addition of the `FooInput` `GraphQLInputObjectType` in this PR). I'm not sure if that's because I'm miss-understanding the way the code executes, or because of a fundamental difference in how the schemas are handled when extending?

### DIFF
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -9,12 +9,14 @@ import { describe, it } from 'mocha';
 import { expect } from 'chai';
 import dedent from '../../jsutils/dedent';
 import { extendSchema } from '../extendSchema';
+import { buildASTSchema } from '../buildASTSchema';
 import { parse, print } from '../../language';
 import { printSchema } from '../schemaPrinter';
 import { Kind } from '../../language/kinds';
 import { graphqlSync } from '../../';
 import {
   GraphQLSchema,
+  GraphQLInputObjectType,
   GraphQLObjectType,
   GraphQLInterfaceType,
   GraphQLUnionType,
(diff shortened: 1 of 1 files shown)


jesstelford:
Sooo, it turns out it was a typo... `type FooInput` !== `input FooInput`.

I do wonder though if a better error message could have been generated here? Should `extendSchema` perhaps validate before extending?

IvanGoncharov:
@jesstelford Thanks a lot for the test it helped a lot 👍
This issue happens because `extendSchema` assume that you can't extend input types because of this check:
https://github.com/graphql/graphql-js/blob/f3ebab8f85e44d74064c557b88a3a44ee80aee92/src/utilities/extendSchema.js#L163-L166

That's why we assume that it safe to just copy args without updating their types, see line 374:
https://github.com/graphql/graphql-js/blob/f3ebab8f85e44d74064c557b88a3a44ee80aee92/src/utilities/extendSchema.js#L369-L374

It's already fixed as part of #1322  which allows to extend input types.
So I think we should focus on merging this PR into the next release.

> I do wonder though if a better error message could have been generated here?

Already handled, see this test:
https://github.com/graphql/graphql-js/blob/f3ebab8f85e44d74064c557b88a3a44ee80aee92/src/type/__tests__/validation-test.js#L1270-L1277

> Should extendSchema perhaps validate before extending?

We try validate schema only once per schema to improve perfomance on big schema and provide better DX (you get all errors at once) so schema validated first time you call `execute`: 
https://github.com/graphql/graphql-js/blob/f3ebab8f85e44d74064c557b88a3a44ee80aee92/src/execution/execute.js#L276-L277

IvanGoncharov:
Fixed in #1373

PR #1357, opened 2018-05-25
instanceOf: Check that one of the names in non-empty string

### LINKED ISSUE #1356: Misleading error in Internet Explorer 11
It throws the following error:

> SCRIPT5022: Cannot use undefined "Boolean" from another module or realm. Ensure that there is only one instance of "graphql" in the node_modules directory. If different versions of "graphql" are the dependencies of other relied on modules, use "resolutions" to ensure only one version is installed. https://yarnpkg.com/en/docs/selective-version-resolutions Duplicate "graphql" modules cannot be used at the same time since different versions may have different capabilities and behavior. The data from one version used in the function from another could produce confusing and spurious results.

Turns out this happens because this code compares two function names:

https://github.com/graphql/graphql-js/blob/f3ebab8f85e44d74064c557b88a3a44ee80aee92/src/jsutils/instanceOf.js#L33-L36

Internet Explorer does not have support to `Function.name`. Thus the comparison will be `undefined === undefined`.

In my case, I fixed that by adding a polyfill. But I think this behavior is, at least, very confusing.

IvanGoncharov:
Fixes #1356

### DIFF
--- a/src/jsutils/instanceOf.js
+++ b/src/jsutils/instanceOf.js
@@ -31,7 +31,7 @@ export default (process.env.NODE_ENV === 'production'
       if (value) {
         const valueClass = value.constructor;
         const className = constructor.name;
-        if (valueClass && valueClass.name === className) {
+        if (className && valueClass && valueClass.name === className) {
           throw new Error(
             `Cannot use ${className} "${value}" from another module or realm.
 


mjmahone:
Thanks!

PR #1359, opened 2018-05-25
validateSchema: validate Input Objects self-references

IvanGoncharov:
Implements https://github.com/facebook/graphql/pull/445
Based on #1352

### DIFF
--- a/src/type/__tests__/validation-test.js
+++ b/src/type/__tests__/validation-test.js
@@ -736,6 +736,120 @@ describe('Type System: Input Objects must have fields', () => {
     ]);
   });
 
+  it('accepts an Input Object with breakable circular reference', () => {
+    const schema = buildSchema(`
+      type Query {
+        field(arg: SomeInputObject): String
+      }
+
+      input SomeInputObject {
+        self: SomeInputObject
+        arrayOfSelf: [SomeInputObject]
+        nonNullArrayOfSelf: [SomeInputObject]!
+        nonNullArrayOfNonNullSelf: [SomeInputObject!]!
+        intermediateSelf: AnotherInputObject
+      }
+
+      input AnotherInputObject {
+        parent: SomeInputObject
+      }
+    `);
+
+    expect(validateSchema(schema)).to.deep.equal([]);
+  });
+
+  it('rejects an Input Object with non-breakable circular reference', () => {
+    const schema = buildSchema(`
+      type Query {
+        field(arg: SomeInputObject): String
+      }
+
+      input SomeInputObject {
+        nonNullSelf: SomeInputObject!
+      }
+    `);
+
+    expect(validateSchema(schema)).to.deep.equal([
+      {
+        message:
+          'Cannot reference Input Object "SomeInputObject" within itself through a series of non-null fields: "nonNullSelf".',
+        locations: [{ line: 7, column: 9 }],
+      },
+    ]);
+  });
+
+  it('rejects Input Objects with non-breakable circular reference spread across them', () => {
+    const schema = buildSchema(`
+      type Query {
+        field(arg: SomeInputObject): String
+      }
+
+      input SomeInputObject {
+        startLoop: AnotherInputObject!
+      }
+
+      input AnotherInputObject {
+        nextInLoop: YetAnotherInputObject!
+      }
+
+      input YetAnotherInputObject {
+        closeLoop: SomeInputObject!
+      }
+    `);
+
+    expect(validateSchema(schema)).to.deep.equal([
+      {
+        message:
+          'Cannot reference Input Object "SomeInputObject" within itself through a series of non-null fields: "startLoop.nextInLoop.closeLoop".',
+        locations: [
+          { line: 7, column: 9 },
+          { line: 11, column: 9 },
+          { line: 15, column: 9 },
+        ],
+      },
+    ]);
+  });
+
+  it('rejects Input Objects with multiple non-breakable circular reference', () => {
+    const schema = buildSchema(`
+      type Query {
+        field(arg: SomeInputObject): String
+      }
+
+      input SomeInputObject {
+        startLoop: AnotherInputObject!
+      }
+
+      input AnotherInputObject {
+        closeLoop: SomeInputObject!
+        startSecondLoop: YetAnotherInputObject!
+      }
+
+      input YetAnotherInputObject {
+        closeSecondLoop: AnotherInputObject!
+        nonNullSelf: YetAnotherInputObject!
+      }
+    `);
+
+    expect(validateSchema(schema)).to.deep.equal([
+      {
+        message:
+          'Cannot reference Input Object "SomeInputObject" within itself through a series of non-null fields: "startLoop.closeLoop".',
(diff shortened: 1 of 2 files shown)


src/type/validate.js
```diff
@@ -578,6 +581,64 @@ function validateInputFields(
   });
 }
 
+function validateInputObjectCircularReferences(
+  context: SchemaValidationContext,
+): void {
+  // Modified copy of algorithm from 'src/validation/rules/NoFragmentCycles.js'.
```
IvanGoncharov: Since this algorithm is not trivial I will try to extract duplicating code into `jsutils`

src/type/validate.js
```diff
@@ -578,6 +581,64 @@ function validateInputFields(
...
+  const visitedTypes = Object.create(null);
+
+  // Array of types nodes used to produce meaningful errors
+  const fieldPath = [];
+
+  // Position in the type path
+  const fieldPathIndexByTypeName = Object.create(null);
+
+  const typeMap = context.schema.getTypeMap();
+  for (const type of objectValues(typeMap)) {
```
IvanGoncharov: It's bad that we need to iterate over every type one more time.
Would be great to integrate this check into the loop inside `validateTypes` function without affecting its readability.

src/type/__tests__/validation-test.js
```diff
@@ -687,6 +687,80 @@ describe('Type System: Input Objects must have fields', () => {
     ]);
   });
 
+  it('accepts an Input Object with breakable circular reference', () => {
```
spawnia: I like how you considered all the possibilities of breaking a circular reference in the first test case, to ensure this does not cause false-positives.

src/type/validate.js
```diff
@@ -578,6 +581,64 @@ function validateInputFields(
...
+
+        fieldPath.push(field);
+        if (cycleIndex === undefined) {
+          detectCycleRecursive(fieldType);
+        } else {
+          const cyclePath = fieldPath.slice(cycleIndex);
+          const fieldNames = cyclePath.map(fieldObj => fieldObj.name);
+          context.reportError(
+            `Cannot reference Input Object "${fieldType.name}" within itself ` +
+              `through a series of non-null field: "${fieldNames.join('.')}".`,
```
spawnia: field -> fields

src/type/validate.js
```diff
@@ -578,6 +581,64 @@ function validateInputFields(
   });
 }
 
+function validateInputObjectCircularReferences(
```
spawnia: Shouldn't this be in a seperate rules file, similar to how NoFragmentCycles is implemented?

src/type/__tests__/validation-test.js
```diff
@@ -687,6 +687,80 @@ describe('Type System: Input Objects must have fields', () => {
...
+          'Cannot reference Input Object "SomeInputObject" within itself through a series of non-null field: "startLoop.nextInLoop.closeLoop".',
+        locations: [
+          { line: 7, column: 9 },
+          { line: 11, column: 9 },
+          { line: 15, column: 9 },
+        ],
+      },
+    ]);
+  });
+
```
spawnia: I suggest to add another test-case:

````javascript
  it('rejects Input Objects with multiple non-breakable circular reference', () => {
    const schema = buildSchema(`
      type Query {
        field(arg: SomeInputObject): String
      }
 
      input SomeInputObject {
        startLoop: AnotherInputObject!
      }
 
      input AnotherInputObject {
        closeLoop: SomeInputObject!
        startSecondLoop: YetAnotherInputObject!
      }
 
      input YetAnotherInputObject {
        closeSecondLoop: AnotherInputObject!
        nonNullSelf: YetAnotherInputObject!
      }
    `);
 
    expect(validateSchema(schema)).to.deep.equal([
      {
        message:
          'Cannot reference Input Object "SomeInputObject" within itself through a series of non-null field: "startLoop.closeLoop".',
        locations: [
          { line: 7, column: 9 },
          { line: 11, column: 9 },
        ],
      },
    ],[
      {
        message:
          'Cannot reference Input Object "AnotherInputObject" within itself through a series of non-null field: "startSecondLoop.closeSecondLoop".',
        locations: [
          { line: 12, column: 9 },
          { line: 16, column: 9 },
        ],
      },
    ],[
      {
        message:
          'Cannot reference Input Object "YetAnotherInputObject" within itself through a series of non-null field: "nonNullSelf".',
        locations: [{ line: 17, column: 9 }],
      },
    ]);
  });
 
````

File: src/type/validate.js (hunk 5 above)
Comment (IvanGoncharov) [reply src/type/validate.js]: @spawnia `src/validate/rules/*` is intended to validate AST only with a focus on the query syntax. You can create the schema using SDL or JS classes like here:
```js
new GraphQLSchema({
  queryType: new GraphQLObjectType({
    name: 'Query',
    fields: {
      foo: {
        type: GraphQLString,
      },
    }
  }),
})
```
Also you construct schema from introspection using `buildClientSchema`.
So schema validation shouldn't be tied to AST and should only use `GraphQL*` classes.

On the other hand, your queries are always AST so they validation is also AST based. 

That's why there is two completely different validation mechanisms: `validate`(AST based, mostly query) and `validateSchema`(`GraphQL*` classes, only for schema).

src/type/validate.js
IvanGoncharov: @spawnia Fixed. 
Should we have a separate error for the case where it's the only one field?

src/type/__tests__/validation-test.js
IvanGoncharov: @spawnia Great test 👍 Added to PR.

src/type/validate.js
spawnia: It would not hurt, but i think this way it is also fine. The error should still be quite obvious.

src/validation/rules/NoFragmentCycles.js
```diff
@@ -34,9 +34,7 @@ export function NoFragmentCycles(context: ValidationContext): ASTVisitor {
   return {
     OperationDefinition: () => false,
     FragmentDefinition(node) {
-      if (!visitedFrags[node.name.value]) {
-        detectCycleRecursive(node);
-      }
+      detectCycleRecursive(node);
```
leebyron: Do you mind factoring the changes to this file to a separate PR?

leebyron:
This is looking great! Let's make sure the spec text is clear as well before merging this

src/validation/rules/NoFragmentCycles.js
IvanGoncharov: @leebyron Done as #1381

spawnia:
@leebyron Anything you would change about the spec change https://github.com/facebook/graphql/pull/445

IvanGoncharov:
Merged since https://github.com/graphql/graphql-spec/pull/445 is on Stage2 and spec is encourage implementing proposal on this stage: 
https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md#stage-2-draft
> GraphQL libraries should implement drafts to provide valuable feedback, though are encouraged not to enable the draft feature without explicit opt-in when possible.

PR #1364, opened 2018-06-04
14.0.0-rc.1

mjmahone:


IvanGoncharov:
@mjmahone I'm working on #1324 which is breaking change and I think it should be in `rc.1`. Can you please wait a day?

IvanGoncharov:
I also wanted to include #1322 since it's almost ready and makes `graphql-js` spec compliant.

mjmahone:
@IvanGoncharov yes it can wait a day. I wanted to put this up to make clear the `14.0` release is coming soon. I'm also OK merging breaking changes until we're off `-rc` versions, so if there's something that we really want that will take more than a day to fix, we can wait a week or so. I'll start canarying the 14.0 release with FB's codebase with each `-rc` though. I'd like to root out any major upgrade issues prior to 14.0.0 (non-rc).

IvanGoncharov:
@mjmahone I caught the flu 🤒 so it will take me more time than I initially planned.
So probably it makes sense to release RC1 as the smoke test for already committed changes.

IvanGoncharov:
@mjmahone My strategy right now is to open PR for all breaking changes on my list so I could get feedback from you and community. I still have a couple items on my list so I focused on opening new PRs.

Sorry for the delay in answering some of your questions on existing PRs they just they require detailed answers and I will try to give them in during next days.

PR #1365, opened 2018-06-04
Reject Infinity supplied as Int or Float value

IvanGoncharov:
Next step toward fixing #1324


riginoommen:
With respective to this change how to fix this error

```js
Float cannot represent non numeric value
````

IvanGoncharov:
@riginoommen Can you please provide full error message?

PR #1367, opened 2018-06-05
[Bug Reproduce] Can't use undefined as enum value

IvanGoncharov:
**Extracted from #1267**

From #836 it looks like both `undefined` and `NaN` could be used as an enum values:
```js
const TestEnum = new GraphQLEnumType({
  name: 'TestEnum',
  values: {
    UNDEFINED: { value: undefined },
    NAN: { value: NaN },
  },
});
```
And run this query:
```graphql
{
  undefined: fieldWithEnumInput(input: UNDEFINED)
  NaN: fieldWithEnumInput(input: NAN)
}
```
It will result in:
```json
{
  "data": {
    "undefined": null,
    "NaN": null
  },
  "errors": [
    {
      "message": "Argument \"input\" has invalid value UNDEFINED.",
      "path": [ "undefined" ]
    }
    {
      "message": "Argument \"input\" has invalid value NAN.",
      "path": [ "NaN" ]
    }
  ]
}
```

### DIFF
--- a/src/execution/__tests__/variables-test.js
+++ b/src/execution/__tests__/variables-test.js
@@ -460,6 +460,7 @@ describe('Execute: Handles inputs', () => {
     it('allows custom enum values as inputs', () => {
       const result = executeQuery(`
         {
+          undefined: fieldWithEnumInput(input: UNDEFINED)
           null: fieldWithEnumInput(input: NULL)
           NaN: fieldWithEnumInput(input: NAN)
           false: fieldWithEnumInput(input: FALSE)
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@mjmahone Yes it known issue with `valueFromAST` since it uses `undefined` return to indicate errors.
Ideally, it should be designed similar to `coerceValue`:
https://github.com/graphql/graphql-js/blob/43c0d46ac9a279a1c3be9c8592381e2a99eec209/src/utilities/coerceValue.js#L26-L45

Correctly handling `undefined` is minor use case for breaking change but it will also makes errors to contain explanation why certain literal value is invalid which is pretty important.

This PR is `[Bug Reproduce]` just to document failing use cases until it will be fixed.

IvanGoncharov:
Resolved in #2148

PR #1369, opened 2018-06-05
Handle NaN as input value

IvanGoncharov:


src/utilities/coerceValue.js
```diff
@@ -46,7 +45,7 @@ export function coerceValue(
 ): CoercedValue {
   // A value must be provided if the type is non-null.
   if (isNonNullType(type)) {
-    if (isNullish(value)) {
+    if (value == null) {
```
IvanGoncharov: `== null` block `null` and `undefined` but not a NaN and simular values

https://github.com/graphql/graphql-js/blob/85971a277c4e7782e9be38d102bd6562dedf7804/src/jsutils/isNullish.js#L13-L15

src/type/definition.js
```diff
@@ -566,9 +565,6 @@ export class GraphQLScalarType {
   // Parses an externally provided value to use as an input.
   parseValue(value: mixed): mixed {
     const parser = this._scalarConfig.parseValue;
-    if (isInvalid(value)) {
-      return undefined;
-    }
```
IvanGoncharov: Safe because it only used here:
https://github.com/graphql/graphql-js/blob/85971a277c4e7782e9be38d102bd6562dedf7804/src/utilities/coerceValue.js#L61-L71

So `null` and `undefined` values are never passed to `parseValue`.
Even though it's not breaking anything in `graphql-js` it's still a minor breaking change since  there is a chance that someone using `parseValue` directly.

PR #1371, opened 2018-06-05
Switch to MJS extension

IvanGoncharov:
Suggested by @Andarist in #1366
Based on: https://github.com/leebyron/iterall/pull/34

@mjmahone I think it's the right moment for this change since there no active PRs from 3rd-party contributors.
Note: It doesn't break Git history since files were moved using `git mv`.


IvanGoncharov:
Looks like coverage is broken 😞

mjmahone:
I'm sorry, I don't understand the point of this (I am not very up to date on the latest happenings in JS). Why do we want to use `.mjs` instead of `.js`? That seems like it will break a lot of third-party tools (like syntax highlighting and other issues).

leebyron/iterall#34 seemed to be to allow in-line flow definitions. But we already have that for this repo, so I'm not sure what's gained here.

Andarist:
Just to clarify - I was by no means suggesting a switch to MJS. Personally I dislike it 😉 

My PR was solely about replacing custom bash script with a little tool that I've made for another project.

> But I think it's better to follow the rest of JS ecosystem and switch *.mjs files.

I'm not sure if rest of JS ecosystem is really going that way, from major tools only webpack has embraced it thus far as first class citizen.

I agree with @mjmahone that it might break a lot of other tooling and personally I would advise against it at this point in time.

IvanGoncharov:
Looks like `node 8` hides this feature under `--experimental-modules` but `node 10` works without it.
So we need to target `node 8` depreciation on December 31, 2019
BTW. I found a good article on a topic: https://medium.com/@giltayar/native-es-modules-in-nodejs-status-and-future-directions-part-i-ee5ea3001f71

Issue #1375, opened 2018-06-08
Pre-Release was published as latest


swernerx:
I think there is an issue happened with the last published package. 14.0.0-rc.1 is clearly a pre-release - even the changelog is saying so, but was published as "latest" which leads some automatic update tools to go to this version. Any chance to get this fixed?

swernerx:
Check yourself here: https://www.npmjs.com/package/graphql?activeTab=versions

IvanGoncharov:
@swernerx I tested `npm install graphql` and `yarn add graphql` and they both installs `14.0.0-rc.1`.
It's pretty bad I think we need to unpublish this package:
https://docs.npmjs.com/cli/unpublish
> `unpublish is only allowed with versions published in the last 24 hours`

@mjmahone Can you please unpublish package?
I think we still have a couple hours.

IvanGoncharov:
@mjmahone Also looks like `14.0.0-rc.2` from #1374 wasn't published to NPM at all.

IvanGoncharov:
Here is an article explaining what happened: https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420

NPM has very strange behavior by default.
As I understand Lee used `npm publish --tag rc` for previous RC releases that's why it `latest` wasn't changed.

IvanGoncharov:
NPM reports that package was published:
> 11 hours ago

So we still have 13 hours to find someone with NPM keys.

IvanGoncharov:
@mjmahone No need to unpublish anything we can simply change NPM tags:
```sh
npm dist-tags add graphql@0.13.2 latest
npm dist-tags add graphql@14.0.0-rc.1 rc
```

See `Whoops, I accidently published without a tag!` from here:
http://jbavari.github.io/blog/2015/10/16/using-npm-tags/

IvanGoncharov:
@mjmahone Can you please set correct tag?
It's a major issue since a lot of developers start to use RC versions unintentional by just doing: `yarn add graphql`.
Also, every new RC triggers update bots/scripts e.g. https://github.com/danger/peril/pull/309

mjmahone:
Shoot sorry I missed this. Trying to update tags now, though I'm not sure whether I have the right permission to do this. Figuring out how things make it to npm has been a bit of a struggle. Thank you @IvanGoncharov for providing much needed resources, and @swernerx for identifying the issue!

mjmahone:
[...]

PR #1376, opened 2018-06-08
Allow interfaces to have no implementors

mjmahone:
Reverts #1277, and implements spec removal from https://github.com/facebook/graphql/pull/459 (which reverts spec change https://github.com/facebook/graphql/pull/424). While it's an open question whether this validation is good, we need to provide an upgrade path for schemas that currently do not satisfy this validation rule.

### DIFF
--- a/src/type/__tests__/validation-test.js
+++ b/src/type/__tests__/validation-test.js
@@ -1205,7 +1205,7 @@ describe('Type System: Interface fields must have output types', () => {
     ]);
   });
 
-  it('rejects an interface not implemented by at least one object', () => {
+  it('accepts an interface not implemented by at least one object', () => {
     const schema = buildSchema(`
       type Query {
         test: SomeInterface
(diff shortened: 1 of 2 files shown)


mjmahone:
As a heads up, I'm going to merge this commit (even without the spec merge) and cut another rc, because this will make it possible for Facebook to upgrade (which means all Facebook's third-party libraries, like Relay, can upgrade too). I'm very open to figuring out some way to optionally include this schema validation, and provide an opt-in mechanism.

IvanGoncharov:
@mjmahone I'm still forming my position on spec changes. So I'm interested in exploring alternatives. 
Switching from a `type` to `interfaces` shouldn't affect client queries.

Can you please check if converting non-implemented interface to a `type` changes Flow typings?
Like converting this:
```graphql
interface Foo {
  foo: String
}
```
To this:
```graphql
type Foo {
  foo: String
}
```

IvanGoncharov:
> cut another rc, 

@mjmahone Can you please set correct npm tag:
https://github.com/graphql/graphql-js/issues/1375#issuecomment-395805384

PR #1377, opened 2018-06-08
RFC: Enforce Interfaces having at least one implementing Object

mjmahone:
It is unclear whether the spec validation from facebook/graphql#424 as implemented by #1277 is what we want to have for the GraphQL Schema long term. This RFC lets us explore that option more, to see whether the breaking schema change is the right path forward.


yaacovCR:
This was reverted from the spec in https://github.com/graphql/graphql-spec/pull/459.

Closing this old PR, feel free to reopen if needed.

PR #1378, opened 2018-06-09
Json serialize types

mjmahone:
Found a bug in `toJSON` not being defined for things like `GraphQLScalarType`. This was a breakage between `v0.13.2` and the latest RC.

This fixes the issue by:
1. Adding tests to make sure we don't break `toJSON` in the future
2. Moving some `toJSON` and `inspect` functions to be defined in the class instead of the prototype.


leebyron:
Moving from definition on the prototype to definition per class instance is definitely going to be a performance regression - have you figured out the root cause of the regression? Ideally we can find a fix that doesn't require the performance regression

mjmahone:
I am not yet 100% sure what's causing the regression. Taking a look now.

leebyron:
After some investigation, it seems like the babel transform changed and broke this, so its awesome to have a test that would catch something like that!

I'll try to patch this onto the same branch so we can land this one

leebyron:
bug fixed!

PR #1382, opened 2018-06-11
BREAKING/BUGFIX Strict coercion of scalar types

### LINKED ISSUE #1324: Spec Compliance: Strict input coercion for scalars

leebyron:
This no longer accepts incoming variable values in a potentially lossy way, mirroring the existing behavior for literals. This fixes an issue with GraphQL.js being not spec compliant.

This is breaking since servers which used to accept incorrect variable values will now return errors to clients.

Serialization of values is not affected in the same way, since this is not a client-visible behavior.

As a bonus, this adds unique serialization and coercion functions for the ID type, allowing to be more restrictive on numeric types and un-stringable object types, while directly supporting valueOf() methods (ala MongoDB). The changes to how the ID type serializes and coerces data could be potentially breaking.

Fixes #1324

### DIFF
--- a/src/type/scalars.js
+++ b/src/type/scalars.js
@@ -113,12 +113,21 @@ export const GraphQLFloat = new GraphQLScalarType({
 });
 
 function serializeString(value: mixed): string {
-  if (Array.isArray(value)) {
-    throw new TypeError(
-      `String cannot represent an array value: ${inspect(value)}`,
-    );
+  // Support serializing objects with custom valueOf() functions - a common way
+  // to represent an complex value which can be represented as a string
+  // (ex: MongoDB id objects).
+  const result =
+    value && typeof value.valueOf === 'function' ? value.valueOf() : value;
+  // Serialize string, number, and boolean values to a string, but do not
+  // attempt to coerce object, function, symbol, or other types as strings.
+  if (
+    typeof result !== 'string' &&
+    typeof result !== 'number' &&
+    typeof result !== 'boolean'
+  ) {
+    throw new TypeError(`String cannot represent value: ${inspect(result)}`);
   }
-  return String(value);
+  return String(result);
 }
 
 function coerceString(value: mixed): string {
@@ -175,11 +184,7 @@ function serializeID(value: mixed): string {
   // Support serializing objects with custom valueOf() functions - a common way
   // to represent an object identifier (ex. MongoDB).
   const result =
-    value &&
-    typeof value.valueOf === 'function' &&
-    value.valueOf !== Object.prototype.valueOf
-      ? value.valueOf()
-      : value;
+    value && typeof value.valueOf === 'function' ? value.valueOf() : value;
   if (
     typeof result !== 'string' &&
     (typeof result !== 'number' || !isInteger(result))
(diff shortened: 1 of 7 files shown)


src/type/scalars.js
```diff
@@ -108,38 +121,84 @@ function coerceString(value: mixed): string {
...
   },
 });
 
+function serializeID(value: mixed): string {
+  // Support serializing objects with custom valueOf() functions - a common way
+  // to represent an object identifier (ex. MongoDB).
+  const result =
+    value &&
+    typeof value.valueOf === 'function' &&
+    value.valueOf !== Object.prototype.valueOf
```
IvanGoncharov: I think we don't need this check since:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/valueOf
> By default, the valueOf method is inherited by every object descended from Object. Every built-in core object overrides this method to return an appropriate value. If an object has no primitive value, valueOf returns the object itself.

So it would be enough to just call `value.valueOf()` and it default one it will fail the next check.

src/type/scalars.js
```diff
@@ -99,7 +112,7 @@ export const GraphQLFloat = new GraphQLScalarType({
   },
 });
 
-function coerceString(value: mixed): string {
+function serializeString(value: mixed): string {
```
IvanGoncharov: Can we make it strict similar to `serializeID` so it would expect only string and object with either `valueOf` or `toString`?
Or aleast throw on `serializeString({})` instead of returning `[object Object]`?

leebyron: Yeah, I agree the [object Object] scenario is useless - it should only string coerce if there's a toString implementation worth calling

src/type/scalars.js
leebyron: You're probably right, I'll double check by running the tests

Issue #1379, opened 2018-06-09
Improve doc: graphl-js does not support interfaces and enums with buildSchema


Franco992:
As far as I understand, it is the recommended approach to use a root object when we build the schema with  buildSchema, because we cannot associate resolver functions when using buildSchema. With this approach the default resolvers are used (AFAIK). The problem is, that we can not use interfaces or union types (defined with SDL), because we can not associate the resolveType function - is that correct? If yes, how can we provide resolveType when using graphql-js with SDL?

The problem here is, that on graphql.org the examples are written for apollo-server, but the graphl-js examples show another approach!

Is there a reason why graphql-js does not use the same approach as graphql-tools (makeExecuteableSchema)? So writing "explicit" resolver functions and not using default resolvers?
What is the recommended approach (or the spec compliant approach)? It seems, that default resolvers are not specified in the spec?

This seems to be very confusing for people, when reading graphql.org and then implementing a graphl-js server.

I appreciate every answer, that clarifies some of my points! Thank you very much!

IvanGoncharov:
> As far as I understand, it is the recommended approach to use a root object when we build the schema with buildSchema, because we cannot associate resolver functions when using buildSchema. With this approach the default resolvers are used (AFAIK).

Right 👍 

> The problem is, that we can not use interfaces or union types (defined with SDL), because we can not associate the resolveType function - is that correct? If yes, how can we provide resolveType when using graphql-js with SDL?

Default `resolveType` is looking for `__typename` inside object so you either assign such property to rootObj or use getter function if you are using ES6 classes.
You actually motivated me to rewrite SWAPI example from this repo see #1384 
https://github.com/APIs-guru/graphql-js/blob/432adaea861d649c47c89b95ad2d0ffd4ab50e37/src/__tests__/starWarsSchema.js#L142-L144

> Is there a reason why graphql-js does not use the same approach as graphql-tools (makeExecuteableSchema)? So writing "explicit" resolver functions and not using default resolvers?
What is the recommended approach (or the spec compliant approach)? It seems, that default resolvers are not specified in the spec?

Default resolver approach used even in `graphql-tools` otherwise you would need to write resolver for every field. My understanding is that `makeExecuteableSchema` was written after `buildSchema` and I didn't find any attempt to contribute such functionality to `graphql-js`.

> What is the recommended approach (or the spec compliant approach)? It seems, that default resolvers are not specified in the spec?

[...]

PR #1384, opened 2018-06-11
SWAPI example: construct schema from SDL

IvanGoncharov:
Motivation #1379

### DIFF
--- a/src/__tests__/starWarsData.js
+++ b/src/__tests__/starWarsData.js
@@ -93,14 +93,9 @@ const droidData = {
  * These are Flow types which correspond to the schema.
  * They represent the shape of the data visited during field resolution.
  */
-export type Character = {
-  id: string,
-  name: string,
-  friends: Array<string>,
-  appearsIn: Array<number>,
-};
+export type CharacterData = HumanData | DroidData;
 
-export type Human = {
+export type HumanData = {
   type: 'Human',
   id: string,
   name: string,
@@ -109,7 +104,7 @@ export type Human = {
   homePlanet: string,
 };
 
-export type Droid = {
+export type DroidData = {
   type: 'Droid',
   id: string,
   name: string,
@@ -121,23 +116,15 @@ export type Droid = {
 /**
  * Helper function to get a character by ID.
  */
-function getCharacter(id) {
+export function getCharacter(id: string): Promise<CharacterData> {
   // Returning a promise just to illustrate GraphQL.js's support.
   return Promise.resolve(humanData[id] || droidData[id]);
 }
 
 /**
- * Allows us to query for a character's friends.
- */
-export function getFriends(character: Character): Array<Promise<Character>> {
-  // Notice that GraphQL accepts Arrays of Promises.
-  return character.friends.map(id => getCharacter(id));
-}
-
-/**
  * Allows us to fetch the undisputed hero of the Star Wars trilogy, R2-D2.
  */
-export function getHero(episode: number): Character {
+export function getHero(episode: number): CharacterData {
   if (episode === 5) {
     // Luke is the hero of Episode V.
     return luke;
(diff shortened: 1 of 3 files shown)


src/__tests__/starWarsSchema.js
```diff
@@ -26,278 +18,173 @@ import { getFriends, getHero, getHuman, getDroid } from './starWarsData';
...
+`);
+
+// FIXME: Patch enum defined in SDL, should be possible to do without hacks :(
+const episodeEnum = (StarWarsSchema.getType('Episode'): any);
+episodeEnum.getValue('NEWHOPE').value = 4;
+episodeEnum.getValue('EMPIRE').value = 5;
+episodeEnum.getValue('JEDI').value = 6;
+for (const enumValue of episodeEnum.getValues()) {
+  episodeEnum._valueLookup.set(enumValue.value, enumValue);
+}
```
IvanGoncharov: There no way to assign enum values after schema constructed.
@leebyron What do you this about adding one more option to `buidlSchema`/`extendSchema`?
```js
const schema = buildSchema(sdl, {
  extraConfig: {
    Episode: {
      values: {
        NEWHOPE: { value: 4 },
        EMPIRE: { value: 5 },
        JEDI: { value: 6 },
      },
    },
  },
});
```

leebyron: Likewise buildSchema does not provide resolvers or resolveType functions - I think this is probably more appropriate for the graphql-tools which provides additional runtime behavior atop buildSchema

src/__tests__/starWarsQuery-test.js
```diff
@@ -170,71 +139,65 @@ describe('Star Wars Query Tests', () => {
 
   describe('Using IDs and query parameters to refetch objects', () => {
     it('Allows us to query for Luke Skywalker directly, using his ID', async () => {
-      const query = `
-        query FetchLukeQuery {
+      const result = await executeQuery(`
```
leebyron: These are intended to be full integration tests, so they intentionally were running the full graphql function instead of just the execute step

src/__tests__/starWarsData.js
```diff
@@ -93,14 +93,9 @@ const droidData = {
  * These are Flow types which correspond to the schema.
  * They represent the shape of the data visited during field resolution.
  */
-export type Character = {
-  id: string,
-  name: string,
-  friends: Array<string>,
-  appearsIn: Array<number>,
-};
+export type CharacterData = HumanData | DroidData;
```
leebyron: I prefer without the `Data` suffix for readability

src/__tests__/starWarsQuery-test.js
IvanGoncharov: @leebyron it's a naming issue `executeQuery` is defined like this:
```js
function executeQuery(query, variableValues) {
  return graphql(StarWarsSchema, query, new Query(), null, variableValues);
}
```
Since now in majority of tests you need to supply 3 args instead of 2:
```js
const result = graphql(StarWarsSchema, query, new Query());
```
I decided to move it into utility function but now I see that it needs a better name. Do you have any suggestions?

guxima:
so these codes for v15??

src/__tests__/starWarsData.js
langpavel: No please, because of:
https://github.com/graphql/graphql-js/blob/eef0878d851f7b8a4392b031550a1080497c018c/src/__tests__/starWarsSchema.js#L12

src/__tests__/starWarsSchema.js
langpavel: I'm implementing this in https://github.com/graphql/graphql-js/pull/1987, it may look like this, if merged:
```js
const schema = buildSchema(sdl, {
  resolvers: {
    Episode: {
      NEWHOPE: 4,
      EMPIRE: 5,
      JEDI: 6,
    },
  },
});
```

yaacovCR:
Closing this old PR, feel free to reopen!

We can track efforts to build a `makeExecutableSchema()` type approach in https://github.com/graphql/graphql-js/pull/3047

Issue #1388, opened 2018-06-12
extend scalar is not supported


leebyron:
The spec describes this extension but this library does not yet implement it. We need this to comply with the spec for the next release, as mentioned in #1383

PR #1390, opened 2018-06-12
Fix `extensionASTNodes` assigment

IvanGoncharov:
Cleanup tests in preparation to #1388 and also found a bug with `extensionASTNodes`

### DIFF
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -9,7 +9,7 @@ import { describe, it } from 'mocha';
 import { expect } from 'chai';
 import dedent from '../../jsutils/dedent';
 import { extendSchema } from '../extendSchema';
-import { parse, print } from '../../language';
+import { parse, print, DirectiveLocation } from '../../language';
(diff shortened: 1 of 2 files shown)


src/utilities/__tests__/extendSchema-test.js
```diff
@@ -884,11 +980,11 @@ describe('extendSchema', () => {
   });
 
   it('does not allow replacing a default directive', () => {
-    const ast = parse(`
+    const sdl = `
```
mjmahone: I like this, but want to make sure I understand our guidelines for `sdl` vs `ast`: `sdl` => used when we have `TypeSystemDefinitionNode` types, and `ast` for `ExecutableDefinitionNode` types?

IvanGoncharov: @mjmahone SDL - is a string containing `TypeSystemDefinitions`
AST - a result of parse being called on any string in GraphQL syntax.

leebyron:
Awesome thanks for spotting and catching this before release!

PR #1391, opened 2018-06-12
Convert 'GraphQL*Config' to exact types

IvanGoncharov:
I was wondering why #1390 happened and I think we need to make config types exact.
This PR dependence on #1390 and should be merged after it.

### DIFF
--- a/src/type/introspection.js
+++ b/src/type/introspection.js
@@ -33,7 +33,6 @@ import type { GraphQLField } from './definition';
 
 export const __Schema = new GraphQLObjectType({
   name: '__Schema',
-  isIntrospection: true,
   description:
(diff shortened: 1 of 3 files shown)


src/type/introspection.js
```diff
@@ -33,7 +33,6 @@ import type { GraphQLField } from './definition';
 
 export const __Schema = new GraphQLObjectType({
   name: '__Schema',
-  isIntrospection: true,
```
IvanGoncharov: Added in 379a3084392179d2cae92c211a4559f9836299c6 but doesn't required now.

IvanGoncharov:
@leebyron Rebased

leebyron:
Makes sense to me. I think the original intent was that this config could hold additional data, but that’s not actually a pattern in use. So this a clear improvement

PR #1400, opened 2018-06-21
Document the contextValue argument

Cito:


leebyron:
Awesome thanks for improving documentation!

Issue #1341, opened 2018-05-08
Is it possible to make findBreakingChange return line numbers?


JCMais:
This would help to create some tools, like github pull request checks.

JCMais:
I will try to work on a POC for this and open a PR to receive feedback

IvanGoncharov:
> I will try to work on a POC for this and open a PR to receive feedback, my plan is to expose both ast nodes, previous one (if any), and new one (if any), depending on the ones that makes sense to the change.

@JCMais Sounds great 👍 I would recommend you to open WIP PR.

> The thing is, the ast nodes are only available when building the schema from a SDL. The tests currently works by creating the types directly, should I create another describe section for the same tests but using SDL with parse? Or create a separated test file (the current one is really big already)?

I think such tests would benefit from using SDL and they were written using `GraphQLSchema` just because SDL was the experimental feature at that time.
Also instead of `buildASTSchema(parse(sdl))` you can call `buildSchema` that accepts SDL string directly.
To make such changes easier to review can you please submit tests conversion as a separate PR.
Please convert a couple of test cases and open a PR so you would get feedback early on.

JCMais:
> To make such changes easier to review can you please submit tests conversion as a separate PR.
will do that this week, thanks for the feedback.

buoyantair:
@leebyron This issue seems to be resolved, close it? :3

JCMais:
it's not resolved, I've only updated the tests to build from SDL instead of using type constructors, the feature mentioned on this issue is still missing, didn't have time to finish it yet

PR #1406, opened 2018-06-26
findBreakingChanges tests: use buildSchema when possible

JCMais:
Instead of generating the schema by using the type constructors, this changes the tests to use `buildSchema` with the correct SDL for each test.

Related to https://github.com/graphql/graphql-js/issues/1341

SDL was generated using the following code:
```js
import { printSchema } from './index';

console.log(
  '    const oldSchema = buildSchema(`\n' +
  printSchema(oldSchema).split('\n').map(s => s.padStart(s.length + 6)).slice(0, -1).join('\n') +
  '\n    `);',
);
console.log('---');
console.log(
  '    const newSchema = buildSchema(`\n' +
  printSchema(newSchema).split('\n').map(s => s.padStart(s.length + 6)).slice(0, -1).join('\n') +
  '\n    `);',
);
```

So it's based solely on how the previous `oldSchema` and `newSchema` of each test case were printed by `printSchema`.

### DIFF
--- a/src/utilities/__tests__/findBreakingChanges-test.js
+++ b/src/utilities/__tests__/findBreakingChanges-test.js
@@ -503,8 +503,20 @@ describe('findBreakingChanges', () => {
     const oldSchema = buildSchema(`
       type Type1 {
         field1(
-          arg1: String, arg2: String, arg3: [String], arg4: String, arg5: String!, arg6: String!, arg7: [Int]!,
-          arg8: Int, arg9: [Int], arg10: [Int!], arg11: [Int], arg12: [[Int]], arg13: Int!, arg14: [[Int]!],
+          arg1: String
+          arg2: String
+          arg3: [String]
+          arg4: String
+          arg5: String!
+          arg6: String!
+          arg7: [Int]!
+          arg8: Int
+          arg9: [Int]
+          arg10: [Int!]
+          arg11: [Int]
+          arg12: [[Int]]
+          arg13: Int!
+          arg14: [[Int]!]
           arg15: [[Int]!]
         ): String
       }
@@ -517,8 +529,21 @@ describe('findBreakingChanges', () => {
     const newSchema = buildSchema(`
       type Type1 {
         field1(
-          arg1: Int, arg2: [String], arg3: String, arg4: String!, arg5: Int, arg6: Int!, arg7: [Int], arg8: [Int]!,
-          arg9: [Int!], arg10: [Int], arg11: [[Int]], arg12: [Int], arg13: [Int]!, arg14: [[Int]], arg15: [[Int!]!]
+          arg1: Int
+          arg2: [String]
+          arg3: String
+          arg4: String!
+          arg5: Int
+          arg6: Int!
+          arg7: [Int]
+          arg8: [Int]!
+          arg9: [Int!]
+          arg10: [Int]
+          arg11: [[Int]]
+          arg12: [Int]
+          arg13: [Int]!
+          arg14: [[Int]]
+          arg15: [[Int!]!]
          ): String
       }
 


src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -1056,215 +714,116 @@ describe('findBreakingChanges', () => {
...
-      locations: [DirectiveLocation.FIELD_DEFINITION],
-      args: {
-        arg1: {
-          name: 'arg1',
-          type: GraphQLNonNull(GraphQLBoolean),
-        },
-      },
-    });
-    const directiveRemovedLocationOld = new GraphQLDirective({
-      name: 'Directive Name',
```
JCMais: This directive had incorrect name, according to the spec, spaces should not be allowed on directive name:
https://facebook.github.io/graphql/draft/#DirectiveDefinition
https://facebook.github.io/graphql/draft/#Name

src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -1307,7 +866,7 @@ describe('findBreakingChanges', () => {
       },
       {
         type: BreakingChangeType.DIRECTIVE_REMOVED,
-        description: 'skip was removed',
+        description: 'DirectiveThatIsRemoved was removed',
```
JCMais: this was necessary since `printSchema` does not print GraphQL default directives, so I've added a directive to test the same scenario.

src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -1330,18 +889,19 @@ describe('findBreakingChanges', () => {
...
-    });
+    const newSchema = buildSchema(`
+      directive @DirectiveThatStays on FIELD_DEFINITION
+    `);
 
     expect(findRemovedDirectives(oldSchema, newSchema)).to.eql([
       {
         type: BreakingChangeType.DIRECTIVE_REMOVED,
-        description: `${GraphQLIncludeDirective.name} was removed`,
+        description: `DirectiveThatIsRemoved was removed`,
```
JCMais: Change made for the same motive than above, `printSchema` does not print GraphQL default directives

src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -1056,215 +714,116 @@ describe('findBreakingChanges', () => {
...
-      },
-    });
-
-    const directiveThatIsRemoved = GraphQLSkipDirective;
-    const directiveThatRemovesArgOld = new GraphQLDirective({
-      name: 'DirectiveThatRemovesArg',
-      locations: [DirectiveLocation.FIELD_DEFINITION],
-      args: {
-        arg1: {
-          name: 'arg1',
```
JCMais: FYI this was giving `undefined` type to `arg1`, so I've added something as value for that arg on the SDL below.

JCMais:
I have not changed the test `'should detect if a directive was implicitly removed'` to use `buildSchema` since it kinda of is testing internal `GraphQLSchema` behaviour. So I've kept it the way it is. 


Another point is that the test `should detect locations removed directives within a schema` has the test coverage for `should detect locations removed from a directive`, so IDK if it's necessary to have both.

src/utilities/__tests__/findBreakingChanges-test.js
IvanGoncharov: Make sense 👍

src/utilities/__tests__/findBreakingChanges-test.js
IvanGoncharov: 👍

src/utilities/__tests__/findBreakingChanges-test.js
IvanGoncharov: 👍

src/utilities/__tests__/findBreakingChanges-test.js
IvanGoncharov: 👍

IvanGoncharov:
@JCMais Thanks for PR 👍 
Everything looks good.
There is something that could be improved but they are unrelated to SDL conversion.
However, there is one important problem that `printSchema` output types sorted by names and that makes the test harder to read.
Types could be rearranged in separate PR but it will complicate commit history.
So it would be great if you rearrange types in order that makes sense when you read SDL top to bottom.

JCMais:
@IvanGoncharov I've changed the ordering of the types, can you check again if it's more readable?

src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -574,215 +451,81 @@ describe('findBreakingChanges', () => {
...
-    });
-
-    const newSchema = new GraphQLSchema({
-      query: queryType,
-      types: [newType],
-    });
+    const oldSchema = buildSchema(`
+      type Type1 {
+        field1(
+          arg1: String, arg2: String, arg3: [String], arg4: String, arg5: String!, arg6: String!, arg7: [Int]!,
```
IvanGoncharov: nit: one argument per line without commas would be consistent with prettier:
https://github.com/prettier/prettier/blob/master/tests/graphql_arguments/__snapshots__/jsfmt.spec.js.snap#L12-L16

mjmahone:
Thanks for making these tests more readable! I had been doing this with my own tests for awhile, so it's great to see this clear improvement in the main repo.

PR #1410, opened 2018-06-30
Validate schema use new extension ast nodes

### LINKED ISSUE #1387: validateSchema doesn't use nodes from new extensionASTNodes properties
As mentioned in #1383 - validateSchema currently misses extended properties added in #1373

IvanGoncharov:
Fixes #1387 + minor cleanup

### DIFF
--- a/src/type/validate.js
+++ b/src/type/validate.js
@@ -169,15 +169,14 @@ function getOperationTypeNode(
 }
 
 function validateDirectives(context: SchemaValidationContext): void {
-  const directives = context.schema.getDirectives();
-  directives.forEach(directive => {
+  for (const directive of context.schema.getDirectives()) {
     // Ensure all directives are in fact GraphQL directives.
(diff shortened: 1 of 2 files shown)


src/type/validate.js
```diff
@@ -169,15 +169,14 @@ function getOperationTypeNode(
 }
 
 function validateDirectives(context: SchemaValidationContext): void {
-  const directives = context.schema.getDirectives();
-  directives.forEach(directive => {
```
mjmahone: Is there a concrete reason to make this change? `.forEach` is how most things are done in this library, and I'd prefer to maintain consistency.

IvanGoncharov: @mjmahone I used `for of` in all new code I wrote for readability and performance reasons but I fully agree with consistency argument so I created #1423 where I converted all other instances of `Array.forEach`.

PR #1411, opened 2018-07-02
Unify optional callbacks validation

IvanGoncharov:


src/type/definition.js
```diff
@@ -757,11 +756,6 @@ function isPlainObj(obj) {
   return obj && typeof obj === 'object' && !Array.isArray(obj);
 }
 
-// If a resolver is defined, it must be a function.
-function isValidResolver(resolver: mixed): boolean {
```
mjmahone: Should we keep `isValidResolver`? I'm not totally sure what the advantage of inlining it is

src/type/definition.js
```diff
@@ -903,12 +897,11 @@ export class GraphQLInterfaceType {
...
     this._fields = defineFieldMap.bind(undefined, config);
     invariant(typeof config.name === 'string', 'Must provide name.');
-    if (config.resolveType) {
-      invariant(
-        typeof config.resolveType === 'function',
-        `${this.name} must provide "resolveType" as a function.`,
-      );
-    }
+    invariant(
+      config.resolveType == null || typeof config.resolveType === 'function',
```
mjmahone: All of these could now become `isValidResolver`

src/type/definition.js
IvanGoncharov: One reason would be that you can't share it without changing the name since `isTypeOf` not a resolver. Plus it also confusing since `null` and `undefined` are not valid resolvers but an explicit absence of resolver.

So the best name that I come up is `isOptionalFunction` which is the same as to just write `resolver == null || typeof resolver === 'function'`.
I think it better to stay explicit.

mjmahone: That's a reasonable argument. I'm convinced.

PR #1412, opened 2018-07-03
Switch GraphQLScalarType methods to callbacks

IvanGoncharov:


src/utilities/extendSchema.js
```diff
@@ -439,9 +439,9 @@ export function extendSchema(
       description: type.description,
       astNode: type.astNode,
       extensionASTNodes,
-      serialize: type._scalarConfig.serialize,
-      parseValue: type._scalarConfig.parseValue,
-      parseLiteral: type._scalarConfig.parseLiteral,
```
IvanGoncharov: At the moment to copy scalar config you should either use `type._scalarConfig.serialize` or `type.serialize.bind(type)`.
This is a big problem for 3rd-party libraries doing schema transformations.

mjmahone:
@IvanGoncharov sorry I feel like I'm missing context on this: it seems like an improvement based purely on the fact that the flow types are better, but other than that I am not sure I see why this is desired? The change is straightforward enough I may merge regardless, but wanted to understand your logic.

IvanGoncharov:
@mjmahone For example, I want to add a prefix to every type in a schema:
```js
function prefixType(type: GraphQLNamedType, prefix: string): GraphQLNamedType {
  if (isScalarType(type)) {
      name: prefix + type.name,
      // ...
      parseLiteral: type.parseLiter,
    });
  }
  // ...
}
```
And `parseLiteral: type.parseLiteral` is a bug because `parseValue`, `parseLiteral` and `serialize` are method and not a callbacks. And it's very different from all other type-related callbacks like `resolve`, `subscribe`, `isTypeOf`, etc.


Moreover you either need to use private property `_scalarConfig` or to create wrapper function using `bind` that would be called every time we `graphql-js` parses literal. Also if we use `bind` that mean newly created scalar has refence to old scalar so they can be garbage collected only together.

src/type/definition.js
```diff
@@ -563,26 +566,6 @@ export class GraphQLScalarType {
     }
   }
 
-  // Serializes an internal value to include in a response.
-  serialize(value: mixed): mixed {
-    const serializer = this._scalarConfig.serialize;
```
IvanGoncharov: >  it seems like an improvement based purely on the fact that the flow types are better

@mjmahone This wrapper function depends on `this` so before proposed change you can't do:
```js
const serialize = GraphQLString.serialize;
serialize(5);
// TypeError: Cannot read property '_scalarConfig' of undefined
```
Or more complicated scenario:
```js

mjmahone:
Great, thanks for the context @IvanGoncharov !

PR #1414, opened 2018-07-10
Scalar coercion cleanup

IvanGoncharov:
Idea is to make coercion code more strict and explicit.

### DIFF
--- a/src/execution/__tests__/schema-test.js
+++ b/src/execution/__tests__/schema-test.js
@@ -91,7 +91,7 @@ describe('Execute: Handles execution with a complex schema', () => {
     function article(id) {
       return {
         id,
-        isPublished: 'true',
+        isPublished: true,
         author: johnSmith,
         title: 'My Article ' + id,
         body: 'This is a post',
--- a/src/type/__tests__/serialization-test.js
+++ b/src/type/__tests__/serialization-test.js
@@ -69,7 +69,7 @@ describe('Type System: Scalar coercion', () => {
       'Int cannot represent non-integer value: Infinity',
     );
     expect(() => GraphQLInt.serialize([5])).to.throw(
-      'Int cannot represent an array value: [5]',
+      'Int cannot represent non-integer value: [5]',
     );
   });
 
@@ -98,7 +98,7 @@ describe('Type System: Scalar coercion', () => {
       'Float cannot represent non numeric value: ""',
     );
     expect(() => GraphQLFloat.serialize([5])).to.throw(
-      'Float cannot represent an array value: [5]',
+      'Float cannot represent non numeric value: [5]',
     );
   });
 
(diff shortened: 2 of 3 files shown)


src/type/__tests__/serialization-test.js
```diff
@@ -126,19 +117,41 @@ describe('Type System: Scalar coercion', () => {
...
+    const badObjValue = {};
+    expect(() => GraphQLString.serialize(badObjValue)).to.throw(
+      'String cannot represent value: {}',
+    );
   });
 
   it('serializes output as Boolean', () => {
-    expect(GraphQLBoolean.serialize('string')).to.equal(true);
-    expect(GraphQLBoolean.serialize('false')).to.equal(true);
-    expect(GraphQLBoolean.serialize('')).to.equal(false);
```
IvanGoncharov: Converting non-empty strings to `true` is confusing because `GraphQLString` coerce  `true` into `"true"`. Also spec doesn't mention this coercion:
http://facebook.github.io/graphql/June2018/#sec-Boolean

src/execution/__tests__/schema-test.js
```diff
@@ -91,7 +91,7 @@ describe('Execute: Handles execution with a complex schema', () => {
     function article(id) {
       return {
         id,
-        isPublished: 'true',
```
IvanGoncharov: one more example why current coercion rules looks strange and confusing:
`"true" => true`
`"false" => false`

src/type/__tests__/serialization-test.js
```diff
@@ -171,8 +185,8 @@ describe('Type System: Scalar coercion', () => {
       'ID cannot represent value: true',
     );
 
-    expect(() => GraphQLID.serialize(-1.1)).to.throw(
-      'ID cannot represent value: -1.1',
```
mjmahone: Oh that's a great change: making the test clear the reason it fails is floating-point coercion not negative-number coercion.

src/type/scalars.js
```diff
@@ -22,13 +22,14 @@ const MAX_INT = 2147483647;
 const MIN_INT = -2147483648;
 
 function serializeInt(value: mixed): number {
-  if (Array.isArray(value)) {
-    throw new TypeError(
-      `Int cannot represent an array value: ${inspect(value)}`,
-    );
+  let num = value;
```
mjmahone: I feel like this logic could be simplified. It's really strange to me to have an early-return in an else block:
```
if (typeof value === 'boolean') {
  return value ? 1 : 0;
}

const num = (typeof value === 'string' && value !== '') ?
  Number(value) : 
  value;

if (!isInteger(num)) {
```

src/type/scalars.js
```diff
@@ -74,13 +75,14 @@ export const GraphQLInt = new GraphQLScalarType({
 });
 
 function serializeFloat(value: mixed): number {
-  if (Array.isArray(value)) {
-    throw new TypeError(
-      `Float cannot represent an array value: ${inspect(value)}`,
-    );
+  let num = value;
```
mjmahone: same idea here

IvanGoncharov:
> This looks solid, I'm happy to merge once the logic is a little bit cleaner.

@mjmahone Thanks for review 👍 Fixed 🔧

PR #1418, opened 2018-07-13
Update deps + switch to fixed versions in package.json

IvanGoncharov:


package.json
```diff
@@ -48,34 +48,34 @@
...
-    "@babel/plugin-transform-classes": "^7.0.0-beta.51",
-    "@babel/plugin-transform-destructuring": "^7.0.0-beta.51",
-    "@babel/plugin-transform-flow-strip-types": "^7.0.0-beta.51",
-    "@babel/plugin-transform-modules-commonjs": "^7.0.0-beta.51",
-    "@babel/plugin-transform-spread": "^7.0.0-beta.51",
-    "@babel/polyfill": "^7.0.0-beta.51",
-    "@babel/preset-env": "^7.0.0-beta.51",
-    "@babel/register": "^7.0.0-beta.51",
-    "babel-eslint": "8.2.4",
+    "@babel/cli": "7.0.0-beta.53",
```
IvanGoncharov: Even though we use `yarn` I think it's better to use explicit versions in `package.json`.

mjmahone: For devDependencies you're probably right. That's definitely not true for regular dependencies, as it can cause people to needlessly download multiple compatible versions of the same library.

PR #1421, opened 2018-07-16
Disable eslint rules that can conflict with prettier

IvanGoncharov:
Continuation of 40f73fd
Rules was disabled using this tool:
https://github.com/prettier/eslint-config-prettier#cli-helper-tool


IvanGoncharov:
Motivated by this change: https://github.com/graphql/graphql-js/pull/1417/files/917c2f606fcdc2182d86868e7f23def7ae02f853#diff-1dc6ee56b778cd91e0327b52aaeaa1b9

.eslintrc
```diff
@@ -42,33 +42,33 @@
   "rules": {
     "prettier/prettier": 2,
 
-    "flowtype/space-after-type-colon": [2, "always"],
-    "flowtype/space-before-type-colon": [2, "never"],
-    "flowtype/space-before-generic-bracket": [2, "never"],
-    "flowtype/union-intersection-spacing": [2, "always"],
+    "flowtype/space-after-type-colon": 0,
```
mjmahone: Do we get the same result if we just remove these lines completely? If so, it makes it a lot easier to parse the file, so let's delete them all.

If instead they default to `1`, then yeah this is the way to go.

mjmahone: Sounds good! Thanks for the context.

PR #1422, opened 2018-07-16
Fragment variables: enforce fragment variable validation

mjmahone:
Treats experimental fragments with variable definitions as their own top-level
"executable definition": this means a query cannot include a variable
definition found solely beneath a fragment-variable definition, and
a fragment must include all variable definitions that exist under it,
excluding other variables defined by other fragments.

Has no repercussions outside of fragment variable definitions, besides allowing recursive fragment lookups. One note: I am not sure how the `NoUnusedFragments` validation is supposed to work: internally we don't use it, as we have many use cases for unused fragments. I suspect fragments with variable definitions are a different definition type, and should always be treated as their own self-contained definition-trees (so we should be doing the same thing as below inside `NoUnusedFragments`). But I'm not positive about that behavior, whereas I am positive about the behavior of unused/undefined variables.

### DIFF
--- a/src/validation/ValidationContext.js
+++ b/src/validation/ValidationContext.js
@@ -155,7 +155,7 @@ export default class ValidationContext {
             const fragment = this.getFragment(fragName);
             if (
               fragment &&
-              !this.isExecutableDefinitionWithVariables(fragment)
+              !isExperimentalFragmentWithVariableDefinitions(fragment)
             ) {
               fragments.push(fragment);
               nodesToVisit.push(fragment.selectionSet);
@@ -244,15 +244,10 @@ export default class ValidationContext {
   getArgument(): ?GraphQLArgument {
     return this._typeInfo.getArgument();
   }
+}
 
-  // All OperationDefinitions, or FragmentDefinitions with variable definitions
-  isExecutableDefinitionWithVariables(
-    definition: ExecutableDefinitionNode,
-  ): boolean {
-    return (
-      definition.kind === Kind.OPERATION_DEFINITION ||
-      (definition.variableDefinitions != null &&
-        definition.variableDefinitions.length > 0)
-    );
-  }
+function isExperimentalFragmentWithVariableDefinitions(
+  ast: FragmentDefinitionNode,
+): boolean %checks {
+  return ast.variableDefinitions != null && ast.variableDefinitions.length > 0;
 }
--- a/src/validation/ValidationContext.js
+++ b/src/validation/ValidationContext.js
@@ -13,6 +13,7 @@ import { visit, visitWithTypeInfo } from '../language/visitor';
 import { Kind } from '../language/kinds';
 import type {
   DocumentNode,
+  ExecutableDefinitionNode,
   OperationDefinitionNode,
   VariableNode,
   SelectionSetNode,
@@ -50,12 +51,12 @@ export default class ValidationContext {
   _fragments: ObjMap<FragmentDefinitionNode>;
   _fragmentSpreads: Map<SelectionSetNode, $ReadOnlyArray<FragmentSpreadNode>>;
   _recursivelyReferencedFragments: Map<
-    OperationDefinitionNode,
+    ExecutableDefinitionNode,
     $ReadOnlyArray<FragmentDefinitionNode>,
   >;
   _variableUsages: Map<NodeWithSelectionSet, $ReadOnlyArray<VariableUsage>>;
   _recursiveVariableUsages: Map<
-    OperationDefinitionNode,
+    ExecutableDefinitionNode,
     $ReadOnlyArray<VariableUsage>,
   >;
 
@@ -129,14 +130,21 @@ export default class ValidationContext {
     return spreads;
   }
 
+  /*
+   * Finds all fragments referenced via the definition, recursively.
+   *
+   * NOTE: if experimentalFragmentVariables are being used, it excludes all
+   * fragments with their own variable definitions: these are considered their
+   * own "root" executable definition.
+   */
   getRecursivelyReferencedFragments(
-    operation: OperationDefinitionNode,
+    definition: ExecutableDefinitionNode,
   ): $ReadOnlyArray<FragmentDefinitionNode> {
-    let fragments = this._recursivelyReferencedFragments.get(operation);
+    let fragments = this._recursivelyReferencedFragments.get(definition);
     if (!fragments) {
       fragments = [];
       const collectedNames = Object.create(null);
-      const nodesToVisit: Array<SelectionSetNode> = [operation.selectionSet];
+      const nodesToVisit: Array<SelectionSetNode> = [definition.selectionSet];
       while (nodesToVisit.length !== 0) {
         const node = nodesToVisit.pop();
         const spreads = this.getFragmentSpreads(node);
@@ -145,14 +153,17 @@ export default class ValidationContext {
           if (collectedNames[fragName] !== true) {
             collectedNames[fragName] = true;
             const fragment = this.getFragment(fragName);
-            if (fragment) {
+            if (
+              fragment &&
+              !this.isExecutableDefinitionWithVariables(fragment)
+            ) {
               fragments.push(fragment);
               nodesToVisit.push(fragment.selectionSet);
             }
           }
         }
       }
-      this._recursivelyReferencedFragments.set(operation, fragments);
+      this._recursivelyReferencedFragments.set(definition, fragments);
     }
     return fragments;
   }
@@ -181,20 +192,27 @@ export default class ValidationContext {
     return usages;
   }
 
+  /*
+   * Finds all variables used by the definition, recursively.
+   *
+   * NOTE: if experimentalFragmentVariables are being used, it excludes all
+   * fragments with their own variable definitions: these are considered their
+   * own "root" executable definition.
+   */
   getRecursiveVariableUsages(
-    operation: OperationDefinitionNode,
+    definition: ExecutableDefinitionNode,
   ): $ReadOnlyArray<VariableUsage> {
-    let usages = this._recursiveVariableUsages.get(operation);
+    let usages = this._recursiveVariableUsages.get(definition);
     if (!usages) {
-      usages = this.getVariableUsages(operation);
-      const fragments = this.getRecursivelyReferencedFragments(operation);
+      usages = this.getVariableUsages(definition);
+      const fragments = this.getRecursivelyReferencedFragments(definition);
       for (let i = 0; i < fragments.length; i++) {
         Array.prototype.push.apply(
           usages,
           this.getVariableUsages(fragments[i]),
         );
       }
-      this._recursiveVariableUsages.set(operation, usages);
+      this._recursiveVariableUsages.set(definition, usages);
     }
     return usages;
   }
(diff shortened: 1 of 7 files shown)


src/validation/rules/NoUndefinedVariables.js
```diff
@@ -20,36 +21,50 @@ export function undefinedVarMessage(varName: string, opName: ?string): string {
...
  *
- * A GraphQL operation is only valid if all variables encountered, both directly
- * and via fragment spreads, are defined by that operation.
+ * A GraphQL definition is only valid if all variables encountered, both
+ * directly and via fragment spreads, are defined by that definition.
+ *
+ * NOTE: if experimentalFragmentVariables are used, then fragments with
+ * variables defined are considered independent "executable definitions".
+ * If a fragment defines at least 1 variable, it must define all recursively
+ * vused ariables, excluding other fragments with variables defined.
```
IvanGoncharov: `vused` typo

src/validation/ValidationContext.js
```diff
@@ -226,4 +244,15 @@ export default class ValidationContext {
   getArgument(): ?GraphQLArgument {
     return this._typeInfo.getArgument();
   }
+
+  // All OperationDefinitions, or FragmentDefinitions with variable definitions
+  isExecutableDefinitionWithVariables(
+    definition: ExecutableDefinitionNode,
+  ): boolean {
```
IvanGoncharov: This function doesn't use `this` it should be a standalone function, not a class method.

IvanGoncharov: Also maybe it worth to add `%checks` so Flow will have better type coverage.

src/validation/rules/NoUnusedVariables.js
```diff
@@ -23,38 +24,53 @@ export function unusedVariableMessage(
...
-      },
+  const executableDefinitionVisitor = {
+    enter(definition: ExecutableDefinitionNode) {
+      if (!context.isExecutableDefinitionWithVariables(definition)) {
+        return;
+      }
+      variableDefs = [];
     },
+    leave(definition: ExecutableDefinitionNode) {
+      if (!context.isExecutableDefinitionWithVariables(definition)) {
```
IvanGoncharov: just check that `variableDefs` is not empty since if it's empty it couldn't be any unused variables.

src/validation/rules/NoUnusedVariables.js
```diff
@@ -23,38 +24,53 @@ export function unusedVariableMessage(
...
-              new GraphQLError(unusedVariableMessage(variableName, opName), [
-                variableDef,
-              ]),
-            );
-          }
-        });
-      },
+  const executableDefinitionVisitor = {
+    enter(definition: ExecutableDefinitionNode) {
+      if (!context.isExecutableDefinitionWithVariables(definition)) {
```
IvanGoncharov: not needed see `leave`.

src/validation/rules/NoUndefinedVariables.js
```diff
@@ -20,36 +21,50 @@ export function undefinedVarMessage(varName: string, opName: ?string): string {
...
-      enter() {
-        variableNameDefined = Object.create(null);
-      },
-      leave(operation) {
-        const usages = context.getRecursiveVariableUsages(operation);
+  const executableDefinitionVisitor = {
+    enter(definition: ExecutableDefinitionNode) {
+      if (!context.isExecutableDefinitionWithVariables(definition)) {
+        return;
+      }
```
IvanGoncharov: not needed since there is nothing bad in resetting value of `variableNameDefined` even on fragments without arguments.

src/validation/rules/NoUndefinedVariables.js
```diff
@@ -20,36 +21,50 @@ export function undefinedVarMessage(varName: string, opName: ?string): string {
...
+      variableNameDefined = Object.create(null);
+    },
+    leave(definition: ExecutableDefinitionNode) {
+      if (!context.isExecutableDefinitionWithVariables(definition)) {
```
IvanGoncharov: As I understand, `variableNameDefined` was added to not depend on `variableDefinitions` property at the same time `isExecutableDefinitionWithVariables` introduces such dependency.
Would be better to replace with:
```js
if (definition.kind ===  Kind.FRAGMENT && Object.keys(variableNameDefined).length === 0)
```

mjmahone: Interesting, that would make sense. Will make the changes

src/validation/rules/NoUndefinedVariables.js
mjmahone: Ah right, because this function only cares about finding undefined variables, and if none are defined, it doesn't matter (for fragments). I think there might be a world in which FragmentWithVariableDefinitionNode is a different AST node than FragmentDefinitionNode, but even in that world this wouldn't be needed :)

src/validation/__tests__/harness.js
```diff
@@ -422,12 +422,20 @@ export const testSchema = new GraphQLSchema({
 });
 
 function expectValid(schema, rules, queryString) {
-  const errors = validate(schema, parse(queryString), rules);
+  const errors = validate(
+    schema,
+    parse(queryString, { experimentalFragmentVariables: true }),
```
IvanGoncharov: There are maintainers of GraphQL 3rd-party implementations that use `graphql-js` test suite as the refence so I don't think we should enable the experimental feature by default for all tests.
It's better to just add `options` as 4th argument, it would make it more explicit and also allow to easily grep for all tests cases with the particular experimental feature.

src/validation/ValidationContext.js
```diff
@@ -129,14 +130,21 @@ export default class ValidationContext {
...
   }
 
+  /*
+   * Finds all fragments referenced via the definition, recursively.
+   *
+   * NOTE: if experimentalFragmentVariables are being used, it excludes all
+   * fragments with their own variable definitions: these are considered their
+   * own "root" executable definition.
+   */
   getRecursivelyReferencedFragments(
```
IvanGoncharov: I'm very concern about changes in this function since it's called `getRecursivelyReferencedFragments` and is part of public API.
+ it's break `NoUnusedFragments`

Alternative solution would be to extract `visitReferencedFragmentsRecursively` that accept visitor function and if you return `false` it doesn't look inside current fragment.

IvanGoncharov:
> I suspect fragments with variable definitions are a different definition type, and should always be treated as their own self-contained definition-trees (so we should be doing the same thing as below inside NoUnusedFragments). But I'm not positive about that behavior, whereas I am positive about the behavior of u

@mjmahone As I understand from #1141 experimental fragments are not executed unless they are referenced inside at least one query so I don't see any difference with normal fragments in that sense.

I also concern with lack of PR/RFC in specification repo because maintainer of 3rd-party libraries use this repo as refence and they also implemented this feature:
https://github.com/webonyx/graphql-php/pull/248/commits/949b85367826b30dd7c19d5578281fed5dc93fd4
https://github.com/sangria-graphql/sangria/commit/89927446c4ba03c0deff994e24ed629e94cd87b0
and probably others.

So I'm worry that we are repeating IDL/SDL mistake then solution that everyone adopted significantly differ from what was added into the spec and entire comunity now goes through painful migration process. This feature could have even bigger effect if people would start sending queries with such fragments to the server. I think it's very important to comminicate the purpose of this feature and it's current status to community members outside of Facebook by openning minimal PR on a spec repo.

src/validation/ValidationContext.js
mjmahone: It only breaks `NoUnusedFragments` if you're using fragments with variable definitions. I *think* this is OK, as fragments with variable definitions are their own independent "unit": they have semantic meaning that is very different from normal fragments without variable definitions. An unused fragment that has variable definitions I see as being "stand-alone" in the same way a query is "stand-alone": you could reasonably figure out a way to persist these to the server and execute them, provided you started on the correct type.

```
fragment A($id: ID!) on Query {
  node(id: $id) {
    name
  }
}
```

is, and in my mind should be, identical to
```
query A($id: ID!) {
  node(id: $id) {
    name
  }
}
```

I'm very OK with breaking people who are actually using fragments with variable definitions.

I can see the appeal to extracting it out, but I'd much prefer taking an opinionated stance about what a fragment with variable definitions is. It will also make it easier to make an RFC in the spec for them.

IvanGoncharov: Ok, in that case, it's outside of scope for `graphql-js` and would make sense to discuss it in PR/RFC.

Just a quick note:
> and execute them, provided you started on the correct type

There is no one to one matching between root types and operation types so you can write:
```graphql
schema {
  query: Type
  mutation: Type
  subscription: Type
}
```

mjmahone:
Fragment variables, and to some extent the SDL/IDL, fall into an interesting problem where we can't know what the right solution is, or even what everything means, until we use it. GraphQL itself was developed before it was formalized. The issue here is that at Facebook we have internal dependencies that would make it impossible for us to experiment with fragment variables (so that we can then drive the future spec RFC explaining what they mean) unless we have support for them within `graphql-js`. Things like Relay have a lot of adoption within Facebook, and we're not going to fork the open source Relay internally just so we can have one version that supports the spec and one that has experimental fragment variable support.

Fragment variables do not belong in the spec. If there was a clean way to depend on a forked-but-constantly-moving-with-open-source version of `graphql-js` in all of our GraphQL repos (including things like Relay), we would have done it.

In terms of fragments with variables being executable: I only think that because I'm seeing how they're actually used, and I think we need a way to describe a collection of executable definitions that is stand-alone, but not a "root" definition. This PR is not for that prupose, it is not changing the nature of fragments with variables, but I think it allows us to move in that direction by making sure people don't implement them without enforcing they define exactly those variables the fragments require. Which I'm hoping to put up as a separate GraphQL-spec RFC, but I don't want to get that wrong. We'll probably have to iterate on the various validation rules internally and through `graphql-js` before opening that up.

src/validation/ValidationContext.js
mjmahone: Yeah after thinking this through I think it does make more sense to add a second function specifically for excluding fragments with variable definitions. I'll make that change.

src/validation/ValidationContext.js
```diff
@@ -181,20 +192,27 @@ export default class ValidationContext {
...
-      usages = this.getVariableUsages(operation);
-      const fragments = this.getRecursivelyReferencedFragments(operation);
+      usages = this.getVariableUsages(definition);
+      const fragments = this.getRecursivelyReferencedFragments(definition);
       for (let i = 0; i < fragments.length; i++) {
         Array.prototype.push.apply(
           usages,
           this.getVariableUsages(fragments[i]),
         );
       }
```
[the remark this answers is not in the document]
mjmahone: Hmmm would it make more sense to create the ValidationContext with a flag indicating we should use fragment variables as a "validation cut" point? Then you're either opting in to it or not, and we could for instance validate "human-written" graphql with fragment variables as well as "transformed" (whatever that means) graphql in two different validation steps, just by creating a new validation context without the flag.

Basically this means people using fragment variables might consume more memory/make two passes, but I'd rather have people on the experimental version have some pain than make things confusing for those on mainline GraphQL.

mjmahone:
I'm going to put this PR on ice for a bit: fragment variables need to go over at least 1-2 more iterations of "what they mean" design: I'm no longer confident this is how fragment variable validation *should* work, so there's no use thrashing more here.

IvanGoncharov:
@mjmahone Thanks for the detailed explanation about the intention of this feature, now I better understand what's happing here and how Facebook depends both on Relay and `graphql-js`.
Sorry for being too meticulous about this change, I just worried that query syntax is the most critical part of GraphQL since we can never do any breaking change to the existing clients.
Ping me if you need any help with this feature or a devil's advocate assistance 👿

PR #1427, opened 2018-07-20
Prevent infinite recursion on invalid unions

IvanGoncharov:
Even though recursive unions are invalid they trigger infinite recursion during buildSchema/extendSchema.
\+ small refactoring

### DIFF
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -280,6 +280,19 @@ describe('extendSchema', () => {
     expect(unionField.type).to.equal(someUnionType);
   });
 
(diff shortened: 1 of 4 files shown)


src/utilities/__tests__/extendSchema-test.js
```diff
@@ -280,6 +280,19 @@ describe('extendSchema', () => {
     expect(unionField.type).to.equal(someUnionType);
   });
 
+  it('allows extension of union by adding itself', () => {
+    const extendedSchema = extendTestSchema(`
```
mjmahone: These are some nifty tests. Thanks for adding them!

PR #1429, opened 2018-07-23
BREAKING: Remove support for deprecated directive locations

IvanGoncharov:
Continuation of #1385


mjmahone:
Thanks for cleaning this up!

PR #1430, opened 2018-07-24
Inline intermediate 'visitUsingRules' function

IvanGoncharov:
After d00fbe9 `visitUsingRules` can't be used outside of `validate.js` so it makes sense to inline it.


mjmahone:
That makes sense. Thank you!

PR #1435, opened 2018-07-27
Validate SDL directives

### LINKED ISSUE #1389: add validation of directives in SDL
From #1383:

> It's very critical to add validation of directives in SDL since as [noted here](https://github.com/facebook/graphql/issues/429#issuecomment-392990065) no one validates them correctly at the moment. That will lead to a situation very similar to the recent `Allow interfaces to have no implementors`. Since spec defines set of validation rules but no one implements them. Probably we shouldn't enforce `Directives Are Unique Per Location` until https://github.com/facebook/graphql/issues/429 resolved but all other validation should be implemented e.g. require the schema to contain applied directives, correct directive arguments, etc.

mjmahone:
Re-uses the existing `validateSchema` to resolve #1389.

This is a sub-optimal solution: ideally we'd be doing validation purely on the AST, as @IvanGoncharov is working on. However, this gets us to a state where people aren't using directives illegally within the SDL, which should allow us to release them.

### DIFF
--- a/src/type/validate.js
+++ b/src/type/validate.js
@@ -7,44 +7,46 @@
  * @flow strict
  */
 
-import {
-  isObjectType,
-  isInterfaceType,
-  isUnionType,
-  isEnumType,
-  isInputObjectType,
-  isNonNullType,
-  isNamedType,
-  isInputType,
-  isOutputType,
-} from './definition';
-import type {
-  GraphQLObjectType,
-  GraphQLInterfaceType,
-  GraphQLUnionType,
-  GraphQLEnumType,
-  GraphQLInputObjectType,
-} from './definition';
-import { isDirective } from './directives';
-import type { GraphQLDirective } from './directives';
-import { isIntrospectionType } from './introspection';
-import { isSchema } from './schema';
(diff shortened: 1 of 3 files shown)


src/type/validate.js
```diff
@@ -165,7 +179,10 @@ function getOperationTypeNode(
   return type.astNode;
 }
 
-function validateDirectives(context: SchemaValidationContext): void {
+function validateDirectiveDefinitions(context: SchemaValidationContext): void {
+  // Ensure no directive is defined multiple times
+  const directiveDefinitions = new Map();
+
```
IvanGoncharov: @mjmahone This change is make sense on it's own since you can manually construct `GraphQLSchema` with duplicated directives.
Can you extract this change into separate PR?

IvanGoncharov: On `Map`: all around this library, we use `Object.create(null)` for maps with strings as keys and `Map` for non-string keys. AFAIK `Map` have significantly lover performance, I'm also not sure about memory usage.

mjmahone:
#1438 is strictly better than this PR, so closing

PR #1437, opened 2018-07-28
[RFC] Allow directives on variable definitions

mjmahone:
This implements directives on variable definitions, as defined in the [GraphQL RFC](https://github.com/facebook/graphql/pull/486)


IvanGoncharov:
@mjmahone Also need to adjust this function:
https://github.com/graphql/graphql-js/blob/e124ca870a57ee85cf73a77863e69889bc11737d/src/validation/rules/KnownDirectives.js#L63

mjmahone:
Thanks @IvanGoncharov: looks like I missed a few other locations as well, such as the introspection query. Added them just now.

mjmahone:
#1438 is strictly better than this PR, so closing.

mjmahone:
Sorry I closed the wrong one.

mjmahone:
Merged https://github.com/facebook/graphql/pull/486, so I'm also merging this.

leebyron:
Please place this behind an experimental flag until we can go through the RFC process for the spec

src/language/visitor.js
```diff
@@ -64,7 +64,7 @@ export const QueryDocumentKeys = {
     'directives',
     'selectionSet',
   ],
-  VariableDefinition: ['variable', 'type', 'defaultValue'],
+  VariableDefinition: ['variable', 'type', 'defaultValue', 'directives'],
```
leebyron: These typically follow the order found in the grammar, but this seems inconsistent

leebyron: My mistake. Consistent for this PR but inconsistent with the spec RFC

mjmahone:
@leebyron will do. Working on that PR now.

mjmahone:
#1454 puts this all behind an experimental flag.

Issue #1221, opened 2018-02-01
0.13-rc.1 notice for maintainers who build tools around graphql-js 


nodkz:
If your tool uses AST type checks something like:
```js
import { LIST_TYPE, NON_NULL_TYPE, DOCUMENT, NAMED_TYPE } from 'graphql/language/kinds';

...
if (inputTypeAST.kind === LIST_TYPE) { ...
```

You need to change your code in following manner:
```diff
- import { LIST_TYPE, NON_NULL_TYPE, DOCUMENT, NAMED_TYPE } from 'graphql/language/kinds';
+ import { Kind } from 'graphql';

...
- if (inputTypeAST.kind === LIST_TYPE) { ...
+ if (inputTypeAST.kind === Kind.LIST_TYPE) { ...
```

According to last [improvements for v0.13.0](https://github.com/graphql/graphql-js/commit/17a0bfd5292f39cafe4eec5b3bd0e22514243b68#diff-fc9da17436b6ffdf301f6e41e53fc207) by @IvanGoncharov you not able to import desired AST types directly from `graphql/language/kinds`. 

And if you want that your tool must **support new and old versions of graphql lib**, please use code above.

IvanGoncharov:
If possible you should use:
```graphql
  import { Kind } from 'graphql';
```

 I didn't think it will break any 3rd-part code since `graphql/*` files are treated like private API.
@nodkz Did you use `graphql/*` files to minimize bundle size or you need to import some private APIs?

nodkz:
It breaks my graphql-compose, already fixed.
And 82 other repos https://github.com/search?l=JavaScript&q=%22graphql%2Flanguage%2Fkinds%22&type=Code&utf8=%E2%9C%93

nodkz:
It's our problem, that we use internals imports. 
So don't worry. Your work is amazing!

leebyron:
Thanks for highlighting this! I'm going to close the issue, but reference it as a potentially breaking change from the release notes!

PR #1438, opened 2018-07-30
[RFC]SDL validation

IvanGoncharov:
Discussed in #1389

### DIFF
--- a/src/utilities/__tests__/extendSchema-test.js
+++ b/src/utilities/__tests__/extendSchema-test.js
@@ -1018,6 +1018,23 @@ describe('extendSchema', () => {
     expect(isScalarType(arg1.type)).to.equal(true);
   });
 
+  it('Rejects invalid SDL', () => {
+    const sdl = `
+      extend schema @unknown
+    `;
+    expect(() => extendTestSchema(sdl)).to.throw(
+      'Unknown directive "unknown".',
+    );
+  });
+
+  it('Allows to disable SDL validation', () => {
+    const sdl = `
+      extend schema @unknown
+    `;
+    extendTestSchema(sdl, { assumeValid: true });
+    extendTestSchema(sdl, { assumeValidSDL: true });
+  });
+
   it('does not allow replacing a default directive', () => {
     const sdl = `
       directive @include(if: Boolean!) on FIELD | FRAGMENT_SPREAD
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -11,6 +11,7 @@ import keyMap from '../jsutils/keyMap';
 import keyValMap from '../jsutils/keyValMap';
 import type { ObjMap } from '../jsutils/ObjMap';
 import { valueFromAST } from './valueFromAST';
+import { assertValidSDL } from '../validation/validate';
 import blockStringValue from '../language/blockStringValue';
 import { TokenKind } from '../language/lexer';
 import { parse } from '../language/parser';
@@ -86,6 +87,13 @@ export type BuildSchemaOptions = {
    * Default: false
    */
   commentDescriptions?: boolean,
+
+  /**
+   * Set to true to assume the SDL is valid.
+   *
+   * Default: false
+   */
+  assumeValidSDL?: boolean,
 };
 
 /**
@@ -112,6 +120,10 @@ export function buildASTSchema(
     throw new Error('Must provide a document ast.');
   }
 
+  if (!options || !options.assumeValidSDL) {
+    assertValidSDL(ast);
+  }
+
   let schemaDef: ?SchemaDefinitionNode;
 
   const typeDefs: Array<TypeDefinitionNode> = [];
@@ -121,9 +133,6 @@ export function buildASTSchema(
     const d = ast.definitions[i];
     switch (d.kind) {
       case Kind.SCHEMA_DEFINITION:
-        if (schemaDef) {
-          throw new Error('Must provide only one schema definition.');
-        }
         schemaDef = d;
         break;
       case Kind.SCALAR_TYPE_DEFINITION:
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -120,7 +120,7 @@ export function buildASTSchema(
     throw new Error('Must provide a document ast.');
   }
 
-  if (!options || !options.assumeValidSDL) {
+  if (!options || !(options.assumeValid || options.assumeValidSDL)) {
     assertValidSDL(ast);
   }
 
--- a/src/validation/ValidationContext.js
+++ b/src/validation/ValidationContext.js
@@ -19,6 +19,7 @@ import type {
   FragmentSpreadNode,
   FragmentDefinitionNode,
 } from '../language/ast';
+import type { ASTVisitor } from '../language/visitor';
 import type { GraphQLSchema } from '../type/schema';
 import type {
   GraphQLInputType,
@@ -63,6 +64,21 @@ export class ASTValidationContext {
     return this._ast;
   }
 }
+
+export class SDLValidationContext extends ASTValidationContext {
+  _schema: ?GraphQLSchema;
+
+  constructor(ast: DocumentNode, schema?: ?GraphQLSchema): void {
+    super(ast);
+    this._schema = schema;
+  }
+
+  getSchema(): ?GraphQLSchema {
+    return this._schema;
+  }
+}
+
+export type SDLValidationRule = SDLValidationContext => ASTVisitor;
 
 export class ValidationContext extends ASTValidationContext {
   _schema: GraphQLSchema;
@@ -234,3 +250,5 @@ export class ValidationContext extends ASTValidationContext {
     return this._typeInfo.getArgument();
   }
 }
+
+export type ValidationRule = ValidationContext => ASTVisitor;
--- a/src/validation/rules/KnownDirectives.js
+++ b/src/validation/rules/KnownDirectives.js
@@ -7,12 +7,15 @@
  * @flow strict
  */
 
-import type { ValidationContext } from '../ValidationContext';
+import type {
+  ValidationContext,
+  SDLValidationContext,
+} from '../ValidationContext';
 import { GraphQLError } from '../../error';
-import find from '../../jsutils/find';
 import { Kind } from '../../language/kinds';
 import { DirectiveLocation } from '../../language/directiveLocation';
 import type { ASTVisitor } from '../../language/visitor';
+import { specifiedDirectives } from '../../type/directives';
 
 export function unknownDirectiveMessage(directiveName: string): string {
   return `Unknown directive "${directiveName}".`;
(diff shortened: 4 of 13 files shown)


src/utilities/__tests__/extendSchema-test.js
```diff
@@ -1286,6 +1302,31 @@ describe('extendSchema', () => {
...
+        }
+
+        type Foo {
+          bar: String
+        }
+      `);
+      schema = extendSchema(schema, ast);
+      const queryType = schema.getQueryType();
+      expect(queryType).to.have.property('name', 'Foo');
+    });
```
IvanGoncharov: @mjmahone This is proposed solution for a situation where library/tools wants to add its directives to user-provided SDL.
Alternatively, you can just prefix user-provided SDL but it will screw line offsets in errors.
Another alternative is to use `concatAST` but there no method to convert `GraphQLDirective` to AST.

This is a pretty big change but in addition to injecting directives it also allows to implement `buildASTSchema` as follows:
```js
function buildASTSchema(ast, options) {
  return extendSchema(new GraphQLSchema({}), ast, options);
}
```

To make review easier tomorrow I will extract this change into separate PR.

IvanGoncharov:
@mjmahone This is working code but it needs more tests and I also want to split out a two PRs: all refactorings and allows to define a schema using `extendSchema`.
I will try to finish it tomorrow.

src/validation/validate.js
```diff
@@ -51,3 +51,39 @@ export function validate(
   visit(documentAST, visitWithTypeInfo(typeInfo, visitor));
   return context.getErrors();
 }
+
+export function validateSDL(
+  documentAST: DocumentNode,
+  schemaToExtend?: ?GraphQLSchema,
```
mjmahone: I like this a lot, and would in fact be a fan of moving all existing schema validation rules into `validateSDL`. Not sure if we want to do that prior to `14.0`, but it might be nice to add them as optional rules for `14.1`, and then have those be the specified rules for `15.0` (6 months down the line? Not sure what our release cadence is supposed to be).

I am very tempted to fix `buildASTSchema(ast, {assumeValid: true})` to not throw on errors that would normally be considered SDL-type errors (i.e. errors that do not prevent us from building a schema, it just would likely fail once built).

src/validation/validate.js
```diff
@@ -51,3 +51,39 @@ export function validate(
...
+  rules?: $ReadOnlyArray<any> = specifiedSDLRules,
```
mjmahone: hmmm, I'm not 100% certain whether or not we should allow people to pass in a `TypeInfo` here. I know internally we have a special `TypeInfo` (centered around dealing with pre-open-source GraphQL oddities) that we always pass into `validate`, but I don't believe it's required for the SDL at all.

src/validation/ValidationContext.js
```diff
@@ -42,11 +42,44 @@ type VariableUsage = {|
  * allowing access to commonly useful contextual information from within a
  * validation rule.
  */
-export default class ValidationContext {
-  _schema: GraphQLSchema;
+export class BaseValidationContext {
```
mjmahone: I like this split, as there's so many rules that don't need a schema at all.

However, I am beginning to think there are three pieces:
- a pure "ErrorReporter". Has no AST knowledge.
- an AST-only context. This is `SDLValidationContext` as it stands. Everything that is a pure `visitor` can use this context. It may need two documents, Schema and Executable ASTs.
- a Schema-aware context. Much like `visitor` can take in a `visitWithTypeInfo`, this would hold a `ASTValidationContext`, and delegate to that one for the ast functionality.

src/validation/ValidationContext.js
```diff
@@ -42,11 +42,44 @@ type VariableUsage = {|
...
+    return this._ast;
+  }
+}
+
+export class SDLValidationContext extends BaseValidationContext {
+  _schema: ?GraphQLSchema;
+
+  constructor(ast: DocumentNode, schema?: ?GraphQLSchema): void {
+    super(ast);
+    this._schema = schema;
```
mjmahone: I don't think we should even allow access to a `GraphQLSchema` here: instead, if you pass in an existing schema, we can convert that schema to AST nodes (a very rough way to do this would be to `parse(printSchema(schema))`).

Ideally, the current `GraphQLSchema` would only be a runtime thing: you need it to validate queries you're trying to persist, but you don't need it to validate or transform itself.

I like the idea that there's a split between "validation that needs a schema" and "validation that does not", and "validation that does not need a schema" is a superset of "SDL validation".

src/utilities/__tests__/extendSchema-test.js
mjmahone: That looks pretty cool! I think we might need to provide a method to convert `GraphQLDirective` to AST.

IvanGoncharov:
@mjmahone 
Update: rebased on top of current master.

src/utilities/buildASTSchema.js
```diff
@@ -86,6 +87,13 @@ export type BuildSchemaOptions = {
    * Default: false
    */
   commentDescriptions?: boolean,
+
+  /**
+   * Set to true to assume the SDL is valid.
+   *
+   * Default: false
+   */
+  assumeValidSDL?: boolean,
```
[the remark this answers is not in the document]
IvanGoncharov: @mjmahone Good catch 🔎 🐞 
Fixed.

src/utilities/buildASTSchema.js
```diff
@@ -112,6 +120,10 @@ export function buildASTSchema(
     throw new Error('Must provide a document ast.');
   }
 
+  if (!options || !(options.assumeValid || options.assumeValidSDL)) {
```
IvanGoncharov: @mjmahone We need separate `assumeValidSDL` check so developers are able to change set of validation rules(add or delete).
```js
const errors = validateSDL(ast, null, customSetOfRules);
if (errors.length !== 0) {
  // ...
}
const schema = buildASTSchema(ast, { assumeValidSDL: true });
```

mjmahone: Ok this makes sense. A little bit annoying to add in that second key, but not the end of the world.

mjmahone: Wait: are you thinking `assumeValid` is a superset of `assumeValidSDL` or disjoint? I would still ideally have it be a superset. We could add `assumeValidExecutableAST` or something that is disjoint with `assumeValidSDL`.

IvanGoncharov: @mjmahone My proposal:
`{}` => `validateSDL` + `validateSchema`
`{ assumeValidSDL: true }` => `validateSchema`
`{assumeValid: true }` => no checks

src/validation/ValidationContext.js
IvanGoncharov: Addressed in #1446

src/validation/validate.js
IvanGoncharov: @mjmahone Are you referring to the `rules` field?
To be more explicit I create and used it here so now it looks like:
```js
rules?: $ReadOnlyArray<SDLValidationRule> = specifiedSDLRules,
```

mjmahone: Oh I'm OK with the rules field. For `validate` there is an extra optional argument, `typeInfo?: TypeInfo` which I do not believe is required.

src/validation/rules/KnownDirectives.js
```diff
@@ -31,29 +34,42 @@ export function misplacedDirectiveMessage(
...
+      }
+    },
     Directive(node, key, parent, path, ancestors) {
-      const directiveDef = find(
-        context.getSchema().getDirectives(),
-        def => def.name === node.name.value,
-      );
-      if (!directiveDef) {
+      const name = node.name.value;
+      const locations = locationsMap[name];
```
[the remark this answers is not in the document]
IvanGoncharov: @mjmahone It's unnecessary complication since you `context` has `getDocument` function. So I just removed `Document` and move this code on the top level.

src/validation/ValidationContext.js
IvanGoncharov: @mjmahone Problem is that conversion from SDL to `GraphQLSchema` is lossless, but not vice versa. For example, you lose all directives which prevent from enforcing the rule that forbids directive duplication in extends.

Also after `parse(printSchema(schema)))` you loose ability to print errors with original locations and formating of nodes.

And there are others many potential scenarious where we loose something by such converion especially if schema that we trying to extend was build from `GraphQL*Types` not from SDL. For example, developer can create custom class derived `GraphQLObjectType` with some additional data and require this additional data inside his custom SDL rules.

And it's not a technical problem, since `GraphQLSchema` is intended to be most complete represenation of schema.

> and "validation that does not need a schema" is a superset of "SDL validation".

"validation that does not need a schema" => ASTValidationContext
And this context suffisient for all rules that scoped to particular defintion.

If rule should check conflicts/dependencies beetween definitions it should handle case when you extending existing schema and use `SDLValidationContext`.

IvanGoncharov:
@mjmahone I updated this PR and answered all review comments.
Can you please take a look one more time?

This PR doesn't contain all necessary rules and tests, e.g. it rule that check duplications in extends:
```
type Query @test;

extend type Query @test;
```
If this PR is ok I will try to finish all other rules tomorrow as a separate PR.

src/validation/ValidationContext.js
mjmahone: We should fix it so it's lossless, but that would be a different PR, and we could probably target 14.1 for that.

> Also after parse(printSchema(schema))) you loose ability to print errors with original locations and formating of nodes.

I think there's a way around this, where we'd do `buildSchemaDocuments(schema): Array<DocumentNode>`, where each `DocumentNode` would contain all definitions from a specific source. That's obviously more complex of a change than what we want to do here, but I think would be worth having. For `GraphQL*Types` we'd have to have some form of "default" or "non-Source" document. But for truly custom classes, you're right we can't make lossless conversions.

> If rule should check conflicts/dependencies beetween definitions it should handle case when you extending existing schema and use SDLValidationContext.

Makes sense, so let's push this forward as-is. But I'd like to switch our mental model to being AST-validation first, and then only using the schema when it's truly necessary. You've mostly accomplished this, but I envision a bunch of people creating validation rules that silently fail if no schema is present (because they assume they always have a schema present). That would be an unfortunate future.

src/validation/rules/LoneSchemaDefinition.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * A GraphQL document is only valid if it contains only one schema definition.
+ */
+export function LoneSchemaDefinition(
+  context: SDLValidationContext,
+): ASTVisitor {
+  const oldSchema = context.getSchema();
+  const alreadyDefined =
+    oldSchema &&
+    (oldSchema.astNode ||
```
mjmahone: It feels weird to check the `astNode` here, but I guess it's needed in case your original schema definition is invalid (for instance, if it has no `query`): we still would want to report this error upon re-defining, too. So this does make sense.

src/validation/rules/LoneSchemaDefinition.js
```diff
@@ -0,0 +1,59 @@
...
+      oldSchema.getQueryType() ||
+      oldSchema.getMutationType() ||
+      oldSchema.getSubscriptionType());
+
+  const schemaNodes = [];
+  return {
+    SchemaDefinition(node) {
+      if (alreadyDefined) {
+        context.reportError(
+          new GraphQLError(canNotDefineSchemaWithinExtension(), [node]),
```
mjmahone: It might be nice to point back to the originally defined schema if provided. But I could be wrong here and that would just be noise.

Also a very small nit provided you're only going to use this node in the error, not sure it's worth changing: 
```
new GraphQLError(canNotDefineSchemaWithinExtension(), node),
```

src/validation/validate.js
```diff
@@ -52,3 +52,39 @@ export function validate(
...
+  const visitors = rules.map(rule => rule(context));
+  visit(documentAST, visitInParallel(visitors));
+  return context.getErrors();
+}
+
+/**
+ * Utility function which asserts a SDL document is valid by throwing an error
+ * if it is invalid.
+ */
+export function assertValidSDL(documentAST: DocumentNode): void {
```
mjmahone: Should we make this API public yet? We don't have a corresponding `assertValidAST` yet, nor `assertValid` (for both SDL and AST). Internally, we do the exact same thing you're doing here internally, but we should probably offer all the functionality (maybe with a single `assertValid` function for now). I'm also not 100% sure the `assertValid` functions belongs here vs. a utility file, and I am not sure we should offer them without offering the ability to pass in your own set of rules.

mjmahone:
@IvanGoncharov once we clear up the `assertValid` functions, I'm ready to get this merged. Everything else looks great!

IvanGoncharov:
> Just want to make sure it makes sense to make assertValidSDL and assertValidSDLExtension functions public with this release. If you really think it's important to, I'm OK with it, but I'd prefer waiting for a future release to play it safe. We also aren't even using them in our test harness, so I don't think it's required just yet.

@mjmahone `graphql-js` officially supports only two kinds of imports:
1. e.g. `import { /* ... */ } from 'graphql';`
1. e.g. `import { /* ... */ } from 'graphql/language';`
See here: https://github.com/graphql/graphql-js/blob/master/src/README.md
Importing from individual files are illegal and all such imports treated as part of private API, e.g. this disscussion: #1221

But I agree it could be confusing so I marked all added exports as `@internal`.

src/validation/rules/LoneSchemaDefinition.js
IvanGoncharov: @mjmahone I was thinking that in some systems (e.g. GraphQL is very popular for CMS) clients can provide their own extension SDL. So was concern that we can leak some internal data in printed errors.

Thinking more about this I think I'm too paranoid since `validateSchema` already leaking this info so if someone decides he doesn't want origin node to be printed he needs to sanitize `GraphQLError`s anyway. Plus such complex systems almost certainly using `GraphQL*Type` classes anyway.

PR #1439, opened 2018-07-31
Improve typings for validation functions

IvanGoncharov:
Split out from #1438


IvanGoncharov:
@mjmahone My idea is to clean up #1438 and merge changes that make sense on its own and we both agree on.

mjmahone:
Great, thank you! This one makes obvious sense.

PR #1441, opened 2018-07-31
Allows to add schema definition missing in the original schema

IvanGoncharov:
Split out from #1438

Allows to inject directives and types into SDL-based schema:

```js
import { GrapQLDateTime } from 'somepackage';
let schema = new GraphQLSchema({
  types: [GrapQLDateTime],
});

const ast = parse(`
  type Query {
    timestamp: DateTime
  }
`);
schema = extendSchema(schema, ast);
```

```js
let schema = new GraphQLSchema({
  directives: [
    new GraphQLDirective({
      name: 'onSchema',
      locations: ['SCHEMA'],
    }),
  ],
});

const ast = parse(`
  schema @onSchema {
    query: Foo
  }
  type Foo {
    bar: String
  }
`);
schema = extendSchema(schema, ast);
```

### DIFF
--- a/src/utilities/extendSchema.js
+++ b/src/utilities/extendSchema.js
@@ -53,6 +53,7 @@ import type {
   DocumentNode,
   DirectiveDefinitionNode,
   SchemaExtensionNode,
+  SchemaDefinitionNode,
 } from '../language/ast';
 
 type Options = {|
@@ -106,6 +107,7 @@ export function extendSchema(
   // have the same name. For example, a type named "skip".
   const directiveDefinitions: Array<DirectiveDefinitionNode> = [];
 
+  let schemaDef: ?SchemaDefinitionNode;
   // Schema extensions are collected which may add additional operation types.
   const schemaExtensions: Array<SchemaExtensionNode> = [];
 
@@ -113,11 +115,20 @@ export function extendSchema(
     const def = documentAST.definitions[i];
     switch (def.kind) {
       case Kind.SCHEMA_DEFINITION:
-        // Sanity check that a schema extension is not defining a new schema
-        throw new GraphQLError(
-          'Cannot define a new schema within a schema extension.',
-          [def],
-        );
+        // Sanity check that a schema extension is not overriding the schema
+        if (
+          schema.astNode ||
+          schema.getQueryType() ||
+          schema.getMutationType() ||
+          schema.getSubscriptionType()
+        ) {
+          throw new GraphQLError(
+            'Cannot define a new schema within a schema extension.',
+            [def],
+          );
+        }
+        schemaDef = def;
+        break;
       case Kind.SCHEMA_EXTENSION:
         schemaExtensions.push(def);
         break;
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@mjmahone Based on the discussion in #1441 I'm not sure if you approved that change or not. Can you please review it separately?

src/utilities/extendSchema.js
```diff
@@ -216,9 +227,9 @@ export function extendSchema(
     subscription: extendMaybeNamedType(schema.getSubscriptionType()),
   };
 
-  // Then, incorporate all schema extensions.
-  for (const schemaExtension of schemaExtensions) {
-    if (schemaExtension.operationTypes) {
+  // Then, incorporate schema definition and all schema extensions.
+  for (const schemaExtension of [schemaDef, ...schemaExtensions]) {
```
mjmahone: oh wait just saw this: can we do instead:
```
if (schemaDef) {
  schemaExtensions.push(schemaDef);
}
for (const schemaExtension of schemaExtensions) {
```

mjmahone: Main worry here: it seems silly to do null checking inside the for loop instead of making sure what we have in the array we're looping over is non-null.

IvanGoncharov: @mjmahone You are right, the current code is confusing 😞 
Sadly, `schemaExtensions` used bellow explicitly for extension nodes.
Moreover `schemaDef` doesn't need `schemaDef.operationTypes` because it always initialized. 
So I decided to just duplicate code since #1435 will reduce this code and we can simplify things in subsequent refactorings.

IvanGoncharov:
@mjmahone Can you please review updated PR?

src/utilities/extendSchema.js
```diff
@@ -184,7 +195,8 @@ export function extendSchema(
     Object.keys(typeExtensionsMap).length === 0 &&
     Object.keys(typeDefinitionMap).length === 0 &&
     directiveDefinitions.length === 0 &&
-    schemaExtensions.length === 0
+    schemaExtensions.length === 0 &&
+    !schemaDef
```
IvanGoncharov: I initially forgot to add this check, so I also change tests to cover it.

src/utilities/extendSchema.js
```diff
@@ -254,9 +275,7 @@ export function extendSchema(
 
   // Then produce and return a Schema with these types.
   return new GraphQLSchema({
-    query: operationTypes.query,
-    mutation: operationTypes.mutation,
-    subscription: operationTypes.subscription,
```
IvanGoncharov: tiny refactoring 😜

src/utilities/extendSchema.js
```diff
@@ -216,19 +228,28 @@ export function extendSchema(
     subscription: extendMaybeNamedType(schema.getSubscriptionType()),
   };
 
-  // Then, incorporate all schema extensions.
+  if (schemaDef) {
+    for (const { operation, type } of schemaDef.operationTypes) {
+      if (operationTypes[operation]) {
+        throw new Error(`Must provide only one ${operation} type in schema.`);
+      }
```
IvanGoncharov: After #1435 I will move this check into SDL validation.

PR #1444, opened 2018-08-02
Update deps

IvanGoncharov:


src/subscription/__tests__/subscribe-test.js
```diff
@@ -882,10 +882,13 @@ describe('Subscription Publish Phase', () => {
...
-      yield { email: { subject: 'Hello' } };
-      throw new Error('test error');
-    }, email => email);
+    const erroringEmailSchema = emailSchemaWithResolvers(
+      async function*() {
+        yield { email: { subject: 'Hello' } };
+        throw new Error('test error');
+      },
+      email => email,
+    );
```
IvanGoncharov: Results from prettier update

package.json
```diff
@@ -48,34 +48,34 @@
...
     "eslint-plugin-flowtype": "2.50.0",
     "eslint-plugin-prettier": "2.6.2",
-    "flow-bin": "0.76.0",
+    "flow-bin": "0.77.0",
     "mocha": "5.2.0",
     "nyc": "12.0.2",
-    "prettier": "1.13.7",
-    "sane": "2.5.2"
+    "prettier": "1.14.0",
+    "sane": "3.0.0"
```
mjmahone: How valuable is the `sane` upgrade? I know we would be stuck on `2.5` for a very long time if we don't upgrade it now, but the release nodes for `3.0` have nothing in them, which makes me think they bumped to `3.0` just so they could stop supporting node v9.

IvanGoncharov: @mjmahone It's only a dev dependency and we use it only for `yarn watch` which is broken anyway. 
So I can easily revert it but as I wrote in #1445 I don't see any problem with dropping Node 9

PR #1445, opened 2018-08-02
Drop Node 9 support

IvanGoncharov:
See: https://github.com/nodejs/Release#end-of-life-releases

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "mocha": "--check-leaks --full-trace --timeout 15000 src/**/__tests__/**/*-test.js"
   },
   "engines": {
-    "node": "6.x || 8.x || 9.x || >= 10.x"
+    "node": "6.x || 8.x || >= 10.x"
   },
   "scripts": {
     "watch": "babel-node ./resources/watch.js",
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: node_js
 # https://github.com/nodejs/Release
 node_js:
   - '10'
-  - '9'
   - '8'
   - '6'
 


IvanGoncharov:
It blocks dependencies update see:
https://travis-ci.org/graphql/graphql-js/jobs/411339001

mjmahone:
Interesting: we must support 8 and 10 but not 9, and only because we didn't end up releasing 14.0 in June :)

IvanGoncharov:
@mjmahone Yes, but effectively it would become impossible to use 9 anyway since all other packages will do the same.
It's not a big problem since `9` was an experimental version and wasn't targeted for production.
+ No sane company will use `9` in production since you wouldn't get any security updates.

IvanGoncharov:
@mjmahone What Node version are you using in Facebook?

package.json
```diff
@@ -18,7 +18,7 @@
     "mocha": "--check-leaks --full-trace --timeout 15000 src/**/__tests__/**/*-test.js"
   },
   "engines": {
-    "node": "6.x || 8.x || 9.x || >= 10.x"
+    "node": "6.x || 8.x || >= 10.x"
```
leebyron: I'm curious to see if this ends up creating issues. This engines line means something quite different than the travis config.

Travis config means we only explicitly test support for a certain set of node versions. This engines line means we require these versions of node to operate - otherwise warnings are emitted during installation.

v9 is not an experimental branch, and there are plenty of companies that sanely use the latest version of node. The difference between odd and even numbered versions is that odd numbered versions have short support cycles while even have long ones.

jure: Well, came here from figuring out an issue. :) For reasons, we're depending on an older version of `nan` which fails to compile in Node v10 (but worked fine in v9). And we're upgrading to `graphql` 14.0.0 to get extended inputs. In any case, I'm not sure why this change is necessary - is v9 missing a feature that `graphql` depends on? Additionally, I don't think I've seen other modules limit engines in this way - i.e. removing a node version because it's no longer in maintenance - but limiting it because of missing features makes sense. 2c.

IvanGoncharov: @jure I removed node 9 because one of our dev dependencies stoped supporting it:
https://github.com/amasad/sane/blob/49f251a95e6db118ad71223b1eca93c876d9aacc/package.json#L47-L49
So the simplest solution was to also drop `v9`.
Because v9 stoped receiving security patches I assumed no one would use it to power GraphQL servers.

So we can enable it back but we wouldn't be able to tests it.
I will try to remove our dependency on `sane` or revert it to the older version  that supports `v9`.

PR #1448, opened 2018-08-05
Remove some unneeded 'any'

IvanGoncharov:


src/utilities/TypeInfo.js
```diff
@@ -139,8 +139,7 @@ export class TypeInfo {
     return this._enumValue;
   }
 
-  // Flow does not yet handle this case.
-  enter(node: any /* ASTNode */) {
```
IvanGoncharov: Flow now support this case 🎉

PR #1449, opened 2018-08-05
Fix "watch" command. Remove 'babel-node'

IvanGoncharov:


package.json
```diff
@@ -21,7 +21,7 @@
     "node": "6.x || 8.x || >= 10.x"
   },
   "scripts": {
-    "watch": "babel-node ./resources/watch.js",
```
IvanGoncharov: Node 6 natively supports all features we use.

PR #1450, opened 2018-08-06
Remove unused dev dependencies

IvanGoncharov:


.babelrc.js
```diff
@@ -14,7 +14,6 @@ module.exports = {
   }]],
   "plugins": [
     "./resources/inline-invariant",
-    "syntax-async-functions",
```
IvanGoncharov: Not required anymore and was moved to archive:
https://github.com/babel/babel-archive/tree/master/packages/babel-plugin-syntax-async-functions

IvanGoncharov:
I also run `yarn build` before and after the change and there no difference in output.

PR #1451, opened 2018-08-06
Generate error per duplicate schema def + Extend tests

IvanGoncharov:


src/validation/rules/LoneSchemaDefinition.js
```diff
@@ -35,25 +35,22 @@ export function LoneSchemaDefinition(
...
-          );
-        }
-      },
+
+      if (schemaDefinitionsCount > 0) {
+        context.reportError(
+          new GraphQLError(schemaDefinitionNotAloneMessage(), node),
+        );
+      }
+      ++schemaDefinitionsCount;
```
IvanGoncharov: Now it generates separate errors on every duplicated schema definition.
It makes the order of errors to match the order of definitions in SDL.
Based on very similar code here:
https://github.com/graphql/graphql-js/blob/c847a45a0efc71f821d777da6121e32edde853e8/src/validation/rules/LoneAnonymousOperation.js#L36-L40

PR #1457, opened 2018-08-08
Use invariants instead of throwing Errors

IvanGoncharov:


src/utilities/buildASTSchema.js
```diff
@@ -116,9 +117,7 @@ export function buildASTSchema(
   ast: DocumentNode,
   options?: BuildSchemaOptions,
 ): GraphQLSchema {
-  if (!ast || ast.kind !== Kind.DOCUMENT) {
-    throw new Error('Must provide a document ast.');
-  }
```
IvanGoncharov: Make it similar to this:
https://github.com/graphql/graphql-js/blob/3fdf240234445789b6b876e39f2bb9ed3a977387/src/utilities/extendSchema.js#L105-L108

mjmahone: Is `invariant` better, or is `GraphQLError` with the `documentAST` better? Internally, we've found converting as many `invariant(<invariant>, msg)` calls to `if (!<invariant>) { throw new GraphQLError(msg, astNode); }` has greatly reduced our tooling support burden: people have mostly stopped asking why the infra failed when their GraphQL document is invalid.

In this case, I'm not sure whether throwing `GraphQLError` would be more or less confusing, though: the callee is passing in null or a bad AST node, so it really is a tool-builder error, not a GraphQL-user error.

src/utilities/buildASTSchema.js
```diff
@@ -342,7 +344,7 @@ export class ASTDefinitionBuilder {
       case Kind.INPUT_OBJECT_TYPE_DEFINITION:
         return this._makeInputObjectDef(def);
       default:
-        throw new Error(`Type kind "${def.kind}" not supported.`);
+        invariant(false, `Type kind "${def.kind}" not supported.`);
```
mjmahone: While the above case should probably be `invariant`, this one should almost definitely me `throw new GraphQLError`: you've got a type definition that somehow parsed but at validate time is invalid. In this instance it's very helpful to see where in your document that error actually is.

IvanGoncharov: @mjmahone You right 👍 especially tanking into account possible Experimental features.

IvanGoncharov: @mjmahone Fixed.

PR #1458, opened 2018-08-08
Fixup: add Experimental note to variable-directive tests

mjmahone:
Added a parsing option to the test harness `expectPassesRule` and `expectFailsRule` functions.

Also added experimental flags on all tests with variable definition directives.

### DIFF
--- a/src/language/__tests__/parser-test.js
+++ b/src/language/__tests__/parser-test.js
@@ -106,7 +106,7 @@ describe('Parser', () => {
     );
   });
 
-  it('parses variable definition directives', () => {
+  it('Experimental: parses variable definition directives', () => {
     expect(() =>
       parse('query Foo($x: Boolean = false @bar) { field }', {
         experimentalVariableDefinitionDirectives: true,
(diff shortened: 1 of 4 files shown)


IvanGoncharov:
@mjmahone Side note: I reviewed #1437 one more time an I concern that we exposing experimental feature in introspection:
https://github.com/graphql/graphql-js/blob/cb0097c70632b2296ff6b36e15be31c46136cee2/src/type/introspection.js#L138-L141
It exposed via `printSchema` and via introspection query:
https://github.com/graphql/graphql-js/blob/1d7efa98e6190c5221106f8261c9bc5adeba97ca/src/utilities/__tests__/schemaPrinter-test.js#L679-L680

mjmahone:
Hmmm that may take a deeper fix. As a note, @leebyron did not have an issue exposing the RFC schema directives in https://github.com/graphql/graphql-js/pull/376 even before we had pushed the GraphQL spec change through (of note, the spec changes weren't merged in until this February, nearly two years later). I think this probably falls into that same category: even if exposed, it's not possible to use unless you add an `experimental` option in your parser.

Given there seems to be little danger to exposing it in introspection, but exposing it behind a feature flag will require significant re-architecting of introspection, I don't think we should hide it from introspection.

PR #1459, opened 2018-08-09
Add Node Predicates

IvanGoncharov:
For SDL validation rules I need to distinguish type definition node from type extension nodes.
\+ I saw same in couple other projects, e.g. `relay`:
https://github.com/facebook/relay/blob/04aac482538b83548c25e878205ecc1a581bcb25/packages/graphql-compiler/core/GraphQLSchemaUtils.js#L198-L203

### DIFF
--- a/src/language/predicates.js
+++ b/src/language/predicates.js
@@ -1,0 +1,90 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import type { ASTNode } from './ast';
+import { Kind } from './kinds';
+
+export function isDefinitionNode(node: ASTNode): boolean %checks {
+  return (
+    isExecutableDefinitionNode(node) ||
+    isTypeSystemDefinitionNode(node) ||
+    isTypeSystemExtensionNode(node)
+  );
+}
+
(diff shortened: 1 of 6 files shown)


src/language/predicates.js
```diff
@@ -0,0 +1,90 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import type { ASTNode } from './ast';
+import { Kind } from './kinds';
+
+export function isDefinitionNode(node: ASTNode): boolean %checks {
```
mjmahone: I am a big fan of these! We hit them a lot, and it's nice to be able to filter through definitions to just get `Executable` or `TypeSystem` definitions!

src/validation/rules/ExecutableDefinitions.js
```diff
@@ -28,10 +29,7 @@ export function ExecutableDefinitions(
   return {
     Document(node) {
       for (const definition of node.definitions) {
-        if (
-          definition.kind !== Kind.OPERATION_DEFINITION &&
-          definition.kind !== Kind.FRAGMENT_DEFINITION
-        ) {
+        if (!isExecutableDefinitionNode(node)) {
```
mjmahone: The flow error here is because you're asking if the Document is an ExecutableDefinitionNode rather than asking if the definition is.

IvanGoncharov: @mjmahone Fixed. Thanks.

PR #1462, opened 2018-08-12
Fix typo in AST predicates

mjmahone:
This was an easy mistake to make: the ordering in the `Kind` object made it seem like `Variable` belonged with `VariableDefinition`, when in fact it belongs with the values (as a variable definition defines that a value can be used).

Additionally, `ObjectField` is not itself a value: it's more of a key-value pair that itself contains a value.

### DIFF
--- a/src/language/kinds.js
+++ b/src/language/kinds.js
@@ -18,7 +18,6 @@ export const Kind = Object.freeze({
   DOCUMENT: 'Document',
   OPERATION_DEFINITION: 'OperationDefinition',
   VARIABLE_DEFINITION: 'VariableDefinition',
-  VARIABLE: 'Variable',
   SELECTION_SET: 'SelectionSet',
   FIELD: 'Field',
   ARGUMENT: 'Argument',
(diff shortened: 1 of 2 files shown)


src/language/kinds.js
```diff
@@ -29,6 +28,7 @@ export const Kind = Object.freeze({
   FRAGMENT_DEFINITION: 'FragmentDefinition',
 
   // Values
+  VARIABLE: 'Variable',
```
mjmahone: Moved this to prevent future confusion

src/language/predicates.js
```diff
@@ -35,15 +35,15 @@ export function isSelectionNode(node: ASTNode): boolean %checks {
...
+    node.kind === Kind.VARIABLE ||
     node.kind === Kind.INT ||
     node.kind === Kind.FLOAT ||
     node.kind === Kind.STRING ||
     node.kind === Kind.BOOLEAN ||
     node.kind === Kind.NULL ||
     node.kind === Kind.ENUM ||
     node.kind === Kind.LIST ||
-    node.kind === Kind.OBJECT ||
-    node.kind === Kind.OBJECT_FIELD
```
mjmahone: `OBJECT_FIELD` is lumped with Values in `kinds.js`, but that's because it doesn't really belong anywhere else. However, if I give an `ObjectFieldNode` to a function expecting a `ValueNode`, we've messed something up.

PR #1465, opened 2018-08-13
Add 'isRequredArgument' & 'isRequredInputField' predicates

IvanGoncharov:


src/validation/rules/ProvidedRequiredArguments.js
```diff
@@ -58,11 +58,7 @@ export function ProvidedRequiredArguments(
         const argNodeMap = keyMap(argNodes, arg => arg.name.value);
         for (const argDef of fieldDef.args) {
           const argNode = argNodeMap[argDef.name];
-          if (
-            !argNode &&
-            isNonNullType(argDef.type) &&
-            argDef.defaultValue === undefined
-          ) {
```
IvanGoncharov: They make code more readable.

Marak:
`required`, not `requred`?

IvanGoncharov:
@Marak Thanks a lot 👍 Fixed.

> is this something you're planning on re-using, or is it just to make this one location more readable?

It's actually a very very painful topic. I didn't want to overburden you with technical details but here is a full story.

So before https://github.com/facebook/graphql/pull/418 `required === NonNull` but after this change, you also need to check `defaultValue`. Lee made necessary changes in all major places including validation rules in #1274. But many other places including all external projects still assume that `required === NonNull`.

That makes me pay extra attention to all `isNonNullType`/`isNullableType` checks and at some point, we need to check all such places and fix all remaining bugs.
For example, here are a few places that are easy to guess:
https://github.com/graphql/graphql-js/blob/7cfd6865cebdca1c1760246a5c132babb791b7d7/src/utilities/findBreakingChanges.js#L250-L257
Adding non-null argument with a default value is not a breaking change.

https://github.com/graphql/graphql-js/blob/7cfd6865cebdca1c1760246a5c132babb791b7d7/src/type/validate.js#L441-L450

Same problem here since according to the spec:
http://facebook.github.io/graphql/June2018/#sec-Objects
> The object field may include additional arguments not defined in the interface field, but any additional argument must not be required, e.g. must not be of a non‐nullable type.

And probably many more places since 
So these functions are not strictly necessary but they solve a lot of confusion and motivate people who don't necessarily know about https://github.com/facebook/graphql/pull/418 change to use correct check in their code.

IvanGoncharov:
@mjmahone I just remembered that Lee made `undefined` a valid value for enums in b013e73084c7c012527a3d44f215f46d1f8465ae and that mean `field.defaultValue === undefined` is invalid check and instead it should use `hasOwnProperty` simular to this place:
https://github.com/graphql/graphql-js/blob/7cfd6865cebdca1c1760246a5c132babb791b7d7/src/type/definition.js#L1140

But I can't use `hasOwnProperty` since a lot of place incorrectly assign `undefined` to `defaultValue`. For example:
https://github.com/graphql/graphql-js/blob/7cfd6865cebdca1c1760246a5c132babb791b7d7/src/utilities/buildASTSchema.js#L311

I think it's even bigger reason to use dedicated function for this check.

PR #1467, opened 2018-08-13
Make error handling consistent in createSourceEventStream

taion:
Currently, errors in `createSourceEventStream` from `buildExecutionContext` will be returned in an `ExecutionResult`, while all other GraphQL errors will be thrown. This is inconsistent with the handling in `subscribe`, and makes it harder to build error handling logic, as there are then two ways to handle the same class of errors.

This makes the error handling consistent and updates the documentation to describe the new logic.

This is technically breaking, but is unlikely to cause problems for any properly-written code, as full error handling logic required handling resolved `ExecutionResult` objects with `errors` anyway.

### DIFF
--- a/src/subscription/__tests__/subscribe-test.js
+++ b/src/subscription/__tests__/subscribe-test.js
@@ -10,6 +10,7 @@ import { describe, it } from 'mocha';
 import EventEmitter from 'events';
 import eventEmitterAsyncIterator from './eventEmitterAsyncIterator';
 import { createSourceEventStream, subscribe } from '../subscribe';
+import { GraphQLError } from '../../error';
 import { parse } from '../../language';
 import {
   GraphQLSchema,
@@ -986,19 +987,13 @@ describe('Subscription Publish Phase', () => {
     });
   });
 
-  it('should pass through error resolved from source event stream', async () => {
+  it('should resolve GraphQL error from source event stream', async () => {
     const erroringEmailSchema = emailSchemaWithResolvers(
       async function*() {
         yield { email: { subject: 'Hello' } };
-        yield new Error('test error');
+        throw new GraphQLError('test error');
       },
-      email => {
-        if (email instanceof Error) {
-          throw email;
-        }
-
-        return email;
-      },
+      email => email,
     );
 
     const subscription = await subscribe(
@@ -1032,14 +1027,9 @@ describe('Subscription Publish Phase', () => {
     expect(payload2).to.deep.equal({
       done: false,
       value: {
-        data: {
-          importantEmail: null,
-        },
         errors: [
           {
             message: 'test error',
-            locations: [{ line: 3, column: 11 }],
-            path: ['importantEmail'],
           },
         ],
       },
--- a/src/subscription/__tests__/subscribe-test.js
+++ b/src/subscription/__tests__/subscribe-test.js
@@ -9,7 +9,7 @@ import { expect } from 'chai';
 import { describe, it } from 'mocha';
 import EventEmitter from 'events';
 import eventEmitterAsyncIterator from './eventEmitterAsyncIterator';
-import { subscribe } from '../subscribe';
+import { createSourceEventStream, subscribe } from '../subscribe';
 import { parse } from '../../language';
 import {
   GraphQLSchema,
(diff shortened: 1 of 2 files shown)


taion:
Sorry, I accidentally deleted the PR branch while cleaning up my fork. The build failed because that branch got deleted while it was running.

Can someone re-run the build?

taion:
Ah, never mind, it's re-running on its own.

IvanGoncharov:
@taion Can you please add test cases that cover behavior that you described?

src/subscription/subscribe.js
```diff
@@ -284,6 +298,8 @@ export function createSourceEventStream(
       );
     });
   } catch (error) {
-    return Promise.reject(error);
+    return error instanceof GraphQLError
+      ? Promise.resolve({ errors: [error] })
+      : Promise.reject(error);
```
IvanGoncharov: @taion Not sure that I fully understand this code, but why need different behavior for `Error` and `GraphQLError`?

taion: This matches the logic for: https://github.com/graphql/graphql-js/blob/b24cbcde325d6e53788debae9818dd8cddebf9e1/src/subscription/subscribe.js#L114-L124

GraphQL errors resolve the promise as an `ExecutionResult` with just `errors`, while other errors reject the promise.

taion: It would be, but is it worth the complexity? What do you think of just adding a comment?

src/subscription/subscribe.js
```diff
@@ -171,19 +172,28 @@ function subscribeImpl(
             reportGraphQLError,
           )
         : ((resultOrStream: any): ExecutionResult),
-    reportGraphQLError,
```
taion: @IvanGoncharov The removal of this "catch" handler means that the existing tests for error handling in `subscribe` already do exercise the new logic. Do you still think explicit new tests are needed?

IvanGoncharov: @taion I mean you change a code to change how certain errors handled and no test cases were broken. It means that there no test cases covering this behaviour and without test cases, we can't be sure it's working as expected.

Plus coverage for `subscribe.js` decreased from 100% in this PR:
https://coveralls.io/builds/18460657/source?filename=src/subscription/subscribe.js#L121

taion: Ah, I didn't see the Coveralls check. I'll add explicit test coverage then.

taion:
I believe I've addressed the PR comments and resolved the coverage hit (which came from an existing code path not being covered by tests).

src/subscription/__tests__/subscribe-test.js
```diff
@@ -933,4 +986,59 @@ describe('Subscription Publish Phase', () => {
       value: undefined,
     });
   });
+
+  it('should resolve GraphQL error from source event stream', async () => {
+    const erroringEmailSchema = emailSchemaWithResolvers(
+      async function*() {
+        yield { email: { subject: 'Hello' } };
+        throw new GraphQLError('test error');
```
taion: this is sort of silly, though.

i wonder if we should just drop `reportGraphQLError` there entirely; the case where the async iterable throws a graphql error seems exceedingly unlikely

IvanGoncharov:
@robzhu @leebyron Can you please take a look at this PR?

taion:
Any updates here?

src/subscription/subscribe.js
```diff
@@ -270,7 +280,11 @@ export function createSourceEventStream(
     return Promise.resolve(result).then(eventStream => {
       // If eventStream is an Error, rethrow a located error.
```
robzhu: This comment needs to be updated.

src/subscription/subscribe.js
```diff
@@ -270,7 +280,11 @@ export function createSourceEventStream(
...
       if (eventStream instanceof Error) {
-        throw locatedError(eventStream, fieldNodes, responsePathAsArray(path));
+        return {
```
robzhu: What if this error is something like, "Redis connection to 1.1.1.8:6379 failed"? Wouldn't this get wrapped into a GraphQLError and returned to the client?

robzhu:
The reason for this seemingly inconsistent error handling behavior in createSourceEventStream is that we wanted to provide more control over error handling. GraphQLErrors are typically communicated back to the client-- that is, errors that contain useful information for the client, similar to the HTTP 400 status code. However, createSourceEventStream would typically depend on some sort of real-time event bus, for which failures are generally going to be 500-class errors (the server's fault, not the client's), and detailed error messages should not be sent to the client. As it's currently implemented, the caller is expected to catch all exceptions coming from createSourceEventStream and can still decide to wrap it in a GraphQLError if appropriate. Does that make sense?

taion:
@robzhu 

I don't understand the difference – doesn't `subscribe` also depend on the same sorts of real-time event buses? Per your comment on https://github.com/graphql/graphql-js/pull/1467#discussion_r235450613, even when using the higher-level API, 500-class errors like failures to connect to e.g. Redis would get reported as GraphQL errors to the client.

But this is also consistent with what would happen for non-subscription fields. If e.g. the GraphQL server were unable to connect to some upstream service in resolving a normal (non-subscription) field, even when this is a 500-class error, it would still get surfaced as a GraphQL error to the client.

The only case right now where this doesn't happen is in the case where `subscribe` on the subscription field doesn't return an async iterable. I'm not totally sure that this is the correct behavior, but at least there isn't an inconsistency there.

robzhu:
> The only case right now where this doesn't happen is in the case where subscribe on the subscription field doesn't return an async iterable. I'm not totally sure that this is the correct behavior, but at least there isn't an inconsistency there.

Ah, my misunderstanding then. If what you're describing is true, then I think this change makes sense.

taion:
To clarify, the behavior of `subscribe` isn't changing at all, which is why none of the existing test cases needed to be changed. Failures in the `subscribe` method on a subscription field continue to surface errors to clients when using the `subscribe` API here.

This purely serves to unify the error handling a bit between when one uses `createSourceEventStream` v. when one uses `subscribe`, and only to remove a case where the `GraphQLError` is thrown rather than returned in an `ExecutionResult` when using the lower-level API. The higher-level API never throws a `GraphQLError` and only returns them in `ExecutionResult` objects.

taion:
Any update here? Can we move forward with this?

taion:
Any updates here?

robzhu:
Looks good to me. @IvanGoncharov any other concerns?

PR #1470, opened 2018-08-15
Benchmark: replace shell scripts with JS code

IvanGoncharov:
In order to have a better discussion around #1463 we need to add benchmarks for `validate` and `validateSDL`. I split out refactoring part into this PR.


mjmahone:
I don't think I have enough context around what we're doing with the benchmarks to give this a proper review. @leebyron might be needed here.

IvanGoncharov:
> We provide two revisions to compare against. Then we run all the benchmark tests to run, and run them one at a time using runBenchmark.
More context and description in my original PR: #1167

@mjmahone Yes, exactly.

> We provide two revisions

It's not limited to two you can provide any number 😉

PR #1471, opened 2018-08-15
Add benchmarks for GQL and SDL validation

IvanGoncharov:
In order to have a better discussion around #1463 we need to add benchmarks for `validate` and `validateSDL`. I split out refactoring part into this PR.
Note: this is PR on top of #1470


resources/benchmark.js
```diff
@@ -56,6 +56,7 @@ function prepareRevision(revision) {
       for file in $(cd "${LOCAL_DIR}src"; find . -path '*/__tests__/*');
         do cp "${LOCAL_DIR}src/$file" "${dir}/src/$file";
       done &&
+      cp -R "${LOCAL_DIR}/src/__fixtures__" "${dir}/src/__fixtures__" &&
```
IvanGoncharov: @mjmahone This makes this PR independent and can be reviewed without #1470

IvanGoncharov:
There is no perfomance difference beetween validating introspection query on `0.13.2` and current master.
image

PR #1473, opened 2018-08-17
Remove unneeded "options" from package.json

IvanGoncharov:


package.json
```diff
@@ -14,9 +14,6 @@
     "type": "git",
     "url": "http://github.com/graphql/graphql-js.git"
   },
-  "options": {
-    "mocha": "--check-leaks --full-trace --timeout 15000 src/**/__tests__/**/*-test.js"
-  },
```
IvanGoncharov: It looks like I added it accidentally in #1338
https://github.com/graphql/graphql-js/pull/1338/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R17

PR #1481, opened 2018-08-25
Update babel & conver "targets" to browserslist format

IvanGoncharov:
I run diff on Babel output and it's identical to the current master.


.babelrc.js
```diff
@@ -1,16 +1,14 @@
...
-    "targets": {
-      "node": 6,
-      "browsers": [
-        "ie 9",
-        "ios 9",
-        "last 2 chrome versions",
-        "last 2 edge versions",
-        "last 2 firefox versions",
-      ]
-    }
```
IvanGoncharov: For details, see https://github.com/babel/babel/pull/8509

mjmahone:
@IvanGoncharov I do not know which browsers Relay supports. @kassens or someone else might, though.

IvanGoncharov:
@kassens Just curious if we really need to support IE9 in `graphql-js`?

PR #1483, opened 2018-08-25
Convert more object types to be exact

IvanGoncharov:
No extremes only input types to help with detecting typos and similar to all other types in this lib.

### DIFF
--- a/src/__tests__/starWarsData.js
+++ b/src/__tests__/starWarsData.js
@@ -100,23 +100,23 @@ export type Character = {
   appearsIn: Array<number>,
 };
 
-export type Human = {
+export type Human = {|
   type: 'Human',
   id: string,
   name: string,
   friends: Array<string>,
   appearsIn: Array<number>,
   homePlanet: string,
-};
+|};
 
-export type Droid = {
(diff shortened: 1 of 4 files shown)


src/language/parser.js
```diff
@@ -70,7 +70,7 @@ import { DirectiveLocation } from './directiveLocation';
 /**
  * Configuration options to control parser behavior
  */
-export type ParseOptions = {
+export type ParseOptions = {|
```
mjmahone: This may be a little dangerous, as I think there are places we do `ParseOptions & { ... }`, such as in [buildASTSchema](https://github.com/graphql/graphql-js/blob/4fd9260efa247d60f74f3ebc259c8a3c2b03af61/src/utilities/buildASTSchema.js#L503).

This would mean I can't create a generic `const REPO_OPTIONS = { ... };` that could be used everywhere I call `parse` or `buildASTSchema`

IvanGoncharov: @mjmahone Good catch 👍
I missed that.
I reverted this change and change to `buildASTSchema` options.

PR #1487, opened 2018-08-27
[flow] Add generics for scalar class

Slessi:
> **Disclaimer**: Was asked to submit a PR here in order to match my PR at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/28419

Basically I want to add the option to specify generics on `GraphQLScalarType`, and not just `GraphQLScalarTypeConfig`

Not sure how to resolve the following issue in flow though

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/type/definition.js:49:5

Cannot use GraphQLScalarType [1] without 0-2 type arguments.

      46│  * These are all of the possible kinds of types.
      47│  */
      48│ export type GraphQLType =
      49│   | GraphQLScalarType
      50│   | GraphQLObjectType
      51│   | GraphQLInterfaceType
      52│   | GraphQLUnionType
        :
 [1] 535│ export class GraphQLScalarType<TInternal = any, TExternal = any> {
```

I would not want to create a change that forces users to add `<>` to their code - Is there a way in flow to make the generics totally optional as in TypeScript?

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -532,7 +532,7 @@ function resolveThunk<+T>(thunk: Thunk<T>): T {
  *     });
  *
  */
-export class GraphQLScalarType<TInternal, TExternal> {
+export class GraphQLScalarType<TInternal = any, TExternal = any> {
   name: string;
   description: ?string;
   serialize: GraphQLScalarSerializer<TExternal>;
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -532,16 +532,16 @@ function resolveThunk<+T>(thunk: Thunk<T>): T {
  *     });
  *
  */
-export class GraphQLScalarType {
+export class GraphQLScalarType<TInternal, TExternal> {
   name: string;
   description: ?string;
-  serialize: GraphQLScalarSerializer<*>;
-  parseValue: GraphQLScalarValueParser<*>;
-  parseLiteral: GraphQLScalarLiteralParser<*>;
+  serialize: GraphQLScalarSerializer<TExternal>;
+  parseValue: GraphQLScalarValueParser<TInternal>;
+  parseLiteral: GraphQLScalarLiteralParser<TInternal>;
   astNode: ?ScalarTypeDefinitionNode;
   extensionASTNodes: ?$ReadOnlyArray<ScalarTypeExtensionNode>;
 
-  constructor(config: GraphQLScalarTypeConfig<*, *>): void {
+  constructor(config: GraphQLScalarTypeConfig<TInternal, TExternal>): void {
     this.name = config.name;
     this.description = config.description;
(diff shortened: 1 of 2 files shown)


sibelius:
@Slessi can you check this as well https://github.com/gtkatakura/fullstack-challenge/commit/36cc1050768e29824f7c7114d22c131d43aefef9

```jsx
export const GraphQLInt: GraphQLScalarType<number>;
export const GraphQLFloat: GraphQLScalarType<number>;
export const GraphQLString: GraphQLScalarType<string>;
export const GraphQLBoolean: GraphQLScalarType<boolean>;
export const GraphQLID: GraphQLScalarType<string>;
```

following this approach we could use mapped types in Typescript to transform GraphQLInt to `number`

Slessi:
@sibelius I don't understand what you are trying to tell me.

I made a PR to `@types/graphql` with the same changes at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/28419 but it was rejected because it needs to match the types here

sibelius:
you PR does not add correct types for GraphQLInt for instance

Slessi:
Yes but what difference does that make in practice? If you're making a custom scalar resolver, you can just make a new type, i.e.

```ts
const ID = new GraphQLScalarType<number, string>({
    name: "ID",
    description: "ID Object",
    serialize: (value) => String(value),
    parseValue,
    parseLiteral: (valueNode, variables) => {
        if (valueNode.kind !== "StringValue") {
            throw new Error("Expected StringValue");
        }

        return parseValue(valueNode.value);
    },
});
```

sibelius:
you can use Mapped Types on Typescript to transform `GraphQLInt` to `number`, no need to retype this on TS as well

sibelius:
tests still failing

I think in flow default of generics is different

<TInternal: any>

or you should try

<TInternal?: any>

try running flow check locally to make it typecheck properly

Slessi:
@sibelius I don't know how to fix it, as mentioned in OP. If you want to try yourself, feel free

gtkatakura-bysoft:
I think this is not going to be possible ...
> I would not want to create a change that forces users to add `<>` to their code - Is there a way in flow to make the generics totally optional as in TypeScript?

At least according to the Flow documentation:
captura de tela 2018-09-13 as 14 27 46

Slessi:
@gtkatakura-bysoft thanks for that, going to close this then as its far too trivial an improvement to merit a breaking change like that.

Unfortunate however that the maintainers of `@types/graphql` won't accept a PR to improve TS types due to a shortcoming of flow

PR #1492, opened 2018-08-29
Correctly detect required/optional args/fields

IvanGoncharov:
Disscussed here: https://github.com/graphql/graphql-js/pull/1465#issuecomment-413699023

### DIFF
--- a/src/utilities/findBreakingChanges.js
+++ b/src/utilities/findBreakingChanges.js
@@ -17,6 +17,8 @@ import {
   isNonNullType,
   isListType,
   isNamedType,
+  isRequiredArgument,
+  isRequiredInputField,
 } from '../type/definition';
 
 import type {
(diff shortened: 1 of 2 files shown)


src/utilities/findBreakingChanges.js
```diff
@@ -42,22 +44,22 @@ export const BreakingChangeType = {
...
 
 export const DangerousChangeType = {
   ARG_DEFAULT_VALUE_CHANGE: 'ARG_DEFAULT_VALUE_CHANGE',
   VALUE_ADDED_TO_ENUM: 'VALUE_ADDED_TO_ENUM',
   INTERFACE_ADDED_TO_OBJECT: 'INTERFACE_ADDED_TO_OBJECT',
   TYPE_ADDED_TO_UNION: 'TYPE_ADDED_TO_UNION',
-  NULLABLE_INPUT_FIELD_ADDED: 'NULLABLE_INPUT_FIELD_ADDED',
-  NULLABLE_ARG_ADDED: 'NULLABLE_ARG_ADDED',
+  OPTIONAL_INPUT_FIELD_ADDED: 'OPTIONAL_INPUT_FIELD_ADDED',
+  OPTIONAL_ARG_ADDED: 'OPTIONAL_ARG_ADDED',
```
IvanGoncharov: It's breaking change but it's very specific API so I don't think it will affect many users especially since most of them just checking for Breaking changes.
Here is few examples:
https://github.com/search?q=NON_NULL_ARG_ADDED+-filename%3AfindBreakingChanges&type=Code

PR #1494, opened 2018-08-29
Exclude __fixtures__ from dist

IvanGoncharov:
Currently `__fixtures__` include in NPM package:
https://github.com/graphql/graphql-js/tree/npm/__fixtures__
Cleanup from #1471
Don't like how it's make `package.json` less readable but it's a normal temporary solution.


IvanGoncharov:
> Thanks for cleaning those up!

It's the cycle of life first you break things and then you fix them and hopefully don't introduce new bugs in the process 😄

PR #1499, opened 2018-09-03
Cleanup 'jsutils/inspect' function

IvanGoncharov:


src/execution/__tests__/executor-test.js
```diff
@@ -1053,7 +1053,7 @@ describe('Execute: Handles basic execution tasks', () => {
       errors: [
         {
           message:
-            'Expected value of type "SpecialType" but got: {value: "bar"}.',
+            'Expected value of type "SpecialType" but got: { value: "bar" }.',
```
IvanGoncharov: Same as standard `util.inspect`:
```
> util.inspect({ a: 1 })
'{ a: 1 }'
```

src/jsutils/inspect.js
```diff
@@ -11,19 +11,25 @@
...
+          return value.inspect();
+        } else if (Array.isArray(value)) {
+          return '[' + value.map(inspect).join(', ') + ']';
+        }
+
+        const properties = Object.keys(value)
+          .map(k => `${k}: ${inspect(value[k])}`)
+          .join(', ');
+        return properties ? '{ ' + properties + ' }' : '{}';
+      }
```
mjmahone: I would personally add a default case rather than falling through, just because it feels a little bit easier to read.

So I would personally do:
```
case 'object':
  if (value) {
    ...
  }
  return String(value);
default:
  return String(value);
```

But I don't have a strong objections if there's a reason to leave it as is.

IvanGoncharov: @mjmahone Agree 👍 
Trying to be cleaver, usually results in stupid bugs 😄

IvanGoncharov: @mjmahone Fixed

PR #1502, opened 2018-09-03
Automatically setup npm publish tag

IvanGoncharov:
Prevents issues similar to #1375
See: https://docs.npmjs.com/files/package.json#publishconfig

### DIFF
--- a/resources/copy-package-json.js
+++ b/resources/copy-package-json.js
@@ -11,8 +11,19 @@
  */
 
 const fs = require('fs');
+const assert = require('assert');
 
 const package = require('../package.json');
 delete package.scripts;
 delete package.devDependencies;
+
+const { version } = package;
+const match = /^[0-9]+\.[0-9]+\.[0-9]+-?([^.]*)/.exec(version);
+assert(match, 'Version does not match semver spec.');
+const tag = match[1];
+assert(!tag || tag === 'rc', 'Only "rc" tag is supported.');
+
+assert(!package.publishConfig, 'Can not override "publishConfig".');
+package.publishConfig = { tag: tag || 'latest' };
+
 fs.writeFileSync('./dist/package.json', JSON.stringify(package, null, 2));


resources/copy-package-json.js
```diff
@@ -11,8 +11,20 @@
...
 const package = require('../package.json');
 delete package.scripts;
 delete package.devDependencies;
+
+const { version } = package;
+const match = /^[0-9]+\.[0-9]+\.[0-9]+-?([^.]*)/.exec(version);
+assert(match, 'Version does not match semver spec.');
+const tag = match[1];
+if (tag) {
+  assert(tag !== 'rc', 'Only "rc" tag is supported.');
```
mjmahone: This seems backwards, if only the 'rc' tag is supported.

IvanGoncharov: `latest` is added by default:
https://docs.npmjs.com/cli/publish#description
> By default, npm publish updates and npm install installs the latest tag

I will add a comment to make it more clear.

IvanGoncharov:
@mjmahone I think I need to explain the reason for my current activities.
After `14.0.0` we need to do the same process for rest of `graphql/*` projects.
I think it makes sense to unify how all these projects build, validate, publish and generate docs. So in a next few weeks, I will try to update `graphql-js` and also implement doc generation as a build step.

Reason for this particular change is that after Greg and Hao switched from GraphQL to other projects many `graphql/*` is stale. Majority of this projects have auto-publish through Travis but they can't publish RC that way.

For example, I also maintain `graphql/express-graphql`(AFAIK Facebook doesn't use it but I depend on it in a few projects) and I want to also make a stable release for this lib. But I can't make RC there since I don't have NPM publish rights. At the same time, I feel like it's not an `expess-graphql` specific issue and all other `graphql/*` projects will benefit from it.

resources/copy-package-json.js
IvanGoncharov: @mjmahone I change the code to explicitly set `latest`.
Can you please review changes.

IvanGoncharov:
@mjmahone Can you please review updated version?

resources/copy-package-json.js
```diff
@@ -11,8 +11,19 @@
...
 
 const package = require('../package.json');
 delete package.scripts;
 delete package.devDependencies;
+
+const { version } = package;
+const match = /^[0-9]+\.[0-9]+\.[0-9]+-?([^.]*)/.exec(version);
+assert(match, 'Version does not match semver spec.');
+const tag = match[1];
+assert(!tag || tag !== 'rc', 'Only "rc" tag is supported.');
```
mjmahone: Shouldn't this be `assert(!tag || tag === 'rc', 'Only "rc" tag is supported.');`? We only support the RC tag, so we don't want to allow non-`rc` tags getting through.

IvanGoncharov: @mjmahone Thanks. Fixed

IvanGoncharov:
@mjmahone Quick update: it working as expected, see here:
https://github.com/graphql/graphql-js/blob/adf4a29b739c3d39fbdc9c35fee9a4e6644d1665/package.json#L27-L29

PR #1503, opened 2018-09-04
Add missing ESLint rules as comments

IvanGoncharov:
Idea is to first add all of them as comment and than enable on per rule
basis.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -38,19 +38,47 @@ rules:
   # `eslint-plugin-flowtype` rule list based on `v2.50.0`
   # https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype
 
+  #flowtype/array-style-complex-type: undecided
+  #flowtype/array-style-simple-type: undecided
+  #flowtype/boolean-style: undecided
   flowtype/define-flow-type: error
+  #flowtype/delimiter-dangle: undecided
+  #flowtype/generic-spacing: undecided
+  #flowtype/newline-after-flow-annotation: undecided
+  #flowtype/no-dupe-keys: undecided
+  #flowtype/no-existential-type: undecided
+  #flowtype/no-flow-fix-me-comments: undecided
+  #flowtype/no-mutable-array: undecided
+  #flowtype/no-primitive-constructor-types: undecided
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
No changes to existing rules I just sorted them in the same order as official docs.
All new rules added as comments.

mjmahone:
So anything that's commented out we want to have on, but we haven't turned it on yet? Any of those that can't be automatically turned on would be great to have as "simple first contribution" tasks for anyone looking to start contributing to the repo.

I think the re-order should be a separate PR: that way, it's clear in this PR which rules we want to add, and we can discuss whether those additions are valuable, before encouraging people to modify the repo to get those rules passing.

IvanGoncharov:
> So anything that's commented out we want to have on, but we haven't turned it on yet?

No, it's just all rules that were added in ESLint. Idea is to keep existing practice of explicitly marking rules as `off` or `error` so when comparing with ESLint list we can be sure that certain rule was evaluated and explicitly rejected.
Not all ESLint rules make sense for this lib, e.g.:
https://eslint.org/docs/rules/max-lines-per-function

I extracted separate PR with rearrangements: #1511

.eslintrc.yml
```diff
@@ -32,182 +32,328 @@ parserOptions:
   unicodeCodePointEscapes: true
 
 rules:
+  # https://github.com/prettier/eslint-plugin-prettier#installation
   prettier/prettier: error
 
+  # `eslint-plugin-flowtype` rule list based on `v2.50.0`
+  # https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype
+
+  #flowtype/array-style-complex-type: undecided
```
IvanGoncharov: @mjmahone now all new rules are marked as `undecided` to prevent confusion

mjmahone: Nice! Thank you

leebyron:
Note that any of these that deal with formatting (most of the flow ones) should be explicitly off if we're going to use prettier

IvanGoncharov:
@leebyron Thanks for the info 👍 
I was planning to use this list https://github.com/prettier/eslint-config-prettier/blob/master/index.js

PR #1508, opened 2018-09-05
update package.json file to support node 9.x

simcoder:
Some of use are using node 9.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "graphql-js"
   ],
   "engines": {
-    "node": "6.x || 8.x || >= 10.x"
+    "node": "6.x || 8.x || 9.x || >= 10.x"
   },
   "scripts": {
     "watch": "node ./resources/watch.js",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "graphql-js"
   ],
   "engines": {
-    "node": "6.x || 8.x || 9.x || >= 10.x"
+    "node": ">= 6.x"
   },
   "scripts": {
     "watch": "node ./resources/watch.js",


IvanGoncharov:
Additional discussion here: https://github.com/graphql/graphql-js/pull/1445#discussion_r215288252

@mjmahone I think we should merge this PR and since master doesn't contain any breaking/feature changes on top of `14.0.0` we can release this fix as `14.0.1`.

And I will try to either remove `sane` dependency or revert it back to the version that supports node v9.

leebyron:
I actually think this should change to “>= 6.x”

The engines field suggests minimum support. I don’t think it’s appropriate to remove the odd versions and create install warnings just because we stop testing them in Travis. 

Travis should test all currently supported versions of node. We don’t need to test older odd versions. 

There’s no need to change the version of sane. That’s a dev dependency and we only expect those to apply in our Travis builds

leebyron:
LTS only means the node versions will get security updates in the future. While a node version out of support may be unwise to use in production, that’s not really our concern or our place to enforce. As long as we’re testing the oldest node version we claim to support and the newest, then we have a very high confidence of coverage for all versions in between. That’s especially true for this library since we do not rely on Node API which can change from version to version. We only rely on JS core API and that’s always backwards compatible.

leebyron:
TL;DR, we should only use the “engines” field to exclude node versions if we think it is unsafe to run this library on those versions. If node < 6 is lacking core JS API or syntax support, then “engines” is an appropriate tool to exclude those.

IvanGoncharov:
@leebyron Agree 👍 I updated PR with your suggestion and merged it.

Since we don't use any Node API or 3rd-party modules (except `iterall` which doesn't have any dependencies) we depend only on JS feature set which should be stay the same for higher version. Plus we are limited by IE anyway 😞

IvanGoncharov:
@mjmahone @leebyron Can you please release the new version?

IvanGoncharov:
I think it should be `14.0.1`.

leebyron:
Yep, I'll start that release now

jure:
Thank you! Also, in terms of continuing to test in v9, it should also be possible to simply ignore `sane`'s engines configuration.

PR #1509, opened 2018-09-06
Prevent infinite loop on invalid introspection

IvanGoncharov:


src/utilities/__tests__/buildClientSchema-test.js
```diff
@@ -844,4 +844,41 @@ describe('Type System: build schema from introspection', () => {
...
+              name: 'Foo',
+              kind: 'OBJECT',
+              fields: [],
+              interfaces: [{ name: 'Foo' }],
+            },
+          ],
+        },
+      };
+      expect(() => buildClientSchema(introspection)).to.throw(
+        'Expected Foo to be a GraphQL Interface type.',
```
IvanGoncharov: Before this change they throw:
> Maximum call stack size exceeded

PR #1514, opened 2018-09-06
ESLint: Disable rules conflicting with Prettier

IvanGoncharov:


.eslintrc.yml
```diff
@@ -40,10 +40,10 @@ rules:
 
   #flowtype/array-style-complex-type: undecided
   #flowtype/array-style-simple-type: undecided
-  #flowtype/boolean-style: undecided
+  flowtype/boolean-style: off
```
IvanGoncharov: Based on this list:
https://github.com/prettier/eslint-config-prettier/blob/master/flowtype.js

.eslintrc.yml
```diff
@@ -226,10 +226,10 @@ rules:
   # Stylistic Issues
   # https://eslint.org/docs/rules/#stylistic-issues
 
-  #array-bracket-newline: undecided
-  #array-bracket-spacing: undecided
-  #array-element-newline: undecided
-  #block-spacing: undecided
+  array-bracket-newline: off
```
IvanGoncharov: Based on this list: https://github.com/prettier/eslint-config-prettier/blob/master/index.js

PR #1520, opened 2018-09-11
Call 'toJSON' if present for ID and String serialize

### LINKED ISSUE #1518: Unable to query for MongoDB ObjectIDs
I recently updated from `0.13.2` to `14.0.2` which includes breaking changes.

This introduced errors with existing queries which include MongoDB Object Ids (probably from #1382):

> ID cannot represent value: { _bsontype: \"ObjectID\", id: <Buffer 5b 96 3d bf 98 0a 04 09 85 c6 6e a1> }

Repository with complete, minimal repeatable example [here](https://github.com/alexmcmillan/graphql-mongodb-query-fail-demo):

```
const Thing = mongoose.model('Thing', new mongoose.Schema({
  id: mongoose.Schema.Types.ObjectId,
  name: String
}));

const ThingType = new GraphQLObjectType({
  name: 'thing',
  fields: function () {
    return {
      id: { type: GraphQLID },
      name: { type: GraphQLString }
    }
  }
});

const RootMutation = new GraphQLObjectType({
  name: 'CreateMutation',
  fields: {
    create: {
      type: ThingType,
      description: 'Create new thing',
      args: {
        name: {
          name: 'Name of Thing',
          type: new GraphQLNonNull(GraphQLString)
        }
      },
      resolve: (root, args) => {
        const newThing = new Thing({ name: args.name });
        newThing.id = newThing._id;
        return new Promise((res, rej) => {
          newThing.save(err => {
            if (err) return rej(err);
            res(newThing);
          });
        });
      }
    }
  }
});
```

IvanGoncharov:
Fixes #1518 

`class ObjectId` doesn't have `valueOf` only `toString` and `toJSON`:
https://github.com/mongodb/js-bson/blob/5acdebf7a63ed02d3bc1eb8481cdcb709fb5c886/lib/objectid.js#L213-L215

We can't use `toString` because:
```js
class A {}
const a = new A();
a.toString()
// '[object Object]'
```

So I added support for `toJSON`.

### DIFF
--- a/src/type/scalars.js
+++ b/src/type/scalars.js
@@ -117,24 +117,39 @@ export const GraphQLFloat = new GraphQLScalarType({
   },
 });
 
-function serializeString(value: mixed): string {
-  // Support serializing objects with custom valueOf() functions - a common way
-  // to represent an complex value which can be represented as a string
-  // (ex: MongoDB id objects).
-  const result =
-    value && typeof value.valueOf === 'function' ? value.valueOf() : value;
+// Support serializing objects with custom valueOf() or toJSON() functions -
+// a common way to represent a complex value which can be represented as
+// a string (ex: MongoDB id objects).
+function serializeObject(value: mixed): mixed {
+  if (typeof value === 'object' && value !== null) {
+    if (typeof value.valueOf === 'function') {
+      const valueOfResult = value.valueOf();
+      if (typeof valueOfResult !== 'object') {
+        return valueOfResult;
+      }
+    }
+    if (typeof value.toJSON === 'function') {
+      return value.toJSON();
+    }
+  }
+  return value;
+}
+
+function serializeString(rawValue: mixed): string {
+  const value = serializeObject(rawValue);
+
   // Serialize string, boolean and number values to a string, but do not
   // attempt to coerce object, function, symbol, or other types as strings.
-  if (typeof result === 'string') {
-    return result;
+  if (typeof value === 'string') {
+    return value;
   }
-  if (typeof result === 'boolean') {
-    return result ? 'true' : 'false';
+  if (typeof value === 'boolean') {
+    return value ? 'true' : 'false';
   }
-  if (isFinite(result)) {
-    return result.toString();
+  if (isFinite(value)) {
+    return value.toString();
   }
-  throw new TypeError(`String cannot represent value: ${inspect(value)}`);
+  throw new TypeError(`String cannot represent value: ${inspect(rawValue)}`);
 }
 
 function coerceString(value: mixed): string {
@@ -190,18 +205,16 @@ export const GraphQLBoolean = new GraphQLScalarType({
   },
 });
 
-function serializeID(value: mixed): string {
-  // Support serializing objects with custom valueOf() functions - a common way
-  // to represent an object identifier (ex. MongoDB).
-  const result =
-    value && typeof value.valueOf === 'function' ? value.valueOf() : value;
-  if (typeof result === 'string') {
-    return result;
+function serializeID(rawValue: mixed): string {
+  const value = serializeObject(rawValue);
+
+  if (typeof value === 'string') {
+    return value;
   }
-  if (isInteger(result)) {
-    return String(result);
+  if (isInteger(value)) {
+    return String(value);
   }
-  throw new TypeError(`ID cannot represent value: ${inspect(value)}`);
+  throw new TypeError(`ID cannot represent value: ${inspect(rawValue)}`);
 }
 
 function coerceID(value: mixed): string {
--- a/src/type/__tests__/serialization-test.js
+++ b/src/type/__tests__/serialization-test.js
@@ -111,11 +111,19 @@ describe('Type System: Scalar coercion', () => {
 
     const stringableObjValue = {
       valueOf() {
-        return 'something useful';
+        return 'valueOf string';
+      },
+      toJSON() {
+        return 'toJSON string';
       },
     };
     expect(GraphQLString.serialize(stringableObjValue)).to.equal(
-      'something useful',
+      'valueOf string',
+    );
+
+    delete stringableObjValue.valueOf;
+    expect(GraphQLString.serialize(stringableObjValue)).to.equal(
+      'toJSON string',
     );
 
     expect(() => GraphQLString.serialize(NaN)).to.throw(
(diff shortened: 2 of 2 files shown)


IvanGoncharov:
@alexmcmillan I don't have access to MongoDB so I can't run: https://github.com/alexmcmillan/graphql-mongodb-query-fail-demo

Can you please test it with attached package:
[graphql-14.0.2.zip](https://github.com/graphql/graphql-js/files/2372034/graphql-14.0.2.zip)
Just unpack it inside `node_modules`.

alexmcmillan:
@IvanGoncharov Can confirm the package you attached does resolve the problem.  Thank you - I hope this gets merged soon!

IvanGoncharov:
@alexmcmillan Thanks 👍 
@mjmahone Can you please review?
`mongoose` is a pretty popular library so I think it's worth to make `14.1.0`.
What do you think?

src/type/scalars.js
```diff
@@ -117,24 +117,36 @@ export const GraphQLFloat = new GraphQLScalarType({
...
 });
 
-function serializeString(value: mixed): string {
-  // Support serializing objects with custom valueOf() functions - a common way
-  // to represent an complex value which can be represented as a string
-  // (ex: MongoDB id objects).
-  const result =
-    value && typeof value.valueOf === 'function' ? value.valueOf() : value;
+// Support serializing objects with custom toJSON() or valueOf() functions -
+// a common way to represent an complex value which can be represented as
```
mjmahone: grammar nit: "a complex value" instead of "an complex value"

IvanGoncharov: Thanks 👍 
Fixed

IvanGoncharov:
> Prior to doing the bump to 14.1.0 I think we'd want @leebyron to help fix our NPM push process: I don't want to accidentally mess up another release, and I don't want Lee to be on the hook for overseeing and validating each patch

Agree, we definitely need to improve [bus factor](https://en.wikipedia.org/wiki/Bus_factor) 👍

> Can the version bump wait until the next WG meeting? 

Yes 👍 

> I'm not sure why this would be a minor release rather than a patch, and I don't currently have the right framework for figuring out that answer.

AFAIK, it covered by semver:
https://semver.org/#spec-item-7
> It MAY be incremented if substantial new functionality or improvements are introduced within the private code.

My personal rule of thumb: if you add something new that clients explicitly depend on it means you need to do `minor` release. So if someone write the resolvers that depend on `toJSON` that mean this will work only on `>= 14.1.0` and if for somereason we release `14.0.3` it wouldn't contain this feature.

It also important for 3rd tools like `graphql-tools` and other graphql middlewares since they wrapping resolvers and may be surprised by object with `toJSON` so they have an option to support only `14.0.x` and be shielded from new features.

So for me this change is not a bug fix but a new functionality especially since `0.13.x` and previous version didn't use `toJSON` (they used `toString`).

src/type/scalars.js
```diff
@@ -117,24 +117,36 @@ export const GraphQLFloat = new GraphQLScalarType({
...
-  // Support serializing objects with custom valueOf() functions - a common way
-  // to represent an complex value which can be represented as a string
-  // (ex: MongoDB id objects).
-  const result =
-    value && typeof value.valueOf === 'function' ? value.valueOf() : value;
+// Support serializing objects with custom toJSON() or valueOf() functions -
+// a common way to represent a complex value which can be represented as
+// a string (ex: MongoDB id objects).
+function serializeObject(value: mixed): mixed {
+  if (typeof value === 'object' && value !== null) {
```
ruiaraujo: ```!= nulll```

IvanGoncharov:
```
> typeof undefined
'undefined'
```
@ruiaraujo So it will fail on the first check.

src/type/scalars.js
```diff
@@ -117,24 +117,36 @@ export const GraphQLFloat = new GraphQLScalarType({
...
+    if (typeof value.toJSON === 'function') {
```
leebyron: This change is actually breaking - there may be objects which define *both* `valueOf` and `toJSON` which provide different values. If those objects previously worked correctly, this PR would cause them to work incorrectly.

If the goal is to also support objects which define `toJSON` but *do not* define `valueOf` then this could be easily solved by just moving this new addition to *below* the existing check for `valueOf`

IvanGoncharov: I did that because: 
```
> typeof ({}).valueOf
'function'
```
But I agree it breaking so I updated this PR to have this instead:
```js
    if (typeof value.valueOf === 'function') {
      const valueOfResult = value.valueOf();
      if (typeof valueOfResult !== 'object') {
        return valueOfResult;
      }
    }
    if (typeof value.toJSON === 'function') {
      return value.toJSON();
    }
```
I also changed tests to explicitly test precidence.
Can you please review new code and tests?

IvanGoncharov:
@mjmahone Since you reviewed this PR and I already addressed @leebyron comments I merging this issue, so people can atleast use our `npm` branch.
Would be great to disscuss the scope for `14.1.0` release, can you please comment in #1543 ?

frankRose1:
> I am having this problem also. Using "graphql": "^14.0.2"
> **"ID cannot represent value: { _bsontype: "ObjectID"**
> In a query resolver such as below, can someone explain what I do to fix it?
> Thanks
> 
> ```
>     user: async (parent, args, context, info) => {
>       let user = await context.UserModel.findOne(args);
>       return user;
>     }
> ```

I know its a bit of a late response but In regards to xabra's post try this:
```javascript
user: async(parent, args, ctx, info){
     let user = await ctx.UserModel.findOne(args);
     return { ...user._doc, _id: user._id.toString() };
}
```

EmiyaYang:
> > I am having this problem also. Using "graphql": "^14.0.2"
> > **"ID cannot represent value: { _bsontype: "ObjectID"**
> > In a query resolver such as below, can someone explain what I do to fix it?
> > Thanks
> > ```
> >     user: async (parent, args, context, info) => {
> >       let user = await context.UserModel.findOne(args);
> >       return user;
> >     }
> > ```
> 
> I know its a bit of a late response but In regards to xabra's post try this:
> 
> ```js
> user: async(parent, args, ctx, info){
>      let user = await ctx.UserModel.findOne(args);
>      return { ...user._doc, _id: user._id.toString() };
> }
> ```

Well, I have no problems to query the _id field. But once I populate something I meet this error.

```ts
  function findOne(id: string) {
    const res = this.nodeModel.findOne({ id }).populate('parent');

    return res;
  }
```

PR #1528, opened 2018-09-20
add testcases for nested list coercion

jjergus:
Add some test cases to cover coercion of list values, especially nested lists where some non-list values are involved. These exercise the following special case from the [spec](http://facebook.github.io/graphql/June2018/#sec-Type-System.List):

> If the value passed as an input to a list type is not a list and not the null value, then the result of input coercion is a list of size one, where the single item value is the result of input coercion for the list’s item type on the provided value (note this may apply recursively for nested lists).

Most of these examples are copied directly from the spec. However, one of them actually coerces to a value different from what the example in the spec says, which I believe to be a mistake in the spec: https://github.com/facebook/graphql/pull/515

@IvanGoncharov what do you think?

### DIFF
--- a/src/utilities/__tests__/coerceValue-test.js
+++ b/src/utilities/__tests__/coerceValue-test.js
@@ -12,6 +12,7 @@ import {
   GraphQLID,
   GraphQLInt,
   GraphQLFloat,
+  GraphQLList,
   GraphQLString,
   GraphQLEnumType,
   GraphQLInputObjectType,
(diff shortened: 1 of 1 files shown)


src/utilities/__tests__/coerceValue-test.js
```diff
@@ -283,4 +284,39 @@ describe('coerceValue', () => {
       ]);
     });
   });
+
+  describe('for nested GraphQLList', () => {
```
IvanGoncharov: @jjergus Thanks for PR 👍 
I think it better to have `describe` for `for GraphQLList` and just implement all test from this table: http://facebook.github.io/graphql/June2018/#sec-Type-System.List

src/utilities/__tests__/coerceValue-test.js
```diff
@@ -283,4 +284,39 @@ describe('coerceValue', () => {
...
+    const TestList = GraphQLList(GraphQLList(GraphQLInt));
+
+    it('correctly coerces a nested list value', () => {
```
IvanGoncharov: I think it's better to use existing naming convention, e.g.
https://github.com/graphql/graphql-js/blob/7bc632a22c358af7746e23231fa657a2adb3c25d/src/utilities/__tests__/coerceValue-test.js#L131-L134
https://github.com/graphql/graphql-js/blob/7bc632a22c358af7746e23231fa657a2adb3c25d/src/utilities/__tests__/coerceValue-test.js#L75-L78

So instead of `correctly coerces a nested list value` it could be `return value for nested list`.
Same goes for other testcases

Issue #1484, opened 2018-08-26
Enable Flow in test files


IvanGoncharov:
Currently, tests are covering only runtime behavior and not testing Flow types.
So it's very easy to break Flow types for public APIs and have it to pass `yarn t`, e.g. https://github.com/graphql/graphql-js/pull/1483#discussion_r212811061

PR #1533, opened 2018-09-23
Explicitly annotate (@flow/@noflow) every JS file

IvanGoncharov:
It's the first step towards #1484 and also enables additional ESLint rules.

### DIFF
--- a/src/__fixtures__/index.js
+++ b/src/__fixtures__/index.js
@@ -1,3 +1,12 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
 import { join } from 'path';
 import { readFileSync } from 'fs';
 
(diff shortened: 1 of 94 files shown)


src/__fixtures__/index.js
```diff
@@ -1,3 +1,12 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
```
IvanGoncharov: New ESLint rules helped to detect missing Copyright.

src/language/printer.js
```diff
@@ -3,6 +3,8 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
+ *
+ * @noflow
```
IvanGoncharov: It's the last non-test file that doesn't cover by Flow.

src/error/__tests__/GraphQLError-test.js
```diff
@@ -3,6 +3,8 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
+ *
+ * @noflow
```
IvanGoncharov: It allows to easily grep for non-Flow files.

PR #1541, opened 2018-10-02
[RFC] Added support for repeatable directives

OlegIlyenko:
This in an implementation for a spec proposal:

* Spec proposal: [[RFC] Limit directive uniqueness to explicitly marked directives](https://github.com/facebook/graphql/pull/472)
* Original issue: [Limiting the scope of "Directives Are Unique Per Location" validation](https://github.com/facebook/graphql/issues/429)

I think I have covered all places:

* SDL AST & parsing
* Schema printer (based on AST as well as based `GraphQL*` types)
* Introspection
* `buildASTSchema` & `buildClientSchema`
* The actual validation rule

Please let me know if I forgot something.


OlegIlyenko:
Sorry, looks like I did my changes on an outdated version of graphql-js (forgot to sync my fork :man_facepalming:). I'm updating it :)

src/validation/rules/UniqueDirectivesPerLocation.js
```diff
@@ -48,7 +70,11 @@ export function UniqueDirectivesPerLocation(
               ]),
             );
           } else {
-            knownDirectives[directiveName] = directive;
+            const repeatable = repeatableMap[directiveName];
```
IvanGoncharov: Can you just do:
```js
for (const directive of directives) {
  const directiveName = directive.name.value;
  const repeatable = repeatableMap[directiveName];
  if (repeatable === undefined || !repeatable) {
    continue;
  }
  // ...
}
```

src/type/directives.js
```diff
@@ -44,12 +44,17 @@ export class GraphQLDirective {
...
   astNode: ?DirectiveDefinitionNode;
+  repeatable: boolean;
 
   constructor(config: GraphQLDirectiveConfig): void {
     this.name = config.name;
     this.description = config.description;
     this.locations = config.locations;
     this.astNode = config.astNode;
+    this.repeatable =
+      config.repeatable === undefined || config.repeatable === null
```
IvanGoncharov: In JS you can do that with a single `config.repeatable == null` and we already using it in couple places.

IvanGoncharov:
> Please let me know if I forgot something.

@OlegIlyenko You also need to handle `repeatable` in `extendSchema`:
https://github.com/graphql/graphql-js/blob/94b38446d1650aef6884e3e16682d5b2266c93ea/src/utilities/extendSchema.js#L317-L323
Also remove of `repeatable` is breaking change so would be great to have it here:
https://github.com/graphql/graphql-js/blob/94b38446d1650aef6884e3e16682d5b2266c93ea/src/utilities/findBreakingChanges.js#L96

src/validation/rules/UniqueDirectivesPerLocation.js
```diff
@@ -48,7 +70,11 @@ export function UniqueDirectivesPerLocation(
...
+
+            if (repeatable === undefined || !repeatable) {
```
IvanGoncharov: Up to discussion: I don't think unknown directives should be treated as non-repeatable. Unknown directives should trigger `KnownDirectives` and other rules shouldn't make any assumption about it.
Previously it was safe since all directives were non-repeatable.

src/type/directives.js
OlegIlyenko: I believe `config.repeatable === undefined` is still necessary since I defined the flow type like this:

```
repeatable?: ?boolean,
```

If I don't do the check, I will get following flow error:

```
Cannot assign `config.repeatable === null ? false : config.repeatable` to `this.repeatable` because:
 - null or undefined [1] is incompatible with boolean [2].
 - undefined [1] is incompatible with boolean [2].
```

IvanGoncharov: @OlegIlyenko Note `== null` (two equal signs) compares both to `null` and `undefined` and Flow should handle it without problems.

> If I don't do the check, I will get following flow error:

It's because you use `===` instead of `==`.
It's safe and very common pattern in JS to do `== null`:
https://eslint.org/docs/rules/eqeqeq#smart

OlegIlyenko: Ah, I see. Haven't noticed this, thanks for the clarification.

I got an impression that the use of `==` is discouraged in javascript community since it relies of a set of coercion rules that many people find error-prone. This is why I tried to avoid using it. I was not aware of the exceptions described in the eslint documentation though. I will update the code and use `==`.

src/validation/rules/UniqueDirectivesPerLocation.js
OlegIlyenko: It is a good point. After thinking about it, I think I would agree with you. In both scenations: directive is not known and it is known but repeatable, it should be excluded from this validation entirely. I will adjust the logic.

src/validation/rules/UniqueDirectivesPerLocation.js
OlegIlyenko: I guess I generally trying to avoid usage of `break` and `continue`, but I can update the code and use it here.

OlegIlyenko:
@IvanGoncharov Thanks for pointing out 2 remaining places! I believe I addressed all review comments now.

src/utilities/introspectionQuery.js
```diff
@@ -33,6 +33,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
           args {
             ...InputValue
           }
+          repeatable
```
IvanGoncharov: @OlegIlyenko As discussed on the working group we can't assume that legacy servers would support `repeatable`. So this change will break GraphiQL and many other tools.

I would suggest making option `directiveRepeatable` similar to this one:
https://github.com/graphql/graphql-js/blob/f8378d0a6f75b69447e251ad305bc006640d091e/src/utilities/introspectionQuery.js#L12-L16

Long term solution would be to implement two-stage introspection as suggested by Lee.

src/validation/rules/UniqueDirectivesPerLocation.js
IvanGoncharov: @OlegIlyenko Thinking about this can you change `repeatableMap` to `uniqueDirectiveMap` with reverse value and do:
```js 
for (const directive of directives) {
  const directiveName = directive.name.value;
  if (uniqueDirectiveMap[directiveName]) {
    // ...
  }
}
```
Plus `uniqueDirectiveMap` match nicely with `UniqueDirectivesPerLocation` rule name.

src/utilities/introspectionQuery.js
```diff
@@ -273,4 +274,5 @@ export type IntrospectionDirective = {|
   +description?: ?string,
   +locations: $ReadOnlyArray<DirectiveLocationEnum>,
   +args: $ReadOnlyArray<IntrospectionInputValue>,
+  +repeatable: boolean,
```
IvanGoncharov: @OlegIlyenko Can you please address this comment in spec RFC:
https://github.com/facebook/graphql/pull/472/files/1ba4e196aa710a10eceaa346d02a970dfd0b2d3a#diff-35fb422eded6e8f4df638c0d159008f6

It blocks this PR but I think we should discuss it in the scope of spec RFC.

src/utilities/introspectionQuery.js
OlegIlyenko: It's a good point, I added an option for it https://github.com/graphql/graphql-js/pull/1541/commits/3b5bf03609b2e92184f511fe7230365628a53182 I also made if disabled by defult in order to maintain backwards compatibility.

OlegIlyenko: Also this one: https://github.com/graphql/graphql-js/pull/1541/commits/3a7fafac79e35384c0c40f888c3abc0a6d9d9679 I forgot to update the introspection flow type

src/utilities/introspectionQuery.js
OlegIlyenko: Could you please describe in more detail what kind of information you would like to add? I'm not 100% sure I understood your comment?

IvanGoncharov: @OlegIlyenko I mean if merge this PR that will expose `repeatable` through introspection and it would be significantly harder to rename it later.
As I wrote in my comment to the spec RFC I think this field should be named `isRepeatable` for consistency with `isDeprecated`.

Everything else in this PR looks good so would be great to discuss this last point but since this PR is just an implementation of spec RFC I think we should discuss it in spec PR.

OlegIlyenko: I see, thanks for the clarification. Sure thing, I will open a discussion on this topic in the RFC PR.

IvanGoncharov: @OlegIlyenko I feel very stupid because in the first comment of this thread I linked review comment I forget to submit (review comments UI is very weird).
I already submitted it in.

IvanGoncharov: @OlegIlyenko Now it visible: https://github.com/facebook/graphql/pull/472/files/1ba4e196aa710a10eceaa346d02a970dfd0b2d3a#diff-35fb422eded6e8f4df638c0d159008f6

OlegIlyenko: 👍 i see, thanks! now it's clear, i was confused about wrong link

src/validation/rules/UniqueDirectivesPerLocation.js
OlegIlyenko: good idea, done https://github.com/graphql/graphql-js/pull/1541/commits/ae9e7f98f73a069096523575de6869d85707c09f

src/utilities/introspectionQuery.js
OlegIlyenko: I did the renaming https://github.com/graphql/graphql-js/pull/1541/commits/56fee7d28b68d44665007e175129839af11ae8d2 I will adjust the spec text accordingly

IvanGoncharov:
@OlegIlyenko I reviewed it one more time and found only one small issue.
Rule description should be adjusted here:
https://github.com/graphql/graphql-js/blob/56fee7d28b68d44665007e175129839af11ae8d2/src/validation/rules/UniqueDirectivesPerLocation.js#L27-L32

OlegIlyenko:
@IvanGoncharov I updated the comment https://github.com/graphql/graphql-js/pull/1541/commits/0dff21fb3508191cc7efbaeff1d2e72135dd9809 Thanks a lot for a fast and careful review! :bowing_man:

src/language/ast.js
```diff
@@ -508,6 +508,7 @@ export type DirectiveDefinitionNode = {
   +description?: StringValueNode,
   +name: NameNode,
   +arguments?: $ReadOnlyArray<InputValueDefinitionNode>,
+  +isRepeatable: boolean,
```
mjmahone: I wonder if this should be:
```
+repeatable?: boolean,
```
If someone never uses repeatable directives, this adds no extra weight to the node.

I also don't think we need the `is` prefix: the fact that it's a boolean is enough information, and the question `repeatable? false` is close-to-valid English as-is (just as understandable as `is repeatable? false`).

src/validation/rules/UniqueDirectivesPerLocation.js
```diff
@@ -22,12 +27,29 @@ export function duplicateDirectiveMessage(directiveName: string): string {
 /**
  * Unique directive names per location
  *
- * A GraphQL document is only valid if all directives at a given location
+ * A GraphQL document is only valid if directives, that are not marked as `isRepeatable`, at a given location
  * are uniquely named.
  */
 export function UniqueDirectivesPerLocation(
-  context: ASTValidationContext,
+  context: ValidationContext | SDLValidationContext,
```
mjmahone: This is technically a breaking change. It's probably OK as I doubt anyone is using this validation explicitly yet, but definitely need to note that in our release notes.

But I think there should be a way, today, to re-order our validation contexts to provide a Schema SDL if possible. I will probably end up forking all of these validation rules to run off an AST-only schema object, because building (and extending) the current GraphQLSchema object is too slow. I won't block this on not using the GraphQLSchema, but I'd like for us to move in that direction.

src/utilities/introspectionQuery.js
```diff
@@ -13,10 +13,15 @@ export type IntrospectionOptions = {|
   // Whether to include descriptions in the introspection result.
   // Default: true
   descriptions: boolean,
+  // Whether to include `isRepeatable` flag on directives.
+  // Default: false
```
mjmahone: Why would this default to false?

src/language/ast.js
OlegIlyenko: The `is` prefix primarily added for consistency with `isDeprecated`, as @IvanGoncharov pointed out https://github.com/facebook/graphql/pull/472/files/1ba4e196aa710a10eceaa346d02a970dfd0b2d3a#diff-35fb422eded6e8f4df638c0d159008f6

I think I would prefer it keep it mandatory. This makes it easier to work with and i feel that the performance overhead is negligible. But I can also make it optional.

WDYT?

src/utilities/introspectionQuery.js
OlegIlyenko: For example, if it would be `true`, then after an update without any adjustments third-party libraries (like GraphiQL) would start to send queries with `isRepeatable` flag to the servers that potentially not updated yet. So default is primarily defined like this to maintain backwards compatibility.

Would you like to change the default and require all downstream libraries to explicitly set this flag? (at least at the initial stage where servers are not yet updated)

src/language/ast.js
IvanGoncharov: @OlegIlyenko In the context of AST, I'm more in favor of @mjmahone proposal since we already have `block` with similar prototype:
https://github.com/graphql/graphql-js/blob/0dff21fb3508191cc7efbaeff1d2e72135dd9809/src/language/ast.js#L320

Plus it looks like JS AST doesn't use `is` prefix, e.g. `async`:
https://github.com/estree/estree/blob/df2290b23a652e1ec354b6775459a3b42e22f108/es2017.md#function

But outside of AST I strongly believe it should be named `isRepeatable` for the consistency with `isDeprecated`.

> This makes it easier to work with and i feel that the performance overhead is negligible.

I'm with you on that point especially since JS parser always preserve boolean flags and they typically work with way bigger files than GraphQL.
Also defacto (in our parser) we always provide values for all keys so why should Flow typings mark that as optional.

That said I think it's outside of scope of this PR and should be disscussed for  AST as a whole and not for individual flags.
So for now I think it should be optional.

src/validation/rules/UniqueDirectivesPerLocation.js
IvanGoncharov: @mjmahone It's not a breaking change since `ASTValidationContext` is not a part of public API. As we agree during `14.0.0` SDL validation is an experimental feature so I didn't expose this and other new class as public API.
ATM only `ValidationContext` is exposed so if library client use only public API he shouldn't be affected by this change in any way.

src/utilities/introspectionQuery.js
IvanGoncharov: @mjmahone Do you remember we discussed it on the last WG in the scope of my 
"schema description" RFC. And the consensus was to keep it backward compatible for now but design two stage introspection as a long-term solution.

src/language/ast.js
OlegIlyenko: I think both alternatives should be fine. I also considered keep `repeatable` name on AST nodes, but then renamed it with `is` prefix across the whole codebase for the sake of the naming consistency.

But I think it should totally fine to have `repeatable` field on AST node and then use `isRepeatable` everywhere else. @mjmahone What do you think?

OlegIlyenko: I renamed `isRepeatable` back to `repeatable` in all places except introspection: https://github.com/graphql/graphql-js/pull/1541/commits/89f5780fa4d5f55ab4028d2136ba7fa14b63bd2c

a-type:
Any movement on this? While it's not a blocker, it would improve syntax for a project I'm experimenting with.

IvanGoncharov:
@a-type It's blocked by introspection changes that I plan to do this week:
https://docs.google.com/presentation/d/1ISS5QhF6D2ncD8ZGChoCv1LnOWj9jryZrJ2dG11i4Mc/edit?usp=sharing

IvanGoncharov:
@a-type I decided that we shouldn't fully block repeatable directives because of required introspection changes so I merged everything except blocked parts as #1965.

spawnia:
@IvanGoncharov anything left in this PR that you have not merged elsewhere?

IvanGoncharov:
@spawnia Thanks for the ping 👍 
Everything is already implemented in `15.0.0` release.
Closing.

PR #1544, opened 2018-10-03
Configure majority of undecided ESLint rules

IvanGoncharov:


.eslintrc.yml
```diff
@@ -71,12 +71,16 @@ rules:
   flowtype/use-flow-type: error
   flowtype/valid-syntax: off
 
+  ##################################################
+  # ESLint builtin rules list based on `v5.6.0`
+  ##################################################
```
IvanGoncharov: @mjmahone Idea here is to add new rules every time we update ESLint in `package.json`.

src/execution/__tests__/sync-test.js
```diff
@@ -28,8 +28,8 @@ describe('Execute: synchronously when possible', () => {
         },
         asyncField: {
           type: GraphQLString,
-          async resolve(rootValue) {
-            return await rootValue;
+          resolve(rootValue) {
+            return Promise.resolve(rootValue);
```
IvanGoncharov: IMO it was very weird construct and explicit promise is a way more readable.

src/language/__tests__/lexer-test.js
```diff
@@ -129,7 +131,7 @@ describe('Lexer', () => {
   it('updates line numbers in error for file context', () => {
     let caughtError;
     try {
-      const str = '' + '\n' + '\n' + '     ?\n' + '\n';
+      const str = ['', '', '     ?', ''].join('\n');
```
IvanGoncharov: It was triggering `no-useless-concat` rule.
@mjmahone If you think it became less readable I can revert it back and just add ESLint comment that will disable `no-useless-concat` rule for this line.

.eslintrc.yml
mjmahone: That makes sense to me. Thanks for adding the comment.

src/execution/__tests__/sync-test.js
mjmahone: I disagree in general, but it's not a big enough deal for me to block this diff.

Personally I much prefer async/await, but I don't love JavaScript, so my opinions are suspect.

src/language/__tests__/lexer-test.js
mjmahone: No I think this is as readable. Both ways are mostly nonsense.

src/execution/__tests__/sync-test.js
IvanGoncharov: @mjmahone I reverted it back and just removed `await`, see 8cefd09bf71911648eb7def1b7d2f2feaa3dfb1f

IvanGoncharov: I reverted it back and just remove useless `await` so now it looks like:
```js
          async resolve(rootValue) {
            return rootValue;
          }
```

PR #1545, opened 2018-10-04
Parser: Add 'skipKeyword' function

IvanGoncharov:
based on 'expectOptionalKeyword' from #1541

### DIFF
--- a/src/language/parser.js
+++ b/src/language/parser.js
@@ -488,7 +488,9 @@ function parseFragment(
 ): FragmentSpreadNode | InlineFragmentNode {
   const start = lexer.token;
   expect(lexer, TokenKind.SPREAD);
-  if (peek(lexer, TokenKind.NAME) && lexer.token.value !== 'on') {
+
+  const hasTypeCondition = skipKeyword(lexer, 'on');
+  if (!hasTypeCondition && peek(lexer, TokenKind.NAME)) {
     return {
       kind: Kind.FRAGMENT_SPREAD,
       name: parseFragmentName(lexer),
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@OlegIlyenko I had a few review comment about `expectOptionalKeyword` from #1541 but didn't want increase the scope of this PR so I extracted it out.
Can you please take a look?

src/language/parser.js
```diff
@@ -1512,21 +1508,31 @@ function expect(lexer: Lexer<*>, kind: TokenKindEnum): Token {
...
 
 /**
- * If the next token is a keyword with the given value, return that token after
- * advancing the lexer. Otherwise, do not change the parser state and return
- * false.
+ * If the next token is a keyword with the given value, return true after advancing
+ * the lexer. Otherwise, do not change the parser state and return false.
  */
-function expectKeyword(lexer: Lexer<*>, value: string): Token {
+function skipKeyword(lexer: Lexer<*>, value: string): boolean {
```
IvanGoncharov: Since we already have skip/expect pair with matching functionality I think we need to have skipKeyword/expectKeyword.

OlegIlyenko:
I found the terminology around "skip" a bit confusing. "skip" by itself does not imply that the token is optional. So when i saw the usage of `skip` function it was a bit misleading.

I personally believe that this part of code might benefit from `skip` + `skipOptional` + keyword and "expect" alternatives. Purely based on the names, I understood "expect" as something to be used in cases where interesting in the token value and "skip" for cases when I just want to jump over the token, but I'm not interested in its value.

Just my 2 cents.

PR #1547, opened 2018-10-05
Parser: Better names for parser util functions

IvanGoncharov:
Based @OlegIlyenko suggestion:
https://github.com/graphql/graphql-js/pull/1545#issuecomment-427196059

### DIFF
--- a/src/language/parser.js
+++ b/src/language/parser.js
@@ -150,9 +150,9 @@ export function parseValue(
 ): ValueNode {
   const sourceObj = typeof source === 'string' ? new Source(source) : source;
   const lexer = createLexer(sourceObj, options || {});
-  expect(lexer, TokenKind.SOF);
+  expectToken(lexer, TokenKind.SOF);
   const value = parseValueLiteral(lexer, false);
-  expect(lexer, TokenKind.EOF);
+  expectToken(lexer, TokenKind.EOF);
   return value;
 }
 
@@ -172,9 +172,9 @@ export function parseType(
 ): TypeNode {
   const sourceObj = typeof source === 'string' ? new Source(source) : source;
   const lexer = createLexer(sourceObj, options || {});
-  expect(lexer, TokenKind.SOF);
+  expectToken(lexer, TokenKind.SOF);
   const type = parseTypeReference(lexer);
-  expect(lexer, TokenKind.EOF);
+  expectToken(lexer, TokenKind.EOF);
   return type;
 }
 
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
> I personally believe that this part of code might benefit from skip + skipOptional + keyword and "expect" alternatives. Purely based on the names, I understood "expect" as something to be used in cases where I'm interested in the token value and "skip" for cases when I just want to jump over the token, but I'm not interested in its value.

@OlegIlyenko I decide to extract it into a separate PR to make review more effective. Can you please review this PR and comment if new names match your suggestion.

OlegIlyenko:
@IvanGoncharov Thanks a lot for considering my feedback and working this PR!

I think it reads much better now :+1: I like how easy it is to identify which tokens are optional and which are not. I also like how occurrences of `skip` and `expect` are unified under the same term. 

Just an idea to play around with: I feel that term "expect" (or even more explicit variations `expectTerm`/`expectKeyword`) would better communicate the intention. But I'm not sure whether it's just me. So I don't want to emphasize it too much. In a current form it is already a good improvement, I think.

IvanGoncharov:
> Just an idea to play around with: I feel that term "expect" (or even more explicit variations expectTerm/expectKeyword) would better communicate the intention. But I'm not sure whether it's just me. So I don't want to emphasize it too much. In a current form it is already a good improvement, I think.

@OlegIlyenko Not sure I fully understand the proposed naming strategy. 
How about we discuss it in term of concrete functions:
* `expect` - if match token `kind` then `advance` and return `token` else `throw`
* `skip` - if match token `kind` then `advance` else `throw`
* `skipOptional` - if match token `kind` then `advance` and return `true` else return `false`
* `skipKeyword` - if match `keyword` then `advance` else `throw`
* `skipOptionalKeyword` - if match `keyword` then `advance` and return `true` else return `false`


Can you please copy this list and add/delete funtions or rename them if needed?

OlegIlyenko:
I mean something like this:

* `expectToken` - if match token kind then advance and return `token` else throw
* `expectOptionalToken` - if match token kind then advance and return `token` else return `null`/`undefined`
* `expectKeyword - if match keyword then advance and return `token` else throw
* `expectOptionalKeyword - if match keyword then advance and return `token` else return `null`/`undefined`

So the idea is to boil it down to only `expect*` functions which can be used in all scenarios. If return value of these functions is not needed, then it can be safely ignored.

mjmahone:
So I think there's still a little confusion about the proposed terms, especially around "skip" which I usually read as closer to "I think this might be here but it means nothing if it isn't". The original naming scheme was still confusing because of this, I think.

I really like @OlegIlyenko's naming scheme on this, though.

IvanGoncharov:
I updated PR according @OlegIlyenko last comment.

PR #1548, opened 2018-10-06
Fix error when printing a `FieldDefinition` AST node without `arguments` field

awfulaxolotl:
I encountered a runtime type error while printing some hand-constructed `FieldDefinition` AST nodes which conform to the type spec.

### Problem

Type spec - declaring it's optional:

https://github.com/graphql/graphql-js/blob/26c9874107e65f19576aae0a32638287820f68aa/src/language/ast.js#L444

Printing implementation - requiring it:

https://github.com/graphql/graphql-js/blob/787422956c9554d12d063a41fe35705335ec6290/src/language/printer.js#L139

### Solution

This PR fixes this discrepancy in the implementation with a default value.

https://github.com/awfulaxolotl/graphql-js/blob/df56ba3eadbfbc282ae6262927dfaeb162662158/src/language/printer.js#L139

### DIFF
--- a/src/language/printer.js
+++ b/src/language/printer.js
@@ -134,7 +134,7 @@ const printDocASTReducer = {
   ),
 
   FieldDefinition: addDescription(
-    ({ name, arguments: args, type, directives }) =>
+    ({ name, arguments: args = [], type, directives }) =>
       name +
       (args.every(arg => arg.indexOf('\n') === -1)
         ? wrap('(', join(args, ', '), ')')
--- a/src/language/printer.js
+++ b/src/language/printer.js
@@ -134,11 +134,11 @@ const printDocASTReducer = {
   ),
 
   FieldDefinition: addDescription(
-    ({ name, arguments: args = [], type, directives }) =>
+    ({ name, arguments: args, type, directives }) =>
       name +
-      (args.every(arg => arg.indexOf('\n') === -1)
-        ? wrap('(', join(args, ', '), ')')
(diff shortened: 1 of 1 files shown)


src/language/printer.js
```diff
@@ -136,7 +136,7 @@ const printDocASTReducer = {
   FieldDefinition: addDescription(
     ({ name, arguments: args, type, directives }) =>
```
IvanGoncharov: @awfulaxolotl you can use default value inside destructuring:
 `({ name: arguments: args = [], directives})`

awfulaxolotl: Hey @IvanGoncharov, thanks for the reminder!

I've pushed a revised version with the cleaner syntax.

IvanGoncharov:
@awfulaxolotl Sorry for the delay.
I reviewed our printer implementation and found another place with an exact same problem.
So I extracted code in question into `hasMultilineItem` function and reuse it in both places.
I pushed my changes on top of your PR and it would be great if you have time to review it.

@mjmahone Can you please review?

IvanGoncharov:
@mjmahone General issue here is that `printer.js` is not covered by Flow but I will try to address this in the separate PR.

awfulaxolotl:
@IvanGoncharov Nice catch. Your extended fix looks good to me! Looking forward to `printer.js` being covered with Flow.

IvanGoncharov:
@mjmahone Can you please review?

PR #1550, opened 2018-10-12
Switch schema parser tests to use dedent

IvanGoncharov:


src/language/__tests__/schema-parser-test.js
```diff
@@ -74,99 +75,102 @@ function inputValueNode(name, type, defaultValue, loc) {
...
+      }
+    `);
 
     expect(toJSONDeep(doc)).to.deep.equal({
       kind: 'Document',
       definitions: [
         {
           kind: 'ObjectTypeDefinition',
-          name: nameNode('Hello', { start: 6, end: 11 }),
+          name: nameNode('Hello', { start: 5, end: 10 }),
```
IvanGoncharov: All offsets are shifted by 1 because template literal contained initial new line after \` but it's striped by `dedent`.

PR #1552, opened 2018-10-14
RFC: Define custom scalars in terms of built-in scalars.

IvanGoncharov:
As @leebyron suggested during last WG I'm adopting #914 and rebasing it on top of current `master`.

Here is the original description:

This proposes an additive change which allows custom scalars to be defined in terms of the built-in scalars. The motivation is for client-side code generators to understand how to map between the GraphQL type system and a native type system. As an example, a `URL` custom type may be defined in terms of the built-in scalar `String`. It could define additional serialization and parsing logic, however client tools can know to treat `URL` values as `String`. Presently, we do this by defining these mappings manually on the client, which is difficult to scale, or by giving up and making no assumptions of how the custom types serialize.

Another real use case of giving client tools this information is GraphiQL: this change will allow GraphiQL to show more useful errors when a literal of an incorrect kind is provided to a custom scalar. Currently GraphiQL simply accepts all values.

To accomplish this, this proposes adding the following:

* A new property when defining `GraphQLScalarType` (`ofType`) which asserts that only built-in scalar types are provided.

* A second type coercion to guarantee to a client that the serialized values match the `ofType`.

* Delegating the `parseLiteral` and `parseValue` functions to those in `ofType` (this enables downstream validation / GraphiQL features)

* Exposing `ofType` in the introspection system, and consuming that introspection in `buildClientSchema`.

* Adding optional syntax to the SDL, and consuming that in `buildASTSchema` and `extendSchema` as well as in `schemaPrinter`.

* Adding a case to `findBreakingChanges` which looks for a scalar's ofType changing.

### DIFF
--- a/src/language/__tests__/schema-kitchen-sink.graphql
+++ b/src/language/__tests__/schema-kitchen-sink.graphql
@@ -75,6 +75,8 @@ scalar CustomScalar
 
 scalar AnnotatedScalar @onScalar
 
+scalar StringEncodedCustomScalar as String
+
 extend scalar CustomScalar @onScalar
 
 enum Site {
--- a/src/language/__tests__/schema-parser-test.js
+++ b/src/language/__tests__/schema-parser-test.js
@@ -756,6 +756,7 @@ type Hello {
           kind: 'ScalarTypeDefinition',
           name: nameNode('Hello', { start: 7, end: 12 }),
           description: undefined,
+          type: undefined,
           directives: [],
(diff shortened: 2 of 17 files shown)


yaacovCR:
Closing this PR in the name of housekeeping. It may be reopened as necessary.

As this PR represents a spec change, it would make sense to reopen with an accompanying spec PR. (There may be one that I have not found!)

PR #1554, opened 2018-10-15
Improve error message for missing required argument in field or directive

everdimension:


src/validation/rules/ProvidedRequiredArguments.js
```diff
@@ -27,7 +27,7 @@ export function missingFieldArgMessage(
 ): string {
   return (
     `Field "${fieldName}" argument "${argName}" of type ` +
-    `"${type}" is required but not provided.`
+    `"${type}" is required, but not provided.`
```
IvanGoncharov: @everdimension Thanks for PR 👍 
I'm not a native speaker myself but `not provided` looks like a dependent sentence for me so it looks like this sentence doesn't need comma based on this article:
https://www.grammarly.com/blog/comma-before-but/

everdimension: Neither am I, but this message seems a bit off for some reason.

After further investigation, it seems that the comma before "but" is not obligatory in the english language when the subject is not repeated in the second clause.

But now that I've had a second look, perhaps the problem is the omitted verb.
So the correct message should sound something like this:

<code>"${type}" is required but <b>was</b> not provided.</code>
or
<code>"${type}" is required, but <b>it was</b> not provided.</code> (with comma, because "it" is repeated in the second clause)

But again, perhaps omitting the auxiliary verb is permissible for system messages like this one.

---

Anyway, this is definitely a nitpick. I can update my PR and add `was` instead of the comma if you agree that this is an improvement. Otherwise feel free to close this

IvanGoncharov: @everdimension I like `was` variant 👍 

Also would be great to also get it reviewed by a native speaker.

mjmahone: The comma should not matter. I think it's a British-ism vs American-ism, and so I don't think it makes a big difference. Despite being an American-English-speaker, commas or not read mostly the same in this case.

I agree that the `was` variant is nicest.

IvanGoncharov:
@everdimension I changed it to `is required, but it was not provided.` and merged 🎉 
Thanks for PR and patience 👍

PR #1557, opened 2018-10-23
Pretty format variable definitions for both operations and fragments.

adek05:
When there is more than a couple argument values, pretty formatting
isn't doing a good job.

Instead I would like to take the following approach:

When there is one argument, print as is, meaning:

`
  query Q ($arg: Type = Default) {
`
When there is more than one do the following:

```
  query Q (
    $arg1: Type = Default,
    $arg2
  ) {
```

An alternative would be to have closing bracket be in the same line
as final argument. I am fine with either, as changes which introduce
another argument on last position always touch two lines - trailing
comma will have to be added.

Edited tests to make sure everything works as expected.

### DIFF
--- a/src/language/__tests__/printer-test.js
+++ b/src/language/__tests__/printer-test.js
@@ -157,11 +157,22 @@ describe('Printer: Query document', () => {
         fragment Foo($a: ComplexType, $b: Boolean = false) on TestType {
           id
         }
+
+        fragment Simple($a: Boolean = true) on TestType {
+          id
+        }
       `,
       { experimentalFragmentVariables: true },
     );
     expect(print(fragmentWithVariable)).to.equal(dedent`
-      fragment Foo($a: ComplexType, $b: Boolean = false) on TestType {
+      fragment Foo(
+        $a: ComplexType,
+        $b: Boolean = false
+      ) on TestType {
+        id
+      }
+
+      fragment Simple($a: Boolean = true) on TestType {
         id
       }
     `);
@@ -174,7 +185,10 @@ describe('Printer: Query document', () => {
 
     expect(printed).to.equal(
       dedent(String.raw`
-      query queryName($foo: ComplexType, $site: Site = MOBILE) @onQuery {
+      query queryName(
+        $foo: ComplexType,
+        $site: Site = MOBILE
+      ) @onQuery {
         whoever123is: node(id: [123, 456]) {
           id
           ... on User @onInlineFragment {
--- a/src/language/__tests__/printer-test.js
+++ b/src/language/__tests__/printer-test.js
@@ -60,7 +60,9 @@ describe('Printer: Query document', () => {
       'query ($foo: TestType) @testDirective { id, name }',
     );
     expect(print(queryAstWithArtifacts)).to.equal(dedent`
-      query ($foo: TestType) @testDirective {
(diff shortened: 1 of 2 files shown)


src/language/printer.js
```diff
@@ -275,3 +275,17 @@ function printBlockString(value, isDescription) {
...
+ * than one argument. Otherwise print inline.
+ */
+function printVariableDefinitions(args) {
+  if (!args || args.length === 0) {
+    return '';
+  }
+  if (args.length === 1) {
+    return '(' + args[0] + ')';
+  }
+  return '(\n' + indent(join(args, ',\n')) + '\n)';
```
mjmahone: We were only using the comma due to having everything on one line. A newline is a perfectly acceptable variable definition separator, just as it is for fields inside a document.

IvanGoncharov:
@adek05 @mjmahone Single line approach works for simple cases but not for something like this:
```
query Q ($arg: Type = Default @someQueryVariableDirective(withArg: "test")) {
```
VS
```
query Q (
  $arg: Type = Default @someQueryVariableDirective(withArg: "test")
) {
```

I think single argument per line is safer option for now but in long run we need to implement something simular to prettier algorithm.

mjmahone:
@IvanGoncharov I wasn't sure whether single variable should be special-cased, but your example convinced me it should not be.

src/language/__tests__/printer-test.js
```diff
@@ -157,11 +157,22 @@ describe('Printer: Query document', () => {
...
+        fragment Simple($a: Boolean = true) on TestType {
+          id
+        }
       `,
       { experimentalFragmentVariables: true },
     );
     expect(print(fragmentWithVariable)).to.equal(dedent`
-      fragment Foo($a: ComplexType, $b: Boolean = false) on TestType {
+      fragment Foo(
+        $a: ComplexType,
```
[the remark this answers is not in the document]
adek05: I can fix the test, that is fine.

src/language/printer.js
adek05: TIL!

adek05:
Changes incorporated, let me know if there is anything else you would like to see or good to go.

src/language/__tests__/printer-test.js
```diff
@@ -60,7 +60,9 @@ describe('Printer: Query document', () => {
       'query ($foo: TestType) @testDirective { id, name }',
     );
     expect(print(queryAstWithArtifacts)).to.equal(dedent`
-      query ($foo: TestType) @testDirective {
+      query (
+        $foo: TestType
+      ) @testDirective {
```
leebyron: This feels worse to me. Why wrap like this?

src/language/printer.js
```diff
@@ -283,3 +283,13 @@ function printBlockString(value, isDescription) {
...
 }
+
+/**
+ * Prints variables one per line.
+ */
+function printVariableDefinitions(args) {
+  if (!args || args.length === 0) {
+    return '';
+  }
+  return '(\n' + indent(join(args, '\n')) + '\n)';
```
leebyron: Happy to see this function factored out to avoid duplication, but I think the additional new lines aren't helping legibility

IvanGoncharov:
Extracted as #3230

PR #1562, opened 2018-10-28
Prevent ReferenceError when running graphql in the browser

bennypowers:
Not every user wants to or should use webpack to bundle graphql and related libs.
Forcing web developers to cludge a global reference to an imaginary `process` in a render-blocking script is not an optimal solution, when libraries can just check it themselves.

### DIFF
--- a/src/jsutils/instanceOf.js
+++ b/src/jsutils/instanceOf.js
@@ -16,9 +16,18 @@ declare function instanceOf(
   constructor: mixed,
 ): boolean %checks(value instanceof constructor);
 
+/* global window */
+// window is undefined on node, let's not trip the linter.
+let glbl;
+try {
+  glbl = global;
+} catch (e) {
+  glbl = window;
+}
+
 // See: https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production
 // See: https://webpack.js.org/guides/production/
-export default (process.env.NODE_ENV === 'production'
+export default (glbl.process && glbl.process.env.NODE_ENV === 'production'
   ? // eslint-disable-next-line no-shadow
     function instanceOf(value: mixed, constructor: mixed) {
       return value instanceof constructor;
--- a/src/jsutils/instanceOf.js
+++ b/src/jsutils/instanceOf.js
@@ -16,18 +16,41 @@ declare function instanceOf(
   constructor: mixed,
 ): boolean %checks(value instanceof constructor);
 
-/* global window */
-// window is undefined on node, let's not trip the linter.
-let glbl;
-try {
-  glbl = global;
-} catch (e) {
-  glbl = window;
+function getGlobal() {
+  /* global window self */
+  if (
+    typeof self !== 'undefined' &&
+    self.self === self &&
+    self.Array === Array &&
+    self.setInterval === setInterval
+  ) {
+    return self;
+  }
+  if (
+    typeof window !== 'undefined' &&
+    window.window === window &&
+    window.Array === Array &&
+    window.setInterval === setInterval
+  ) {
+    return window;
+  }
+  if (
+    typeof global !== 'undefined' &&
+    global.global === global &&
+    global.Array === Array &&
+    global.setInterval === setInterval
+  ) {
+    return global;
+  }
+  throw new Error('Cannot find the global object');
 }
+
+const production =
+  getGlobal().process && getGlobal().process.env.NODE_ENV === 'production';
 
 // See: https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production
 // See: https://webpack.js.org/guides/production/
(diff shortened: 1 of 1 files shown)


Neitsch:
Will `glbl.process` be undefined on the client, so that it always evaluates to false? I understand where this is coming from. Let's make sure that it's a safe change :)

bennypowers:
@Neitsch would you prefer to also check for `glbl.process.env` and `global.process.env.NODE_ENV` ?

Neitsch:
Oops, sorry, I was thinking about something wrong. Yeah, this looks like a good idea

bennypowers:
Yes. I want to run this in the browser, and I don't want to since this problem with added tooling.

bennypowers:
:wave: Any reasons not to merge this?

Here's an example use case:

IvanGoncharov:
@bennypowers Sorry for delay and thanks for the ping.
> Here's an example use case:

Example code is missing in your comment for some reason.

> Yes. I want to run this in the browser, and I don't want to solve this problem with added tooling.

Just to better understand your use case how you handle `require` inside graphql JS files without webpack or other bundlers?

bennypowers:
> Example code is missing in your comment for some reason.

*whoopsie!* sleep is good, I promise.

sorry for the long story here but it's genuinely convoluted.

https://lit-apollo-subscriptions.herokuapp.com uses [`@apollo-elements/lit-apollo`](https://github.com/apollo-elements/apollo-elements) to define and render components.

`lit-apollo` depends on `apollo-client`, which has a peer dependency for `graphql`

so when `instanceOf.mjs` loads up in the browser, it hits that ternary and crashes the app.

I handle `require` with `rollup-plugin-commonjs`, but prefer to use es modules for treeshaking and bundle size reasons.

So, I could solve this by importing builtins from `rollup-plugin-globals` or something, but I think that's a poor choice for many reasons:
1. it pointlessly inflates the bundle size
1. it doesn't add value to the app in any way asides from preventing this error
1. it strikes me as odd, and way overkill to apply that level of tooling when a simple check would suffice
1. even if some preferred tool would come along and solve the above, why limit users to a specific toolchain? surely the goal should be to reduce the amount of *required* tooling

So that leaves me with the solution here:

https://gitlab.com/bennyp/demo-lit-apollo-subscriptions/blob/master/src/index.html#L32

Which makes me a little sad. Why assume node-everywhere when, for the price of 44 characters of straightforward, good-old-fashioned runtime javascript, you can cover node and web cases.

I hope that makes sense.

IvanGoncharov:
@bennypowers Thanks for detailed explanation now it totally makes sense.
One problem though is that we spend a few iterations to make this code work properly with different bundlers and other tools.
AFAIK many such tools rely on static code analysis to figure out code patterns so it's hard to predict if we break something or not. 

The ideal solution would be to find some popular library that faces the same problems and copy-paste code from here. This way we can ensure that this code is battle tested with the majority of setups/stacks without putting `graphql-js` community through a new cycle of releases that break stuff that worked previously.

For example, I did a quick check for `react-dom` and it looks like they have a separate build for NPM that have an exact same check as we have ATM:
https://github.com/facebook/react/blob/baff5cc2f69d30589a5dc65b089e47765437294b/packages/react-dom/npm/index.js#L31
I never used React without webpack so I not sure how you would use React in such case.

So if you have the time it would be great if you find out how other libraries solve this issue and save us from reinventing the wheel.

bennypowers:
in the node context this will have the exact same behavior as before

It will hit the `try` block, then evaluate the expression
```js
glbl = global;
```
when it hits the `if` statement, `glbl.process && glbl.process.env.NODE_ENV` will evaluate the exact same as `process.env`

[proof](https://repl.it/@bennypowers/node-global-proof-for-graphql-pr)

The difference here is that it won't cause a ReferenceError on the web.

[proof of error](https://stackblitz.com/edit/js-e1jkcy) (open console)
[proof of fix](https://stackblitz.com/edit/js-fmshh5?file=index.html)

bennypowers:
But here are some more comprehensive thoughts on the matter
https://github.com/tc39/proposal-global#rationale

bennypowers:
I pushed a commit which uses the expanded `getGlobal` from `es6-shim` 
https://github.com/paulmillr/es6-shim/commit/2367e0953edd01ae9a5628e1f47cf14b0377a7d6#commitcomment-27989605

In the future, I suppose `globalThis` would do the trick.

Neitsch:
As a hypothetical: What is the advantage of your solution over just doing `process && process.env && process.env.NODE_ENV == "production"? To use production mode you can set the global variable appropriately.

On a side note, I would strongly advise declaring `process.env.NODE_ENV = 'production'`, since this can reduce bundle size.

bennypowers:
I originally did that, but a more comprehensive solution was sought

IvanGoncharov:
Superseded by #2409

PR #1566, opened 2018-11-06
Add Extra Tests for findArgChanges

helloqiu:
Add test for running `findArgChanges` with schema which field's default value is an array.
This PR also reproduces a bug of `findArgChanges`.


mjmahone:
Thank you for finding this bug and producing a test, @helloqiu ! This is a problem for any non-scalar input default values, and I should get a commit with a fix for this up shortly.

PR #1570, opened 2018-11-13
Upgrade to flow 0.86.0

kassens:


KevinGrandon:
Thank you for upgrading! 👏 Any chance we could ask for a version bump or pre-release so we can upgrade dependent packages to the latest flow as well?

PR #1574, opened 2018-11-16
Don't check for specified scalars in removed types

mfix22:
This PR includes a broken test (that I will update if approved). 

Not sure if you want to go this direction (please feel free to close if not), but having a specified scalar type removed from a schema is not an issue on it's own, and the breakage is already typically covered by a type changing on a field.

For example, changing a schema from 
```graphql
type Query {
  version: String
}
```
to 
```graphql
type Query {
  version: Int
}
```

Will include `"String was removed."` _and_ `"Query.version changed type from String to Int."` I think the first message is benign, where the second is not 🤷‍♂️ 

I can understand the reasoning for keeping it as is, since a lot of these checks could get super-complicated, but this change seems rather useful 🙂

### DIFF
--- a/src/utilities/findBreakingChanges.js
+++ b/src/utilities/findBreakingChanges.js
@@ -35,6 +35,8 @@ import keyMap from '../jsutils/keyMap';
 import type { ObjMap } from '../jsutils/ObjMap';
 import type { DirectiveLocationEnum } from '../language/directiveLocation';
 
+import { isSpecifiedScalarType } from '../type/scalars';
+
 export const BreakingChangeType = {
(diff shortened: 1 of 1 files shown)


mjmahone:
I think we do actually need the scalar removal messaging. If someone is depending on some sort of String value existing via the schema introspection query, making this change changes the result of introspection to no longer have that scalar type, which would be equivalent to deleting a user-defined scalar type from the schema. I agree in some cases this may not be a breaking change, but I know of other situations (like when you remove a type that's used as an input value) where this will cause formerly valid queries to start failing.

I'd rather have us telling people all the ways they could be causing problems than limiting it to a single reason.

mfix22:
Great 👍

PR #1593, opened 2018-12-01
Non-Scalar default dangerous change: Fix for #1566

mjmahone:
@helloqiu identified in #1566 that our dangerous changes check was incorrectly saying any non-scalar default value was a dangerous change, even if the default value hadn't actually changed. This is because we need to check the default value change for deep equality, rather than just whether the objects are identical.

### DIFF
--- a/src/utilities/findBreakingChanges.js
+++ b/src/utilities/findBreakingChanges.js
@@ -232,8 +232,10 @@ export function findArgChanges(
                 `${oldArgDef.type.toString()} to ${newArgDef.type.toString()}`,
             });
           } else if (
-            oldArgDef.defaultValue !== undefined &&
-            oldArgDef.defaultValue !== newArgDef.defaultValue
+            !isChangeSafeForDefaultValue(
+              oldArgDef.defaultValue,
+              newArgDef.defaultValue,
+            )
           ) {
             dangerousChanges.push({
               type: DangerousChangeType.ARG_DEFAULT_VALUE_CHANGE,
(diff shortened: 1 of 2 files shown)


src/utilities/findBreakingChanges.js
```diff
@@ -496,6 +498,41 @@ function isChangeSafeForInputObjectFieldOrFieldArg(
...
+    if (Array.isArray(oldDefaultValue) !== Array.isArray(newDefaultValue)) {
+      return false;
+    }
+
+    for (const key in oldDefaultValue) {
+      if (
+        !isChangeSafeForDefaultValue(oldDefaultValue[key], newDefaultValue[key])
+      ) {
+        return false;
+      }
```
IvanGoncharov: @mjmahone According to current implementation adding properties to object is safe.
For example `isChangeSafeForDefaultValue({}, {a: 1}) => true` or `isChangeSafeForDefaultValue({a: undefined}, {a: 1}) => true`.

IvanGoncharov:
@mjmahone I think it's better to add `isDeepEqual` to `jsutils` folder.
We can use https://github.com/substack/node-deep-equal as basis and remove support for all weird types (e.g. `Buffer`, `arguments`, etc.)

src/utilities/findBreakingChanges.js
Neitsch: I second this. Can we add a couple more testcases to make sure that check works correctly?

mjmahone: That should be correct: adding a previously undefined value is OK, but removing a previously undefined value is not (recursively).

mjmahone:
This will not make it into 14.1 milestone: there's a little bit too much clean-up work required here. I'll try to fast-follow with a patch to get this included, but I don't want to hold up the 14.1 release on a bugfix.

PR #1610, opened 2018-12-15
add type name to information

### LINKED ISSUE #1609: improve debug experience for schema
when there are missing types we just get an error message like
```
Error: Introspection must provide output type for fields.
```
it would help a lot if the message lists the missing type.

OneCyrus:
fixes #1609

now it lists the type which fails the validation

### DIFF
--- a/src/utilities/buildClientSchema.js
+++ b/src/utilities/buildClientSchema.js
@@ -143,7 +143,7 @@ export function buildClientSchema(
     const type = getType(typeRef);
     invariant(
       isInputType(type),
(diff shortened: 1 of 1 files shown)


Neitsch:
Can you provide some example outputs, please?

It would also be nice, if the output of inspect is included in the sentence: 
```
"Introspection must provide output type for fields, but received: " + inspect(type)
```

OneCyrus:
updated the message. the message looks like this:

```
Introspection must provide output type for fields, but received: [Volume]
```

PR #1628, opened 2018-12-24
Add stripIgnoredTokens feature to remove insignificant whitespace

rybon:
Solves #1523.


rybon:
@IvanGoncharov can you review this one? Thanks.

IvanGoncharov:
@rybon Thanks for PR 👍 and Happy New Year 🎄 

Thanks a lot for writting bunch of test cases.
Here is a few desing question to move this PR forward:

-----

Proposed functionality doesn't work with AST so I don't see any benefits in having it as part of `print`. Moreover it's hard to use with `printSchema`:
```js
const schema = buildClientSchema(introspection);
let sdl = printSchema(schema);
const ast = parse(sdl);
sdl = print(ast, { condense: true });
```
I think we should make 'condense' a standalone function inside `utils`.

-----

RegExp are good for PoC but it's a big maintainance problem in the future.
Plus the are huge number of edge cases, e.g. simplest one is:
```graphql
{
  compile(jsCode: "function () { console.log('Valid JS code inside GraphQL string'); }")
}
```
You just can't handle such use cases with RegExp and should use lexer to correctly implement this functionality:
https://github.com/graphql/graphql-js/blob/fc3e2e31050d08853259b426a1eb8fb061d7fb8e/src/language/__tests__/lexer-test.js#L726-L740

-----

Last issue for me is the name, I understand why you can't call it `minify` (names of variables, fragments and operations stay unchanged) but I personally think it's imposible to figure out what `condense` do just from its name.
Since you just remove [ignored tokens](https://facebook.github.io/graphql/draft/#sec-Appendix-Grammar-Summary.Ignored-Tokens) how about renaming it to `removeIgnoredTokens`?

@rybon What do you think?

rybon:
@IvanGoncharov Happy New Year to you too :-). Thanks for the review.

I've put this feature in the `printer` module as that is the module that 'prints' an AST to a string. But if that isn't the right place to put it and we should move it to `utils` instead, fine by me. I don't really have an opinion on that. I'd rather have something in place in this library than nothing.

I agree that RegExp is kinda hacky and not an ideal solution. There are indeed a number of edge cases that are very hard to catch with this approach. And if the GraphQL spec expands, one would need to expand the RegExp as well. I don't have any experience with ASTs and lexers, so I don't know how to implement this feature with that approach instead. But I'm willing to try it out. Would you (or anyone else) be willing to help me get started on this?

I'm not really attached to the name `condense`. Naming things is hard. I was trying to come up with a name that is descriptive of the feature (https://en.wiktionary.org/wiki/condense) and `minify` didn't quite feel right. But another name is fine by me.

IvanGoncharov:
@rybon Sorry, I accidentally screw this PR trying to add commit on top of it.

>  I don't have any experience with ASTs and lexers, so I don't know how to implement this feature with that approach instead. But I'm willing to try it out. Would you (or anyone else) be willing to help me get started on this?

I implemented it here:
https://github.com/graphql/graphql-js/commit/be79c4e1fd42a93a6be1f90beb4eaa4f1fd40eeb
You just need to document and test it.
I can't reopen this PR :(
Please open a new one based on `removeIgnoredTokens` and make necessary changes.

IvanGoncharov:
@rybon I screw up `master` at your GitHub repo and can't revert it 😭 
Can you please force push local `master` with: `git push -f origin master`?

rybon:
Sure. You want me to override your latest changes to my branch?

IvanGoncharov:
@rybon Thanks 👍 
I reopen PR and added commit on top of it.

IvanGoncharov:
@rybon I tested it based on your example:
```js
const {
  stripIgnoredTokens
} = require('./dist/utilities/stripignoredtokens');

const result = stripIgnoredTokens(`
query SomeQuery($foo: String!, $bar: String) {
  someField(foo: $foo, bar: $bar) {
    a
    b {
      c
      d
    }
  }
} 
`);

console.log(result);
// query SomeQuery($foo:String!$bar:String){someField(foo:$foo bar:$bar){a b{c d}}}
```

rybon:
Looks good.

I'll revert the changes in `printer` and move the tests to `utilities`.

IvanGoncharov:
@rybon Great 👍 
Please also add tests for SDL.
Some introspection results are huge and storing them as condense SDL is valid use case.

rybon:
@IvanGoncharov done. I've added tests for the Query document and SDL document.

IvanGoncharov:
> I've synced with master to make use of the fixtures and added documentation.

@rybon Thanks a lot.

>  What still needs to be done in this PR before merging to master?

Please enable `@flow strict` inside tests and also export this function inside `index.js` both in `src` and `src/utilities`.

I will try to review tests in the next few days, feel free to ping me.

rybon:
@IvanGoncharov done. With `@flow strict` I get errors when using `String.raw`. Flow doesn't like it very much apparently. See https://github.com/facebook/flow/issues/2616 and https://github.com/facebook/flow/issues/5705.

rybon:
@IvanGoncharov feel free to review.

IvanGoncharov:
@rybon Looking at this right now.
I will make a commit on top for you to review.

src/utilities/__tests__/stripIgnoredTokens-test.js
```diff
@@ -0,0 +1,461 @@
...
+
+  it('produces helpful error messages', () => {
+    const badAst1 = { random: 'Data' };
+    expect(() => stripIgnoredTokens(print(badAst1))).to.throw(
+      'Invalid AST Node: { random: "Data" }',
+    );
+  });
+
+  it('correctly prints non-query operations without name', () => {
+    const queryAstShorthanded = parse('query { id, name }');
```
IvanGoncharov: @rybon You basically duplicated all tests from `print`.
Sadly most of them not relevant to this functionality and you basically test `print` one more time and not covering edge cases specific to `stripIgnoredTokens`.
These test shouldn't contain `print` at all since `stripIgnoredTokens` work with string, not with AST.

I will work on this tomorrow.

mjmahone:
@rybon we are trying to increase the cadence of our minor releases, so that feature additions don't have to wait months to make it into the release.

Once this feature is complete, we should already have a release bump scheduled shortly thereafter. It will not take nearly as long to go from `14.1 => 14.2` as it took to go from `14.0 => 14.1`.

langpavel:
I like this idea, really. 

But code coverage should be increased — especially new features should be deeply tested.

BTW Can this be implemented for AST too? — like `print` is now?

I always have AST (processed, validated, etc) when I wish write it down for next tooling.

rybon:
@langpavel OK fair enough. But keep in mind that I'm not very familiar with this codebase, so I would need some help to get that done. Could you help me out?

langpavel:
@rybon I think that all you need is altered copy of [src/language/printer.js](https://github.com/graphql/graphql-js/blob/master/src/language/printer.js)

rybon:
That is where the implementation was originally done (see earlier commits). What needs to be done in that copy?

langpavel:
@rybon Ok, I see first commit.
Point is that you used regex replaces, why? All the source code is baked here from AST, so if you rewrite [this functions](https://github.com/graphql/graphql-js/blob/master/src/language/printer.js#L241..L265) (`join`, `block`, `indent`) and `printDocASTReducer`, you will have everythig for free.

BTW I like latest tiny implementation. But having only AST version make more sense to me.
@IvanGoncharov What do you think?

rybon:
The RegExp implementation was done because I wasn't aware of an alternative approach at the time. I didn't decrease coverage on purpose, but I see it decreased by 0.007%. Are we striving for 100% coverage in this project? If so, how do we achieve that (have every single line covered)?

OK, I can rewrite these functions, but they would have to be parameterized to deal with this use case. What should the implementation be in that case? I have no experience with ASTs.

langpavel:
@rybon let's wait for @IvanGoncharov.

> OK, I can rewrite these functions, but they would have to be parameterized to deal with this use case. What should the implementation be in that case? I have no experience with ASTs.

Better way is create new `printDocASTReducer`, say `printDenseASTReducer` which will define custom serializers.

You can learn about AST here: https://astexplorer.net/

IvanGoncharov:
@rybon Situation is like this I agree with the current implementation.
It contains some minor hack: testing token for value to distinguish between punctuation tokens and all other but it can be a fixed latter.

The real problem here is that `graphql-js` has a pretty high standard for tests.
Not only in coverage (ATM 98% and we slowly moving to 100%) but in testing functionality.
It's very easy to fully cover this function since it short and doesn't contain a lot of `if` statements.

The idea here is that it simply not enough to use tests for `print` they are designed for a totally different case in mind.
You can just pass the output of `print` through `stripIgnoredTokens` and test result since `stripIgnoredTokens` is the standalone function and should be tested as such.
For example `print` never return tabs so how do you test that tabs are removed.
Or how we test for Unicode BOM which is also ignored token.

I'm fully supporting this functionality but currently I need to finish a couple of other long standing issues with this lib. I can promise that we will include it in upcoming `14.2.0`. But I need some time to finish other projects that I already started.

langpavel:
@rybon What is not covered is:
- [ ] input handling:
   - [x] input is string
   - [ ] input is `Source`
   - [ ] throws in other cases
- [ ] throwing out comments from input

rybon:
@langpavel done. Only line I can't seem to get covered is the `continue;` statement, could you help me with that?

rybon:
@IvanGoncharov anything else that needs to be done for this PR to land?

IvanGoncharov:
@rybon Sorry for the delay.
Lexer implementation that I added previously was just a proof of concept and not really future-proof.
Also based on feedback #1523 I think we should support striping of indentation inside block string. 
Plus even though current test do 100% percent coverage it is still based on print tests some of don't make sense for this functionality and missing many other checks.

I totally understand that adding this feature taking too much time so I stoped development of all other features until I merge this one. I'm working on it in my free time but I expect it will take a couple days at most.

P.S. I need to rebase some commit and also add new ones so to prevent stupid situation with github (like what I did with Lexer commit) I will open separate PR.

rybon:
Just saw your PR. Great work! Thank you for your efforts, I really appreciate it.

PR #1637, opened 2018-12-28
Fix: Handle new lines inside block string correctly

### LINKED ISSUE #1636: [lexer] Block string does not advance line
Lexer does not advance line number when tokenizing block strings.

Line numbers are then shifted on all following tokens

langpavel:
This fixes #1636

### DIFF
--- a/src/language/__tests__/lexer-test.js
+++ b/src/language/__tests__/lexer-test.js
@@ -21,6 +21,12 @@ function lexOne(str) {
   return lexer.advance();
 }
 
+function lexSecond(str) {
+  const lexer = createLexer(new Source(str));
+  lexer.advance();
+  return lexer.advance();
+}
+
 function expectSyntaxError(text, message, location) {
   expect(() => lexOne(text))
     .to.throw('Syntax Error: ' + message)
@@ -356,6 +362,24 @@ describe('Lexer', () => {
       start: 0,
       end: 68,
       value: 'spans\n  multiple\n    lines',
+    });
+  });
+
+  it('advance line after lexing block string', () => {
+    expect(
+      lexSecond(`"""
+
+        spans
+          multiple
+            lines
+
+        """ second_token`),
+    ).to.contain({
+      kind: TokenKind.NAME,
+      start: 69,
+      end: 81,
+      line: 7,
+      value: 'second_token',
     });
   });
 
--- a/src/language/__tests__/lexer-test.js
+++ b/src/language/__tests__/lexer-test.js
@@ -365,7 +365,7 @@ describe('Lexer', () => {
     });
   });
 
-  it('advance line after lexing block string', () => {
+  it('advance line after lexing multiline block string', () => {
     expect(
       lexSecond(`"""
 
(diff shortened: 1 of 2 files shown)


langpavel:
Hi @IvanGoncharov 
Can I kindly ask you for review?
Thanks!

src/language/lexer.js
```diff
@@ -668,8 +668,22 @@ function readBlockString(source, start, line, col, prev): Token {
...
 
-    // Escape Triple-Quote (\""")
-    if (
+    if (code === 10) {
+      // new line
+      ++position;
+      ++lexer.line;
+      lexer.lineStart = position;
+    } else if (code === 13) {
+      // carriage return
```
IvanGoncharov: @langpavel Can you please add test cases that cover `\r\n` and `\n\r`?

src/language/__tests__/lexer-test.js
```diff
@@ -359,6 +365,24 @@ describe('Lexer', () => {
...
 
+  it('advance line after lexing block string', () => {
+    expect(
+      lexSecond(`"""
+
+        spans
+          multiple
+            lines
+
+        """ second_token`),
```
IvanGoncharov: To cover all cases in one test you can do something like this:
```js
const str = [
  '""" \n',
  'spans \r\n',
  'multiple \n\r',
  'lines \n\n',
  '""" second_token'
].join('')
```

IvanGoncharov:
@langpavel Thanks for PR 👍 
Please expand tests and I will merge this PR.

P.S. It's a good temporary solution but I don't like duplication of new line code handling and having special handling of block strings.
I think in long run we need to implement a common function for advancing lexer to the next character. Maybe even unify with `CharacterStream` from language server:
https://github.com/graphql/graphql-language-service/blob/master/packages/parser/src/CharacterStream.js

langpavel:
Ok, I will try refactor to new lexer method `advanceLine`, existing tests should cover this so no need to expand tests for block string

langpavel:
@IvanGoncharov I extracted `consumeNewline` function so newlines are handled on one place.

src/language/lexer.js
```diff
@@ -361,6 +361,34 @@ function unexpectedCharacterMessage(code) {
   return `Cannot parse the unexpected character ${printCharCode(code)}.`;
 }
 
+function consumeNewline(
+  startPosition: number,
+  code: number,
+  codeahead: number,
+  lexer: Lexer<*>,
+): [boolean, number] {
```
IvanGoncharov: @langpavel It complicates code :(
Just revert to previous commit.
My point was that sometime in future we need to refactor Lexer, and your previous code was good until then.

langpavel: Ok, I will force push up to previous commit

IvanGoncharov: @langpavel Sorry for confusing you

langpavel: Last commit deleted

IvanGoncharov:
> existing tests should cover this so no need to expand tests for block string

@langpavel Existing tests don't check that `\r\n` or `\n\r` advance line counter by one.

src/language/__tests__/lexer-test.js
langpavel: This case will not cover edge cases because various combinations can annihilate

langpavel:
Ok, I will add more tests

langpavel:
@IvanGoncharov I added your test case and also added `column` assertion

src/language/lexer.js
langpavel: Added

src/language/__tests__/lexer-test.js
langpavel: Added, exactly like you suggested

IvanGoncharov:
@langpavel Thanks 👍 
Merged 🎉

langpavel:
Thanks! Happy new year :tada:

PR #1646, opened 2019-01-05
Remove unnecessary use of hasOwnProperty

IvanGoncharov:


src/utilities/coerceValue.js
```diff
@@ -147,58 +148,53 @@ export function coerceValue(
...
-    // Ensure every provided field is defined.
-    for (const fieldName in value) {
-      if (hasOwnProperty.call(value, fieldName)) {
-        if (!fields[fieldName]) {
-          const suggestions = suggestionList(fieldName, Object.keys(fields));
-          const didYouMean =
-            suggestions.length !== 0
-              ? `did you mean ${orList(suggestions)}?`
-              : undefined;
+    for (const field of objectValues(fields)) {
```
IvanGoncharov: No changes to algorithm just 
```js
for (const fieldName in fields) {
  if (hasOwnProperty.call(fields, fieldName)) {
    // ...
  }
}
```
into
```js
for (const field of objectValues(fields)) {
  // Exact same code :)
}
```

PR #1667, opened 2019-01-16
Update travis api_key for npm deploys

mjmahone:
This api key was out of date preventing us from doing automatic version deployments.

Created the api key by running "travis encrypt": https://docs.travis-ci.com/user/deployment/npm/#npm-auth-token

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,17 @@
 language: node_js
+
+# https://github.com/nodejs/Release
 node_js:
-- '11'
-- '10'
-- '8'
-- '6'
+  - '11'
+  - '10'
+  - '8'
+  - '6'
+
 git:
   depth: 5
+
 cache: yarn
+
 notifications:
   irc:
     use_notice: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,52 +1,45 @@
...
+    node_js: lts/*
+    deploy:
+    - provider: script
+      script: npm run gitpublish
+      skip_cleanup: true
+      on:
+        branch: master
+    - provider: npm
+      skip_cleanup: true
+      email: lee@leebyron.com
```
IvanGoncharov: @mjmahone I think you need to write email that you registered in NPM:
image

.travis.yml
```diff
@@ -1,52 +1,45 @@
...
+      api_key:
+        secure: VrDxiDK9vhe0F8WulJ+HcAkMwuvqIhtXYethlJBwkS9N57F4RuxzCJE0VuDDS90qMAewA450lfT5caCazir4opvpZebB16iy8/9J2ymhBj6bJs9rm/B3Xg/xQSLOWAmdq1H4lPF6CihQ5LR57BZbi9Rx3f5JOeXgiqQNASJH+mrXiQh3qZ9Ry8YJXhvcUDQNX57ooSBeGYXf5PcKsEUF0Qs6XLUCUdOztMrtA+sIZkQxAaEBBG7tkE6LdF3OJl2ORNMig05juyfeevcGk6g4XnoIA4gvnW1KHToLvuGWdoCHAW1N269JAaH4bL2JfcA5A7KXbZVwif0jIJz47dmOsnIMlMdSqBpw7zSRSq2Ikv1Vb5XRPeVk0CK0d2kUZej6Jb9+/Jed39chfyAAMVM4Wba8UcTKbOIqv7Kl1SmmgBOGveK7k4fpffbWdofi9avuqzeftdFrM1FULsX5Wne9n3xuJDvVLW0YRB2+FVmfwGNwzW5X/bzgLDRWB5W5JL60FXt91Hrk5bIi0GcPx6n0Ls7IJ9U6r5kWfcuTYl0gxOiWQDJ0OCXBXA5taIO6/AX/d9af4VhGc/EVxuIpUaW52nZOL49KG/3tYx39mDCqRf7w7/658RVXNO/yLy6b/QPY7+3Lua9y88xLxru3bI9gpNUdnYUFIehktuiAkuunWoA=
```
IvanGoncharov: @mjmahone Can you please copy this token into old `.travis.yml`?
All other changes in the file are results of `travis` CLI and shouldn't influence publishing.
It will keep GitHub history clean.

.travis.yml
mjmahone: Hmmm I'm not sure but it's required, as we both have publish access, but that would make sense. Also probably better for the person whose auth token is being used to get any emails associated with it.

PR #1669, opened 2019-01-16
Remove usage of Flow placeholder argument + cleanup

IvanGoncharov:


src/utilities/buildASTSchema.js
```diff
@@ -277,19 +279,31 @@ export class ASTDefinitionBuilder {
...
       description: getDescription(field, this._options),
-      args: field.arguments && this._makeInputValues(field.arguments),
+      args: keyByNameNode(field.arguments || [], arg => this.buildArg(arg)),
       deprecationReason: getDeprecationReason(field),
       astNode: field,
     };
   }
 
-  buildInputField(value: InputValueDefinitionNode): GraphQLInputField {
+  buildArg(value: InputValueDefinitionNode): GraphQLArgumentConfig {
```
mjmahone: We probably shouldn't expose this, as we're not building a schema `GraphQLArgument` but just a `GraphQLArgumentConfig`.

Having `buildFoo` sometimes return a `GraphQLFoo` and sometimes a `GraphQLFooConfig` is confusing.

mjmahone: Oh I see we're already doing this with `buildField`. That is quite confusing but is fine then. I wish `buildDirective` was also building a GraphQLDirectiveConfig too, but that ship has sailed. Maybe a wishlist for v15?

IvanGoncharov: @mjmahone The entire class is internal it's just a bunch of functions that `buildSchema`/`extendSchema` are using.
So it's just cleanup of internal API.

Issue #1413, opened 2018-07-09
ObjectType implementing InterfaceType doesn't end up in the schema


CrocoDillon:
Consider this code:

```js
const {
  GraphQLID,
  GraphQLInterfaceType,
  GraphQLNonNull,
  GraphQLObjectType,
  GraphQLSchema,
  printSchema,
} = require('graphql')

const NodeType = new GraphQLInterfaceType({
  name: 'Node',
  fields: {
    id: {
      type: new GraphQLNonNull(GraphQLID),
    },
  },
})

class Car {} // Kept brief for demo purpose

const CarType = new GraphQLObjectType({
  name: 'Car',
  interfaces: [NodeType],
  fields: {
    id: {
      type: new GraphQLNonNull(GraphQLID),
    },
  },
  isTypeOf: value => value instanceof Car,
})

const schema = new GraphQLSchema({
  query: new GraphQLObjectType({
    name: 'Query',
    fields: {
      node: {
        type: NodeType,
        args: {
          id: {
            type: new GraphQLNonNull(GraphQLID),
          },
        },
        resolve: () => new Car(),
      },
    },
  }),
})

console.log(printSchema(schema))
```

Which gives this output:

```gql
interface Node {
  id: ID!
}

type Query {
  node(id: ID!): Node
}
```

This is a very simplified demo of the problem I'm running in to. Cars (in this case) don't have their own queries so they don't end up in the schema even though you can query for them by the interface they implement.

How can I make them appear in the printed schema?

IvanGoncharov:
> How can I make them appear in the printed schema?

@CrocoDillon You need to write this:
```js
const schema = new GraphQLSchema({
  types: [CarType],
  query: new GraphQLObjectType({
```

CrocoDillon:
Thanks, do you think it might be a good idea to add this to the documentation? I couldn't find it and I looked everywhere (except in the code itself, where it is pretty obvious... silly me)

For example here https://graphql.org/graphql-js/type/#graphqlschema

If so, I can make a PR for [`graphql/graphql.github.io`](https://github.com/graphql/graphql.github.io/blob/source/site/graphql-js/APIReference-TypeSystem.md#graphqlschema)

Either way thanks a lot!

IvanGoncharov:
@CrocoDillon Yes, it would be extremely helpful.
One problem through `graphql.org/graphql-js` was handwritten for the first release of this lib and severely outdated.
[...]

PR #1679, opened 2019-01-21
Documentation for explicitly mentioning ObjectTypes

buoyantair:
Based on the discussion at #1413

### DIFF
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -82,6 +82,15 @@ export function assertSchema(schema: mixed): GraphQLSchema {
  *       ...
  *       directives: specifiedDirectives.concat([ myCustomDirective ]),
  *     })
+
+ * Note: ObjectTypes which do not have their own queries in the Schema will not end up
+ * in the final Schema. To achieve this, You need to explicitly mention the ObjectType
+ * while creating the Schema. Example:
+ *
+ *      const schema = new GraphQLSchema({
+ *        types: [CarType],
+ *        ...
+ *      }) 
  *
  */
 export class GraphQLSchema {


src/type/schema.js
```diff
@@ -83,6 +83,15 @@ export function assertSchema(schema: mixed): GraphQLSchema {
  *       directives: specifiedDirectives.concat([ myCustomDirective ]),
  *     })
  *
+ * Note: ObjectTypes which do not have their own queries in the Schema will not end up
+ * in the final Schema. To achieve this, You need to explicitly mention the ObjectType
+ * while creating the Schema. Example:
```
IvanGoncharov: @buoyantair Thanks for PR 👍 

> which do not have their own queries in the Schema will not end up in the final Schema

It sounds confising for me. Idea is that schema will include only types that are reachable by traversing root types and only through direct references.
But I'm not sure how to better formulate it.
@Neitsch Can you please help?

I few minor suggestion:
  - Any type not only ObjectTypes so please change `ObjectTypes` => `Types`.
  - would be great to have an example based around interface scenario since it most popular one you can use this code as a basis and remove all unnecessary details but include simplified `Query` definition:
https://github.com/graphql/graphql-js/blob/bdfd1c63e0b45fc48b9c52c2b652680049d76322/src/__tests__/starWarsSchema.js#L299-L302 
  - Can you please move it before note about directive.

buoyantair: @IvanGoncharov Updated, do review.

src/type/schema.js
```diff
@@ -72,6 +72,62 @@ export function assertSchema(schema: mixed): GraphQLSchema {
  *       mutation: MyAppMutationRootType,
  *     })
  *
+ * Note: When the schema is constructed, only the types that are reachable through
+ * traversing the root types & those that are explicitly referenced are included.
```
IvanGoncharov: How about?
> Note: When the schema is constructed by default it includes only types that are reachable through traversing root types by following only explicit references.

src/type/schema.js
```diff
@@ -72,6 +72,62 @@ export function assertSchema(schema: mixed): GraphQLSchema {
...
+ * Example:
+ *      const {
+ *        GraphQLID,
+ *        GraphQLInterfaceType,
+ *        GraphQLNonNull,
+ *        GraphQLObjectType,
+ *        GraphQLSchema,
+ *        printSchema,
+ *      } = require('graphql')
```
IvanGoncharov: I think it should be shorter + most of our examples are based on SWAPI example.
So how about this?
```js
const characterInterface = new GraphQLInterfaceType({
  name: 'Character',
  ...
});

const humanType = new GraphQLObjectType({
  name: 'Human',
  interfaces: [characterInterface],
  ...
});

const droidType = new GraphQLObjectType({
  name: 'Droid',
  interfaces: [characterInterface],
  ...
});

    name: 'Query',
    fields: {
      hero: { type: characterInterface, ... },
    }
  }),
  ...
  // Since this schema reference only `Character` interface it's nessary to
  // explicitly list types that implement it.
  types: [humanType, droidType],
})
```

src/type/schema.js
buoyantair: @IvanGoncharov Does that mean that it will only include types that are mentioned in query field's type or does it mean it will only include those types in fields too only if they are explicitly referenced in the `types` property of schema config?

src/type/schema.js
buoyantair: @IvanGoncharov You're right. This would keep things more coherent.

buoyantair: @IvanGoncharov Wouldn't the sentence in the schema example only be valid if the user explicitly wants those types (`humanType`, `driodType`) to be included in the final schema?

IvanGoncharov: > In which case I would like to rewrite it as such:

@buoyantair Excellent correction 👍

src/type/schema.js
IvanGoncharov: By default, (without `types`) type should be reachable to be included.
Basically, entire schema reduced to the list of types using https://github.com/graphql/graphql-js/blob/bdfd1c63e0b45fc48b9c52c2b652680049d76322/src/type/schema.js#L300
And root types + `config.types` is used as starting set for reduce:
https://github.com/graphql/graphql-js/blob/bdfd1c63e0b45fc48b9c52c2b652680049d76322/src/type/schema.js#L141-L158

Not sure how to comunicate that in clear form without causing even more confusion.

buoyantair: @IvanGoncharov How about: 

> Note: When the schema is constructed, by default only the types that are reachable by traversing the root types are included, other types must be explicitly referenced.

IvanGoncharov: Looks good! 👍

src/type/schema.js
```diff
@@ -83,6 +120,7 @@ export function assertSchema(schema: mixed): GraphQLSchema {
  *       directives: specifiedDirectives.concat([ myCustomDirective ]),
  *     })
  *
+ *
```
IvanGoncharov: @buoyantair Can please remove it?

buoyantair: @IvanGoncharov Resolved

ab-pm:
Unfortunately this change didn't arrive in the documentation at https://graphql.org/graphql-js/type/#graphqlschema. Is there something else that needs to be done for that?

PR #1756, opened 2019-02-28
Speedup parser (~50%) by using 'slice' and 'charCodeAt' directly

IvanGoncharov:
image

### DIFF
--- a/src/language/lexer.js
+++ b/src/language/lexer.js
@@ -137,9 +137,6 @@ export function getTokenDesc(token: Token): string {
   return value ? `${token.kind} "${value}"` : token.kind;
 }
 
-const charCodeAt = String.prototype.charCodeAt;
-const slice = String.prototype.slice;
-
 /**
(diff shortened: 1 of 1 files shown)


src/language/lexer.js
```diff
@@ -137,9 +137,6 @@ export function getTokenDesc(token: Token): string {
   return value ? `${token.kind} "${value}"` : token.kind;
 }
 
-const charCodeAt = String.prototype.charCodeAt;
-const slice = String.prototype.slice;
-
```
IvanGoncharov: This code was added in the initial commit(b5ed31e6e3136accba13c3a4a5b7f61c36d7dcb4) so I missing context why it was added in the first place and is it safe to remove.
Even thought parsing is not the most critical function 50% speed up is very big.

@leebyron @mjmahone @Neitsch Can you please review and comment on whenever it is safe to use `slice` and `charCodeAt` directly?

PR #1769, opened 2019-03-04
add Chinese README file

chenmf6:
add README-ZH.md

### DIFF
--- a/README-ZH.md
+++ b/README-ZH.md
@@ -1,0 +1,123 @@
+# GraphQL.js
+
+GraphQL.js是GraphQL的JavaScript参考实现。GraphQL是由Facebook开发的一组API查询语言。
+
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@chenmf6 Thanks for PR 👍 
And as a non-English speaker myself I understand the benefits of translated docs.
On the other hand, we can't maintain docs in multiple languages inside a single repo.
The proper solution would be to have separate `ch.graphql.org` similar to how it's done in React: https://reactjs.org/blog/2019/02/23/is-react-translated-yet.html
More details here: https://github.com/facebook/react/issues/8063

I think we should have a general solution for GraphQL spec, `graphql.org` and `graphql-js` and I think we should discuss possible solutions in a single place: https://github.com/facebook/graphql/issues/331

PR #1771, opened 2019-03-04
Limit maximum output of 'inspect'

### LINKED ISSUE #1753: Large memory allocations and very slow execution when type errors occur in complex inputs
Hi,

We have a rather large complex object as an input. A single request with hundreds of type errors in the input took 9 seconds in `graphql.execute`. The majority of the time spent (>95%) was spent in the `inspect()` call, specifically its recursive object string construction. There were huge amount of memory allocations. With 20 concurrent requests, we can consistently cause node to run out of memory. Changing `inspect()` to return an empty string lowered the total execution time of our process to low hundreds of milliseconds and the process was stable. I don't know the exact latency of `graphql.execute` with `inspect` effectively disabled. Given this finding, I am proposing three changes.

1. `inspect()` should never be in the execution path as currently written. At the very least, it should not be recursive and/or have a maximum string length. In our case, each `inspect` grew to be over 100KB.
2. There should be a maximum amount of errors in `getVariableValues()`. If the number of errors reaches the maximum, no additional varDefNodes are processed. This grew to be ~1500. Cap at 5? 
3. If `inspect()` must be kept, there should also be a maximum total amount of error lengths between error messages in `getVariableValues()`. If the limit is reached, no more varDefNodes should be processed. This would prevent large single errors from multiplying by the cap.

Given that we had ~1500 errors, each error was ~100KB and there were 20 concurrent requests, that is 3GB of memory. That is more than the default max-old-space-size.

Item 2 is an easy fix but 1 and 3 will make the execution path much more robust. We are in an air-gapped environment so I cannot easily provide a reproducible case.

```
        const properties = Object.keys(value)
          .map(k => `${k}: ${inspect(value[k])}`)
          .join(', ');
```

graphql 14.1.1
node 8.11.3

IvanGoncharov:
It's a partial fix for #1753 
I basically copied how Node.js `inspect` limits output but with different limits:
 * `10` array items instead of `100`
 * `2` levels of recursion instead of `3`

Note: Node.js `inspect` has no limits on the max number of properties inside the object so I didn't add it either.

@Cito AFAIK Python version also has this problem. So it would be great if you also review it.

Note: It doesn't fully fix #1753 since you can still produce a huge number of errors but at least those error messages become shorter in most of the cases.

### DIFF
--- a/src/jsutils/inspect.js
+++ b/src/jsutils/inspect.js
@@ -10,15 +10,16 @@
 import nodejsCustomInspectSymbol from './nodejsCustomInspectSymbol';
 
 const MAX_ARRAY_LENGTH = 10;
+const MAX_RECURSIVE_DEPTH = 2;
 
 /**
  * Used to print values in error messages.
  */
 export default function inspect(value: mixed): string {
-  return formatValue(value);
+  return formatValue(value, 0);
 }
 
-function formatValue(value) {
+function formatValue(value, recurseTimes) {
   switch (typeof value) {
     case 'string':
       return JSON.stringify(value);
@@ -33,12 +34,12 @@ function formatValue(value) {
           const customValue = customInspectFn.call(value);
           return typeof customValue === 'string'
             ? customValue
-            : formatValue(customValue);
+            : formatValue(customValue, recurseTimes);
         } else if (Array.isArray(value)) {
-          return formatArray(value);
+          return formatArray(value, recurseTimes);
         }
 
-        return formatObject(value);
+        return formatObject(value, recurseTimes);
       }
 
       return String(value);
@@ -47,23 +48,31 @@ function formatValue(value) {
   }
 }
 
-function formatObject(object) {
+function formatObject(object, recurseTimes) {
   const keys = Object.keys(object);
   if (keys.length === 0) {
     return '{}';
   }
 
+  if (recurseTimes === MAX_RECURSIVE_DEPTH) {
+    return '[' + getObjectTag(object) + ']';
+  }
+
   const properties = keys.map(key => {
-    const value = formatValue(object[key]);
+    const value = formatValue(object[key], recurseTimes + 1);
     return key + ': ' + value;
   });
 
   return '{ ' + properties.join(', ') + ' }';
 }
 
-function formatArray(array) {
+function formatArray(array, recurseTimes) {
   if (array.length === 0) {
     return '[]';
+  }
+
+  if (recurseTimes === MAX_RECURSIVE_DEPTH) {
+    return '[Array]';
   }
 
   const len = Math.min(MAX_ARRAY_LENGTH, array.length);
@@ -71,7 +80,7 @@ function formatArray(array) {
   const items = [];
 
   for (let i = 0; i < len; ++i) {
-    items.push(formatValue(array[i]));
+    items.push(formatValue(array[i], recurseTimes + 1));
   }
 
   if (remaining === 1) {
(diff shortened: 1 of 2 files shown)


Cito:
@IvanGoncharov Right, this needs to be fixed in Python as well. I had first used `repr()` which at least took care of recursion, but then switched to `inspect()`. Will have a look at your fix and maybe add a few more tests later this week.

slicejunk:
in the current stable release, inspect causes RangeError in this line from "values.js"
```
       error.message =
                `Variable "$${varName}" got invalid value ${inspect(value)}; ` +
                error.message;
```

when ```value``` contains an object such as a file to upload, maybe due to its circular structure

please merge this fix!

Cito:
@IvanGoncharov Looks good to me. Maybe for the sake of safety and consistency we should still restrict the number of object properties even if node,js doesn't do this? This could be implemented very similar to how you do it with arrays (i.e. also giving a hint how many entries were left out). Also, I wonder whether we should truncate long strings as well (those returned by String() and custom inspect functions)? Again, there should be an ellipsis with a hint how many chars were left out when this happens.

Cito:
Probably you should also increment `recurseTimes` when the `customInspectFn` is not a string. Otherwise objects like `const a = { inspect: () => a }` can cause an infinite recursion.

IvanGoncharov:
> Probably you should also increment recurseTimes when the customInspectFn is not a string. Otherwise objects like const a = { inspect: () => a } can cause an infinite recursion.

@Cito Addressed in 8f2c38333a47e63a12d0e9e77e485154dabc018b 
I also added detection of Circular objects similar to how it's done in Node.js `inspect`.
Can you please take a look?

Cito:
@IvanGoncharov with that check you can only avoid direct recursions, but not larger circles like this:
```javascript
a = {}
b = {}
a.inspect = () => b
b.inspect = () => a
inspect(a) // -> Maximum call stack size exceeded
```

IvanGoncharov:
> with that check you can only avoid direct recursions, but not larger circles like this:

Thanks for the great test case 👍 
I added it in fadbc78 and it's correctly detected as `[Circular]`.
The big benefit in copying Node.js `inspect` is that it's already battle-tested.

Cito:
Nice. You only mentioned 8f2c38333a47e63a12d0e9e77e485154dabc018b which was not sufficient, but the actual fix is 2700388936350f1a1495de743ce3c65ef1fd9546 which does the job and I agree is the best solution - both keeping track of seen objects and limiting the track record (recursion depth).

Btw, the `inspect` property seems to be deprecated in favor of the symbol in node.js. Should we also print a deprecation message?

IvanGoncharov:
> Btw, the inspect property seems to be deprecated in favor of the symbol in node.js. Should we also print a deprecation message?

@Cito In theory entire `graphql-js` should be usable in the browser (in practice people mostly use `parse`, `visit`, `buildClientSchema`, etc.) printing debug/deprecation messages on the client side is usually a bad idea at least in production builds.

In the future, we can add detection of dev/prod builds and do something similar to `React` warnings but it's topic for the separate PR.

src/jsutils/inspect.js
```diff
@@ -9,40 +9,102 @@
 
 import nodejsCustomInspectSymbol from './nodejsCustomInspectSymbol';
 
+const MAX_ARRAY_LENGTH = 10;
```
mjmahone: Should these be customizeable values? I realize these match, for instance, the current `console.log` max depth. For now it is probably better to hard-code, as we can always open up a customizeable option in the future, whereas if we supply that option now we can't easily take it back.

IvanGoncharov:
@mjmahone Thanks for the review 👍 
> I probably would have picked the same limits as node, but this seems like a good enough improvement that I'm happy with it as-is!

Since DoS scenario described in #1753 based on `MEMORY = NUM_ERRORS * ERROR_SIZE`.
We should either limit the number of errors or potential size of the error.
I think if limit number of errors it will be more negatively for DX.

PR #1772, opened 2019-03-05
fix out of range return check for charCodeAt in lexer (NaN instead of null)

superhawk610:
According to the [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt#Return_value), `String.prototype.charCodeAt` will only ever return a `number` or `NaN`, never `null`.

This PR updates that check in each of the 4 places that it occurs.


IvanGoncharov:
@superhawk610 Thanks for PR 👍 
Looks good and definitely improve code quality.

PR #1786, opened 2019-03-12
Change printError() to mark location before using ↓ instead of ^ after.

jaydenseric:
Queries, particularly [when compressed](https://twitter.com/jaydenseric/status/1103801032675549184), can have a long line length. In narrow viewports (small Terminal windows, etc.) this results in the query wrapping over multiple lines.

Previously the error line location was marked on a line after with sufficient padding and a `^` character. The problem is, when the line above wraps the marker points up at an incorrect location.

By putting the marker on a line before, not after, there is less chance it will point at the wrong location on the following line due to wrapping. At least if the error occurred anywhere in the first line wrap (fairly likely) the marker will be correct.

Also, the `↓` symbol on the line before is more intuitive because people read from top to bottom and the symbol itself is more semantic.

Before (incorrect locations marked):

Screen Shot 2019-03-12 at 2 58 54 pm

After (correct locations marked):

Screen Shot 2019-03-12 at 2 57 47 pm

### DIFF
--- a/src/error/__tests__/printError-test.js
+++ b/src/error/__tests__/printError-test.js
@@ -28,8 +28,8 @@ describe('printError', () => {
       Single digit line number with no padding
 
       Test (9:1)
+         ↓
       9: *
-         ^
     `);
 
     const doubleDigit = new GraphQLError(
@@ -42,8 +42,8 @@ describe('printError', () => {
       Left padded first line number
 
       Test (9:1)
+          ↓
        9: *
-          ^
       10: 
     `);
   });
@@ -87,14 +87,14 @@ describe('printError', () => {
 
       SourceA (2:10)
       1: type Foo {
+                  ↓
       2:   field: String
-                  ^
       3: }
 
       SourceB (2:10)
       1: type Foo {
+                  ↓
       2:   field: Int
-                  ^
       3: }
     `);
   });
--- a/src/error/printError.js
+++ b/src/error/printError.js
@@ -64,8 +64,8 @@ function highlightSourceAtLocation(
     printPrefixedLines([
       // Lines specified like this: ["prefix", "string"],
       [`${lineNum - 1}: `, lines[lineIndex - 1]],
+      ['', whitespace(columnNum - 1) + '↓'],
       [`${lineNum}: `, lines[lineIndex]],
-      ['', whitespace(columnNum - 1) + '^'],
       [`${lineNum + 1}: `, lines[lineIndex + 1]],
     ])
   );
--- a/src/language/__tests__/lexer-test.js
+++ b/src/language/__tests__/lexer-test.js
@@ -129,8 +129,8 @@ describe('Lexer', () => {
 
       GraphQL request (3:5)
       2: 
+             ↓
       3:     ?
-             ^
       4: 
     `);
   });
@@ -149,8 +149,8 @@ describe('Lexer', () => {
 
       foo.js (13:6)
       12: 
+               ↓
       13:      ?
-               ^
       14: 
     `);
   });
@@ -167,8 +167,8 @@ describe('Lexer', () => {
       Syntax Error: Cannot parse the unexpected character "?".
 
       foo.js (1:5)
+             ↓
       1:     ?
-             ^
     `);
   });
 
--- a/src/language/__tests__/parser-test.js
+++ b/src/language/__tests__/parser-test.js
@@ -55,8 +55,8 @@ describe('Parser', () => {
       Syntax Error: Expected Name, found <EOF>
 
       GraphQL request (1:2)
+          ↓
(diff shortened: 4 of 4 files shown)


Cito:
Interesting idea. Two things you should consider, though: 1) Unicode is often not shown properly in logs and on the console. 2) GraphQL.js (to which we want to be compatible in graphql-core) and Python are both using the caret in error messages, probably because of reason 1) and people are accustomed to this notation.

IvanGoncharov:
@jaydenseric I agree with @Cito Unicode symbols come with its own set of problems.
Moreover, it's not a full solution to the core problem with long lines since it will work only until the first wrap. So it doesn't solve one of the most common mistake: missing closing curly brace.

I don't think we should reinvent the wheel so if you want errors to be correctly mapped to minimize query you need to implement something similar to the source map but for graphql.

Also, it's sims like everyone is using `^`. For example Node and Python:
image

jaydenseric:
@IvanGoncharov 

> Unicode symbols come with its own set of problems.

The symbol is not the main point of this PR, it can be substituted if this is a genuine issue.

> So it doesn't solve one of the most common mistake: missing closing curly brace.

I disagree that's a particularly common problem. I can't remember having that issue personally. The vast majority of errors are uncaught exceptions in top level query/mutation resolvers. The screenshots I shared above are of real errors I had to fix yesterday.

> if you want errors to be correctly mapped to minimize query

This PR is not *only* for compressed queries. Uncompressed query lines can also end up wrapping in narrow terminals.

This PR is a very simple solution that solves the problem for most cases, which is better than none.

> it's sims like everyone is using `^`

In this instance there is no requirement to conform to what other environments are doing. Maybe they will end up being inspired by this improvement.

Another idea that could happen in addition, is to only show an excerpt of characters before and after the error position, instead of a very long full line.

IvanGoncharov:
@jaydenseric As I said before I don't think we should reinvent the wheel. This is not GraphQL specific problem and almost any programming language or data format can have long lines, e.g. even JSON/XML/HTML have this problem.
So to move this issue forward please provide a few examples on how other popular languages or data formats solve this problem.

jaydenseric:
Honestly I'm feeling frustrated by the negativity.

> I don't think we should reinvent the wheel.

It's not "reinventing the wheel". It's a low hanging fruit; one line shifted and a single character change.

> So to move this issue forward please provide a few examples on how other popular languages or data formats solve this problem.

Why?

You are insinuating that if no one else is doing it (not that that is really the case here) it's by default a bad idea and not worth considering. By that logic, nothing would ever be invented, created or improved. That thinking is the opposite of the OSS ethos.

I've literally gone to the effort to make a working PR that solves the problem for most cases. It's impossible to solve for all, but this is an improvement with no downsides.

> This is not GraphQL specific problem and almost any programming language or data format can have long lines, e.g. even JSON/XML/HTML have this problem.

Not relevant to approving this PR. I would suggest any code is best annotated from above than below.

Cito:
@jaydenseric sorry for sounding negative (and a bit confusing - actually believed to be in the graphql-core issue tracker when responding). We certainly appretiate your proposal, and *are* considering it. That's hwat we are doing here, and that's how OSS works - by discussing everything in the open, collecting arguments and opinions from the community about the pros and cons, and then making a decision. In the end it's the project maintainer who is left holding the bag after changes, and you need to understand why project maintainers sometimes seem to be too conservative and reluctant in accepting change requests.

The reasons for being reluctant in this particular issue is that error messages are consumed, stored and processed in various ways, and for that matter are often assumed to be 7-bit-ascii - non-ascii may be not processed or displayed on the console correctly. For instance, on Windows the character set is often cp-1252 instead of utf-8. Btw, the × you suggest as alternative is also not 7-bit-ascii (and ambigous in 8-bit-"ascii": 158 is code page 850, but in unicode and cp-1252 it is 215). 

I'm also not completely convinced that annotating the line from above is really better, particularly when it's a single line (most see error messages are in debug mode where I don't compress queries). I understand your view, but you can also argue that people are accustomed to getting error messages displayed the old-school way on the console, and that the most important information should come first, which is the content, and annotation should come next.

Cito:
Of course, you can also argue that graphql strings are assumed to be utf-8 per the graphql and json specs, so there is no ambiguity regarding encoding and we can safely assume the down arrow displays properly everywhere in 2019.

Maybe one day we will even put facepalm emojis in error messages when the AI detects that it was a particulary stupid error...

IvanGoncharov:
@jaydenseric Sorry but we can't merge this PR based on problems outlined by @mjmahone @Cito and me.
That said, I'm still open to the PR based on a proven solution from other popular projects:
> So to move this issue forward please provide a few examples on how other implementations of popular languages or data formats solve this problem.

Issue #1791, opened 2019-03-14
Duplicate fields in deep inputs not validated properly


jbblanchet:
When using a simple input, setting the same key twice will give an error: `There can be only one input field named "y".`

When using an input in an input, that second level input will not give the same error.

```javascript
var { graphql, buildSchema } = require('graphql');

// Construct a schema, using GraphQL schema language
var schema = buildSchema(`
  type Query {
    deepInput(input: InputA): String,
    singleInput(input: InputB): String
  }

  input InputA {
    x: InputB
  }

  input InputB {
    y: String
  }
`);

var root = {
  singleInput: () => {
    return 'Hello single world!';
  },
  deepInput: () => {
    return 'Hello deep world!';
  },
};

// This will five the error that `y` is defined twice
graphql(schema, '{ singleInput(input: { y: "foo", y: "bar" }) }', root).then((response) => {
  console.log(response);
}, (error) => {
  console.log(error)
});

// But this won't
graphql(schema, '{ deepInput(input: { x: { y: "foo", y: "bar" } }) }', root).then((response) => {
  console.log(response);
}, (error) => {
  console.log(error)
});
```

IvanGoncharov:
@jbblanchet Thanks for reporting 👍 
Already fixed by @Cito in #1792 🚀 and will be included in the next release.

PR #1792, opened 2019-03-14
Duplicate fields in deep inputs not validated properly (#1791)

Cito:
When checking input fields, also descend into nested fields.


IvanGoncharov:
@Cito Thanks a lot 👍 
Merged 🎉

jbblanchet:
Thanks for the quick fix!

PR #1798, opened 2019-03-22
Rename 'MaybePromise' to 'PromiseOrValue'

IvanGoncharov:
Suggested by @martijnwalraven
`MaybePromise` name confuses users into thinking: `execute`/`graphql` returns either Promise or null/undefined.
Also in our codebase, we frequently use `[Mm]aybe*` to represent optional things,
e.g.:
https://github.com/graphql/graphql-js/blob/8c96dc8276f2de27b8af9ffbd71a4597d483523f/src/language/visitor.js#L353
https://github.com/graphql/graphql-js/blob/8c96dc8276f2de27b8af9ffbd71a4597d483523f/src/language/printer.js#L244

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -19,7 +19,7 @@ import memoize3 from '../jsutils/memoize3';
 import promiseForObject from '../jsutils/promiseForObject';
 import promiseReduce from '../jsutils/promiseReduce';
 import type { ObjMap } from '../jsutils/ObjMap';
-import type { MaybePromise } from '../jsutils/MaybePromise';
+import type { PromiseOrValue } from '../jsutils/PromiseOrValue';
 
 import { getOperationRootType } from '../utilities/getOperationRootType';
 import { typeFromAST } from '../utilities/typeFromAST';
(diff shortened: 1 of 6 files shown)


IvanGoncharov:
@Cito What do you think? Maybe you have a better naming proposal?

IvanGoncharov:
Note: not breaking change, since it's part of `jsutils`:
https://github.com/graphql/graphql-js/blob/master/src/jsutils/README.md
> These functions are not part of the module interface and are subject to change.

Cito:
The name MaybePromise is also used by [others](https://www.npmjs.com/package/maybe-promise), but @martijnwalraven has a point when he says Maybe already means "or null/undefined" in GraphQL.js elsewhere, and Flow also has the concept of [Maybe Types](https://flow.org/en/docs/types/maybe/) for optional values. PromiseOrValue sounds good to me. I found that name also in the typescript definition for [Promise.all()](https://definitelytyped.org/docs/when--when/modules/when.d.when.html#all). So renaming *maybe* (pun intended) makes sense.

IvanGoncharov:
> Maybe already means "or null/undefined" in GraphQL.js elsewhere, and Flow also has the concept of Maybe Types for optional values. PromiseOrValue sounds good to me. I found that name also in the typescript definition for Promise.all().

Thanks for research 👍 
I think naming is important for `graphql-js` as the reference implementation since readability is one of the top priorities.

Issue #1821, opened 2019-04-13
Should `parseLiteral` throw a TypeError when provided an invalid value? 


danielrearden:
When calling a `GraphQLScalarType`'s `parseLiteral` or `parseValue` methods, those methods can either return `undefined` or throw an error -- in either case, the value is considered invalid and an error is included in the response. However, returning `undefined` results in a generic error message about the type being invalid, while throwing a TypeError allows us to provide additional details as to why the provided value was invalid.

For all built-in scalars, `parseValue` throws a TypeError when the value is invalid, while `parseLiteral` always returns undefined. This results in inconsistent messaging when validating literal inputs compared to variables. For example:

> Expected type Int, found 2147483648.

compared with

> Variable \"$first\" got invalid value 2147483648; Expected type Int; Int cannot represent non 32-bit signed integer value: 2147483648

Anywhere `parseLiteral` is actually called, it's already wrapped with a try/catch and any error is handled appropriately -- namely [here](https://github.com/graphql/graphql-js/blob/8c96dc8276f2de27b8af9ffbd71a4597d483523f/src/validation/rules/ValuesOfCorrectType.js#L185-L207) and [here](https://github.com/graphql/graphql-js/blob/8c96dc8276f2de27b8af9ffbd71a4597d483523f/src/utilities/valueFromAST.js#L158-L162).

I know this is fairly trivial, but is there any reason to avoid throwing a TypeError inside `parseLiteral` when the value is invalid? Even if `parseLiteral` and `parseValue` can't share the exact same logic, it might make sense to reconcile them, at least in that regard. Would you accept a PR to that affect?

IvanGoncharov:
@danielrearden Thanks for bringing this up and especially for deep research 👍 

> I know this is fairly trivial, but is there any reason to avoid throwing a TypeError inside parseLiteral when the value is invalid?

As I understand, it's like that because initially, it was problematic to parse SDL with complex default values. So returning `undefined` was introduced as a temporary workaround but the current `valueFromAST` implementation already solves this problem and makes workaround unnecessary.

>  Would you accept a PR to that affect?

Yes, it would be great. One note though: `parseLiteral` is part of public API so technically it would be a breaking change so we can't merge it in `14.x` line. 
But we plan to start working on `15.0.0` next month and can merge PR at the beginning of this process.

IvanGoncharov:
This enhancement is tracked in #1827

Issue #500, opened 2016-09-22
Scalar's serialize, parseValue and parseLiteral


adriano-di-giovanni:
Please, correct me if I'm wrong

`serialize` gets invoked when we query data (i.e. we format a Date from MongoDB to ISO 8601)
`parseValue` gets invoked when we mutate data (i.e. we get a value from an `InputType` and try to coerce it to a Javascript value.

What about `parseLiteral`?

Thanks

helfer:
serialize: gets invoked when serializing the result to send it back to a client.

parseValue: gets invoked to parse client input that was passed through variables.

parseLiteral: gets invoked to parse client input that was passed inline in the query.

leebyron:
Correct.

parseValue gets a plain JS object, and parseLiteral gets a value AST.

adriano-di-giovanni:
Sorry if I'm re-opening this but I'm working again on scalars and there's something I can't understand event reading your code:

* should `serialize` always return a `String`value?
* should `parseValue` and `parseLiteral` always return the same value?

Thanks,
Adriano

amir-s:
@adriano-di-giovanni I know it is from long time ago. I came here by search, so I thought I might answer.

> should `serialize` always return a `String` value?

No. The response in GraphQL is JSON. So, you could return anything that can be represented as such. Number, String, Array or simply JS Object.

> should `parseValue` and `parseLiteral` always return the same value?
Yes, I believe. In a resolver function, we don't care whether the value of a type was inline or came from `variables` object. It should be the same. However since there are two ways of reading the input and the type of the input are different in those two ways, we need two methods for them.

pscanf:
> > should serialize always return a String value?
>
> No. The response in GraphQL is JSON. So, you could return anything that can be represented as such. Number, String, Array or simply JS Object.

Looking at the [`src/utilities/astFromValue.js` source code](https://github.com/graphql/graphql-js/blob/master/src/utilities/astFromValue.js#L131-L175) it seems like `serialize`  can only return nullish values, booleans, numbers or strings.

This constraint doesn't seem to be enforced [during serialization](https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L314-L318) though, so problems occur only when `astFromValue` is invoked, ie during introspection.

Reading around the spec and the source code I wasn't able to figure out what's the intended behaviour though. Should scalars ultimately resolve to either ~nulls, booleans, numbers or strings?

shtse8:
[...]

PR #1827, opened 2019-04-25
Update built-in scalar's parseLiteral method to throw TypeError

danielrearden:
This addresses #1821

### DIFF
--- a/src/type/scalars.js
+++ b/src/type/scalars.js
@@ -7,6 +7,7 @@ import inspect from '../jsutils/inspect';
 import isObjectLike from '../jsutils/isObjectLike';
 
 import { Kind } from '../language/kinds';
+import { print } from '../language/printer';
 
 import { GraphQLScalarType, isScalarType } from './definition';
 
@@ -62,13 +63,18 @@ export const GraphQLInt = new GraphQLScalarType({
   serialize: serializeInt,
   parseValue: coerceInt,
   parseLiteral(ast) {
-    if (ast.kind === Kind.INT) {
-      const num = parseInt(ast.value, 10);
-      if (num <= MAX_INT && num >= MIN_INT) {
-        return num;
-      }
+    if (ast.kind !== Kind.INT) {
+      throw new TypeError(
+        'Int cannot represent non-integer value: ' + print(ast),
+      );
     }
-    return undefined;
+    const num = parseInt(ast.value, 10);
+    if (num > MAX_INT || num < MIN_INT) {
+      throw new TypeError(
+        'Int cannot represent non 32-bit signed integer value: ' + ast.value,
+      );
+    }
+    return num;
   },
 });
 
@@ -105,9 +111,12 @@ export const GraphQLFloat = new GraphQLScalarType({
   serialize: serializeFloat,
   parseValue: coerceFloat,
   parseLiteral(ast) {
-    return ast.kind === Kind.FLOAT || ast.kind === Kind.INT
-      ? parseFloat(ast.value)
-      : undefined;
+    if (ast.kind !== Kind.FLOAT && ast.kind !== Kind.INT) {
+      throw new TypeError(
+        'Float cannot represent non numeric value: ' + print(ast),
+      );
+    }
+    return parseFloat(ast.value);
   },
 });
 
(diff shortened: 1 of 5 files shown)


src/type/scalars.js
```diff
@@ -66,13 +66,19 @@ export const GraphQLInt = new GraphQLScalarType({
   serialize: serializeInt,
   parseValue: coerceInt,
   parseLiteral(ast) {
-    if (ast.kind === Kind.INT) {
-      const num = parseInt(ast.value, 10);
-      if (num <= MAX_INT && num >= MIN_INT) {
-        return num;
-      }
+    const value = ast.value || '';
```
IvanGoncharov: @danielrearden It means you can get `Int cannot represent non-integer value: ""` for some node types (e.g. null nodes) which would be very confusing.
It's better to write it like this:
```js
  parseLiteral(ast) {
    if (ast.kind !== Kind.INT) {
      throw new TypeError(
        'Int cannot represent non-integer value' +
          (ast.value ? ': ' + inspect(ast.value) : '.'),
      );
    }
    const num = parseInt(ast.value, 10);
    // ...
  },
```
Can you please change it here and in other places?
Also please write add tests for such cases (AST nodes without values) so we would maintain same test coverage:
https://codecov.io/gh/graphql/graphql-js/commit/92d58e750257cabb84361fb64e9b6306ba91e57c#D2-69

IvanGoncharov:
Note: enum class also has `parseLiteral` that needs to be fixed:
https://github.com/graphql/graphql-js/blob/dfcdce789aaf7ba599a5b6e6c6de25f1c8cda934/src/type/definition.js#L1200-L1205
It can be done in this or in separate PR.

src/type/scalars.js
danielrearden: @IvanGoncharov Sorry for the late response. Finally had a bit of time to circle back around to this PR. I'll make the necessary changes, but after taking another pass at it, I'm wondering if we could be better around reducing the code duplication between the two `parse` methods.

I'm thinking we could convert the `ValueNode` that's passed into `parseLiteral` into a regular JS value and then pass that to `parseValue`. That way there's a single function responsible for input validation. There's already a `valueFromASTUntyped` utility function available, so it could be as simple as:

```
parseLiteral: (ast, variables) => coerceString(valueFromASTUntyped(ast, variables))
```

The main gotcha I can think of is that `1.0` would get coerced into `1` so we'd have to still explicitly throw on float values for `GraphQLInt` and `GraphQLID`.

IvanGoncharov: @danielrearden Let's keep this PR around replacing `undefined` with `throw`.
On a topic why we need both `parseLiteral` and `parseValue` as separate function see this comment: https://github.com/graphql/graphql-js/issues/500#issuecomment-478193989

danielrearden:
@IvanGoncharov This PR should be good to go now. 🤞 I can fix `GraphQLEnum` too (maybe as a separate PR) but I'm lacking a bit of context as to where these methods are actually used. Unlike `serialize`, as far as I can tell, we only use `parseLiteral`/`parseValue` with scalars and never actually call these methods on enum types?

src/type/scalars.js
```diff
@@ -66,13 +66,20 @@ export const GraphQLInt = new GraphQLScalarType({
...
   parseLiteral(ast) {
-    if (ast.kind === Kind.INT) {
-      const num = parseInt(ast.value, 10);
-      if (num <= MAX_INT && num >= MIN_INT) {
-        return num;
-      }
+    if (ast.kind !== Kind.INT) {
+      throw new TypeError(
+        `Int cannot represent non-integer value${
+          ast.value === undefined ? '.' : ': ' + inspect(ast.value)
```
IvanGoncharov: since this expression is at the end of the string, I think simple string concatenation with `+` would be more readable.

danielrearden: Fixed

IvanGoncharov: Thanks

IvanGoncharov:
@danielrearden Sorry for the long delay.
We finally started working on `15.0.0` so I merged this breaking change.

PR #1781, opened 2019-03-08
Add directive predicates tests and fix test names

Cito:
The predicates for directives were not yet tested.

PR #1837, opened 2019-04-29
isSpecifiedDirective should not assume Directive type

Cito:
isSpecifiedDirective should not just check the name property, but also the type

### DIFF
--- a/src/type/directives.js
+++ b/src/type/directives.js
@@ -187,10 +187,13 @@ export const specifiedDirectives: $ReadOnlyArray<*> = [
   GraphQLDeprecatedDirective,
 ];
 
-export function isSpecifiedDirective(
-  directive: GraphQLDirective,
-): boolean %checks {
-  return specifiedDirectives.some(
-    specifiedDirective => specifiedDirective.name === directive.name,
+export function isSpecifiedDirective(directive: mixed): boolean %checks {
+  return (
+    isDirective(directive) &&
+    // Would prefer to use specifiedDirectives.some(), however %checks needs
+    // a simple expression.
+    (directive.name === GraphQLIncludeDirective.name ||
+      directive.name === GraphQLSkipDirective.name ||
+      directive.name === GraphQLDeprecatedDirective.name)
   );
 }
--- a/src/type/__tests__/predicate-test.js
+++ b/src/type/__tests__/predicate-test.js
@@ -12,6 +12,11 @@ import { expect } from 'chai';
 
 import {
   GraphQLScalarType,
+  GraphQLBoolean,
+  GraphQLID,
+  GraphQLInt,
+  GraphQLFloat,
+  GraphQLString,
   GraphQLEnumType,
   GraphQLInputObjectType,
   GraphQLInterfaceType,
@@ -19,13 +24,13 @@ import {
   GraphQLUnionType,
   GraphQLList,
   GraphQLNonNull,
-  GraphQLString,
   GraphQLDirective,
   GraphQLIncludeDirective,
   GraphQLSkipDirective,
   GraphQLDeprecatedDirective,
   isType,
   isScalarType,
+  isSpecifiedScalarType,
   isObjectType,
   isInterfaceType,
   isUnionType,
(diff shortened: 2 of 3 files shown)


IvanGoncharov:
@Cito Thanks for PR 👍 
I just trying to understand what's the problem with the current check.
Idea is to have non-overlapping checks so if you have value and want to enusure it's a directive you would use:
https://github.com/graphql/graphql-js/blob/ce832eef31ee136140757f397c2478e696d10fcb/src/type/directives.js#L32-L34

Also, note that `isSpecifiedDirective` take `directive` of type `GraphQLDirective`
https://github.com/graphql/graphql-js/blob/ce832eef31ee136140757f397c2478e696d10fcb/src/type/directives.js#L190-L192

So if you need to test if some value is standard directive you would do:
```js
if (isDirective(value) && isSpecifiedDirective(value)) {
  // ...
}
```
What do you think?

Cito:
> I just trying to understand what's the problem with the current check.

The problem was that it was unclear to me that this should be considered a non-overlapping, additional check, and not as a standalone check. To me a standalone check would be more intuitive. I was also looking at these [four tests](https://github.com/graphql/graphql-js/blob/ce832eef31ee136140757f397c2478e696d10fcb/src/type/__tests__/predicate-test.js#L642) which run the predicate against various garbage. These tests should be removed as superfluous and misleading if the input is supposed to be a directive anyway. Otherwise, a test with `{name: 'skip'}` as garbage should be added as well.

I just had a look at `isSpecifiedScalarType` for comparison, but that is even more confusing, since it makes an `isNamedType` check internally. When it's supposed to be a non-overlapping check it should assume that it's a ScalarType already, right (and otherwise, it should check more speficifally for ScalarType, not NamedType).

IvanGoncharov:
> I was also looking at these four tests which run the predicate against various garbage.

You added them in #1781 😄 

> Just had a look at isSpecifiedScalarType for comparison, which is even more unclear, since it makes an isNamedType check internally.

`isIntrospectionType` also has it 😞 

```js
export function isIntrospectionType(type: mixed): boolean %checks {
  return (
    isNamedType(type) &&
```

> The problem was that it was unclear to me that this should be considered a non-overlapping, additional check, and not as a standalone check. 

My idea was that type checks (e.g. `isInputType`, `isScalarType`, etc.) should be standalone but entity checks (e.g. `isSpecifiedDirective`) should be guarded by type checks.

But since `isIntrospectionType` and `isSpecifiedScalarType` have different symantics it makes sense to change `isSpecifiedDirective`.

> since it makes an isNamedType check internally.

Can you please change it to `isScalarType` and add tests.

src/type/directives.js
```diff
@@ -190,7 +190,10 @@ export const specifiedDirectives: $ReadOnlyArray<*> = [
 export function isSpecifiedDirective(
   directive: GraphQLDirective,
```
IvanGoncharov: It should be `mixed`.

Cito:
> You added them in #1781 😄

Good that git always reminds us of our past sins 😂

I've pushed a new commit that hopefully is better. Both predicates now work as standalone checks.

Cito:
Btw, there is another issue here that I don't quite understand.

These checks could all be simplified by using `specifiedTypes.some(...)`, but a comment says this is not possible because " %checks needs a simple expression", so I did not make that simplification.

But what exactly would break here when using `some()`? It compiles fine, and all tests pass. Maybe it would not be usable as a type guard? But I could not come up with a working example, e.g. this one does not work anyway, for example:

    function example(d: GraphQLDirective) {
      if (isSpecifiedDirective(d)) {
        const name: 'include' | 'skip' | 'deprecated' = d.name;  // flow error
      }
    }

IvanGoncharov:
@Cito Merged 🎉 Thanks for PR 👍 

> But what exactly would break here when using some()? It compiles fine, and all tests pass. Maybe it would not be usable as a type guard? 

By default flow assumes that function arguments are mutable so it can be sure that `type` argument wasn't changed before `some` callback called.
But since we switched to `@flow strict` that ensure arguments are always const Flow is finally happy.
As an experiment, you can change `@flow strict` => `@flow` and see error messages.
Fixed #1848, thanks for investigating this 👍

Cito:
Thanks for the explanation. Normally you would expect to see *more* errors when adding "strict", but you're right, in this case the error disappears because with "strict" flow can assume function parameters are const. Setting "strict" seems to pay off, good decision.  👍

PR #1850, opened 2019-05-03
Add comment to make test better understandable

Cito:
At first I didn't understand why Boolean was not tested,
therefore added a small comment with an explanation.

(Actually Boolean is also included via the standard directives).


IvanGoncharov:
@Cito Thanks 👍

PR #1853, opened 2019-05-07
Enables graphql-js to ESModules without .mjs

deinok:


IvanGoncharov:
@deinok I can't find this field in official `package.json` docs:
https://docs.npmjs.com/files/package.json
Can you please provide a reference to something that explains this `type` field?

deinok:
Sure, Its pretty new.
http://2ality.com/2019/04/nodejs-esm-impl.html#filename-extensions

IvanGoncharov:
@deinok Thanks for the link 👍 
Problem is that we need to support older versions of Node 8+ and even for Node 12 we can't depend on `--experimental-modules` flag being enabled.
So we still need to ship two versions of the same file: CJS (with `require`) and MJS (with `import`).

PR #1872, opened 2019-05-19
Replace all "var" with "const" & "let" in "resources/*.js" files

IvanGoncharov:


BrunoScheufler:
This looks promising, is refactoring variables an ongoing effort or will this change be the last one? I'd be interested in what's planned in terms of refactoring 😄

IvanGoncharov:
@BrunoScheufler It's just small refactoring of "internal" scripts. Main code base in "src" folder already using modern ES6 for years :)

PR #1878, opened 2019-05-20
Add "version" string and parsed "versionInfo"

### LINKED ISSUE #1726: Export library version number
Developers shouldn't write their own hacks to detect library version:
https://github.com/graphql-compose/graphql-compose/blob/0ac3cc16948e7dce098705ff1163801d1f284a0a/src/utils/graphqlVersion.js

IvanGoncharov:
Fixes #1726


IvanGoncharov:
Since `14.0.0` was advertised as a stable release but we plan to release `15.0.0` with some breaking changes. I think some 3rd-party libs/tools will be forced to support both `14.x.x` and `15.x.x` in the same code base.
For example, `graphql-compose` supports `0.13.0` and `14.0.0` using ugly version detection hack: https://github.com/graphql-compose/graphql-compose/blob/0ac3cc16948e7dce098705ff1163801d1f284a0a/src/utils/graphqlVersion.js

@mjmahone @Neitsch @nodkz Can you please take a look?

resources/build.js
```diff
@@ -26,14 +27,21 @@ if (require.main === module) {
...
     if (filepath.endsWith('.js')) {
       buildJSFile(filepath);
     }
   }
+
+  const packageJSON = buildPackageJSON();
+  assert(
+    packageJSON.version === require('../dist/version').version,
+    'Version in package.json and version.js should match',
+  );
```
IvanGoncharov: This check guarantees that `package.json` and `version.js` are always in sync.

resources/build.js
```diff
@@ -52,13 +60,14 @@ function buildPackageJSON() {
...
 
-  const { version } = packageJSON;
-  const match = /^[0-9]+\.[0-9]+\.[0-9]+-?([^.]*)/.exec(version);
-  assert(match, 'Version does not match semver spec.');
-  const tag = match[1];
-  assert(!tag || tag === 'rc', 'Only "rc" tag is supported.');
+  const { preReleaseTag } = parseSemver(packageJSON.version);
+  if (preReleaseTag != null) {
+    const [tag] = preReleaseTag.split('.');
+    assert(tag === 'rc', 'Only "rc" tag is supported.');
```
martijnwalraven: I see the code had this check before, but I'm just curious why only `rc` is supported?

martijnwalraven:
This looks great! Really happy it pre-parses the version components so we don't have to do that at runtime.

nodkz:
Looks great! Thanks a lot!

resources/build.js
IvanGoncharov: @martijnwalraven Thanks for the review 👍 
It was added to prevent issues similar to https://github.com/graphql/graphql-js/issues/1375
These scripts are run by Travis and package automatically published so I'm trying to use "fail fast" design for them. Especially since I don't have NPM keys to do a manual fix if something goes wrong.
If we decide to use other tags in future it will be a matter of just adding them here.

IvanGoncharov:
Thanks for review 👍
Merged 🎉

PR #1948, opened 2019-06-04
add mergeAST from GraphiQL, refactor using visit() [WIP]

acao:
Accomplishes https://github.com/graphql/graphiql/issues/836, originally introduced via https://github.com/graphql/graphiql/pull/762 earlier this year, `mergeAST` is a handy utility for in-lining fragments for complex queries.

As per @IvanGoncharov's suggestion, I refactored it to use `language/visitor`'s `visit` function.  Good thing I'd been working with AST and visitors on another project using antlr4 earlier this year, or I would've been so lost here, haha. That said, GraphQL's `visit()` is so powerful, I'm blown away at how simple this was.

Thank you @RamonSaboya for the inspiration, and some solid unit tests, and an excellent idea all around.

I'm wondering if a `Set` would be a more appropriate data structure here?

**This is a WIP**, just learning about `visit()` and this is just a learning excercise for me. 
Not a huge priority to any part of the ecosystem currently, and a lot of important edge cases are still broken here.

### DIFF
--- a/src/jsutils/uniqueBy.js
+++ b/src/jsutils/uniqueBy.js
@@ -1,0 +1,31 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+/**
+ * Returns an array of unique values based on iteratee
+ * which is invoked for each element in array to generate
+ * the criterion by which uniqueness is computed.
+ *
+ * Simiar to _.uniqBy from lodash.
+ */
+export function uniqueBy(
+  array: $ReadOnlyArray<any>,
+  iteratee: (item: any) => any,
+) {
+  const FilteredMap = new Map();
+  const result = [];
+  for (const item of array) {
+    const uniqeValue = iteratee(item);
+    if (!FilteredMap.has(uniqeValue)) {
+      FilteredMap.set(uniqeValue, true);
+      result.push(item);
+    }
+  }
+  return result;
+}
--- a/src/utilities/inlineNamedFragments.js
+++ b/src/utilities/inlineNamedFragments.js
@@ -1,0 +1,52 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import { type ObjMap } from '../jsutils/ObjMap';
+import { uniqueBy } from '../jsutils/uniqueBy';
+import { visit } from '../language/visitor';
+import {
+  type DocumentNode,
+  type FragmentDefinitionNode,
+} from '../language/ast';
+
+/**
+ * Given a document AST, inline all named fragment definitions
+ */
+export function inlineNamedFragments(documentAST: DocumentNode): DocumentNode {
+  const fragmentDefinitions: ObjMap<FragmentDefinitionNode> = Object.create(
+    null,
+  );
+
+  for (const definition of documentAST.definitions) {
+    if (definition.kind === 'FragmentDefinition') {
+      fragmentDefinitions[definition.name.value] = definition;
+    }
+  }
+
+  return visit(documentAST, {
+    FragmentSpread(node) {
+      return {
+        ...fragmentDefinitions[node.name.value],
+        kind: 'InlineFragment',
+      };
+    },
+    SelectionSet(node) {
+      return {
+        ...node,
+        selections: uniqueBy(
+          node.selections,
+          selection => selection.name.value,
+        ),
+      };
+    },
+    FragmentDefinition() {
+      return null;
+    },
+  });
+}
--- a/src/index.js
+++ b/src/index.js
@@ -397,6 +397,8 @@ export {
   isValidLiteralValue,
   // Concatenates multiple AST together.
   concatAST,
+  // Inline named fragments from AST
+  inlineNamedFragments,
   // Separates an AST into an AST per Operation.
   separateOperations,
   // Strips characters that are not significant to the validity or execution
--- a/src/utilities/__tests__/inlineNamedFragments-fixture.js
+++ b/src/utilities/__tests__/inlineNamedFragments-fixture.js
@@ -1,0 +1,189 @@
+/**
+ *  Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ *
+ */
+
+export const fixtures = [
+  {
+    desc: 'does not modify query with no fragments',
+    query: `
+      {
+        id
+      }`,
+    resultQuery: `
+      {
+        id
+      }
+      `,
+  },
+  {
+    desc: 'inlines simple nested fragment',
+    query: `
+      {
+        ...Fragment1
+      }
+
+      fragment Fragment1 on Test {
+        id
+      }`,
+    resultQuery: `
+      {
+        ... on Test {
+          id
+        }
+      }
+      `,
+  },
+  {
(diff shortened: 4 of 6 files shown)


acao:
I found a failing case that wasn't in the original unit tests:
```graphql
      query Test1 {
        ...Fragment4
        ...Fragment5
      }
  
      fragment Fragment5 on Test1 {
        ...Fragment4
      }

      fragment Fragment4 on Test1 {
        id
      }
```
does not give me

```graphql
 query Test1 {
        ...on Test1 {
          id
        }
        ...on Test1 {
          ...on Test1 {
            id
          }
        }
      }
```
instead, I get:
```graphql
 query Test1 {
        ...on Test1 {
          id
        }
        ...on Test1 {
          id
        }
      }
```
this works in the original! hmmm 🤔

src/utilities/mergeAST.js
```diff
@@ -0,0 +1,55 @@
...
+export function mergeAST(ast: DocumentNode): DocumentNode {
+  const fragmentDefinitions: { [name: string]: FragmentDefinitionNode } = {};
+  const inlinedFragments: { [name: string]: FragmentSpreadNode } = {};
+
+  visit(ast, {
+    FragmentDefinition(node) {
+      // Collect the existing FragmentDefinition nodes.
+      fragmentDefinitions[node.name.value] = node;
+    },
+  });
```
IvanGoncharov: You don't need `visit` here, simple `for of` over `definitions` would be enough.

src/utilities/__tests__/mergeAST-test.js
```diff
@@ -0,0 +1,27 @@
...
+import { mergeAST } from '../mergeAST';
+
+import { fixtures } from './mergeAST-fixture';
+
+describe('MergeAST', () => {
+  fixtures.forEach(fixture => {
+    it(fixture.desc, () => {
+      const result = print(mergeAST(parse(fixture.query))).replace(/\s/g, '');
+      expect(result).to.equal(fixture.mergedQuery.replace(/\s/g, ''));
+    });
```
IvanGoncharov: We don't use fixtures please unroll it here.
Use `dedent` to remove indentation from strings.
And defined `expectMerged` wrapper to not repeat `print(mergeAST(parse`

src/utilities/__tests__/mergeAST-test.js
```diff
@@ -0,0 +1,27 @@
...
+ *  LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ *
+ */
+
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { parse, print } from '../../index';
```
IvanGoncharov: Use direct imports from appropriate files.

src/utilities/mergeAST.js
```diff
@@ -0,0 +1,55 @@
...
+} from '../language/ast';
+
+/**
+ * Given a document AST, merge all fragment definitions into the
+ * provided operations using inline fragments.
+ */
+
+export function mergeAST(ast: DocumentNode): DocumentNode {
+  const fragmentDefinitions: { [name: string]: FragmentDefinitionNode } = {};
+  const inlinedFragments: { [name: string]: FragmentSpreadNode } = {};
```
IvanGoncharov: Please use `Object.create(null)` here and use `ObjMap` for types.

src/utilities/mergeAST.js
```diff
@@ -0,0 +1,55 @@
...
+  type DocumentNode,
+  type FragmentDefinitionNode,
+  type FragmentSpreadNode,
+} from '../language/ast';
+
+/**
+ * Given a document AST, merge all fragment definitions into the
+ * provided operations using inline fragments.
+ */
+
```
IvanGoncharov: Remove empty line

src/utilities/mergeAST.js
```diff
@@ -0,0 +1,55 @@
...
+    FragmentDefinition(node) {
+      // Collect the existing FragmentDefinition nodes.
+      fragmentDefinitions[node.name.value] = node;
+    },
+  });
+
+  return visit(ast, {
+    FragmentSpread(node) {
+      // No repeats!
+      if (inlinedFragments[node.name.value]) {
```
IvanGoncharov: This is incorrect check, same fragment can be used in multiple places
```graphql
{
  person: {
    ...ProfileInfo
    friend: {
      ...ProfileInfo
    }
  }
}
```
Also please cover in tests.

IvanGoncharov:
> I found a failing case that wasn't in the original unit tests:

I think I know the reason, I will fix this.

src/utilities/mergeAST.js
acao: covered in unit tests now :)

src/jsutils/uniqueBy.js
```diff
@@ -0,0 +1,31 @@
...
+ * which is invoked for each element in array to generate
+ * the criterion by which uniqueness is computed.
+ *
+ * Simiar to _.uniqBy from lodash.
+ */
+export function uniqueBy(
+  array: $ReadOnlyArray<any>,
+  iteratee: (item: any) => any,
+) {
+  const FilteredMap = new Map();
```
langpavel: Why not `Set`?

acao:
@langpavel good questions!

src/jsutils/uniqueBy.js
acao: `Set` makes a lot more sense

src/utilities/inlineNamedFragments.js
```diff
@@ -0,0 +1,52 @@
...
+      fragmentDefinitions[definition.name.value] = definition;
+    }
+  }
+
+  return visit(documentAST, {
+    FragmentSpread(node) {
+      return {
+        ...fragmentDefinitions[node.name.value],
+        kind: 'InlineFragment',
+      };
```
IvanGoncharov: Please review type definitions before you use them or convert them into each other.
Fragment definition has way more fields than inline fragments.

src/utilities/inlineNamedFragments.js
```diff
@@ -0,0 +1,52 @@
...
+    },
+    SelectionSet(node) {
+      return {
+        ...node,
+        selections: uniqueBy(
+          node.selections,
+          selection => selection.name.value,
```
IvanGoncharov: You also removing duplicated fields, including fields with different aliases, args, etc.

src/jsutils/uniqueBy.js
IvanGoncharov: `Map`, `Set` should be used only for non-string keys, `Object.create(null)` is perfectly fine for working with string keys.

src/utilities/inlineNamedFragments.js
acao: for sure, adding more test cases, it can't be this simple

src/utilities/__tests__/mergeAST-test.js
IvanGoncharov: Please write it like this:
```js
function expectInlined(query) {
  return expect(print(inlineNamedFragments(parse(query))));
}

describe('inlineNamedFragments', () => {
  it('does not modify query with no fragments', () => {
    expectInlined(dedent`
      {
        id
      }
    `).to.equal(dedent`
      {
        id
      }
    `);
  });
```

acao:
disabling until I've done more research

src/jsutils/uniqueBy.js
langpavel: @IvanGoncharov `Set` is much better *primitive* for checking presence/uniqueness than `Object.create(null)`, both by math sense and by the fact — `Map`s and `Set`s are well tested and perfectly tuned for purpose. Misusing objects as hash maps, well, it's only relique. Polyfilling is not needed today.

benjie:
I have a more thorough implementation of this here: https://github.com/graphql/graphiql/pull/1542

Let me know if you're interested into a PR to GraphQL.js

IvanGoncharov:
@benjie I think it would make sense to test it in GraphiQL to refine API and implementation and latter move to `graphql-js`.
Also, the relay compiler supports this and several similar transformations so we can learn from them.

PR #1949, opened 2019-06-04
[printError] Make location formatting IDE friendly

alloy:
This is the format used by jest, flow, typescript, and many others.

As an example, this change [when used in relay](https://github.com/facebook/relay/pull/2752) changes the experience like so:

### Before

relay-ide-friendly-errors-before

### After

relay-ide-friendly-errors-after

PR #1985, opened 2019-06-15
Print minified location

IvanGoncharov:
Context: https://github.com/graphql/graphql-js/pull/1786

### DIFF
--- a/src/language/__tests__/printLocation-test.js
+++ b/src/language/__tests__/printLocation-test.js
@@ -7,7 +7,48 @@ import dedent from '../../jsutils/dedent';
 import { Source } from '../../language';
 import { printSourceLocation } from '../printLocation';
 
-describe('printLocation', () => {
+describe('printSourceLocation', () => {
+  it('prints minified documents', () => {
+    const minifiedSource = new Source(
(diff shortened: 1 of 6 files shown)


src/language/__tests__/printLocation-test.js
```diff
@@ -7,7 +7,48 @@ import dedent from '../../jsutils/dedent';
...
+
+    const firstLocation = printSourceLocation(minifiedSource, {
+      line: 1,
+      column: minifiedSource.body.indexOf('FIRST_ERROR_HERE') + 1,
+    });
+    expect(firstLocation + '\n').to.equal(dedent`
+      GraphQL request:1:53
+      1 | query SomeMiniFiedQueryWithErrorInside($foo:String!=FIRST_ERROR_HERE$bar:String)
+        |                                                     ^
+        | {someField(foo:$foo bar:$bar baz:SECOND_ERROR_HERE){fieldA fieldB{fieldC fieldD.
```
IvanGoncharov: > Queries, particularly when compressed, can have a long line length. In narrow viewports (small Terminal windows, etc.) this results in the query wrapping over multiple lines.

@jaydenseric I think I finally figured out a generic solution to the problem that you described in #1786. This test (+ 2 test cases below) shows the new format handles error location in the beggining, in the middle and in the end of the minified document.
What do you think?

IvanGoncharov:
@alloy Since we already changed error format in #1949 and assume some of GraphQL tools/libraries will probably need to update their test suites, I decided to also address this issue to batch all changes in a single release.
Can you please review this PR and check if this change will be compatible with Relay?

alloy:
@IvanGoncharov Follow-up is at https://github.com/facebook/relay/pull/2784 and works well 👍

Issue #1858, opened 2019-05-09
Add utility function function for generating fully-featured schema from SDL


danielrearden:
With SDL [officially](https://github.com/graphql/graphql-spec/pull/90) part of the specification, would it be worthwhile to provide a way to build a build a fully-featured schema  from a Document AST?

Currently, schemas built using `buildASTSchema`/`buildSchema` lack several features that are otherwise available if constructed programmatically:

* Specifying `resolve` functions for fields
* Specifying `resolveType` functions for unions and interfaces and `isTypeOf` functions for types
* Specifying enum value values
* Adding custom scalars

Workarounds exist for some of these limitations (functions can be passed through the root object for root type fields, abstract types can be handled by explicitly providing a `__typename` property, etc.). In practice, though, if I'm starting a new project and want to use SDL to define my schema, I'm forced to use `graphql-tools` or another third-party package. This was completely understandable prior to the 2018 version of the spec, but now it may be appropriate for this functionality to be part of the core library.

IvanGoncharov:
@danielrearden It's one of the goals of `buildSchema`/`extendSchema` refactoring I'm working in the background. I still like that simplicity of `buildSchema` so ATM my long term plan is to create a better version of `transformSchema` from #1199.
And based on it create a high-level function to attach resolvers to any schema (produced by `buildSchema`, `extendSchema`, `buildClientSchema`, `new GraphQLSchema`, etc.).

BTW, You can use SDL together with ES6 classes so you don't need to attach resolvers to schema: https://github.com/IvanGoncharov/swapi-demo/blob/master/src/index.ts
That said I still think it extremely important to support the pattern of attaching static resolvers since it's the widest spread pattern in GraphQL JS ecosystem.

IvanGoncharov:
Here is a great article on the topic: https://medium.com/@samerbuna/the-api-i-wish-javascript-graphql-implementations-supported-91eecfb96106 and accompanying library https://github.com/jscomplete/graphql-makers

yaacovCR:
Related: #597

PR #1987, opened 2019-06-21
`buildSchema` should be able to create fully featured schema

langpavel:
### DONE: :tada: 

- [x] can lookup for type field resolvers https://github.com/graphql/graphql-js/commit/9157529987f7f33ff7a1b0aedecace4e19e677dc
- [x] `enum` value lookup (exactly like previous) https://github.com/graphql/graphql-js/commit/1b0f68a8ba6b87e6b38a70424f63a7668ce864e7
- [x] tests https://github.com/graphql/graphql-js/pull/1987/commits/84adacf05643e11e46e67335b7e3369977a5e476 (object field resolver, enum values)
- [x] custom scalars + test https://github.com/graphql/graphql-js/pull/1987/commits/396098e7d9608f40d3f2c2f0e0e7fa341f62fcc4

### TO DO, need help:

- `isTypeOf` on ObjectType (should I follow `graphql-tools` with `__isTypeOf`?)
- `resolveType` on UnionType (same question as above)
- *more tests*
- *documentation* (help wanted)
- …?

Related: #1384 #1858

/cc @IvanGoncharov, @leebyron

### DIFF
--- a/src/utilities/buildASTSchema.js
+++ b/src/utilities/buildASTSchema.js
@@ -46,6 +46,7 @@ import {
   type GraphQLArgumentConfig,
   type GraphQLEnumValueConfig,
   type GraphQLInputFieldConfig,
+  type GraphQLFieldResolver,
   GraphQLScalarType,
   GraphQLObjectType,
   GraphQLInterfaceType,
@@ -72,6 +73,10 @@ import {
   GraphQLSchema,
 } from '../type/schema';
 
+export type TypeFieldResolverMap = ObjMap<
+  ObjMap<GraphQLFieldResolver<mixed, mixed, mixed>>,
+>;
+
 export type BuildSchemaOptions = {
   ...GraphQLSchemaValidationOptions,
 
@@ -91,6 +96,13 @@ export type BuildSchemaOptions = {
    * Default: false
    */
   assumeValidSDL?: boolean,
+
+  /**
+   * Object map of object maps to resolver funtions.
+   *
+   * Default: undefined
+   */
+  resolvers?: TypeFieldResolverMap,
 
   ...
 };
@@ -243,7 +255,18 @@ export class ASTDefinitionBuilder {
     });
   }
 
-  buildField(field: FieldDefinitionNode): GraphQLFieldConfig<mixed, mixed> {
+  buildField(
+    field: FieldDefinitionNode,
+    typeName?: string,
+  ): GraphQLFieldConfig<mixed, mixed> {
+    const resolve =
+      (typeName &&
+        this._options &&
+        this._options.resolvers &&
+        this._options.resolvers[typeName] &&
+        this._options.resolvers[typeName][field.name.value]) ||
+      undefined;
+
     return {
       // Note: While this could make assertions to get the correctly typed
       // value, that would throw immediately while type system validation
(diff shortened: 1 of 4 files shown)


src/utilities/buildASTSchema.js
```diff
@@ -72,6 +73,11 @@ import {
   GraphQLSchema,
 } from '../type/schema';
 
+export type TypeFieldResolverMap = ObjMap<
+  | ObjMap<GraphQLFieldResolver<mixed, mixed, mixed>> /* type and interface */
+  | ObjMap<any> /* enum */,
```
langpavel: I don't like this much because `any` for enum will eat resolvers for object types.
Should I wrap enum values to object with key one key `values` or `enum`?
It may works better for flow and TS checking but feels weird :grimacing: @IvanGoncharov what do you think?

IvanGoncharov:
@langpavel Thanks, for PR 👍 
It's definitely a problem we need to address.
But ATM I'm focused on fixing last issues in `14.x.x` and start working on `15.x.x`.

Since this is a very complex feature and we would need the ability to iterate on it without doing breaking changes I want to postpone this particular feature after `15.x.x` and include it in experimental `16.0.0-alpha.1` line.

Meanwhile, I need to think about the general API design and how to handle a thing like this:
```graphql
scalar CustomScalarWithValidationRule
directive @SomeDirective(arg: CustomScalarWithValidationRule) on FIELD_DEFINITION

type Query {
  foo: String @SomeDirective(arg: "<Invalid value for custom scalar>")
  bar(arg: CustomScalarWithValidationRule = "<Invalid value for custom scalar>")
}
```
We need ability to provide `parseLiteral` for `CustomScalarWithValidationRule` before SDL validation started.

langpavel:
Hi @IvanGoncharov 

> It's definitely a problem we need to address.

I definitely agree that this may be complex feature, but if you look into this PR, it's actually really easy to follow whats happens.
Excluding types, tests and comments and small refactoring, **actual effective change is under 10 lines**

> But ATM I'm focused on fixing last issues in `14.x.x` and start working on `15.x.x`.

Ok, this ones: [`v14.x.x`](https://github.com/graphql/graphql-js/milestones/v14.x.x) and [`v15.0.0`](https://github.com/graphql/graphql-js/milestones/v15.0.0)

> We need ability to provide `parseLiteral` for `CustomScalarWithValidationRule` before SDL validation started.

This is already partially covered here: https://github.com/graphql/graphql-js/pull/1987/commits/396098e7d9608f40d3f2c2f0e0e7fa341f62fcc4
You can pass scalars (or scalars config, interface is same) with `resolvers`, so you can access every scalar value like this: 
```js
const scalarConfig = config.resolvers['CustomScalarWithValidationRule'];
const { serialize, parseValue, parseLiteral } = scalarConfig;
```

Because config can be passed to validation, you can use actual scalar implementation to test values.

Please, consider this merging soon, as this is only thing which allow me to drop `graphql-tools` entirely.
I don't like `graphql-tools` because they mutate actual schema.

yaacovCR:
@langpavel in terms of comment above about graphql tools mutating schema, graphql-tools-fork gives you a cloneSchema method that allows you to avoid that if/when desired.

https://www.npmjs.com/package/graphql-tools-fork

yaacovCR:
Related: https://github.com/ardatan/graphql-tools/pull/1456 changes graphql-tools to no longer modify original schema in place and instead return a new rewired schema.

yaacovCR:
Thank you so much for this PR pointing the way.

Almost 7 years later, #4803 now implements this on top of 17.x.x ...

PR #1988, opened 2019-06-23
Add test for didYouMean with sub-message

Cito:


IvanGoncharov:
@Cito Thanks a lot 👍 
Merged 🎉

PR #2009, opened 2019-07-03
Defensively verify that Symbol.for is available

### LINKED ISSUE #2007: undefined is not a function in nodejsCustomInspectSymbol
is it feasible to add another check, `typeof Symbol.for === 'function'` to nodejsCustomInspectSymbol?

we get some cases of `TypeError: undefined is not a function` originating from https://github.com/graphql/graphql-js/blob/master/src/jsutils/nodejsCustomInspectSymbol.js#L5

in other words: there are cases were `Symbol` is a `function`, but `Symbol.for` is `undefined` :~(

unfortunately i have no idea how this can happen and no idea how to reproduce. we use core-js and no other polyfills.

interestingly, the bug only occurs exclusively in a few user agents:

- Chrome 74.0.3729.169 
- Chrome 74.0.3729.157
- Samsung Browser 9.2

if you don't want to add another check, i'll investigate further. this *might* be a bug or conflict related to core-js...

jaynetics:
closes #2007

### DIFF
--- a/src/jsutils/nodejsCustomInspectSymbol.js
+++ b/src/jsutils/nodejsCustomInspectSymbol.js
@@ -1,7 +1,7 @@
 // @flow strict
 
 const nodejsCustomInspectSymbol =
-  typeof Symbol === 'function'
+  typeof Symbol === 'function' && typeof Symbol.for === 'function'
     ? Symbol.for('nodejs.util.inspect.custom')
     : undefined;
 


IvanGoncharov:
@jaynetics Thanks for PR 👍
Merged 🎉
Released as 14.4.2 📦

jaynetics:
@IvanGoncharov awesome, thanks for the uncomplicated process and quick release!

PR #2062, opened 2019-07-29
Limits errors in getVariableValues()

IvanGoncharov:
Based on #2037

### DIFF
--- a/src/utilities/coerceInputValue.js
+++ b/src/utilities/coerceInputValue.js
@@ -1,0 +1,214 @@
+// @flow strict
+
+import { forEach, isCollection } from 'iterall';
+
+import objectValues from '../polyfills/objectValues';
+
+import inspect from '../jsutils/inspect';
+import invariant from '../jsutils/invariant';
+import didYouMean from '../jsutils/didYouMean';
+import isObjectLike from '../jsutils/isObjectLike';
+import suggestionList from '../jsutils/suggestionList';
+import printPathArray from '../jsutils/printPathArray';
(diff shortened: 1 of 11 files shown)


src/utilities/coerceInputValue.js
```diff
@@ -0,0 +1,209 @@
...
+type OnErrorCB = (
+  path: $ReadOnlyArray<string | number>,
+  invalidValue: mixed,
+  error: GraphQLError,
+) => void;
+
+/**
+ * Coerces a JavaScript value given a GraphQL Input Type.
+ */
+export function coerceInputValue(
```
IvanGoncharov: I decided to create a new `coerceInputValue` function sine `coerceValue` is part of public API so we can't change its behavior.

In addition to that the new name is less confusing since GraphQL Spec defines `Input coercion` and `Result coercion`:
https://graphql.github.io/graphql-spec/draft/#sec-Scalars.Input-Coercion

IvanGoncharov:
It's draft PR (I still need to work on tests) but it would be great to get feedback on a new `coerceInputValue` API function and a general approach to limiting the number of errors.

@SoyYoRafa @Cito Can you please take a look?

SoyYoRafa:
I took a look after the fact, looks good. I thought about doing a similar thing but since I am not a maintainer, I didn't want to just rewrite parts of the code.

IvanGoncharov:
@SoyYoRafa Thanks for the review 👍 
Since this change introduces new API I need to release it as `14.5.0` next week.

PR #2073, opened 2019-08-03
Simplify parser, avoid extra function call

Cito:


IvanGoncharov:
@Cito Thanks for PR 👍 
Merged 🎉

PR #2084, opened 2019-08-13
RFC: Allow interfaces to implement other interfaces

mike-marcacci:
This is an implementation of the RFC described by spec PR https://github.com/graphql/graphql-spec/pull/373

This is a successor to PR #1218 which had accumulated too many merge conflicts to bother fixing.

### DIFF
--- a/src/__fixtures__/schema-kitchen-sink.graphql
+++ b/src/__fixtures__/schema-kitchen-sink.graphql
@@ -7,7 +7,7 @@ schema {
 This is a description
 of the `Foo` type.
 """
-type Foo implements Bar & Baz {
+type Foo implements Bar & Baz & Two {
   "Description of the `one` field."
   one: Type
   """
@@ -50,11 +50,17 @@ interface AnnotatedInterface @onInterface {
 
 interface UndefinedInterface
 
-extend interface Bar {
+extend interface Bar implements Two {
   two(argument: InputType!): Type
 }
 
 extend interface Bar @onInterface
+
+interface Baz implements Bar & Two {
+  one: Type
+  two(argument: InputType!): Type
+  four(argument: String = "string"): String
+}
 
 union Feed =
   | Story
--- a/src/execution/__tests__/union-interface-test.js
+++ b/src/execution/__tests__/union-interface-test.js
@@ -19,20 +19,32 @@ import { execute } from '../execute';
 class Dog {
   name: string;
   barks: boolean;
+  mother: ?Dog;
+  father: ?Dog;
+  progeny: Array<Dog>;
 
   constructor(name, barks) {
     this.name = name;
     this.barks = barks;
+    this.mother = null;
+    this.father = null;
+    this.progeny = [];
   }
 }
 
 class Cat {
   name: string;
   meows: boolean;
+  mother: ?Cat;
+  father: ?Cat;
+  progeny: Array<Cat>;
 
   constructor(name, meows) {
     this.name = name;
     this.meows = meows;
+    this.mother = null;
+    this.father = null;
+    this.progeny = [];
   }
 }
 
@@ -55,23 +67,46 @@ const NamedType = new GraphQLInterfaceType({
   },
 });
 
+const LifeType = new GraphQLInterfaceType({
+  name: 'Life',
+  fields: () => ({
+    progeny: { type: GraphQLList(LifeType) },
+  }),
+});
+
+const MammalType = new GraphQLInterfaceType({
+  name: 'Mammal',
+  interfaces: [LifeType],
+  fields: () => ({
+    progeny: { type: GraphQLList(MammalType) },
+    mother: { type: MammalType },
+    father: { type: MammalType },
+  }),
+});
+
 const DogType = new GraphQLObjectType({
   name: 'Dog',
-  interfaces: [NamedType],
-  fields: {
+  interfaces: [MammalType, LifeType, NamedType],
+  fields: () => ({
     name: { type: GraphQLString },
     barks: { type: GraphQLBoolean },
-  },
+    progeny: { type: GraphQLList(DogType) },
+    mother: { type: DogType },
+    father: { type: DogType },
+  }),
   isTypeOf: value => value instanceof Dog,
 });
 
 const CatType = new GraphQLObjectType({
   name: 'Cat',
-  interfaces: [NamedType],
-  fields: {
+  interfaces: [MammalType, LifeType, NamedType],
+  fields: () => ({
     name: { type: GraphQLString },
     meows: { type: GraphQLBoolean },
-  },
+    progeny: { type: GraphQLList(CatType) },
+    mother: { type: CatType },
+    father: { type: CatType },
+  }),
   isTypeOf: value => value instanceof Cat,
 });
 
@@ -90,12 +125,15 @@ const PetType = new GraphQLUnionType({
 
 const PersonType = new GraphQLObjectType({
   name: 'Person',
-  interfaces: [NamedType],
-  fields: {
+  interfaces: [NamedType, MammalType, LifeType],
+  fields: () => ({
     name: { type: GraphQLString },
     pets: { type: GraphQLList(PetType) },
     friends: { type: GraphQLList(NamedType) },
-  },
+    progeny: { type: GraphQLList(PersonType) },
+    mother: { type: PersonType },
+    father: { type: PersonType },
+  }),
   isTypeOf: value => value instanceof Person,
 });
 
@@ -105,7 +143,13 @@ const schema = new GraphQLSchema({
 });
 
 const garfield = new Cat('Garfield', false);
+garfield.mother = new Cat("Garfield's Mom", false);
+garfield.mother.progeny = [garfield];
+
 const odie = new Dog('Odie', true);
+odie.mother = new Dog("Odie's Mom", true);
+odie.mother.progeny = [odie];
+
 const liz = new Person('Liz');
 const john = new Person('John', [garfield, odie], [liz, odie]);
 
@@ -114,6 +158,15 @@ describe('Execute: Union and intersection types', () => {
     const ast = parse(`
       {
         Named: __type(name: "Named") {
+          kind
+          name
+          fields { name }
+          interfaces { name }
+          possibleTypes { name }
+          enumValues { name }
+          inputFields { name }
+        }
+        Mammal: __type(name: "Mammal") {
           kind
           name
           fields { name }
@@ -140,7 +193,16 @@ describe('Execute: Union and intersection types', () => {
           kind: 'INTERFACE',
           name: 'Named',
           fields: [{ name: 'name' }],
-          interfaces: null,
+          interfaces: [],
+          possibleTypes: [{ name: 'Person' }, { name: 'Dog' }, { name: 'Cat' }],
+          enumValues: null,
+          inputFields: null,
+        },
+        Mammal: {
+          kind: 'INTERFACE',
+          name: 'Mammal',
+          fields: [{ name: 'progeny' }, { name: 'mother' }, { name: 'father' }],
+          interfaces: [{ name: 'Life' }],
           possibleTypes: [{ name: 'Person' }, { name: 'Dog' }, { name: 'Cat' }],
           enumValues: null,
           inputFields: null,
@@ -178,8 +240,16 @@ describe('Execute: Union and intersection types', () => {
         __typename: 'Person',
         name: 'John',
         pets: [
-          { __typename: 'Cat', name: 'Garfield', meows: false },
-          { __typename: 'Dog', name: 'Odie', barks: true },
+          {
+            __typename: 'Cat',
+            name: 'Garfield',
+            meows: false,
+          },
+          {
+            __typename: 'Dog',
+            name: 'Odie',
+            barks: true,
+          },
         ],
       },
     });
@@ -210,8 +280,16 @@ describe('Execute: Union and intersection types', () => {
         __typename: 'Person',
         name: 'John',
         pets: [
-          { __typename: 'Cat', name: 'Garfield', meows: false },
-          { __typename: 'Dog', name: 'Odie', barks: true },
+          {
+            __typename: 'Cat',
+            name: 'Garfield',
+            meows: false,
+          },
+          {
+            __typename: 'Dog',
+            name: 'Odie',
+            barks: true,
+          },
         ],
       },
     });
@@ -259,6 +337,20 @@ describe('Execute: Union and intersection types', () => {
           ... on Cat {
             meows
           }
+
+          ... on Mammal {
+            mother {
+              __typename
+              ... on Dog {
+                name
+                barks
+              }
+              ... on Cat {
+                name
+                meows
+              }
+            }
+          }
         }
       }
     `);
@@ -268,8 +360,17 @@ describe('Execute: Union and intersection types', () => {
         __typename: 'Person',
         name: 'John',
         friends: [
-          { __typename: 'Person', name: 'Liz' },
-          { __typename: 'Dog', name: 'Odie', barks: true },
+          {
+            __typename: 'Person',
+            name: 'Liz',
+            mother: null,
+          },
+          {
+            __typename: 'Dog',
+            name: 'Odie',
+            barks: true,
+            mother: { __typename: 'Dog', name: "Odie's Mom", barks: true },
+          },
         ],
       },
     });
@@ -280,7 +381,14 @@ describe('Execute: Union and intersection types', () => {
       {
         __typename
         name
-        pets { ...PetFields }
+        pets {
+          ...PetFields,
+          ...on Mammal {
+            mother {
+              ...ProgenyFields
+            }
+          }
+        }
         friends { ...FriendFields }
       }
 
@@ -306,6 +414,12 @@ describe('Execute: Union and intersection types', () => {
           meows
         }
       }
+
+      fragment ProgenyFields on Life {
+        progeny {
+          __typename
+        }
+      }
     `);
 
     expect(execute(schema, ast, john)).to.deep.equal({
@@ -313,12 +427,29 @@ describe('Execute: Union and intersection types', () => {
         __typename: 'Person',
         name: 'John',
         pets: [
-          { __typename: 'Cat', name: 'Garfield', meows: false },
-          { __typename: 'Dog', name: 'Odie', barks: true },
+          {
+            __typename: 'Cat',
+            name: 'Garfield',
+            meows: false,
+            mother: { progeny: [{ __typename: 'Cat' }] },
+          },
+          {
+            __typename: 'Dog',
+            name: 'Odie',
+            barks: true,
+            mother: { progeny: [{ __typename: 'Dog' }] },
+          },
         ],
         friends: [
-          { __typename: 'Person', name: 'Liz' },
-          { __typename: 'Dog', name: 'Odie', barks: true },
+          {
+            __typename: 'Person',
+            name: 'Liz',
+          },
+          {
+            __typename: 'Dog',
+            name: 'Odie',
+            barks: true,
+          },
         ],
       },
     });
--- a/src/language/__tests__/schema-parser-test.js
+++ b/src/language/__tests__/schema-parser-test.js
@@ -169,7 +169,7 @@ describe('Schema Parser', () => {
     });
   });
 
-  it('Extension without fields', () => {
+  it('Object extension without fields', () => {
     const doc = parse('extend type Hello implements Greeting');
 
     expect(toJSONDeep(doc)).to.deep.equal({
@@ -188,7 +188,25 @@ describe('Schema Parser', () => {
     });
   });
 
-  it('Extension without fields followed by extension', () => {
+  it('Interface extension without fields', () => {
+    const doc = parse('extend interface Hello implements Greeting');
+    expect(toJSONDeep(doc)).to.deep.equal({
+      kind: 'Document',
+      definitions: [
+        {
+          kind: 'InterfaceTypeExtension',
+          name: nameNode('Hello', { start: 17, end: 22 }),
+          interfaces: [typeNode('Greeting', { start: 34, end: 42 })],
+          directives: [],
+          fields: [],
+          loc: { start: 0, end: 42 },
+        },
+      ],
+      loc: { start: 0, end: 42 },
+    });
+  });
+
+  it('Object extension without fields followed by extension', () => {
     const doc = parse(`
       extend type Hello implements Greeting
 
@@ -226,7 +244,51 @@ describe('Schema Parser', () => {
     });
   });
 
-  it('Extension do not include descriptions', () => {
+  it('Interface extension without fields followed by extension', () => {
+    const doc = parse(`
+      extend interface Hello implements Greeting
+
+      extend interface Hello implements SecondGreeting
+    `);
+    expect(toJSONDeep(doc)).to.deep.equal({
+      kind: 'Document',
+      definitions: [
+        {
+          kind: 'InterfaceTypeExtension',
+          name: nameNode('Hello', { start: 24, end: 29 }),
+          interfaces: [typeNode('Greeting', { start: 41, end: 49 })],
+          directives: [],
+          fields: [],
+          loc: { start: 7, end: 49 },
+        },
+        {
+          kind: 'InterfaceTypeExtension',
+          name: nameNode('Hello', { start: 74, end: 79 }),
+          interfaces: [typeNode('SecondGreeting', { start: 91, end: 105 })],
+          directives: [],
+          fields: [],
+          loc: { start: 57, end: 105 },
+        },
+      ],
+      loc: { start: 0, end: 110 },
+    });
+  });
+
+  it('Object extension without anything throws', () => {
+    expectSyntaxError('extend type Hello').to.deep.equal({
+      message: 'Syntax Error: Unexpected <EOF>.',
+      locations: [{ line: 1, column: 18 }],
+    });
+  });
+
+  it('Interface extension without anything throws', () => {
+    expectSyntaxError('extend interface Hello').to.deep.equal({
+      message: 'Syntax Error: Unexpected <EOF>.',
+      locations: [{ line: 1, column: 23 }],
+    });
+  });
+
+  it('Object extension do not include descriptions', () => {
     expectSyntaxError(`
       "Description"
       extend type Hello {
@@ -239,6 +301,27 @@ describe('Schema Parser', () => {
 
     expectSyntaxError(`
       extend "Description" type Hello {
+        world: String
+      }
+    `).to.deep.equal({
+      message: 'Syntax Error: Unexpected String "Description".',
+      locations: [{ line: 2, column: 14 }],
+    });
+  });
+
+  it('Interface extension do not include descriptions', () => {
+    expectSyntaxError(`
+      "Description"
+      extend interface Hello {
+        world: String
+      }
+    `).to.deep.equal({
+      message: 'Syntax Error: Unexpected Name "extend".',
+      locations: [{ line: 3, column: 7 }],
+    });
+
+    expectSyntaxError(`
+      extend "Description" interface Hello {
         world: String
       }
     `).to.deep.equal({
@@ -339,6 +422,31 @@ describe('Schema Parser', () => {
     });
   });
 
+  it('Simple interface inheriting interface', () => {
+    const doc = parse('interface Hello implements World { field: String }');
+    expect(toJSONDeep(doc)).to.deep.equal({
+      kind: 'Document',
+      definitions: [
+        {
+          kind: 'InterfaceTypeDefinition',
+          name: nameNode('Hello', { start: 10, end: 15 }),
+          description: undefined,
+          interfaces: [typeNode('World', { start: 27, end: 32 })],
+          directives: [],
+          fields: [
+            fieldNode(
+              nameNode('field', { start: 35, end: 40 }),
+              typeNode('String', { start: 42, end: 48 }),
+              { start: 35, end: 48 },
+            ),
+          ],
+          loc: { start: 0, end: 50 },
+        },
+      ],
+      loc: { start: 0, end: 50 },
+    });
+  });
+
   it('Simple type inheriting interface', () => {
     const doc = parse('type Hello implements World { field: String }');
 
@@ -394,6 +502,34 @@ describe('Schema Parser', () => {
     });
   });
 
+  it('Simple interface inheriting multiple interfaces', () => {
+    const doc = parse('interface Hello implements Wo & rld { field: String }');
+    expect(toJSONDeep(doc)).to.deep.equal({
+      kind: 'Document',
+      definitions: [
+        {
+          kind: 'InterfaceTypeDefinition',
+          name: nameNode('Hello', { start: 10, end: 15 }),
+          description: undefined,
+          interfaces: [
+            typeNode('Wo', { start: 27, end: 29 }),
+            typeNode('rld', { start: 32, end: 35 }),
+          ],
+          directives: [],
+          fields: [
+            fieldNode(
+              nameNode('field', { start: 38, end: 43 }),
+              typeNode('String', { start: 45, end: 51 }),
+              { start: 38, end: 51 },
+            ),
+          ],
+          loc: { start: 0, end: 53 },
+        },
+      ],
+      loc: { start: 0, end: 53 },
+    });
+  });
+
   it('Simple type inheriting multiple interfaces with leading ampersand', () => {
     const doc = parse('type Hello implements & Wo & rld { field: String }');
 
@@ -420,6 +556,36 @@ describe('Schema Parser', () => {
         },
       ],
       loc: { start: 0, end: 50 },
+    });
+  });
+
+  it('Simple interface inheriting multiple interfaces with leading ampersand', () => {
+    const doc = parse(
+      'interface Hello implements & Wo & rld { field: String }',
+    );
+    expect(toJSONDeep(doc)).to.deep.equal({
+      kind: 'Document',
+      definitions: [
+        {
+          kind: 'InterfaceTypeDefinition',
+          name: nameNode('Hello', { start: 10, end: 15 }),
+          description: undefined,
+          interfaces: [
+            typeNode('Wo', { start: 29, end: 31 }),
+            typeNode('rld', { start: 34, end: 37 }),
+          ],
+          directives: [],
+          fields: [
+            fieldNode(
+              nameNode('field', { start: 40, end: 45 }),
+              typeNode('String', { start: 47, end: 53 }),
+              { start: 40, end: 53 },
+            ),
+          ],
+          loc: { start: 0, end: 55 },
+        },
+      ],
+      loc: { start: 0, end: 55 },
     });
   });
 
@@ -478,6 +644,7 @@ describe('Schema Parser', () => {
           kind: 'InterfaceTypeDefinition',
           name: nameNode('Hello', { start: 10, end: 15 }),
           description: undefined,
+          interfaces: [],
           directives: [],
           fields: [
             fieldNode(
--- a/src/type/__tests__/schema-test.js
+++ b/src/type/__tests__/schema-test.js
@@ -181,6 +181,12 @@ describe('Type System: Schema', () => {
       const SomeInterface = new GraphQLInterfaceType({
         name: 'SomeInterface',
         fields: {},
+        interfaces: () => [AnotherInterface],
+      });
+
+      const AnotherInterface = new GraphQLInterfaceType({
+        name: 'AnotherInterface',
+        fields: {},
       });
 
       const SomeSubtype = new GraphQLObjectType({
@@ -192,6 +198,7 @@ describe('Type System: Schema', () => {
       const schema = new GraphQLSchema({ types: [SomeSubtype] });
 
       expect(schema.getType('SomeInterface')).to.equal(SomeInterface);
+      expect(schema.getType('AnotherInterface')).to.equal(AnotherInterface);
       expect(schema.getType('SomeSubtype')).to.equal(SomeSubtype);
     });
 
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -32,6 +32,7 @@ import {
   type GraphQLNamedType,
   type GraphQLAbstractType,
   type GraphQLObjectType,
+  type GraphQLInterfaceType,
   isObjectType,
   isInterfaceType,
   isUnionType,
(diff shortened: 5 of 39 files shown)


mike-marcacci:
It looks like I can't request particular reviewers, so I'm just tagging @mjmahone and @michaelstaib in case you weren't notified already. The PR is incredibly simple but touches many pieces of code (especially tests).

src/type/schema.js
```diff
@@ -234,7 +248,24 @@ export class GraphQLSchema {
...
+
+    const types = this._objectImplementations[abstractType.name];
+    if (!types) {
+      return (this._objectImplementations[abstractType.name] = []);
+    }
+
+    return types;
+  }
+
+  getImplementations(
```
mjmahone: This should be a separate PR: this functionality is  useful even without interfaces implementing other interfaces, and should be considered on its own merits. Especially as you're not using it in this PR directly.

src/type/schema.js
```diff
@@ -254,6 +285,13 @@ export class GraphQLSchema {
     return Boolean(possibleTypeMap[abstractType.name][possibleType.name]);
   }
 
+  isImplementation(
```
mjmahone: Same here, though it's used once in this diff: we could either add an internal-only utility function for this for this PR, or land a PR with this included before this.

src/language/__tests__/schema-parser-test.js
```diff
@@ -480,6 +639,7 @@ describe('Schema Parser', () => {
           kind: 'InterfaceTypeDefinition',
           name: nameNode('Hello', { start: 10, end: 15 }),
           description: undefined,
+          interfaces: [],
```
mjmahone: Do we have an integration test (probably not under the `parser`, but in `buildASTSchema`) for an AST having `interfaces` as undefined?

I know we ran into problems in the past when hand-modeling the AST where, unless you explicitly set `directives: []`, you'd end up getting an undefined reference, even though the flow type allowed it.

src/type/schema.js
mike-marcacci: Thanks so much for the excellent review! I just went to split these out, but their language and types _do_ assume that interfaces are potentially valid implementors. Are these assumptions appropriate to break out into an unrelated? I'm more than happy to do it, but I would be cautions of actually merging them without the rest, as the assumptions are only true in the context of this RFC.

src/type/schema.js
mike-marcacci: (Same as above.) This function actually does call `getImplementations`; I broke them into two primarily to respect the pattern used by `getPossibleTypes`/`isPossibleType`.

src/language/__tests__/schema-parser-test.js
mike-marcacci: I just added an assertion for this case in dfc2da9d937987a7a23baf22ab203a8dea228087.

src/type/schema.js
mike-marcacci: @mjmahone just to bump this and make sure I'm clear on the rationale for breaking this into a separate PR: is the thought that this is changing the public API of this type in a way that is useful even without the interface RFC? I am a bit reluctant to do this, because in a two-PR scenario I would introduce a non-breaking change in the first one (adding the methods, which only ever return concrete types simply because those are the only thing that can currently implement an interface); then introduce a potentially-breaking change in the second (these methods would now return both concrete types _and_ interfaces). Introducing these methods in the form written here makes sure there's only one, non-breaking change.

IvanGoncharov: @mjmahone @mike-marcacci I don't think it's worth to extract this into separate PR since `getImplementations(interaceType)` is fully queal to `getPossibleTypes(interfaceType)` without this PR being merged.

src/type/schema.js
```diff
@@ -123,7 +124,8 @@ export class GraphQLSchema {
   _subscriptionType: ?GraphQLObjectType;
   _directives: $ReadOnlyArray<GraphQLDirective>;
   _typeMap: TypeMap;
-  _implementations: ObjMap<Array<GraphQLObjectType>>;
+  _implementations: ObjMap<Array<GraphQLObjectType | GraphQLInterfaceType>>;
+  _objectImplementations: ObjMap<Array<GraphQLObjectType>>;
```
IvanGoncharov: @mike-marcacci It would be better to have a single object map:
```
_implementations: ObjectMap<{ objects: Array<GraphQLObjectType>, interfaces: Array<GraphQLInterfaceType> }>
```

mike-marcacci: Hey @IvanGoncharov, I was implementing this just now, and I actually disagree here. Currently, `_implementations` contains a map of objects _and_ interfaces, while `_objectImplementations` only contains objects. Changing this so that interface implementations are stored separate from objects would make us have to recombine then in `getImplementations()` which would then return new arrays each time. This has the disadvantage of creating additional unnecessary objects (which probably doesn't matter much) but also breaks any object identity comparisons (`schema.getImplementations(myInterface) === schema.getImplementations(myInterface)`).

src/__fixtures__/github-schema.json
```diff
@@ -56052,7 +56052,7 @@
             },
             {
               "name": "INTERFACE",
-              "description": "Indicates this type is an interface. `fields` and `possibleTypes` are valid fields.",
+              "description": "Indicates this type is an interface. `fields`, `interfaces`, and `possibleTypes` are valid fields.",
```
IvanGoncharov: It's introspection snapshot of public GitHub API that we use for perf testing.
Please revert.

mike-marcacci: Hmm, should this be reverted even though [the source line](https://github.com/mike-marcacci/graphql-js/blob/6f85d6b33c20c0de6556d153cd723dac4324472a/src/type/introspection.js#L401) has changed?

IvanGoncharov: @mike-marcacci It's literally the result of introspection query against GitHub public API. Next time someone updates it this change will be lost.

src/__fixtures__/schema-kitchen-sink.graphql
```diff
@@ -56,6 +56,12 @@ extend interface Bar {
 
 extend interface Bar @onInterface
 
+interface Baz implements Bar {
```
IvanGoncharov: Since it's a kitchen sink it's better to implement couple of interfaces e.g. `implements Foo & Bar`.
Also please add an example of interface extension that adds a new interface

src/utilities/schemaPrinter.js
```diff
@@ -194,9 +194,13 @@ function printObject(type: GraphQLObjectType, options): string {
 }
 
 function printInterface(type: GraphQLInterfaceType, options): string {
+  const interfaces = type.getInterfaces();
```
IvanGoncharov: please extract it into `printImplementedInterfaces` and reuse for both objects and interfaces, similar to `printFields`

src/utilities/buildClientSchema.js
```diff
@@ -253,9 +253,17 @@ export function buildClientSchema(
   function buildInterfaceDef(
     interfaceIntrospection: IntrospectionInterfaceType,
   ): GraphQLInterfaceType {
+    if (!interfaceIntrospection.interfaces) {
+      throw new Error(
+        'Introspection result missing interfaces: ' +
+          inspect(interfaceIntrospection),
+      );
+    }
+
```
IvanGoncharov: Please extract it into a separate function and reuse for building object type, similar to `buildFieldDefMap`

src/utilities/findBreakingChanges.js
```diff
@@ -324,6 +324,30 @@ function findObjectTypeChanges(
...
+      description: `${newInterface.name} added to interfaces implemented by ${oldType.name}.`,
+    });
+  }
+
+  for (const oldInterface of interfacesDiff.removed) {
+    schemaChanges.push({
+      type: BreakingChangeType.INTERFACE_REMOVED_FROM_OBJECT,
+      description: `${oldType.name} no longer implements interface ${oldInterface.name}.`,
+    });
+  }
```
IvanGoncharov: Please extract into a separate function and reuse for the object type, similar to `findFieldChanges`

src/__fixtures__/github-schema.json
mike-marcacci: Ah, gotcha. Updated!

src/__fixtures__/schema-kitchen-sink.graphql
mike-marcacci: Done and done.

src/utilities/schemaPrinter.js
mike-marcacci: 👍

src/utilities/buildClientSchema.js
mike-marcacci: What part here is reusable? Do you just want to consolidate the creation of this object?

```js
{
  name: interfaceIntrospection.name,
  description: interfaceIntrospection.description,
  interfaces: () => interfaceIntrospection.interfaces.map(getInterfaceType),
  fields: () => buildFieldDefMap(interfaceIntrospection),
}
```

If so, I'm not sure the extra redirection here is worth the DRY...

IvanGoncharov: @mike-marcacci I'm referring to:
```
    if (!interfaceIntrospection.interfaces) {
      throw new Error(
        'Introspection result missing interfaces: ' +
          inspect(interfaceIntrospection),
      );
    }
    return interfaceIntrospection.interfaces.map(getInterfaceType)
```

mike-marcacci: Ah got it; good call. Fixed.

src/utilities/findBreakingChanges.js
```diff
@@ -190,7 +190,7 @@ function findTypeChanges(
     } else if (isObjectType(oldType) && isObjectType(newType)) {
       schemaChanges.push(...findObjectTypeChanges(oldType, newType));
     } else if (isInterfaceType(oldType) && isInterfaceType(newType)) {
-      schemaChanges.push(...findFieldChanges(oldType, newType));
+      schemaChanges.push(...findInterfaceTypeChanges(oldType, newType));
```
IvanGoncharov: Don't think it's worth to have `findObjectTypeChanges` and `findInterfaceTypeChanges` as wrapper functions. Please change it to:
```
schemaChanges.push(
  ...findImplementedInterfacesChanges(oldType, newType),
  ...findFieldChanges(oldType, newType),
);
```
and do the same for object type.

src/utilities/schemaPrinter.js
```diff
@@ -181,22 +181,27 @@ function printScalar(type: GraphQLScalarType, options): string {
...
+  return interfaces.length
     ? ' implements ' + interfaces.map(i => i.name).join(' & ')
     : '';
+}
+
+function printObject(type: GraphQLObjectType, options): string {
   return (
     printDescription(options, type) +
-    `type ${type.name}${implementedInterfaces}` +
+    `type ${type.name}${printImplementedInterfaces(type)}` +
```
IvanGoncharov: Please don't wrap it with `${}` use, please use simple concatenation same as `printFields` on a line bellow.

src/utilities/schemaPrinter.js
```diff
@@ -181,22 +181,27 @@ function printScalar(type: GraphQLScalarType, options): string {
...
     printFields(options, type)
   );
 }
 
 function printInterface(type: GraphQLInterfaceType, options): string {
   return (
     printDescription(options, type) +
-    `interface ${type.name}` +
+    `interface ${type.name}${printImplementedInterfaces(type)}` +
```
IvanGoncharov: same here

src/utilities/findBreakingChanges.js
mike-marcacci: done

src/utilities/schemaPrinter.js
mike-marcacci: done

src/utilities/schemaPrinter.js
mike-marcacci: done

mike-marcacci:
@IvanGoncharov I've added logic to prevent cyclic implementations. Because we already require flattening of transitively implemented interfaces, this ended up being trivial and didn't require any recursion. I believe this is in its final state.

I'm going to write up a blog post today, but this PR can probably be merged without it. Does anyone have a preference on how/where I publish the post? I can publish it myself on Medium after this is merged, or create a PR [to the GraphQL blog](https://github.com/graphql/graphql.github.io/tree/source/site/blog).

IvanGoncharov:
> Does anyone have a preference on how/where I publish the post? I can publish it myself on Medium after this is merged, or create a PR to the GraphQL blog

@mike-marcacci I think publishing it on personal account would be enough.

Also can you please update TS typings in `tstypes` folder?

> I believe this is in its final state.

Great 🎉 I'm currently in the middle of review:
image
But I will try to finish it tomorrow.

mike-marcacci:
Wonderful! Great catch on the TS types; I just added those.

src/utilities/__tests__/buildASTSchema-test.js
```diff
@@ -285,6 +285,12 @@ describe('Schema Builder', () => {
     const sdl = dedent`
       interface EmptyInterface
     `;
+
+    const definition = parse(sdl).definitions[0];
+    expect(
+      definition.kind === 'InterfaceTypeDefinition' && definition.interfaces,
+    ).to.deep.equal([], 'The interfaces property must be an empty array.');
+
```
IvanGoncharov: @mike-marcacci Why do you need this check?
This should be checked in parser tests.

mike-marcacci: Adding this test was requested by @mjmahone [in this comment](https://github.com/graphql/graphql-js/pull/2084#discussion_r313964490).

IvanGoncharov:
@mike-marcacci Currently we are in feature freeze for `14.x.x` so this feature should be included in the upcoming `15.0.0`. I plan to release `15.0.0-alpha.1` in the next few days and hopefully include this feature.

I branch of `14.x.x` and merged `15.x.x` into `master`. Can you please rebase this PR on top of `master`?

mike-marcacci:
Thanks so much for the thorough review @IvanGoncharov! I just rebased this on top of master. 👍

I also added a `// prettier-ignore` to prevent the addition of a trailing comma in one of the typedefs.

IvanGoncharov:
@mike-marcacci Sorry for the delay I was traveling to a conference and speaking today.
I will 100% finish the review tomorrow.

mike-marcacci:
@IvanGoncharov no worries at all! You're an absolute hero for all your work here :)

src/type/validate.js
```diff
@@ -410,19 +429,53 @@ function validateObjectImplementsInterface(
...
   }
 }
 
+function validateImplementsAncestors(
+  context: SchemaValidationContext,
+  implementing: GraphQLObjectType | GraphQLInterfaceType,
+  implemented: GraphQLInterfaceType,
+): void {
+  const implementingInterfaces = implementing.getInterfaces();
+  implemented.getInterfaces().forEach(transitive => {
```
IvanGoncharov: @mike-marcacci We don't use `forEach` can please use `for of` similar to other cycles in this file?

src/execution/__tests__/union-interface-test.js
```diff
@@ -140,7 +175,16 @@ describe('Execute: Union and intersection types', () => {
...
+          inputFields: null,
+        },
+        Mammal: {
+          kind: 'INTERFACE',
+          name: 'Mammal',
+          fields: [{ name: 'progeny' }, { name: 'mother' }, { name: 'father' }],
+          interfaces: [{ name: 'Life' }],
           possibleTypes: [{ name: 'Person' }, { name: 'Dog' }, { name: 'Cat' }],
           enumValues: null,
           inputFields: null,
```
IvanGoncharov: Can you please also extend other tests bellow with named and inline fragments on intermidiate interfaces.

src/language/__tests__/schema-parser-test.js
```diff
@@ -226,13 +244,59 @@ describe('Schema Parser', () => {
...
+  it('Interface extension without anything throws', () => {
+    expectSyntaxError('extend interface Hello').to.deep.equal({
+      message: 'Syntax Error: Unexpected <EOF>.',
+      locations: [{ line: 1, column: 23 }],
+    });
+  });
+
+  it('Object extension do not include descriptions', () => {
+    expectSyntaxError(
+      `
```
IvanGoncharov: Can you please change to 
```
expectSyntaxError(`
"Description"
...
`).to.deep.equal({
```

src/language/__tests__/schema-parser-test.js
```diff
@@ -247,6 +311,31 @@ describe('Schema Parser', () => {
     });
   });
 
+  it('Interface extension do not include descriptions', () => {
+    expectSyntaxError(
+      `
```
IvanGoncharov: same here

src/type/__tests__/schema-test.js
```diff
@@ -183,24 +191,35 @@ describe('Type System: Schema', () => {
...
+      const SomeImplementingInterface = new GraphQLInterfaceType({
+        name: 'SomeImplementingInterface',
         fields: {},
         interfaces: () => [SomeInterface],
       });
 
+      const SomeImplementingObject = new GraphQLObjectType({
+        name: 'SomeImplementingObject',
+        fields: {},
+        interfaces: () => [SomeInterface, SomeImplementingInterface],
```
IvanGoncharov: No need to modify this test since you explicitly listing both interfaces here

src/type/validate.js
```diff
@@ -314,94 +317,110 @@ function validateFields(
   }
 }
 
-function validateObjectInterfaces(
+function validateInterfaces(
   context: SchemaValidationContext,
-  object: GraphQLObjectType,
+  implementing: GraphQLObjectType | GraphQLInterfaceType,
```
IvanGoncharov: Can you please change the name to `type` similar to `validateFields`:
https://github.com/graphql/graphql-js/blob/5873fd26f60e88c473dea678a7723888f766658b/src/type/validate.js#L259-L262

src/type/validate.js
```diff
@@ -314,94 +317,110 @@ function validateFields(
...
 ): void {
   const implementedTypeNames = Object.create(null);
-  for (const iface of object.getInterfaces()) {
```
IvanGoncharov: It's unnecessary to rename `iface` to `implemented`.
It will remove unnecessary diff.

src/type/validate.js
```diff
@@ -314,94 +317,110 @@ function validateFields(
...
       context.reportError(
-        `Type ${object.name} can only implement ${iface.name} once.`,
-        getAllImplementsInterfaceNodes(object, iface),
+        `Type ${implementing.name} can only implement ${implemented.name} once.`,
+        getAllImplementsInterfaceNodes(implementing, implemented),
       );
       continue;
     }
-    implementedTypeNames[iface.name] = true;
-    validateObjectImplementsInterface(context, object, iface);
```
IvanGoncharov: It's better to use `validateTypeImplementsInterface`.

src/type/validate.js
```diff
@@ -314,94 +317,110 @@ function validateFields(
...
   }
 }
 
-function validateObjectImplementsInterface(
+function validateImplementsInterface(
   context: SchemaValidationContext,
-  object: GraphQLObjectType,
-  iface: GraphQLInterfaceType,
+  implementing: GraphQLObjectType | GraphQLInterfaceType,
+  implemented: GraphQLInterfaceType,
```
IvanGoncharov: Same here, I think `type` and `iface` is perfectly fine.
This also will minimize diff.
I agree that renaming stuff may be beneficial but it should be done in separate PR and for the entire file so we have consistency.

IvanGoncharov: With all renaming going on it's extremely hard to spot actual changes :(

src/type/validate.js
```diff
@@ -410,19 +429,53 @@ function validateObjectImplementsInterface(
...
+  context: SchemaValidationContext,
+  implementing: GraphQLObjectType | GraphQLInterfaceType,
+  implemented: GraphQLInterfaceType,
+): void {
+  const implementingInterfaces = implementing.getInterfaces();
+  implemented.getInterfaces().forEach(transitive => {
+    if (!implementingInterfaces.includes(transitive)) {
+      if (transitive === implementing) {
+        context.reportError(
+          `${isObjectType(implementing) ? 'Object' : 'Interface'} type ${
```
IvanGoncharov: Please move this into
```
const typeKindStr = isObjectType(implementing) ? 'Object' : 'Interface';
```
so the rest of the string can be a single line

src/type/validate.js
```diff
@@ -410,19 +429,53 @@ function validateObjectImplementsInterface(
...
 
+function validateImplementsAncestors(
+  context: SchemaValidationContext,
+  implementing: GraphQLObjectType | GraphQLInterfaceType,
+  implemented: GraphQLInterfaceType,
+): void {
+  const implementingInterfaces = implementing.getInterfaces();
+  implemented.getInterfaces().forEach(transitive => {
+    if (!implementingInterfaces.includes(transitive)) {
+      if (transitive === implementing) {
```
IvanGoncharov: This check is independent of the previous one so why did you nested it or am I missing something here?

src/type/validate.js
```diff
@@ -314,94 +317,110 @@ function validateFields(
...
+  implementing: GraphQLObjectType | GraphQLInterfaceType,
+  implemented: GraphQLInterfaceType,
 ): void {
-  const objectFieldMap = object.getFields();
-  const ifaceFieldMap = iface.getFields();
+  const implementingFieldMap = implementing.getFields();
+  const implementedFieldMap = implemented.getFields();
+
+  // Assert each ancestor interface is explicitly implemented.
+  validateImplementsAncestors(context, implementing, implemented);
```
IvanGoncharov: No need to make call stack deeper, please move it into `validateInterfaces` before `validateImplementsInterface` call.

src/type/schema.js
```diff
@@ -32,6 +32,7 @@ import {
   type GraphQLNamedType,
   type GraphQLAbstractType,
   type GraphQLObjectType,
```
IvanGoncharov: You forget to synchronize this file with `schema.d.ts`

src/type/validate.js
mike-marcacci: I'm not 100% sure I'm following here (this removes those lines, and `validateTypeImplementsInterface` doesn't exist).

src/type/validate.js
mike-marcacci: Thi follows the language in the spec, which is very helpful when verifying correctness.

src/type/validate.js
mike-marcacci: This is simply to switch which error message is generated (is the interface missing from the list of implemented interfaces due to a circular reference, or because it was never declared). I'm changing this to a ternary to make this logic more clear.

src/type/validate.js
mike-marcacci: With that in mind, would you still prefer this be in a separate PR?

src/type/schema.js
mike-marcacci: Good catch!

src/type/__tests__/schema-test.js
mike-marcacci: Hmm, unless I'm missing something, the only real changes here are to the names, which make them more symmetrical and consistent w/ the spec and the language used in the implementations.

mike-marcacci:
@IvanGoncharov thanks SO much for the great review. I pushed back on 2 of your comments RE name changes, since the language I used here mirrors the spec, and I found it much easier to follow when comparing the two. If you have a strong preference despite this, let me know and I'll modify the names here.

I did also need some clarification on one of the other comments.

Let me know about these and I'll get any other changes made ASAP!

src/type/validate.js
IvanGoncharov: I mean if we rename `validateObjectImplementsInterface` to something `validateTypeImplementsInterface` is better follows existing naming schema than `validateImplementsAncestors`

mike-marcacci: Ah, gotcha. As these are separate functions with the same signature, should these both be changed from:

```js
validateImplementsAncestors(context, type, iface);
validateImplementsInterface(context, type, iface);
```

to:
```js
validateTypeImplementsAncestors(context, type, iface);
validateTypeImplementsInterface(context, type, iface);
```

mike-marcacci: (Previously `validateImplementsAncestors ` was only called from inside `validateImplementsInterface`, but this was changed to avoid an extra level of depth in the stack.)

src/language/__tests__/schema-parser-test.js
```diff
@@ -247,6 +309,29 @@ describe('Schema Parser', () => {
...
+      message: 'Syntax Error: Unexpected Name "extend".',
+      locations: [{ line: 3, column: 7 }],
+    });
+
+    expectSyntaxError(
+      `
+      extend "Description" interface Hello {
+        world: String
+      }
+    `,
```
IvanGoncharov: Can you please also reformat this one?

src/type/validate.js
mike-marcacci: I've gone ahead and renamed these per your suggestion.

src/type/validate.js
mike-marcacci: I've gone ahead and renamed these functions and variable names; hopefully these look better.

src/type/__tests__/schema-test.js
mike-marcacci: I updated all the names to match your suggestions... I'm not sure why GitHub's diffs aren't showing these changes. I'm going to force them to refresh.

mike-marcacci:
@IvanGoncharov I've gone ahead and made all your suggested changes – we can debate the variable naming later. Let me know if you have any other requests, and I'll address them ASAP!

mike-marcacci:
Rebased on top of master again.

tstypes/type/definition.d.ts
```diff
@@ -387,6 +387,7 @@ export class GraphQLObjectType<
   toConfig(): GraphQLObjectTypeConfig<any, any> & {
     interfaces: GraphQLInterfaceType[];
     fields: GraphQLFieldConfigMap<any, any>;
+    extensions: Maybe<Readonly<Record<string, any>>>;
```
IvanGoncharov: @mike-marcacci Thanks for fixing I went ahead so and fixed all similar stuff in #2202 sorry for rebases but it will keep cleaner diff in this PR.

mike-marcacci: No worries – it's easy as long as I don't let it get too far out of date :)

src/type/__tests__/schema-test.js
IvanGoncharov: @mike-marcacci Just to make sure we both on the same page: This test testing that all interfaces returned by `interfaces` Thunk included in the resulting schema. Since you all list all interfaces we can't test if schema traverse interfaces or not so add transitive interfaces doesn't test anything and just makes this test bigger.

src/type/validate.js
IvanGoncharov: So there are two problems here:
1. Since now it validates both interface and objects but in your version, it matches only interface algorithm.
2. Maybe it's only me but `implementing` and `implemented` looks too similar when I'm scanning through the source code. Maybe it's because I'm not a native speaker or because of some mild dyslexia.

src/utilities/__tests__/buildASTSchema-test.js
IvanGoncharov: So Matt's comment was about testing if `buildSchema` supports both empty array and undefined:
> Do we have an integration test (probably not under the parser, but in buildASTSchema) for an AST having interfaces as undefined?

In its current form, you just duplicating parser tests since it's just checking `interfaces` value is an array and Matt suggested to modify it to `undefined` and test that result is the same.

At the moment we don't have any such test so think it should be a generic solution so I created #2203

src/type/validate.js
mike-marcacci: 1. Yes, originally I had two different functions: one for objects and one for interfaces. This allowed me to use language that matched the spec in both places. You're right, though, that since we combined them into one function, only one will have wording that matches the spec.
2. Yes, this is a bit hard to read, but in the context of interfaces implementing other interfaces, these were the most clear words I could think to use. I agree that while that might be a good choice in the spec (for the sake of correctness) it's much easier to follow here, and probably the right choice given point 1.

src/utilities/__tests__/buildASTSchema-test.js
mike-marcacci: That's a great point; I assume I should just leave this as is and we can remove this test in a separate PR that fixes #2203?

src/type/__tests__/schema-test.js
mike-marcacci: Aah, I see what you mean; point taken. I've reverted this file to match master.

src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -572,6 +572,27 @@ describe('findBreakingChanges', () => {
...
+
+    const newSchema = buildSchema(`
+      interface Interface1
+
+      interface Interface2
+    `);
+
+    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([
+      {
+        type: BreakingChangeType.INTERFACE_REMOVED_FROM_OBJECT,
```
IvanGoncharov: It's a problem that enum value is called `INTERFACE_REMOVED_FROM_OBJECT` because it was removed from interface. Since we targeting `15.0.0` we can do breaking change so I propose to rename it to `IMPLEMENTED_INTERFACE_REMOVED` similar to `FIELD_REMOVED`:
https://github.com/graphql/graphql-js/blob/21de98ea4855df8c5146580611fda8ad76f52a2f/src/utilities/findBreakingChanges.js#L45

src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -1000,6 +1021,30 @@ describe('findDangerousChanges', () => {
...
+    const newSchema = buildSchema(`
+      interface OldInterface
+      interface NewInterface
+
+      interface Interface1 implements OldInterface & NewInterface
+    `);
+
+    expect(findDangerousChanges(oldSchema, newSchema)).to.deep.equal([
+      {
+        type: DangerousChangeType.INTERFACE_ADDED_TO_OBJECT,
```
IvanGoncharov: same here

src/utilities/__tests__/findBreakingChanges-test.js
mike-marcacci: Great point! I've renamed this to `IMPLEMENTED_INTERFACE_REMOVED`.

src/utilities/__tests__/findBreakingChanges-test.js
mike-marcacci: I've renamed this to `IMPLEMENTED_INTERFACE_ADDED` following the other types.

src/type/schema.js
```diff
@@ -250,6 +281,13 @@ export class GraphQLSchema {
     return this._possibleTypeMap[abstractType.name][possibleType.name] != null;
   }
 
+  isImplementation(
+    interfaceType: GraphQLInterfaceType,
+    possibleType: GraphQLObjectType | GraphQLInterfaceType,
+  ): boolean {
+    return this.getImplementations(interfaceType).includes(possibleType);
```
IvanGoncharov: Includes is not supported in IE11 so we can't use it until we drop support for it :(

IvanGoncharov: Also please check other places for functions not supported in IE11

mike-marcacci: I didn't realize that IE11 was a target for this library. I removed all instances of `Array.prototype.includes`, including one that wasn't introduced in this PR. I know that with TS you can specify the library available in the target... this is something we should make sure to set correctly during the move to TS to catch any of these cases.

mike-marcacci: I'll keep my new changes from `.includes` to `.indexOf` for now, but reading through the babel configs, it looks like `.includes` should be fine, because it is rewritten with a polyfill when targeting IE11.

mike-marcacci: Err... it looks like the other language features that would require an IE11-targeting app to include `babel-polyfill` are only used in the tests, so my claim here doesn't hold true; I reverted my change to the code not introduced by this PR, since it was in a test.

src/type/validate.js
```diff
@@ -410,19 +425,41 @@ function validateObjectImplementsInterface(
...
 
+function validateTypeImplementsAncestors(
+  context: SchemaValidationContext,
+  type: GraphQLObjectType | GraphQLInterfaceType,
+  iface: GraphQLInterfaceType,
+): void {
+  const implementingInterfaces = type.getInterfaces();
+  for (const transitive of iface.getInterfaces()) {
+    if (implementingInterfaces.indexOf(transitive) === -1) {
+      const typeKindStr = isObjectType(type) ? 'Object' : 'Interface';
```
IvanGoncharov: We have other similar errors and they simply refer to type as `Type` I don't think this error should be special:
https://github.com/graphql/graphql-js/blob/21de98ea4855df8c5146580611fda8ad76f52a2f/src/type/validate.js#L268
https://github.com/graphql/graphql-js/blob/21de98ea4855df8c5146580611fda8ad76f52a2f/src/type/validate.js#L325
Fixed in https://github.com/mike-marcacci/graphql-js/pull/1

src/type/validate.js
```diff
@@ -410,19 +425,41 @@ function validateObjectImplementsInterface(
...
 }
 
+function validateTypeImplementsAncestors(
+  context: SchemaValidationContext,
+  type: GraphQLObjectType | GraphQLInterfaceType,
+  iface: GraphQLInterfaceType,
+): void {
+  const implementingInterfaces = type.getInterfaces();
+  for (const transitive of iface.getInterfaces()) {
+    if (implementingInterfaces.indexOf(transitive) === -1) {
```
IvanGoncharov: This check doesn't for self-referencing interfaces:
```
interfaces Foo implements Foo {}
```
Fixed in https://github.com/mike-marcacci/graphql-js/pull/1

src/type/validate.js
```diff
@@ -314,94 +317,106 @@ function validateFields(
...
-        `Interface field ${iface.name}.${fieldName} expected but ${object.name} does not provide it.`,
-        [ifaceField.astNode, ...getAllNodes(object)],
+        `Interface field ${iface.name}.${fieldName} expected but ${type.name} does not provide it.`,
+        [implementedField.astNode, ...getAllNodes(type)],
       );
       continue;
     }
 
-    // Assert interface field type is satisfied by object field type, by being
+    // Assert interface field type is satisfied by implementing field type, by being
```
IvanGoncharov: After the change, this comment doesn't make sense `is satisfied by implementing field type` is different from `is satisfied by object field type`.
Arguments here is the same as in this thread:
https://github.com/graphql/graphql-js/pull/2084#discussion_r328241103
So can we revert to the previous state and have `typeField` and `ifaceField`?
and thus minimize changes in this function.

IvanGoncharov:
Fixed type collection from interace's interfaces here:
https://github.com/mike-marcacci/graphql-js/pull/1/commits/887219c4dcb1bee87caa943e123e3e7aa43caf8b

src/type/schema.js
```diff
@@ -232,7 +246,24 @@ export class GraphQLSchema {
     if (isUnionType(abstractType)) {
       return abstractType.getTypes();
     }
-    return this._implementations[abstractType.name] || [];
+
+    const types = this._objectImplementations[abstractType.name];
+    if (!types) {
+      return (this._objectImplementations[abstractType.name] = []);
+    }
```
IvanGoncharov: What is the motivation for this change?
It's increasing the size `_objectImplementations` and also allocates memory during execution. If you feel like `|| []` is a problem (if so can you please explain why?) you can put an empty array inside the constructor to have predictable performance less chance to hit performance cliff due to V8 deoptimization.

src/utilities/buildClientSchema.js
```diff
@@ -232,19 +232,28 @@ export function buildClientSchema(
...
-  function buildObjectDef(
-    objectIntrospection: IntrospectionObjectType,
-  ): GraphQLObjectType {
-    if (!objectIntrospection.interfaces) {
+  function buildImplementationsList(
+    implementingIntrospection:
+      | IntrospectionObjectType
+      | IntrospectionInterfaceType,
+  ) {
+    if (!implementingIntrospection.interfaces) {
```
IvanGoncharov: We need to have legacy support for servers that would return `null` as `interfaces` value.
Fixed in https://github.com/mike-marcacci/graphql-js/pull/1/commits/07d8a2261f654768143a921507a1c96ce1cf645a

src/type/validate.js
mike-marcacci: Great catch!

src/type/schema.js
mike-marcacci: While this does add some trivial allocation of empty arrays, it prevents a new array from being created every time `getPossibleTypes` is called

mike-marcacci: I'm certainly willing to move this to the constructor, but I'm not convinced that the resulting behavior would be an improvement. If this were a realtime language like rust or C, I would understand a preference for time-predictable calls here over total allocation, and then we could avoid memoization in general. However, given that JS is garbage collected and a memoization pattern was already employed here, I think this is the "minimal change" that actually provides correct behavior.

mike-marcacci: Perhaps an alternative which would reduce allocations while still preservice object identity for comparison purposes would be to simply hoist the empty array out of the function. I'll implement that instead.

mike-marcacci: I've changed my mind here RE the importance of object identity. While it's a goal that I appreciate, I don't want to force on a project where it's not a shared by everyone involved, as it would be too easy for a consuming app to rely on object identity for the sake of comparisons, then break on a minor/patch update if that isn't considered part of the API. I'm going to take your recommendation here and just use `|| []`.

mike-marcacci:
@IvanGoncharov I just rebased on top of master to resolve a small merge conflict. I also squashed all the commits in this PR to make the rebase easier and so this can enter the history as a single atomic change. Just to make sure, is there anything you're waiting on from me?

IvanGoncharov:
@mike-marcacci Sorry for the delay. 
I reviewed almost all files and everything looks good.
But I want to try a couple of different approaches to `GraphQLSchema` changes.
I will try my hard to finish it tomorrow.

src/type/schema.js
IvanGoncharov: I also changed my mind since interfaces and unions without subtypes should be pretty rare I moved this code into the constructor:
https://github.com/APIs-guru/graphql-js/blob/102a8d9eda29b6be52f37f4c17c857ae810895bb/src/type/schema.js#L335-L337

IvanGoncharov:
@mike-marcacci I created https://github.com/mike-marcacci/graphql-js/pull/2 with proposed changes:
1. `isPossibleType` & `isImplementation` merged into single `isSubType` function and deprecate `isPossibleType`.
2. `getImplementations` now returns object with `objects` and `interfaces` fields. That way we don't have two separate maps with overlapping arrays (one for objects and another for objects and interfaces).
Can you please review it?

It looks like it was the last major issue I spotted.

mike-marcacci:
I just merged your suggestions – thanks for making those as a PR :)

I agree that having a single map with wrapping objects is probably a better approach.

Issue #1527, opened 2018-09-20
RFC: Extra property on field definition to pass extra metadata


JCMais:
I'm currently adding extra properties to some graphql object field definitions, like the following:

```js
const MutationType = new GraphQLObjectType({
  name: 'Mutation',
  fields: () => ({
    AddSomething: {
      // ... normal field properties
      somethingElse: {},
    }
  }),
})
```

And then using them later on via the `info` argument inside some middlewares (using [`graphql-middleware`](https://github.com/prisma/graphql-middleware/)):
```js
const mutationField = info.schema.getMutationType().getFields()[info.fieldName];
console.log(mutationField.somethingElse);
```

For more details, see the following medium post: [graphql mutation arguments validation using yup](https://itnext.io/graphql-mutation-arguments-validation-with-yup-using-graphql-middleware-645822fb748)

---

The thing is, this is relying on internal behavior.
The following code spreads all properties given to the field:
https://github.com/graphql/graphql-js/blob/81719749e01f030cfb3a01a97e7e4bfc534bb08f/src/type/definition.js#L720-L724

Is that something expected to not change? If yes, then no need for any other extra property or for this issue. 😄 

But if this is something that can change in future versions, I would love the possibility of having an extra field for that extra metadata.

I'm available to work on adding this, if it's approved.

JCMais:
+1 for `extensions` 

Ideally libraries using this would allow to customize the key name, to reduce possibility of key collisions.

JCMais:
what can be done to move this forward?

sibelius:
I think the first part is to remove exact types on flow definitions

scpwebdev:
+1 this could help fix issues with join-monster that requires extra metadata for fields

flipxfx:
+1 this could help fix issues with join-monster that requires extra metadata for fields

19majkel94:
I would treat this exact Flow definition as a bug because this quote from @leebyron (https://github.com/graphql/graphql-js/issues/1262#issuecomment-368110891):

> You can include any arbitrary data on a GraphQLObjectType that you like - it's just a JS object after all. 

So I treat this pattern as a recomended way to have some metadata on types/fields like directives allows use to do in SDL.

IvanGoncharov:
> I would treat this exact Flow definition as a bug because this quote 

This quote is about actual `GraphQL*Type` not about config types like `GraphQL*TypeConfig`.

[...]

PR #2089, opened 2019-08-14
definition: replace object spread with explicit assignments

IvanGoncharov:
In preparation for #1527

### DIFF
--- a/src/type/__tests__/definition-test.js
+++ b/src/type/__tests__/definition-test.js
@@ -192,6 +192,10 @@ describe('Type System: Objects', () => {
       isDeprecated: true,
       name: 'bar',
       args: [],
+      astNode: undefined,
+      resolve: undefined,
+      subscribe: undefined,
+      description: undefined,
     });
   });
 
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
Also, it strangely produces significant speedup (almost 50%) for `buildSchema` and `buildClientSchema` functions:
image
image

taion:
@IvanGoncharov 

This is sort of a breaking change in practice. Because of the older spread-through, it was possible to define arbitrary other properties on fields, e.g. for https://github.com/4Catalyzer/graphql-validation-complexity#usage. This now no longer works.

It's true that this was undocumented behavior, but I also don't think it was quite that rare in practice.

IvanGoncharov:
@taion Can you please open an issue?
Can you switch to `extensions` mechanism added in #2097?

> but I also don't think it was quite that rare in practice.

If there are devs stuck with this hack and can't switch to `extensions` I don't have a problem to temporary revert it until `15.0.0`.
But it is 100% happening in `15.0.0` and everyone will need to switch to `extensions` anyway.

PR #2091, opened 2019-08-15
definitions: make constructed fields non-optional

IvanGoncharov:
Reflects changes in #2089

### DIFF
--- a/src/type/__tests__/predicate-test.js
+++ b/src/type/__tests__/predicate-test.js
@@ -573,77 +573,87 @@ describe('Type predicates', () => {
   });
 
   describe('isRequiredArgument', () => {
-    it('returns true for required arguments', () => {
-      const requiredArg = {
+    function buildArg(config) {
+      return {
         name: 'someArg',
-        type: GraphQLNonNull(GraphQLString),
+        description: undefined,
+        defaultValue: undefined,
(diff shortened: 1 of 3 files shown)


mjmahone:
Not sure I understand the logic here: it seems to make it strictly more difficult to construct AST nodes by hand, as you now need to explicitly include undefined fields.

IvanGoncharov:
@mjmahone It's not AST it's `GraphQLFields` and `GraphQLArgs`.
They not intended to be constructed by hand, for example, you construct `GraphQLFieldConfig` by hand and pass it to `GraphQLObjectType` constructor that convert it to `GraphQLField` internally.
So these changes shouldn't affect any external code since all fields (except introspection fields that are constructed by the library itself) should belong to some type and thus should be created by `GraphQLObjectType` or `GraphQLInterfaceType` constructor.

Why it's important: Having a stable set of properties significantly speed up execution since it allows V8 to reuse shadow classes: https://github.com/graphql/graphql-js/pull/2089#issuecomment-521247378

Since I plan to add `extensions` fields inside every `GraphQL*` type as part of #1527 I want to minimize the performance impact of this feature but having one stable set of properties per `GraphQL*` type.

mjmahone:
Thanks!

PR #2097, opened 2019-08-20
Add 'extensions' to all Type System objects

### LINKED ISSUE #1527: RFC: Extra property on field definition to pass extra metadata

IvanGoncharov:
Fixes #1527

### DIFF
--- a/src/jsutils/ObjMap.js
+++ b/src/jsutils/ObjMap.js
@@ -1,3 +1,9 @@
 // @flow strict
 
 export type ObjMap<T> = { [key: string]: T, __proto__: null, ... };
+export type ObjMapLike<T> = ObjMap<T> | { [key: string]: T, ... };
+
+export type ReadOnlyObjMap<T> = { +[key: string]: T, __proto__: null, ... };
+export type ReadOnlyObjMapLike<T> =
+  | ReadOnlyObjMap<T>
+  | { +[key: string]: T, ... };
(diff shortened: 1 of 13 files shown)


glen-84:
Is this documented anywhere?

PR #2105, opened 2019-08-22
Sync tstypes/errors with flow

JacksonKearl:
One remaining diff is that in TS, we define `formatError` as
```ts
export interface GraphQLFormattedError<
  TExtensions extends Record<string, any> = Record<string, any>
> {
  readonly message: string;
  readonly locations?: ReadonlyArray<SourceLocation>;
  readonly path?: ReadonlyArray<string | number>;
  readonly extensions?: TExtensions;
}
``` 

whereas in flow, we define it like
```
export type GraphQLFormattedError = {|
  +message: string,
  +locations: $ReadOnlyArray<SourceLocation> | void,
  +path: $ReadOnlyArray<string | number> | void,
  +extensions?: { [key: string]: mixed, ... },
|};
```

(comments stripped)

Don't have much flow expereince, but playing in the [playground](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBTAHgBzgJwBcxCBPHDMAcXwEMcALARQBkAxAgW1sMIwBMAovnwEAPABVBWPgDsAzgEs4CgFxgA3mADaAawyl18wvkWyA5gF11nRVgEAaMADpXYAL5gAvJp37DYMamFtZgtvb8Tq7OHmAAfGDemgA+qIkA1JwY8vK05hhGJmbmDmlg6fAAxjzKamAAJABKGLT8APKyMKQAgiK0pGLhAgnJYABucIqRZek4PAzqTS3tnT19A0HFYKOyAK6cAEYY+CPjk9MZ2HJKKvIA-OpSMhgKtfKlye4A3GWJf-8AwFA4EgxKoGAYYi0dQ0ejMdhcHh8IQiAjoCHEA4wuiMVgcfDcXgCYSifBiOKoIA), its saying that adding a type parameter is a breaking change (must use `GraphQLFormattedError<>`, even though the type has a defualt)

### DIFF
--- a/src/error/formatError.js
+++ b/src/error/formatError.js
@@ -22,9 +22,31 @@ export function formatError(error: GraphQLError): GraphQLFormattedError {
     : { message, locations, path };
 }
 
+/**
+ * @see https://github.com/graphql/graphql-spec/blob/master/spec/Section%207%20--%20Response.md#errors
+ */
 export type GraphQLFormattedError = {|
+  /**
+   * A short, human-readable summary of the problem that **SHOULD NOT** change
+   * from occurrence to occurrence of the problem, except for purposes of
+   * localization.
+   */
   +message: string,
+  /**
+   * If an error can be associated to a particular point in the requested
+   * GraphQL document, it should contain a list of locations.
+   */
   +locations: $ReadOnlyArray<SourceLocation> | void,
+  /**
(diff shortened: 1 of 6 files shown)


IvanGoncharov:
@JacksonKearl Merged 🎉 
Thanks a lot for separating `.d.ts` comments into a separate commit 👍

PR #2106, opened 2019-08-22
Sync execution TS definitions with Flow.

JacksonKearl:
Specifically:
Create `tstypes/jsutils/Path.d.ts` and pull some utility functions out to there (as is done in the implementation)
Add typeResolver options


IvanGoncharov:
@JacksonKearl Excellent work 👍 
Merged 🎉

PR #2108, opened 2019-08-22
Sync subscription TS definitions with Flow

JacksonKearl:


IvanGoncharov:
@JacksonKearl Thanks 👍 Merged 🎉

PR #2109, opened 2019-08-23
Sync type TS definitions with Flow

JacksonKearl:
Specifically:
Add `extensions` field in lots of places
Reorder fields

Both this and #2106 add Path.d.ts, but they're the same. If theres a merge conflict, just pick one.

### DIFF
--- a/tstypes/type/definition.d.ts
+++ b/tstypes/type/definition.d.ts
@@ -288,7 +288,7 @@ export class GraphQLScalarType {
   serialize: GraphQLScalarSerializer<any>;
   parseValue: GraphQLScalarValueParser<any>;
   parseLiteral: GraphQLScalarLiteralParser<any>;
-  extensions: Maybe<Record<string, any>>;
+  extensions: Maybe<Readonly<Record<string, any>>>;
   astNode: Maybe<ScalarTypeDefinitionNode>;
   extensionASTNodes: Maybe<ReadonlyArray<ScalarTypeExtensionNode>>;
   constructor(config: GraphQLScalarTypeConfig<any, any>);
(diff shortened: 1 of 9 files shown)


JacksonKearl:
This will need to wait until @2106 is in, then we can rebase off of it. Currently `ResponsePath` is undefined.

tstypes/type/scalars.d.ts
```diff
@@ -8,4 +8,4 @@ export const GraphQLID: GraphQLScalarType;
 
 export const specifiedScalarTypes: ReadonlyArray<GraphQLScalarType>;
 
-export function isSpecifiedScalarType(type: GraphQLScalarType): boolean;
+export function isSpecifiedScalarType(type: unknown): type is GraphQLScalarType;
```
IvanGoncharov: I thought `unknown` available only in TS 3.0?

tstypes/type/introspection.d.ts
```diff
@@ -40,6 +30,8 @@ export const SchemaMetaFieldDef: GraphQLField<any, any>;
 export const TypeMetaFieldDef: GraphQLField<any, any>;
 export const TypeNameMetaFieldDef: GraphQLField<any, any>;
 
-export const introspectionTypes: ReadonlyArray<any>;
+export const introspectionTypes: ReadonlyArray<
+  GraphQLObjectType | GraphQLEnumType
```
IvanGoncharov: I think it's better to make it just `GraphQLType` since we can latter add different types.

tstypes/type/definition.d.ts
```diff
@@ -285,13 +288,16 @@ export class GraphQLScalarType {
   serialize: GraphQLScalarSerializer<any>;
   parseValue: GraphQLScalarValueParser<any>;
   parseLiteral: GraphQLScalarLiteralParser<any>;
+  extensions: Maybe<Record<string, any>>;
```
IvanGoncharov: Can we make them `ReadOnly`?

JacksonKearl: Yep!

tstypes/type/introspection.d.ts
JacksonKearl: Got ti

tstypes/type/scalars.d.ts
JacksonKearl: Moved it to `any`

JacksonKearl:
@IvanGoncharov should be good to go now

IvanGoncharov:
@JacksonKearl Merged 🎉

PR #2115, opened 2019-08-23
Sync utilities TS definitions with flow

JacksonKearl:
Removed a ton from `tstypes/utilities/findBreakingChanges.d.ts`.


IvanGoncharov:
@JacksonKearl Great job 👍

PR #2116, opened 2019-08-23
Copy non-standard properites from field's config

### LINKED ISSUE #2114: #2089 makes a breaking change to field config behavior
By no longer passing through other unknown properties to field configurations, #2089 introduces a small breaking change to behavior.

Previously, integrations like https://github.com/4Catalyzer/graphql-validation-complexity#usage could enable additional field configuration just by specifying other fields. Now, against 14.5.0, this no longer works.

It would be great if we could temporarily revert this change, then increment the major version to v15 to make this change.

IvanGoncharov:
Fixes #2114
Partly reverts #2089

### DIFF
--- a/src/type/definition.js
+++ b/src/type/definition.js
@@ -791,6 +791,7 @@ function defineFieldMap<TSource, TContext>(
     }));
 
     return {
+      ...fieldConfig,
       name: fieldName,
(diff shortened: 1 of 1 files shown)


taion:
Thanks!

PR #2120, opened 2019-08-24
Fix relative imports

JacksonKearl:
Mistakenly imported some items using incorrect relative imports in the pervious TS merger. This fixes that.


IvanGoncharov:
@JacksonKearl Thanks for such a quick fix.
Interesting question: Why did `dtslint` didn't detect this on CI?

PR #2135, opened 2019-08-26
tstypes: Use `any` as BREAK type.

JacksonKearl:
Otherwise, TS will reposrt errors when a visit function only conditionally returns BREAK (not all paths return a value error)


IvanGoncharov:
@JacksonKearl Thanks 👍

jLouzado:
was setting the type or `null` not an option? Then your visit function return type would be `<type> |  null`

PR #2136, opened 2019-08-26
fix(flatMap): suppress flow error from second reference to flatMap

jedwards1211:
Gives me an error on Flow 0.84.0 at least


IvanGoncharov:
@jedwards1211 Thanks for reporting looking into it right now.

PR #2137, opened 2019-08-27
Flow: fix errors in 'flatMap.js' on old Flow versions

IvanGoncharov:
Fixes #2136

### DIFF
--- a/src/polyfills/flatMap.js
+++ b/src/polyfills/flatMap.js
@@ -5,11 +5,14 @@ declare function flatMap<T, U>(
   fn: (item: T, index: number) => $ReadOnlyArray<U> | U,
 ): Array<U>;
 
+// Workaround to make older Flow versions happy
+const flatMapMethod = (Array.prototype: any).flatMap;
+
 /* eslint-disable no-redeclare */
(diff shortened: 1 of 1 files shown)


src/polyfills/flatMap.js
```diff
@@ -5,11 +5,14 @@ declare function flatMap<T, U>(
   fn: (item: T, index: number) => $ReadOnlyArray<U> | U,
 ): Array<U>;
 
+// Workaround to make older Flow versions happy
+const flatMapMethod = (Array.prototype: any).flatMap;
```
IvanGoncharov: @jedwards1211 We enabled Flow warnings in config:
https://github.com/graphql/graphql-js/blob/8ab65246bbd91527f28b5e98c3ab6946701ecea4/.flowconfig#L35
So flow generate warnings for ignored `$FlowFixMe`.
It also possible that other projects using `include_warnings=true`
So I decided to just use `any` since it works in all Flow versions.

IvanGoncharov: I released `14.5.4` :package: with this fix included.
Sorry for the delay.

PR #2142, opened 2019-08-28
tstypes: Mark `options` property optional in `getVariableValues`

JacksonKearl:


IvanGoncharov:
@JacksonKearl Thank you 👍

Issue #869, opened 2017-05-19
printSchema loses @directive information


jzimmek:
Hi, the printSchema function in "graphql/utilities/schemaPrinter" seems to forget the directives on field definitions.

```
import {buildSchema} from  "graphql/utilities/buildASTSchema"
import {printSchema} from  "graphql/utilities/schemaPrinter"

const schema = buildSchema(`
  type Person {
    id: ID!
    name: String @skip(if: true)
  }
  type Query {
    people: [Person]
  }
`)

console.info(printSchema(schema))
```

current output:

```
type Person {
  id: ID!
  name: String
}

type Query {
  people: [Person]
}
```

expected output:

```
type Person {
  id: ID!
  name: String @skip(if: true)
}
type Query {
  people: [Person]
}
```

Tested against graphql version: 0.9.6

I already digged a bit deeper into schemaPrinter and it seems that there is some special handling for directives going on:

1.) printSchema filters directives down to "skip", "include" and "deprecated"
2.) @ deperected seems to be handled 

This is maybe better addressed in a separate issue, but I would like to add my own custom directives and get them also properly handled by printSchema.

Would you consider a PR for proper handling of spec directives as well as custom ones?

leebyron:
This is correct behavior. Directives are artifacts of the GraphQL language, and not properties of a GraphQL schema. While directives are allowed to be used in the GraphQL schema language, they exist as language-level metadata which different tools may interpret in different ways.

For example, as you pointed out `@deprecated` is a directive which the schema generation can use to fill in the actual field deprecation information in the schema. Tools like graph.cool or other code generation service often use these directives to describe what kind of resolvers or data storage to use when treating this document as a source of truth.

However the opposite is not the case. Since a schema is an in-memory data structure artifact, and not simply a representation of the GraphQL language, it does not have a concept of directives on its own. Though utilities like `printSchema()`, which produce GraphQL language from a GraphQL schema, may use directives to express other concepts, the one that it currently uses is `@deprecated`.

However to your example, `@skip` applied to a type definition has no semantic meaning in GraphQL (in fact, if you ran `validate()` on this document, it would likely point this out), and thus is not used when producing the GraphQL schema. When you then use `printSchema()` to produce the text again, since the directive had no semantic meaning and wasn't used, it doesn't appear in the printed output.

mrjj:
[...]

Issue #2020, opened 2019-07-08
Allow schemaPrinter to be customized / extended.


j:
Some libraries are having to copy the schemaPrinter code to do custom stuff to the output, (`@apollo/federation`, `type-graphql`).  These libraries require outputting type and field level directives since this isn't supported in this library.  Is there any support to do so, or can the schemPrinter just export the other functions so we don't have to copy every function or, even better, create a SchemaPrinter class so everything can be over-ridden, then `printSchema` options can set the printer for BC:

```javascript

type Options = {
    commentDescriptions?: boolean,
    printer?: SchemaPrinter
};

const defaultPrinter = new SchemaPrinter();

export function printSchema(schema: GraphQLSchema, options?: Options): string {
    const printer = options.printer || defaultPrinter;

    return printer.print(
        schema,
        n => !isSpecifiedDirective(n),
        isDefinedType,
        options,
    );
}

export function printIntrospectionSchema(
    schema: GraphQLSchema,
    options?: Options,
): string {
    const printer = options.printer || defaultPrinter;

    return printer.print(
        schema,
        isSpecifiedDirective,
        isIntrospectionType,
        options,
    );
}
```

mjmahone:
Instead of making an extensible printer, would it solve your issue if directives on schema definitions were included in the print step?

I believe this is actually a bug where we forgot to add directive printing for `printSchema` when we added directives on type and field definitions. I'd welcome both a PR for test cases that are currently missing and a PR for fixing the issue in the printer.

IvanGoncharov:
@j @mjmahone I don't think we should print directives based on `astNode` properties since directives are tight to SDL context. For example:
```graphql
type Foo {
  foo: String
}

type Bar @copyFields(fromType: "Foo") {
  bar: String
}
```
Assuming `@copyFields` directive successfully modified the schema and assuming we implemented your proposal output will be like that:
```graphql
type Foo {
  foo: String
}

type Bar @copyFields(fromType: "Foo") {
  foo: String
  bar: String
}
```
That doesn't make sense in the context of `@copyFields` or any other directive that do something more complicated than attaching metadata.

I think the correct solution would be to design `GraphQLSDLDirective` (derived from `GraphQLDirective`) and allow you to register callbacks for transforming types/fields (including `extensions` as described in #1527) and another callback that called during schema printing where directive can decide to return `DirectiveNode` for particular type/field (probably base on `extensions`).

[...]

PR #2145, opened 2019-08-31
export function printFilteredSchema

Zemnmez:
It is useful to be able to print directives that are part of graphql's representation of the schema. Previous discussions on this have [noted](https://github.com/graphql/graphql-js/issues/869#issuecomment-309899380) that printing directives as part of a schema doesn't make sense for a publicly accessible schema as they're not true parts of the 'graph' graphql represents.

However, I'm currently working with code generation tools, and not being able to print the full in-memory schema into canonical form is a deal breaker. `printFilteredSchema` already exposes the ability to adjust the way the schema is printed so this PR simply exposes it.

### DIFF
--- a/src/utilities/schemaPrinter.js
+++ b/src/utilities/schemaPrinter.js
@@ -79,7 +79,7 @@ function isDefinedType(type: GraphQLNamedType): boolean {
   return !isSpecifiedScalarType(type) && !isIntrospectionType(type);
 }
 
-function printFilteredSchema(
+export function printFilteredSchema(
   schema: GraphQLSchema,
   directiveFilter: (type: GraphQLDirective) => boolean,
   typeFilter: (type: GraphQLNamedType) => boolean,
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@Zemnmez Issue that you referenced is about printing directive usage, not directive definitions.
The only thing that `printFilteredSchema` can do is to additionally print is definitions of `@skip`, `@include` and `@deprecate`:
https://github.com/graphql/graphql-js/blob/0d2220f0a64238473f4e1c9aa8f73f891d0fd3e0/src/utilities/__tests__/schemaPrinter-test.js#L508-L532

So are you sure that you want directive definitions and not directive usage?

Zemnmez:
@IvanGoncharov oh... you're right. Is there a better way of doing what I'm trying to do?

IvanGoncharov:
@Zemnmez Proper solution is discussed in #2020 
Currently, we are working on porting this library to TypeScript so we in a feature freeze until it's done. 

As a temporary workaround you can do (**NOT TESTED**):
```js
const sdl = printSchema(schema);
let currentType;
const modifiedAST = visit(parse(sdl), {
  ObjectDefinition(node) {
    currentType = schema.getType(node.name.value);
  },
  Field(node) {
    const fieldDef = currentType.getFields()[node.name.value];
    return {
      ...node,
      directives: fieldDef.astNode,
    };
  },
});
const sdlWithDirectives = print(modifiedAST);
```

Zemnmez:
thanks!

PR #2151, opened 2019-09-02
findDangerousChanges: sort fields inside 'defaultValue'

### LINKED ISSUE #2150: findDangerousChanges doesn't work as expected after lexicographic sort and deserialization
Hello! My colleague reported a bug where `findDangerousChanges` reports invalid dangerous change after serializing the schema with `lexicographicSortSchema` and then deserializing it again. I tried to reproduce the error here: https://github.com/kiwicom/graphql-bc-checker/commit/f5fa6be30ab5f0569e6418f75f9425f724a0b7a1#diff-b3245f2dbca2b4c14d7bc1ff7eef9c87

Note that we are using one schema which has unsorted args, then printing it into a string and later trying to compare this string (potentially saved on disk) with the actual schema. It should not report any dangerous changes but it does report this:

```json
{
  "description": "RootQueryType.xxx arg yyy has changed defaultValue from {aaa: \\"aaa\\", bbb: \\"bbb\\"} to {bbb: \\"bbb\\", aaa: \\"aaa\\"}.",
  "type": "ARG_DEFAULT_VALUE_CHANGE",
}
```

Thanks for checking it! :)

IvanGoncharov:
Fixes #2150

### DIFF
--- a/src/utilities/findBreakingChanges.js
+++ b/src/utilities/findBreakingChanges.js
@@ -7,6 +7,7 @@ import inspect from '../jsutils/inspect';
 import invariant from '../jsutils/invariant';
 
 import { print } from '../language/printer';
+import { visit } from '../language/visitor';
 
 import { type GraphQLSchema } from '../type/schema';
 import {
@@ -395,6 +396,9 @@ function findArgChanges(
           description: `${oldType.name}.${oldField.name} arg ${oldArg.name} defaultValue was removed.`,
         });
       } else {
+        // Since we looking only for client's observable changes we should
+        // compare default values in the same representation as they are
+        // represented inside introspection.
         const oldValueStr = stringifyValue(oldArg.defaultValue, oldArg.type);
         const newValueStr = stringifyValue(newArg.defaultValue, newArg.type);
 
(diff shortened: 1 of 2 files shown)


src/utilities/findBreakingChanges.js
```diff
@@ -395,6 +396,9 @@ function findArgChanges(
           description: `${oldType.name}.${oldField.name} arg ${oldArg.name} defaultValue was removed.`,
         });
       } else {
+        // Since we looking only for client's observable changes we should
+        // compare default values in the same representation as they are
+        // represented inside introspection.
```
IvanGoncharov: > For input type default values we apparently just compare it's stringified value instead of doing a deep comparison.

@Neitsch It's intentional since we want to detect only changes observable by the client.
So we should compare `defaultValues` in the same representation as used by introspection:
https://github.com/graphql/graphql-js/blob/0d2220f0a64238473f4e1c9aa8f73f891d0fd3e0/src/type/introspection.js#L337-L345

It's important because `defaultValue` is always specified in "internal" format (same format as returned by `parseLiteral`/`parseValue` and passed into `serialize`) that mean it can be for example instances of `Date` class and `===` wouldn't work for them but if you pass both values into `serialize` you can compare resulting strings without any problems.

IvanGoncharov:
@mrtnzlml @Neitsch Thanks for a quick review 👍

mrtnzlml:
Thanks for the fix. It was really fast! :)

IvanGoncharov:
@mrtnzlml Released in `v14.5.5` 🎉 sorry for the delay.

PR #2169, opened 2019-09-14
Make onError optional in SDLValidationContext

Cito:
The onError argument was probably meant to be optional,
just like in ASTValidationContext and ValidationContext.


IvanGoncharov:
Thanks 👍Merged 🎉 
Note: in `15.0.0` it will become required see:
https://github.com/graphql/graphql-js/pull/2130/files

Cito:
Because the error collection will not be done automatically any more, so you will always need this, right?

PR #2181, opened 2019-09-17
Replace 'find{Breaking,Dangerous}Changes' with generic 'findSchemaChanges'

wtrocki:
# Motivation

Currently to find potentially Breaking and Dangerous changes developers need to call 2 separate methods and execute logic twice. This change simply exposes the underlying method that will give access to both with single execution logic.

### DIFF
--- a/src/utilities/__tests__/findBreakingChanges-test.js
+++ b/src/utilities/__tests__/findBreakingChanges-test.js
@@ -14,11 +14,10 @@ import { buildSchema } from '../buildASTSchema';
 import {
   BreakingChangeType,
   DangerousChangeType,
-  findBreakingChanges,
-  findDangerousChanges,
+  findSchemaChanges,
 } from '../findBreakingChanges';
 
-describe('findBreakingChanges', () => {
+describe('findSchemaChanges', () => {
   it('should detect if a type was removed or not', () => {
     const oldSchema = buildSchema(`
       type Type1
@@ -28,13 +27,13 @@ describe('findBreakingChanges', () => {
     const newSchema = buildSchema(`
       type Type2
     `);
-    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
       {
         type: BreakingChangeType.TYPE_REMOVED,
         description: 'Type1 was removed.',
       },
     ]);
-    expect(findBreakingChanges(oldSchema, oldSchema)).to.deep.equal([]);
+    expect(findSchemaChanges(oldSchema, oldSchema)).to.deep.equal([]);
   });
 
   it('should detect if a type changed its type', () => {
@@ -49,7 +48,7 @@ describe('findBreakingChanges', () => {
       union TypeWasInterfaceBecomesUnion
       input TypeWasObjectBecomesInputObject
     `);
-    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
       {
         type: BreakingChangeType.TYPE_CHANGED_KIND,
         description:
@@ -119,7 +118,7 @@ describe('findBreakingChanges', () => {
       }
     `);
 
-    const changes = findBreakingChanges(oldSchema, newSchema);
+    const changes = findSchemaChanges(oldSchema, newSchema);
     expect(changes).to.deep.equal([
       {
         type: BreakingChangeType.FIELD_REMOVED,
@@ -216,7 +215,7 @@ describe('findBreakingChanges', () => {
       }
     `);
 
-    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
       {
         type: BreakingChangeType.FIELD_REMOVED,
         description: 'InputType1.field2 was removed.',
@@ -281,11 +280,21 @@ describe('findBreakingChanges', () => {
       }
     `);
 
-    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
       {
         type: BreakingChangeType.REQUIRED_INPUT_FIELD_ADDED,
         description:
           'A required field requiredField on input type InputType1 was added.',
+      },
+      {
+        description:
+          'An optional field optionalField1 on input type InputType1 was added.',
+        type: 'OPTIONAL_INPUT_FIELD_ADDED',
+      },
+      {
+        description:
+          'An optional field optionalField2 on input type InputType1 was added.',
+        type: 'OPTIONAL_INPUT_FIELD_ADDED',
       },
     ]);
   });
@@ -306,7 +315,11 @@ describe('findBreakingChanges', () => {
       union UnionType1 = Type1 | Type3
     `);
 
-    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
+      {
+        description: 'Type3 was added to union type UnionType1.',
+        type: 'TYPE_ADDED_TO_UNION',
+      },
       {
         type: BreakingChangeType.TYPE_REMOVED_FROM_UNION,
         description: 'Type2 was removed from union type UnionType1.',
@@ -331,7 +344,11 @@ describe('findBreakingChanges', () => {
       }
     `);
 
-    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
+      {
+        description: 'VALUE3 was added to enum type EnumType1.',
+        type: 'VALUE_ADDED_TO_ENUM',
+      },
       {
         type: BreakingChangeType.VALUE_REMOVED_FROM_ENUM,
         description: 'VALUE1 was removed from enum type EnumType1.',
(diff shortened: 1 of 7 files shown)


src/utilities/findBreakingChanges.js
```diff
@@ -102,7 +102,11 @@ export function findDangerousChanges(
   return ((dangerousChanges: any): Array<DangerousChange>);
 }
 
-function findSchemaChanges(
+/**
+ * Given two schemas, returns an Array containing descriptions of all the types
+ * of potentially breaking change and dangerous change
+ */
+export function findSchemaChanges(
```
IvanGoncharov: @wtrocki Totally reasonable 👍 
Moreover, I think it's confusing to have multiple functions that basically do the same so 
I think `findSchemaChanges` should be a function to rule them all.

Since we working on new major release `15.0.0` I think its good time to remove `findDangerousChanges` and `findBreakingChanges`. And merge `BreakingChange | DangerousChange` into single `SchemaChange` type.

Can you please do that?

Also, don't forget to update TS typings in `tstypes` folder.
Another thing all public APIs should be exported through `src/index.js` and `src/utilities/index.js`.
https://github.com/graphql/graphql-js/blob/master/src/README.md

IvanGoncharov:
It would also allow merging #1127 by @eapache without adding new API functions.

wtrocki:
Working on this now

wtrocki:
Done

wtrocki:
What is the policy on documenting breaking changes/providing migration paths?

IvanGoncharov:
@wtrocki These functions are used mostly in the build scripts and CLI tools so migration path should be pretty easy. We have a script that generates changelog that will list all breaking changes. Would be great if you write a small note about this change and how to implement `findBreakingChange`/`findDangerousChange` on top of `findSchemaChange` and after release, I will manually add this note into the generated changelog.

src/utilities/findBreakingChanges.js
```diff
@@ -62,50 +62,20 @@ export const DangerousChangeType = Object.freeze({
...
-export type BreakingChange = {
-  type: $Keys<typeof BreakingChangeType>,
-  description: string,
-  ...
-};
-
-export type DangerousChange = {
-  type: $Keys<typeof DangerousChangeType>,
+export type SchemaChange = {
+  type: $Keys<typeof BreakingChangeType> | $Keys<typeof DangerousChangeType>,
```
IvanGoncharov: It's better to create common `SchemaChangeType`:
```js
export const SchemaChangeType = Object.freeze({
  ...BreakingChangeType,
  ... DangerousChangeType,
});
```

src/utilities/__tests__/findBreakingChanges-test.js
```diff
@@ -281,12 +280,22 @@ describe('findBreakingChanges', () => {
       }
     `);
 
-    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
```
IvanGoncharov: @wtrocki Since you effectively merging tests for `findBreakingChanges` and `findDangerousChanges` please adjust tests accordingly.
For example, rename this one to `should detect if a field is added to an input type` and remove the duplicating test:
https://github.com/graphql/graphql-js/blob/06ef5db069cf2f896c51126973c6100a33e1ea7c/src/utilities/__tests__/findBreakingChanges-test.js#L1026-L1047

IvanGoncharov: @wtrocki Please do the same for other tests where you modify snapshot.

src/utilities/findBreakingChanges.js
wtrocki: Perfect. I was thinking about it

mjmahone:
I want to add a note of warning: the purpose of "findBreakingChanges" is to find only those changes to the schema that could potentially break old versions of  clients that are shipped and can never be updated.

A function called `findSchemaChanges` needs to have different behavior: it needs to say whether a type was added or removed, whether there's a new field, etc.

This is OK, and even desireable to  have. But I believe we should keep `findDangerousChanges`, or have a DANGEROUS_CHANGES const that we can use with `findSchemaChanges` to preserve the functionality of identifying only those changes that will cause backwards-compatibility issues with existing, in production applications. Also something to note: the "potentially dangerous" changes subset basically never causes us problems, at least at Facebook. We build our applications and our infrastructure surrounding GraphQL in a way where they do not cause breakages, though if you get it wrong (for instance, using exhaustive switch statements with enums), these class of  changes will cause problems.

If the intention continues to be finding only those changes that *may* cause problems, we should have `findSchemaChanges` act more like our validation rules: you can pass in a list of changes to find  (with the "dangerous" set, or the  "known to be breaking" set, or "all" set). This is something  I'd heartily support. But again, if we name it `findSchemaChanges`, it needs to be able to find *all* schema changes, not just those that might be breakages.

IvanGoncharov:
> This is OK, and even desirable to have. But I believe we should keep findDangerousChanges, or have a DANGEROUS_CHANGES const that we can use with findSchemaChanges to preserve the functionality 

@mjmahone Thanks for review 👍 It's exactly the plan since we already expose constants for both breaking and dangerous changes:
https://github.com/graphql/graphql-js/blob/06ef5db069cf2f896c51126973c6100a33e1ea7c/src/utilities/findBreakingChanges.js#L38-L63

So it's just a matter of writing one-liner like this:
```js
import { BreakingChangeType, findSchemaChanges } from 'graphql';

function findBreakingChanges(oldSchema, newSchema) {
  return findSchemaChanges(oldSchema, newSchema)
    .filter(({type}) => type in BreakingChangeType);
}
```

Also both constant is already part of public API and exposed through main `index.js`:
https://github.com/graphql/graphql-js/blob/06ef5db069cf2f896c51126973c6100a33e1ea7c/src/index.js#L403-L405
And there is no plans to ever remove those constants.

@mjmahone @Neitsch Does this migration path address your concerns?

wtrocki:
The awesome point from @mjmahone
To extend it there are two use cases:

1) Comparing changes and see what was added/removed (with metadata about what and where)
2) Finding Breaking and Dangerous Changes 

My initial intention was to expose a single function that can be used to perform option nr 2 by calling a single function, so people can get both at the same time with single method call and then decide on their own what they define as dangerous etc.

As far I checked both enums do not comprehensively represent all possible changes that can happen so option nr1 will require separate method. I have done some research how such implementation could look like and found out that there is already a community package that does both:
https://github.com/kamilkisiela/graphql-inspector

With a generic change interface:
https://github.com/kamilkisiela/graphql-inspector/blob/33ca4f9f37c5b6bedba6b6ef190440777050cc9c/packages/core/src/diff/changes/change.ts#L1-L59

IvanGoncharov:
@wtrocki So the idea is that we need to merge other types of schema changes in future, e.g. #1127. Classifying all changes into non-overlapping categories can be tricky so instead, we can simply do: 
```js
export const SchemaChangeType = Object.freeze({
  ...BreakingChangeType,
  ... DangerousChangeType,
  TYPE_ADDED: 'TYPE_ADDED',
  // All other type of changes
});
```

> return findSchemaChanges(oldSchema, newSchema, {types: BeakingChangeTypes})

This approach has a few problems:
1) For very big schemas it preferable to get all changes in single iteration and it will look awkward:
```js
const changes = findSchemaChanges(oldSchema, newSchema, {types: {
   ...BeakingChangeTypes,
   ...DangerousChangeTypes,
}})
for (const change of changes) {
  if (change in BreakingChangeTypes) {
    reportError(change);
  } else if (change in DangerousChangeTypes) {
    reportWarning(change);
  }
}
```

2) We need to put every change in some category and changing it from one category into another would be a breaking change.

wtrocki:
I will wait for consensus and then apply the required changes.

mjmahone:
I'm a big fan of this having an almost identical API as `validate`, i.e. `(original: GraphQLSchema, updated: GraphQLSchema, findChanges: $ReadOnlyArray<FindChangeFn> = ALL_SCHEMA_CHANGES)`. Though if we went that far, it might make sense to have it operate over an SDL instead of over a GraphQLSchema. We could always have `findSchemaChanges` be an internal function to start, and re-implement `findBreakingChanges` and `findDangerousChanges` using `findSchemaChanges`.

In short: I'm happy with adding a `findSchemaChanges` function that can be used in the future to find all schema changes. I think  we should *definitely* push this new function into the repo, even if it's only internal, so it's easier to iterate on. I don't have strong opinions on what *exactly* the API is, but I do think it should have the business logic for finding changes, and at least allow you the flexibility to choose whether you want  to  find all changes, breaking-only changes, or dangerous-only changes. Having a solid  API that provides even more flexibility would be even more desireable, but we don't, IMO, need to block getting something in until we have that ideal API.

tstypes/utilities/index.d.ts
```diff
@@ -104,10 +104,8 @@ export { assertValidName, isValidNameError } from './assertValidName';
 export {
   BreakingChangeType,
   DangerousChangeType,
-  findBreakingChanges,
```
mjmahone: A lot of tooling currently relies on only getting the schema changes. I think all  the code changes from above are great, but until we make `findSchemaChanges` more flexible, we should keep these utility functions, to avoid forcing people to rewrite their code significantly in order to upgrade.

IvanGoncharov:
> I'm happy with adding a findSchemaChanges function that can be used in the future to find all schema changes. I think we should definitely push this new function into the repo, even if it's only internal, so it's easier to iterate on. I don't have strong opinions on what exactly the API is, but I do think it should have the business logic for finding changes, and at least allow you the flexibility to choose whether you want to find all changes, breaking-only changes, or dangerous-only changes.

@mjmahone So just to clarify you oppose exposing `findSchemaChanges` in its current form?
(This is a core change in this PR, all the rest is just consequences of removing `findBreakingChanges` and `findDangerousChanges`).

Personally I think there is no big difference in filtering after `findSchemaChanges` and splitting change detection into separate functions will increase significantly code size/complexity and significantly slow down change detection on big schemas(since you need to do whole schema iteration in every rule).

But I'm totally ok with an alternative implementation if code size/performance will stay the same.

JoviDeCroock:
Implemented in #4235

PR #2186, opened 2019-09-19
Consistent error messages in buildClientSchema

spawnia:
Always:
- put a space after :
- terminate with a .


IvanGoncharov:
@spawnia Not sure why Travis doesn't update status but build is broken on this PR:
https://travis-ci.org/graphql/graphql-js/jobs/587147923#L235
Can you please run `npm run prettier` and also `npm run test`?

src/utilities/buildClientSchema.js
```diff
@@ -238,7 +240,8 @@ export function buildClientSchema(
     if (!objectIntrospection.interfaces) {
       throw new Error(
         'Introspection result missing interfaces: ' +
-          inspect(objectIntrospection),
+          inspect(objectIntrospection) +
+          '.',
```
IvanGoncharov: It's better to have it like `Introspection result missing interfaces: ${inspect(objectIntrospection)}.` but only if it will fit in 80 chars limit and prettier will keep it as a single line. 
If not you can do:
```
const objectInstrospectionStr = inspect(objectIntrospection);
throw new Error(`Introspection result missing interfaces: ${objectInstrospectionStr}.`);
```
Please do it for all other error messages you changed except for `devCheck` since long term plan is to remove it in production build so it should be self-contained.

spawnia:
I went all in on template strings and extracted where needed to keep the line count low and prettier happy.

spawnia:
@IvanGoncharov Resolved the conflicts and modified the PR according to your suggestions.

src/utilities/buildClientSchema.js
```diff
@@ -381,13 +381,14 @@ export function buildClientSchema(
   }
 
   function buildDirective(directiveIntrospection) {
-    const directiveIntrospectionStr = inspect(directiveIntrospection);
```
IvanGoncharov: @spawnia `inspect` is not a zero-cost some we shouldn't call it unless we throw an error.
I think in this case is better to duplicate code.

spawnia: Good point, makes sense.

src/utilities/__tests__/buildClientSchema-test.js
```diff
@@ -604,12 +604,12 @@ describe('Type System: build schema from introspection', () => {
     it('throws when introspection is missing __schema property', () => {
       // $DisableFlowOnNegativeTest
       expect(() => buildClientSchema(null)).to.throw(
-        'Invalid or incomplete introspection result. Ensure that you are passing "data" property of introspection response and no "errors" was returned alongside: null',
+        'Invalid or incomplete introspection result. Ensure that you are passing "data" property of introspection response and no "errors" was returned alongside: null.',
```
IvanGoncharov: @spawnia I change tests to check for a dot at the end of sentences.

src/utilities/__tests__/buildClientSchema-test.js
```diff
@@ -29,10 +29,10 @@ import { introspectionFromSchema } from '../introspectionFromSchema';
 /**
  * This function does a full cycle of going from a string with the contents of
  * the SDL, build in-memory GraphQLSchema from it, produce a client-side
- * representation of the schema by using "buildClientSchema"and then finally
- * printing that that schema into the SDL
+ * representation of the schema by using "buildClientSchema" and then
+ * returns that schema printed as SDL.
```
IvanGoncharov: Not a big deal: but for future PRs it would be great to keep grammar fixes as separate PR.

IvanGoncharov:
@spawnia Merged 🎉
Sorry for the delay had a really busy month 😞 
I will definitely react more quickly to the next PRs.

PR #2194, opened 2019-09-23
Correct isSpecifiedScalarType refinement, fix #2153

### LINKED ISSUE #2153: TS: isSpecifiedScalarType incorrectly refines type

tgriesser:
Fixes #2153

Ran into this when looking to upgrade to the latest in [graphql-nexus](https://github.com/prisma/nexus). To reproduce:

```ts
if (isScalarType(type)) {
  if (isSpecifiedScalarType(type)) {
    rootTypeMap[type.name] = SpecifiedScalars[type.name]
  } else {
    rootTypeMap[type.name] = "any"
  }
}
```

Before the changes in this PR, `type` in the `else` block is inferred to be `never`, because `isSpecifiedScalarType` is currently narrows to all scalars rather than just the "specified" ones, and as such `type.name` is invalid:

Screen Shot 2019-09-22 at 11 21 55 PM

By adding the `& { name: '...' }` for each of the specified scalars, we give enough context to know that the `else` block is a valid codepath for all "non-specified" scalars.

### DIFF
--- a/tstypes/type/scalars.d.ts
+++ b/tstypes/type/scalars.d.ts
@@ -13,6 +13,6 @@ export type SpecifiedScalarType =
   | typeof GraphQLBoolean
   | typeof GraphQLID;
 
-export const specifiedScalarTypes: ReadonlyArray<GraphQLScalarType>;
+export const specifiedScalarTypes: ReadonlyArray<SpecifiedScalarType>;
 
 export function isSpecifiedScalarType(type: any): type is SpecifiedScalarType;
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
Had a deep and long conversation with @tgriesser that helped me to formulate my position so to summaries it here:
* It's typescript specific refinement and it's not about syntax, entire idea of refining possible value of `string` to `Int | Float | ...` is TS specific. Which is problematic for `graphql-js` goal of being a reference implementation that should be easily portable into other languages.
* It gives a false sense of security: if we add new scalar (e.g. `GraphQLDateTime` discussed in https://github.com/graphql/graphql-spec/issues/579) TS code that was validated against older versions will fail even through adding new standard scalar is a non-breaking change.

PR #2195, opened 2019-09-24
tstypes: fix typings for 'isSpecifiedDirective'/'isSpecifiedScalarType'

### LINKED ISSUE #2153: TS: isSpecifiedScalarType incorrectly refines type

IvanGoncharov:
Fixes #2153


IvanGoncharov:
Since we still discussing a long-term solution in #2194 vs #2196 for `14.x.x` branch it would be best to just revert to `boolean` since it was what we had in `DefinetlyTyped` anyway:
https://github.com/graphql/graphql-js/blob/d1391eea86caf030f1945a518aeadb6e2d3e7c20/tstypes/type/scalars.d.ts#L11

PR #2196, opened 2019-09-24
Change 'isSpecified*' & 'isIntrospectionType' accept only correct types

### LINKED ISSUE #2153: TS: isSpecifiedScalarType incorrectly refines type

IvanGoncharov:
Fixes #2153
Previous disscussion: https://github.com/graphql/graphql-js/pull/1837#issuecomment-488250420

### DIFF
--- a/src/type/__tests__/predicate-test.js
+++ b/src/type/__tests__/predicate-test.js
@@ -157,35 +157,6 @@ describe('Type predicates', () => {
     it('returns false for custom scalar', () => {
       expect(isSpecifiedScalarType(ScalarType)).to.equal(false);
     });
-
-    it('returns false for scalar class (rather than specified instance)', () => {
-      expect(isSpecifiedScalarType(GraphQLScalarType)).to.equal(false);
-    });
-
-    it('returns false for wrapped specified scalar', () => {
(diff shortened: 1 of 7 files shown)


IvanGoncharov:
An alternative solution to #2194 
I feel like we made a mistake in designing `isSpecifiedDirective`, `isSpecifiedScalarType` and `isIntrospectionType` predicates because we mixed type refinement and value refinement. For example in case of `isSpecifiedScalarType` we are mixing `type instanceof GraphQLScalarType`(refines type) and `type.name === ...`(refines value). 
So I think #2153 is not unique to TS but a symptom of the general problem here.

@tgriesser Would be great to get your opinion?
@Cito We already discussed `isSpecifiedDirective` in #1837 would be great to get your opinion on this change?

IvanGoncharov:
Also, we have other value predicates for example `isRequeiredArgument`:
https://github.com/graphql/graphql-js/blob/ed0dfe8afb92d942e20a9b254c87dee01181a0f3/src/type/definition.js#L944-L946

https://github.com/graphql/graphql-js/blob/ed0dfe8afb92d942e20a9b254c87dee01181a0f3/src/type/definition.js#L1464-L1468

PR #2198, opened 2019-09-25
Add two more tests to KnownArgumentNames validation rule

Cito:
Tests that arguments passed to an argument-less directive are properly reported as an error.

The test suite did not catch that this did not work properly in the Python port, so I added these tests.


IvanGoncharov:
@Cito Thanks a lot 👍

PR #2205, opened 2019-09-27
Document private functions with @internal tag

### LINKED ISSUE #2183: Annotate all internal functions with '@internal'
Even though some functions are expored from `src/**/*.js` files they are private.
The only function exported by `src/index.js` and `src/*/index.js` are considered to be part of the public API. Please see:
https://github.com/graphql/graphql-js/blob/master/src/README.md

It would be great to have all private API marked as such, e.g.:
https://github.com/graphql/graphql-js/blob/6f341a3ae0d6b3cc4d9136a58fd9272c615c2dba/tstypes/validation/validate.d.ts#L32-L33

Also, it would be great to find all public APIs without JS doc comments.

craicoverflow:
Closes #2183

Manually added an `@internal` JSDoc tag to exported functions that are not part of the public API.

I began by by making an ESLint rule which lives in the `resources` directory.

Although this helped me to identify the exported functions which are not exported in `src/index.js`, the rule is not yet stable enough to be included and I may look to follow this up with a separate PR, dealing exclusively with the ESLint rule.

### DIFF
--- a/src/__tests__/starWarsData.js
+++ b/src/__tests__/starWarsData.js
@@ -122,6 +122,8 @@ function getCharacter(id) {
 
 /**
  * Allows us to query for a character's friends.
+ * 
+ * @internal
  */
 export function getFriends(character: Character): Array<Promise<Character>> {
   // Notice that GraphQL accepts Arrays of Promises.
@@ -130,6 +132,8 @@ export function getFriends(character: Character): Array<Promise<Character>> {
 
 /**
  * Allows us to fetch the undisputed hero of the Star Wars trilogy, R2-D2.
+ * 
+ * @internal
  */
 export function getHero(episode: number): Character {
   if (episode === 5) {
@@ -142,6 +146,8 @@ export function getHero(episode: number): Character {
 
 /**
  * Allows us to query for the human with the given id.
+ * 
+ * @internal
  */
 export function getHuman(id: string): Human {
   return humanData[id];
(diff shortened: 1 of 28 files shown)


wtrocki:
I can confirm the same values locally. I just did not add internal flags to `src/__tests__/starWarsData.js`

src/__tests__/starWarsData.js
```diff
@@ -122,6 +122,8 @@ function getCharacter(id) {
 
 /**
  * Allows us to query for a character's friends.
+ *
+ * @internal
```
IvanGoncharov: @craicoverflow Thanks a lot for PR 👍 
One I didn't mention is that `__tests__` & `__fitures__` are not part of NPM package so all function exported from these folders are private by definition.
So can you please revert all changes under `__tests__` folder?

src/jsutils/Path.js
```diff
@@ -7,6 +7,8 @@ export type Path = {|
 
 /**
  * Given a Path and a key, return a new Path containing the new key.
+ *
+ * @internal
```
IvanGoncharov: Same goes for `jsutils` folder:
https://github.com/graphql/graphql-js/blob/master/src/jsutils/README.md
> These functions are not part of the module interface and are subject to change.

craicoverflow: Sure, I can change that :+1: 

It's maybe worth noting that in this same file, `pathToArray` is exported as public in `src/index.js` (aliased as `responsePathAsArray`).

https://github.com/graphql/graphql-js/blob/83d5011df8d5fc466d08517a2d13f9d3fb527161/src/index.js#L287

IvanGoncharov: @craicoverflow Good catch 👍 I need to think about it 🤔 💭

PR #2224, opened 2019-10-16
Fix typescript definition for getVariableValues

ivome:
This fixes the typescript definition to match flow and prevent errors when passing a read-only array.

### DIFF
--- a/src/execution/values.d.ts
+++ b/src/execution/values.d.ts
@@ -29,7 +29,7 @@ type CoercedVariableValues =
  */
 export function getVariableValues(
   schema: GraphQLSchema,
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@ivome Thanks a lot for PR 👍 
At the moment we are working on `15.0.0` release so this change would be released as `15.0.0-alpha.2`.
Do you need this fix to be backported in `14.x.x` or you can wait for `15.0.0` release?

ivome:
Thanks for merging this! I can wait for the `15.x` release. I am doing the type conversions manually in my codebase for now...

PR #2233, opened 2019-10-20
parser: convert Location to be an ES6 class

IvanGoncharov:


mjmahone:
Note on this: we forgot to update the exports, so it's impossible for someone to create a new Location now. We only export Location as a type.

PR #2248, opened 2019-11-06
buildSchema/extendSchema: add support for extensions

### LINKED ISSUE #922: buildASTSchema() should throw on encountering `extend` definition

IvanGoncharov:
Fixes #922

### DIFF
--- a/src/utilities/__tests__/buildASTSchema-test.js
+++ b/src/utilities/__tests__/buildASTSchema-test.js
@@ -6,6 +6,7 @@ import { describe, it } from 'mocha';
 import dedent from '../../jsutils/dedent';
 import invariant from '../../jsutils/invariant';
 
+import { Kind } from '../../language/kinds';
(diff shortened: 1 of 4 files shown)


IvanGoncharov:
Current state:
* Missing support for schema extensions
* Missing tests for Enums, Input Object, and Unions
* Missing tests for `extendSchema`

Sonamp:
Looks like it is not supported anymore. @IvanGoncharov could you please confirm if the extend type support is still there?

yaacovCR:
What version are you using? Can you post a link to a reproduction?

PR #2252, opened 2019-11-07
docs: follow express' best practice

Ardeshir81:
There were lots of `console.logs` that were coded like synchronous code, but express' `listen` function provides callback and is best practice to provide the callback for post-listen operations.

It was  already correct in this documentation: https://github.com/graphql/graphql-js/blob/master/docs/Tutorial-Mutations.md


IvanGoncharov:
@Ardeshir81 Merged 🎉 Thanks 👍

PR #2254, opened 2019-11-12
Vary the test case for building client schemas with enum

spawnia:
Instead of repeating basically the same enum value five times, i changed
the test to be less repetitive but have more variance in test scenarios.

### DIFF
--- a/src/utilities/__tests__/buildClientSchema-test.js
+++ b/src/utilities/__tests__/buildClientSchema-test.js
@@ -341,20 +341,11 @@ describe('Type System: build schema from introspection', () => {
           value: 1,
         },
         FRUITS: {
-          description: 'Foods that are fruits.',
           value: 2,
         },
         OILS: {
-          description: 'Foods that are oils.',
+          deprecationReason: 'Too fatty',
           value: 3,
-        },
-        DAIRY: {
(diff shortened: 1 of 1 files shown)


src/utilities/__tests__/buildClientSchema-test.js
```diff
@@ -341,21 +341,12 @@ describe('Type System: build schema from introspection', () => {
           value: 1,
         },
         FRUITS: {
-          description: 'Foods that are fruits.',
           value: 2,
         },
         OILS: {
-          description: 'Foods that are oils.',
+          deprecationReason: 'Too fatty',
```
IvanGoncharov: nit-pick: Can you put 'deprecationReason' after value to match the order in output?

spawnia:
Possible breaking change: Non-vegan options have been removed from `Food`.

On the plus side, cutting down on the `OILS` is probably more healthy.

IvanGoncharov:
@spawnia Thanks 👍 Dietary improvement always appreciated 🥕 

P.S. Have a personal OCD about the order of fields I always try to order in type definition 😜
 https://github.com/graphql/graphql-js/blob/96132b22758df89c28969e77565225f5cffdccf9/src/type/definition.js#L1330-L1336

PR #2264, opened 2019-11-19
chore(readme): improve contributing

AugustoCalaca:


saihaj:
Hey @AugustoCalaca this project uses `npm` and we shouldn't suggest `yarn` it may cause conflicts.

PR #2268, opened 2019-11-25
Pass down context info to serialize method of leaf nodes

willhl:
This came from a requirement to change the serialization behaviour based on what arguments were supplied to the field in the query. Specifically, a custom scalar was developed to convert values to the units requested in a "unit" field argument, e.g. request length values in meters instead of feet.

With this change this now possible by inspecting the `fieldNodes` and its arguments. Having the `info` passed down too is required to find the default value of the field arguments.

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -850,7 +850,7 @@ function completeValue(
   // If field type is a leaf type, Scalar or Enum, serialize to a valid value,
   // returning null if serialization is not possible.
   if (isLeafType(returnType)) {
-    return completeLeafValue(returnType, result);
+    return completeLeafValue(returnType, result, fieldNodes, info);
   }
 
   // If field type is an abstract type, Interface or Union, determine the
@@ -935,8 +935,8 @@ function completeListValue(
  * Complete a Scalar or Enum by serializing to a valid value, returning
  * null if serialization is not possible.
  */
-function completeLeafValue(returnType: GraphQLLeafType, result: mixed): mixed {
-  const serializedResult = returnType.serialize(result);
+function completeLeafValue(returnType: GraphQLLeafType, result: mixed, fieldNodes: $ReadOnlyArray<FieldNode>, info: GraphQLResolveInfo): mixed {
+  const serializedResult = returnType.serialize(result, fieldNodes, info);
   if (isInvalid(serializedResult)) {
     throw new Error(
       `Expected a value of type "${inspect(returnType)}" but ` +


IvanGoncharov:
@willhl Thanks for PR 👍

But scalar serialization behavior should be independent of field argument.
Scalars should be used both by input/output types and also inside SDL, e.g.:
```gql
scalar Distance

input Dimensions {
  x: Distance = 1 # We can't provide args inside SDL
  y: Distance = 1 # so we can't specify the unit of measurements
  z: Distance = 1 
}
```

A possible solution would be to create wrapper function for your resolvers:
```js
function changeScalarBasedOnArgs(resolver) {
   return (source, args, context, info) => {
     const result = resolver(source, args, context, info);
     return { value: result, unitOfMeasurement: args.unitOfMeasurement };
   }
}
```

```js
new GraphQLScalarType({
  name: 'Distance',
  serialize: ({ value, unitOfMeasurement }) => {
    switch(unitOfMeasurement) {
      case 'feets':
        return value * 3.281;
      case 'meters':
      default:
        return value;
    }
  }
});
```

Does it solves your problem?
If not can you please provide more details about your use case?

willhl:
Thanks for your response @IvanGoncharov, I suspected there would be an issue around SoC. I didn't come to this solution lightly and did have a think about other methods (not the one you've suggested though).

The problem with specifying "Distance" without args is it's not possible to infer from introspection what units it is returning by default. In your example it's meters, but there's no way to know this without looking into the code. I guess you could provide this info in separate documentation but it would be nice for introspection to provider everything you need to know.

So the use cases I was trying to solve:

- Allow the units to be inferred via introspection, for both queries and mutations
- Allow queries to ask for values to be returned in other applicable units. I know this is an application level problem to solve, but in reality this is going to be a very common thing to need to in my case, so why not have one implementation to do it for you? Sometimes breaking the rules is ok?

I'm also using neo4j-graphql-js where mutations are generated automatically from the types, so this makes it really convenient to infer the units expected where these fields appear in mutations. 

Your solution looks promising though, I'll have another think about the idea based on your suggestion and see where it takes me.

IvanGoncharov:
> The problem with specifying "Distance" without args is it's not possible to infer from introspection what units it is returning by default. 

@willhl Not sure I understand this point, so can you please provide an example of how you would achieve this with this PR?

> In your example it's meters, but there's no way to know this without looking into the code. I guess you could provide this info in separate documentation but it would be nice for introspection to provider everything you need to know.

You can specify default values for arguments:
```graphql
type Dimensions {
  height(unitOfMeasurement: UnitOfMeasurement = METERS)
  length(unitOfMeasurement: UnitOfMeasurement = METERS)
} 
```

Also, do you need to switch units for every scalar field separately or can it be done once per query?

```graphql
query MyQuery(
  $measurementSystem: MeasurmentSystemName = METRIC
) @measurementSystem(name: $measurementSystem) {
  # ...
}
```

IvanGoncharov:
@willhl Closing this PR.
Please feel free to open a new issue and describe your use case in more detail.

PR #2273, opened 2019-11-27
Add per-field thunk support

### LINKED ISSUE #277: Support thunks on a field itself?
Currently, the entire `fields` property supports a thunk. Would it be worthwhile to also support thunks on a per-field basis, so code like the following would work ...

``` js
let destinationTeam = () => ({
  type: new GraphQLNonNull(TeamType),
  description: "The team to which the invite is extended.",
  resolve: source => Sql.Team.getById(source.destination_team)
});

export default new GraphQLObjectType({
  name: "ScrimInvite",
  fields: {
    destinationTeam
  }
});
```

This might allow more flexibility in code formatting and cleanliness on really unwieldy types.

caseywebdev:
Closes #277

It'd be nice to not have to use helper functions around all `fields` definitions that resolve per-field thunks that are imported from other files. That's currently the only way to avoid cyclic import issues when a field is defined in a separate file from the object type.

Tests were failing locally because of a seemingly unrelated `Identifier 'xxx' has already been declared` issue, so hoping CI tests will work...

### DIFF
--- a/docs/APIReference-TypeSystem.md
+++ b/docs/APIReference-TypeSystem.md
@@ -294,8 +294,10 @@ type GraphQLArgumentConfig = {
 }
 
 type GraphQLFieldConfigMap = {
-  [fieldName: string]: GraphQLFieldConfig;
+  [fieldName: string]: GraphQLFieldConfigThunk | GraphQLFieldConfig;
 };
+
+type GraphQLFieldConfigThunk = () => GraphQLFieldConfig;
 ```
 
 Almost all of the GraphQL types you define will be object types. Object types
--- a/src/type/__tests__/definition-test.js
+++ b/src/type/__tests__/definition-test.js
@@ -213,12 +213,35 @@ describe('Type System: Objects', () => {
     });
   });
 
-  it('accepts an Object type with a field function', () => {
+  it('accepts an Object type with a fields function', () => {
     const objType = new GraphQLObjectType({
       name: 'SomeObject',
       fields: () => ({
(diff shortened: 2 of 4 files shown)


caseywebdev:
Whelp I'm losing the battle with Flow here. Maybe someone with Flow experience can make Flow happy with this one-line change?

caseywebdev:
@IvanGoncharov (pinging because you seem most active here) any idea how to satisfy Flow (or is this TypeScript?) here. I've never used either, but the only real change I made is very simple

https://github.com/graphql/graphql-js/pull/2273/commits/61cf8e267090148a4f3ca0cf7a0e892ea75273ad#diff-15f6b6c92ff954f80995d31726f89851R765-R766

and the added test passes. The only blocker here is getting these types right. I assumed simply wrapping `GraphQLFieldConfig` with `Thunk<>` would be sufficient but that ends up breaking a ton of other flow checks. Any help in pushing this through would be appreciated!

IvanGoncharov:
> The only blocker here is getting these types right. I assumed simply wrapping GraphQLFieldConfig with Thunk<> would be sufficient but that ends up breaking a ton of other flow checks. Any help in pushing this through would be appreciated!

Really strange behavior, I tried to figure out what's happing for ~3hours with no success.
I would suggest to either ping some Flow expert or wait until convert codebase to TS.

caseywebdev:
> some Flow expert

Happen to know anyone? I feel like since TS has blown up I don't hear much about Flow anymore.

PR #2276, opened 2019-12-06
Add `@specifiedBy` directive

m14t:
### Description

This in an implementation for a spec proposal:

* Spec proposal: [[RFC] Custom Scalar Specification URIs](https://github.com/graphql/graphql-spec/pull/649)
* Original issue: [[RFC] Custom Scalar Specification URIs](https://github.com/graphql/graphql-spec/issues/635)

#### Source
```javascript
const { graphql, buildSchema } = require("graphql");

const schema = buildSchema(`
  """
  A UUID
  """
  scalar UUID @specified(by: "https://tools.ietf.org/html/rfc4122")
  
  type Query {
    someUUID: UUID
  }
`);

const root = {
  someUUID: () => {
    return "750fce86-fc2f-4b25-a7df-fd0941d3fa08";
  }
};

const query = `
{
  __type(name: "UUID") {
    kind
    name
    description
    specifiedBy
  }
}`;

graphql(schema, query, root).then(response => {
  console.log(JSON.stringify(response, null, 4));
});
```

#### Output
```
{
    "data": {
        "__type": {
            "kind": "SCALAR",
            "name": "UUID",
            "description": "A UUID",
            "specifiedBy": "https://tools.ietf.org/html/rfc4122"
        }
    }
}
```

### Notes
Please let me know if I forgot something.

### DIFF
--- a/src/type/__tests__/introspection-test.js
+++ b/src/type/__tests__/introspection-test.js
@@ -63,6 +63,7 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'SCALAR',
@@ -72,6 +73,7 @@ describe('Introspection', () => {
               interfaces: null,
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'SCALAR',
@@ -81,6 +83,7 @@ describe('Introspection', () => {
               interfaces: null,
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
@@ -185,6 +188,7 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
@@ -218,6 +222,17 @@ describe('Introspection', () => {
                 },
                 {
                   name: 'description',
+                  args: [],
+                  type: {
+                    kind: 'SCALAR',
+                    name: 'String',
+                    ofType: null,
+                  },
+                  isDeprecated: false,
+                  deprecationReason: null,
+                },
+                {
+                  name: 'specifiedByUrl',
                   args: [],
                   type: {
                     kind: 'SCALAR',
@@ -358,6 +373,7 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'ENUM',
@@ -408,6 +424,7 @@ describe('Introspection', () => {
                 },
               ],
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
@@ -508,6 +525,7 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
@@ -570,6 +588,7 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
@@ -632,6 +651,7 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
@@ -729,6 +749,7 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
             },
             {
               kind: 'ENUM',
@@ -834,6 +855,7 @@ describe('Introspection', () => {
                 },
               ],
               possibleTypes: null,
+              specifiedByUrl: null,
             },
           ],
           directives: [
@@ -871,6 +893,26 @@ describe('Introspection', () => {
                     ofType: {
                       kind: 'SCALAR',
                       name: 'Boolean',
+                      ofType: null,
+                    },
+                  },
+                },
+              ],
+            },
+            {
+              name: 'specifiedBy',
+              isRepeatable: false,
+              locations: ['SCALAR'],
+              args: [
+                {
+                  defaultValue: null,
+                  name: 'url',
+                  type: {
+                    kind: 'NON_NULL',
+                    name: null,
+                    ofType: {
+                      kind: 'SCALAR',
+                      name: 'String',
                       ofType: null,
                     },
                   },
--- a/src/type/__tests__/introspection-test.js
+++ b/src/type/__tests__/introspection-test.js
@@ -30,8 +30,8 @@ describe('Introspection', () => {
     });
     const source = getIntrospectionQuery({
       descriptions: false,
-      directiveIsRepeatable: true,
       specifiedByUrl: true,
+      directiveIsRepeatable: true,
     });
 
     const result = graphqlSync({ schema, source });
@@ -657,6 +657,7 @@ describe('Introspection', () => {
             {
               kind: 'OBJECT',
               name: '__Directive',
+              specifiedByUrl: null,
               fields: [
                 {
                   name: 'name',
@@ -750,7 +751,6 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'ENUM',
@@ -901,6 +901,22 @@ describe('Introspection', () => {
               ],
             },
             {
+              name: 'deprecated',
+              isRepeatable: false,
+              locations: ['FIELD_DEFINITION', 'ENUM_VALUE'],
+              args: [
+                {
+                  defaultValue: '"No longer supported"',
+                  name: 'reason',
+                  type: {
+                    kind: 'SCALAR',
+                    name: 'String',
+                    ofType: null,
+                  },
+                },
+              ],
+            },
+            {
               name: 'specifiedBy',
               isRepeatable: false,
               locations: ['SCALAR'],
@@ -916,22 +932,6 @@ describe('Introspection', () => {
                       name: 'String',
                       ofType: null,
                     },
-                  },
-                },
-              ],
-            },
-            {
-              name: 'deprecated',
-              isRepeatable: false,
-              locations: ['FIELD_DEFINITION', 'ENUM_VALUE'],
-              args: [
-                {
-                  defaultValue: '"No longer supported"',
-                  name: 'reason',
-                  type: {
-                    kind: 'SCALAR',
-                    name: 'String',
-                    ofType: null,
                   },
                 },
               ],
--- a/src/type/__tests__/introspection-test.js
+++ b/src/type/__tests__/introspection-test.js
@@ -31,6 +31,7 @@ describe('Introspection', () => {
     const source = getIntrospectionQuery({
       descriptions: false,
       directiveIsRepeatable: true,
+      specifiedByUrl: true,
     });
 
     const result = graphqlSync({ schema, source });
--- a/src/type/__tests__/introspection-test.js
+++ b/src/type/__tests__/introspection-test.js
@@ -46,6 +46,7 @@ describe('Introspection', () => {
             {
               kind: 'OBJECT',
               name: 'QueryRoot',
+              specifiedByUrl: null,
               fields: [
                 {
                   name: 'onlyField',
@@ -63,31 +64,31 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'SCALAR',
               name: 'String',
+              specifiedByUrl: null,
               fields: null,
               inputFields: null,
               interfaces: null,
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'SCALAR',
               name: 'Boolean',
+              specifiedByUrl: null,
               fields: null,
               inputFields: null,
               interfaces: null,
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
               name: '__Schema',
+              specifiedByUrl: null,
               fields: [
                 {
                   name: 'description',
@@ -188,11 +189,11 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
               name: '__Type',
+              specifiedByUrl: null,
               fields: [
                 {
                   name: 'kind',
@@ -373,11 +374,11 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'ENUM',
               name: '__TypeKind',
+              specifiedByUrl: null,
               fields: null,
               inputFields: null,
               interfaces: null,
@@ -424,11 +425,11 @@ describe('Introspection', () => {
                 },
               ],
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
               name: '__Field',
+              specifiedByUrl: null,
               fields: [
                 {
                   name: 'name',
@@ -525,11 +526,11 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
               name: '__InputValue',
+              specifiedByUrl: null,
               fields: [
                 {
                   name: 'name',
@@ -588,11 +589,11 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
               name: '__EnumValue',
+              specifiedByUrl: null,
               fields: [
                 {
                   name: 'name',
@@ -651,7 +652,6 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
-              specifiedByUrl: null,
             },
             {
               kind: 'OBJECT',
@@ -754,6 +754,7 @@ describe('Introspection', () => {
             {
               kind: 'ENUM',
               name: '__DirectiveLocation',
+              specifiedByUrl: null,
               fields: null,
               inputFields: null,
               interfaces: null,
@@ -855,7 +856,6 @@ describe('Introspection', () => {
                 },
               ],
               possibleTypes: null,
-              specifiedByUrl: null,
             },
           ],
           directives: [
--- a/src/type/definition.d.ts
+++ b/src/type/definition.d.ts
@@ -291,7 +291,7 @@ export type Thunk<T> = (() => T) | T;
 export class GraphQLScalarType {
   name: string;
   description: Maybe<string>;
-  specifiedByUrl?: Maybe<string>;
+  specifiedByUrl: Maybe<string>;
   serialize: GraphQLScalarSerializer<any>;
   parseValue: GraphQLScalarValueParser<any>;
   parseLiteral: GraphQLScalarLiteralParser<any>;
--- a/src/type/definition.d.ts
+++ b/src/type/definition.d.ts
@@ -297,6 +297,7 @@ export class GraphQLScalarType {
   extensions: Maybe<Readonly<Record<string, any>>>;
   astNode: Maybe<ScalarTypeDefinitionNode>;
   extensionASTNodes: Maybe<ReadonlyArray<ScalarTypeExtensionNode>>;
+  specifiedByUrl?: Maybe<string>;
 
   constructor(config: Readonly<GraphQLScalarTypeConfig<any, any>>);
 
@@ -306,6 +307,7 @@ export class GraphQLScalarType {
     parseLiteral: GraphQLScalarLiteralParser<any>;
     extensions: Maybe<Readonly<Record<string, any>>>;
     extensionASTNodes: ReadonlyArray<ScalarTypeExtensionNode>;
+    specifiedByUrl: Maybe<string>;
   };
 
   toString(): string;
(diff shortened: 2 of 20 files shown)


src/type/directives.d.ts
```diff
@@ -54,6 +54,12 @@ export const GraphQLIncludeDirective: GraphQLDirective;
  */
 export const GraphQLSkipDirective: GraphQLDirective;
 
+/**
+ * Used to provide an RFC3986-compliant URI for specifying the behaviour of
+ * custom scalar definitions.
+ */
```
IvanGoncharov: Should match comment in `directive.js`

m14t: Resolved by 2a1973a5.

m14t:
@IvanGoncharov Thanks for the review!

While trying to improve the test coverage here, I wanted pass an SDL that used `@specified` to `cycleSDL`, but ran into https://github.com/graphql/graphql-js/issues/2020.

I'm not entirely sure of the implications of this, but if more work is need and given some direction, I'd be happy to continue to work on this.

cc: @eapache

src/type/directives.js
```diff
@@ -167,6 +167,21 @@ export const GraphQLSkipDirective = new GraphQLDirective({
   },
 });
 
+/**
+ * Used to provide URL for specifying the behaviour of custom scalar definitions.
```
eapache: nit: Used to provide *a* URL

src/type/introspection.js
```diff
@@ -221,6 +222,20 @@ export const __Type = new GraphQLObjectType({
...
+        resolve: type => {
+          if (!isScalarType(type) || !type.astNode) {
+            return null;
+          }
+
+          const specified = getDirectiveValues(
+            GraphQLSpecifiedDirective,
+            type.astNode,
+          );
+          return specified && specified.by;
```
IvanGoncharov: @m14t You shouldn't do that, `astNode` should be used only for error reporting.
Please replicate how `@deprecate` is converted into `deprecationReason`.

IvanGoncharov:
> While trying to improve the test coverage here, I wanted pass an SDL that used @specified to cycleSDL, but ran into #2020.

@m14t It should be handled the same way as `@deprecate`.
#2020 is about printing user-provided directives and there is nothing wrong in hardcoding standard directives for a time being.
Also please review tests that mention `@deprecate` or `deprecatedReason` and replicate ones that make sense for `@specified`/`specifiedBy`.

src/type/directives.js
m14t: Fixed by a115024c

m14t:
@IvanGoncharov thank you for the feedback.

I've made the requested changes, and feel better about this.  I still need to check that I have all of the test cases, but think that the implementation is done.

src/utilities/__tests__/buildClientSchema-test.js
```diff
@@ -530,6 +530,18 @@ describe('Type System: build schema from introspection', () => {
     expect(cycleIntrospection(sdl)).to.equal(sdl);
   });
 
+  it('builds a schema with specified by', () => {
+    const sdl = dedent`
+      type Query {
+        someUUID: UUID
+      }
+
+      scalar UUID @specified(by: "https://tools.ietf.org/html/rfc4122")
```
IvanGoncharov: If at some point we decide to add `UUID` scalar into this library it would be a confusing example.
It's better to use some abstract scalar like `Foo` and `https://example.com/foo_spec`.

src/utilities/buildASTSchema.js
```diff
@@ -441,6 +447,7 @@ export class ASTDefinitionBuilder {
           name,
           description,
           astNode,
+          specifiedBy: getSpecifiedBy(astNode),
```
IvanGoncharov: Please make sure that this also works in `extendSchema`:
```gql
extend scalar Foo @specified(by: 'https://example.com/foo_spec')
```

src/utilities/schemaPrinter.js
```diff
@@ -317,6 +321,21 @@ function printDeprecated(fieldOrEnumVal) {
...
+  const by = scalar.specifiedBy;
+  const byAST = astFromValue(by, GraphQLString);
+  if (!byAST) {
+    invariant(
+      false,
+      'Unexpected null value returned from `astFromValue` for specifiedBy',
+    );
+  }
+  return ' @specified(by: ' + print(byAST) + ')';
+}
```
IvanGoncharov: Can you please add tests for this code?

IvanGoncharov:
> I still need to check that I have all of the test cases, but think that the implementation is done.

@m14t You should have 100% coverage for diff ATM it's 97.56%:
image

src/utilities/buildASTSchema.js
m14t: I added [this test](https://github.com/graphql/graphql-js/pull/2276/files#diff-c766829f03b82fbf1ebc1ea0926dd286R341-R368), but was unsure about the preferred implementation when there were multiple extensions (some of which could be unrelated to `specifiedBy`).

[Right now](https://github.com/graphql/graphql-js/pull/2276/files#diff-a2222d77ff884acdf97c5f295babf27cR330-R343) I am creating an array with the base `.specifiedBy` value, and calling `getSpecifiedBy` on each extension, filtering out the undefined values with `.filter(Boolean)`, and then getting the "last value".  With this if there are no `specifiedBy` values, it will try to get the `-1` element of the array, which is always `undefined`, and valid, but feels a bit weird.

Open to suggestions.

IvanGoncharov:
@m14t Can please update this PR to be in sync with https://github.com/graphql/graphql-spec/pull/649/commits/b418b6086a351197fcd3479122daf226641b6772 ?

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -338,6 +338,34 @@ describe('extendSchema', () => {
...
+      }
+
+      scalar Foo
+
+      directive @foo on SCALAR
+    `);
+    const extensionSDL = dedent`
+      extend scalar Foo @foo
+
+      extend scalar Foo @specified(by: "https://first.com/foo_spec")
```
m14t: @IvanGoncharov Thank you for the link to [`UniqueDirectivesPerLocation.js`](https://github.com/graphql/graphql-js/blob/master/src/validation/rules/UniqueDirectivesPerLocation.js#L59-L60).  That is indeed working as expected, and verified by this new test case I just added: d0606579

Apologies for misremember during today's working group call, but the actual situation that I was concerned about is demonstrated here, where a scalar type is extended multiple times.

I just wanted to call this out and to get confirmation that this was valid, and the expected result, as well as ask for set of yes on this specific part of the [implementation](https://github.com/m14t/graphql-js/blob/specified-directive/src/utilities/extendSchema.js#L330-L346)

Thanks again.

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -338,6 +338,36 @@ describe('extendSchema', () => {
...
+      scalar Foo
+
+      directive @foo on SCALAR
+    `);
+    const extensionSDL = dedent`
+      extend scalar Foo @foo
+
+      extend scalar Foo @specifiedBy(url: "https://first.com/foo_spec")
+
+      extend scalar Foo @specifiedBy(url: "https://second.com/foo_spec")
```
m14t: @IvanGoncharov Thank you for the link to [`UniqueDirectivesPerLocation.js`](https://github.com/graphql/graphql-js/blob/master/src/validation/rules/UniqueDirectivesPerLocation.js#L59-L60).  That is indeed working as expected, and verified by this new test case I just added: d0606579

Apologies for misremember during today's working group call, but the actual situation that I was concerned about is demonstrated here, where a scalar type is extended multiple times.

I just wanted to call this out and to get confirmation that this was valid, and the expected result, as well as ask for set of yes on this specific part of the [implementation](https://github.com/m14t/graphql-js/blob/specified-directive/src/utilities/extendSchema.js#L330-L346)

Thanks again.

m14t: @IvanGoncharov I wanted to see if you had any thoughts on this.

src/utilities/__tests__/buildASTSchema-test.js
```diff
@@ -763,6 +772,39 @@ describe('Schema Builder', () => {
...
+
+      type Query {
+        foo: Foo @deprecated
+      }
+    `;
+
+    expect(() => buildSchema(sdl)).to.throw(
+      'The directive "@specifiedBy" can only be used once at this location.',
+    );
+  });
```
IvanGoncharov: No need to test for test this error specifically for `@specifiedBy` since it's a general rule for all directives and already tested inside `src/validation/__tests__/`.

src/utilities/__tests__/extendSchema-test.js
IvanGoncharov: @m14t No worries. Thanks for adding the detail test.
Sorry for the delayed reply and thanks for pinging 👍 
I'm traveling through Asia ATM so can disappear from time to time please feel free to ping me if I don't respond in a reasonable time

I'm trying to finish a few last features for `15.0.0` release and then I will review this PR in detail. Since it's not a breaking change I want to include it in `15.1.0` (if `15.0.0` don't result in any major problems we can release `15.1.0` in 1 or 2 weeks).

m14t: Sounds good. Holding on this till `15.1.0` makes sense to me.

 I'll keep an eye out for `15.0.0` and if I don't hear from you within a week after that, I'll ping again.  Enjoy your travels!

m14t: FWIW: this has been rebased off of `v15.0.0-rc.2`

src/utilities/extendSchema.js
```diff
@@ -324,9 +327,15 @@ export function extendSchemaImpl(
     const config = type.toConfig();
     const extensions = typeExtensionsMap[config.name] ?? [];
 
+    const specifiedByUrl = [
+      config.specifiedByUrl,
+      ...extensions.map(getSpecifiedByUrl),
+    ].filter(Boolean)[0];
```
m14t: RFC: Is this code clear enough?  Any suggestions what would be more clear?

We want to:
1. use `config.specifiedByUrl` if present
2. else, check if any of the extensions include a `GraphQLSpecifiedByDirective`, and use the `url` value from there if present.

Since a Scalar could be extended multiple times, each of which could include multiple directives (but only one `GraphQLSpecifiedByDirective`), this was the best balance I could find between being clear and not overly verbose.


**Alternative 1: With ternary and `.filter(Boolean)`**
```suggestion
    const specifiedByUrl =
      config.specifiedByUrl != null
        ? config.specifiedByUrl
        : extensions.map(getSpecifiedByUrl).filter(Boolean)[0];
```

**Alternative 2: With ternary and double call to `getSpecifiedByUrl`**
```suggestion
    const specifiedByUrl =
      config.specifiedByUrl != null
        ? config.specifiedByUrl
        : getSpecifiedByUrl(extensions.find(getSpecifiedByUrl) || {});
```

src/type/introspection.js
```diff
@@ -195,7 +195,7 @@ export const __DirectiveLocation = new GraphQLEnumType({
 export const __Type = new GraphQLObjectType({
   name: '__Type',
   description:
-    'The fundamental unit of any GraphQL Schema is the type. There are many kinds of types in GraphQL as represented by the `__TypeKind` enum.\n\nDepending on the kind of a type, certain fields describe information about that type. Scalar types provide no information beyond a name and description, while Enum types provide their values. Object and Interface types provide the fields they describe. Abstract types, Union and Interface, provide the Object types possible at runtime. List and NonNull types compose other types.',
+    'The fundamental unit of any GraphQL Schema is the type. There are many kinds of types in GraphQL as represented by the `__TypeKind` enum.\n\nDepending on the kind of a type, certain fields describe information about that type. Scalar types provide no information beyond a name, description and optional specifiedByUrl, while Enum types provide their values. Object and Interface types provide the fields they describe. Abstract types, Union and Interface, provide the Object types possible at runtime. List and NonNull types compose other types.',
```
chrisbutcher:
```suggestion
    'The fundamental unit of any GraphQL Schema is the type. There are many kinds of types in GraphQL as represented by the `__TypeKind` enum.\n\nDepending on the kind of a type, certain fields describe information about that type. Scalar types provide no information beyond a name, description and optional `specifiedByUrl`, while Enum types provide their values. Object and Interface types provide the fields they describe. Abstract types, Union and Interface, provide the Object types possible at runtime. List and NonNull types compose other types.',
```

Minor, but it looks like backticks are used when referencing type and field names.

chrisbutcher:
Hey @m14t, great work! One question / potential suggestion:

`specifiedBy` lets schema maintainers indicate what specs a given custom scalar should conform to.

Should implementations of `specifiedBy` itself (like this PR) validate that the spec URL provided by schema maintainers is be valid, as specified by [RFC 1738](https://tools.ietf.org/html/rfc1738)? I had a look at the diff here, but don't see any such validation being performed.

m14t:
@chrisbutcher Thanks for the review.

I believe validating URLs was discussed in one of the working group sessions, though I wasn't able to find this with a quick search of the meeting notes.

My recollection was that this not something that would be enforced.  If it is, we should probably update https://github.com/graphql/graphql-spec/pull/649 to include this restriction.

CC: @eapache / @IvanGoncharov / @leebyron incase any of you have a different/better memory than myself.

eapache:
I remember discussing whether the spec should say "URL" or "URL as defined by RFC..." everywhere and we decided that just "URL" was unambiguous. But the spec still declares that the argument is a URL, and doesn't leave a lot of room for it to be some other string. I don't know if we'd want to require this validation in the spec, but it seems like a good idea to include in the reference implementation? Though maybe that means we should just clarify the spec.

The WG meeting is tomorrow, we can discuss then I guess :)

src/type/definition.d.ts
```diff
@@ -292,6 +292,7 @@ export class GraphQLScalarType {
   extensions: Maybe<Readonly<Record<string, any>>>;
   astNode: Maybe<ScalarTypeDefinitionNode>;
   extensionASTNodes: Maybe<ReadonlyArray<ScalarTypeExtensionNode>>;
+  specifiedByUrl?: Maybe<string>;
```
IvanGoncharov: Can you please move it between `description` and `serialize` in all types.

src/type/definition.js
```diff
@@ -599,6 +601,14 @@ export class GraphQLScalarType {
         `${this.name} must provide both "parseValue" and "parseLiteral" functions.`,
       );
     }
+
+    if (config.specifiedByUrl != null) {
+      devAssert(
+        typeof config.specifiedByUrl === 'string',
```
IvanGoncharov:
```suggestion
        config.specifiedByUrl != null && typeof config.specifiedByUrl === 'string',
```

src/utilities/__tests__/buildASTSchema-test.js
```diff
@@ -770,6 +779,25 @@ describe('Schema Builder', () => {
...
+        foo: Foo @deprecated
+      }
+    `;
+    expect(cycleSDL(sdl)).to.equal(sdl);
+
+    const schema = buildSchema(sdl);
+
+    const foo = assertScalarType(schema.getType('Foo'));
+
+    expect(foo).to.include({
```
IvanGoncharov:
```suggestion
    expect(schema.getType('Foo')).to.include({
```

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -359,6 +359,33 @@ describe('extendSchema', () => {
...
+    const extensionSDL = dedent`
+      extend scalar Foo @foo
+
+      extend scalar Foo @specifiedBy(url: "https://example.com/foo_spec")
+    `;
+
+    const extendedSchema = extendSchema(schema, parse(extensionSDL));
+    const foo = assertScalarType(extendedSchema.getType('Foo'));
+
+    expect(foo.toConfig().specifiedByUrl).to.equal(
```
IvanGoncharov: You don't need `toConfig` here, just do:
```suggestion
    expect(foo).to.include({
```

src/utilities/__tests__/extendSchema-test.js
```diff
@@ -359,6 +359,33 @@ describe('extendSchema', () => {
...
 
+  it('extends scalars by adding specifiedBy directive', () => {
+    const schema = buildSchema(`
+      type Query {
+        foo: Foo
+      }
+
+      scalar Foo
+
+      directive @foo on SCALAR
```
IvanGoncharov: Why do you need `@foo` here?
Since it's already tested in the above test case.

m14t: @IvanGoncharov Good question.  The reason I added this was to validate the use-case when there were multiple `extend`s.

If we remove this directive, it would be possible for `src/utilities/extendSchema.js`'s `extendScalarType` method to have a naive implementation that only looked at the first extension and could miss this.  This is mostly testing the `extensions.map`/`extensions.find` part of `extendScalarType`.

How would you like to proceed?  I am happy to:
1. Simplify this test and remove this as suggested
2. Leave as is
3. Make this more robust by adding another directive to validate that the `extendScalarType` isn't simply looking at the last extension

src/utilities/__tests__/extendSchema-test.js
m14t: Fixed in b44e2c9e

src/utilities/__tests__/buildASTSchema-test.js
m14t: Fixed in 407b1930

src/type/definition.js
m14t: Fixed in bd6d4968

src/type/definition.d.ts
m14t: Fixed in 4c7d339d and bd6d4968.  There were some instances where `specifiedByUrl` and `serialize` were both present , but not `description`.  In those cases I moved `specifiedByUrl` immediately before `serialize`.

eapache:
(as discussed at [the working group meeting](https://github.com/graphql/graphql-wg/blob/master/agendas/2020-03-05.md), we're not going to be validating URLs for now)

m14t:
Hi @IvanGoncharov,

I just wanted to check in if there were any other changes you were looking for ahead of the working group session next week.

I believe all of your March 5th comments have been resolved with the exception of one question that I replied to with three potential options.

This branch is up to date with `v15.0.0-rc.2`.  Please let me know if you would prefer it rebased off of `master`.

src/utilities/getIntrospectionQuery.js
```diff
@@ -58,6 +58,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
       kind
       name
       ${descriptions}
+      specifiedByUrl
```
IvanGoncharov, line 61: @m14t Since older GraphQL servers don't know about this field it will result in the entire query being rejected.
I'm actively working on a two-stage introspection solution that is a generic solution to this problem.
But in the mean time, you need to flag similar one for schema description:
https://github.com/graphql/graphql-js/pull/2419/files#diff-72ccdfc3a6c358028739a30a14edddce
Also please add tests:
https://github.com/graphql/graphql-js/pull/2419/files#diff-80e66acfef544cd667aab53b088cc9ed

src/type/definition.d.ts
```diff
@@ -286,6 +286,7 @@ export type Thunk<T> = (() => T) | T;
 export class GraphQLScalarType {
   name: string;
   description: Maybe<string>;
+  specifiedByUrl?: Maybe<string>;
```
IvanGoncharov, line 289:
```suggestion
  specifiedByUrl: Maybe<string>;
```
Since you always assign it inside constructor

src/type/__tests__/introspection-test.js
```diff
@@ -63,6 +63,7 @@ describe('Introspection', () => {
               interfaces: [],
               enumValues: null,
               possibleTypes: null,
+              specifiedByUrl: null,
```
IvanGoncharov, line 66: nit: Better to keep the same order of fields as they specified in introspection query, so `specifiedBy` should be after `name`.

m14t: Fixed in d6e4f37a61f0f4597a008280822cec3a13e11794

src/type/definition.d.ts
m14t: Fixed in 22524b102198e2b9c5103ebdf989eb6d069d2943

src/utilities/getIntrospectionQuery.js
m14t: Done in 18689db4cd0cb9c031ea7054b55da3ebe796e034.

Should this default to `true` or `false`?  This commit defaults it to `true`.

m14t:
@IvanGoncharov Sorry for the delay here, but this is ready for another review when you have a chance.

src/utilities/getIntrospectionQuery.js
IvanGoncharov: @m14t `false` otherwise it would be a breaking change.

m14t: Done in 4dcf49be.

m14t: Apologies, changing the default from `true` to `false` required passing this argument as `true` in two other tests that was missed.  Fixed in 17d4ad38.

IvanGoncharov:
@m14t Sorry for the delay will try to review, merge, release before tomorrow's WG.

src/utilities/__tests__/extendSchema-test.js
IvanGoncharov: @m14t Thank you for an explanation, now it totally makes sense 👍

src/utilities/extendSchema.js
IvanGoncharov: @m14t I converted it to a simple `for` loop.

src/type/definition.js
```diff
@@ -568,6 +568,7 @@ function undefineIfEmpty<T>(arr: ?$ReadOnlyArray<T>): ?$ReadOnlyArray<T> {
 export class GraphQLScalarType {
   name: string;
   description: ?string;
+  specifiedByUrl: ?string;
```
mjmahone, line 571: Just a note: this is technically a breaking change. If anyone was using `GraphQLScalarType` in their code, they now need to add `specifiedByUrl` in order to satisfy the type checker.

src/utilities/getIntrospectionQuery.js
```diff
@@ -166,6 +175,7 @@ export type IntrospectionScalarType = {|
   +kind: 'SCALAR',
   +name: string,
   +description?: ?string,
+  +specifiedByUrl: ?string,
```
mjmahone, line 178: This is technically a breaking change: code that was using `IntrospectionScalarType` no longer type-checks unless `specifiedByUrl` is explicitly set.

Issue #1819, opened 2019-04-13
Export clean ESM modules


sebdeckers:
Using graphql@14.2.1

`/node_modules/graphql/jsutils/instanceOf.mjs` line 16

```js
// See: https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production
// See: https://webpack.js.org/guides/production/
export default process.env.NODE_ENV === 'production' ? // eslint-disable-next-line no-shadow
```

```
Uncaught ReferenceError: process is not defined
    at instanceOf.mjs:16
```

The check `process.env.NODE_ENV === 'production'` breaks in the browser when using ESM version of `graphql` package. Is it possible to remove before publishing to NPM? Without it the ESM files work perfectly.

I am not using Webpack/RollUp/Babel. Simply copying the ESM files from node_modules to my frontend folder using [unbundle](https://www.npmjs.com/package/unbundle).

Related but different:
- https://github.com/graphql/graphql-js/issues/1536
- https://github.com/graphql/graphql-js/issues/1293
- https://github.com/graphql/graphql-js/issues/1217

abdonrd:
Also related but different: https://github.com/apollographql/graphql-tag/issues/229

Carglass:
Hello, 

Id like to try to help with this issue. Using Unbundle, do you have a possibility to differentiate production from development code? I could not find a way to do it without using "bundlers". If there is not, the main issue, is that this export returns an optimized function for production, so that in the absence of the process.env, there is a choice to make about using the optimized function or the more helpful for development function. Any thoughts on this?

dandv:
Not sure if this will get any traction before Node 12 drops the --experimental-modules flag, because they've asked the community to [not publish any ES modules until then](https://twitter.com/sheerun/status/1122894214646370304).

sebdeckers:
> While we are aware that the community has embraced the `"module"` field, it is unlikely that Node.js will adopt that field as many of the packages published using `"module"` include ES module JavaScript that may not evaluate in Node.js (because extensions are left off of filenames, or the code includes `require` statements, etc.). Please do not publish any ES module packages intended for use by Node.js until this is resolved.

1. I am looking to run the ESM on the web, not Node.js. (Universal ESM would be nice tho.)
1. The `module` field (et al.) is a de-facto practice. Supported by WebPack, RollUp, and Pika (afaik).

[...]

PR #2277, opened 2019-12-06
ESM build for the browser

lukejacksonn:
Hey 👋 I'm a big advocate of buildstep-less architectures on the web, taking advantage of ES modules to eschew the need for bundling ([es-react](https://github.com/lukejacksonn/es-react) for example). This PR addresses https://github.com/graphql/graphql-js/issues/1819.

### 💭 Rationale

Recently I wanted to use [urql](https://github.com/formidablelabs/urql) in one of my projects that falls into this build free category. So I set about picking apart the dependencies of the package to see how far it was from being ESM ready. Probably unsurprisingly this package is the biggest dependency of urql so I started with fixing it up and this is what I ended up working with.

### 👨‍🔧 Changes

- https://github.com/graphql/graphql-js/pull/2277/commits/316a69e39d3da3e441b80ec068b0ec07f85616ec Created a new build `esm` with a custom babel extension that appends `.es.js` to import/export statement sources. The differences between this and `mjs` build are subtle but this decision was made as not to disturb any existing builds. Explicit extensions are required according to the browser es module spec (also the most recent node es module spec).
- https://github.com/graphql/graphql-js/pull/2277/commits/8e7cc02775e77c3603049c58115540da2e350d1f Checks `typeof process` in `instanceOf.js` to prevent `Cannot find property .env of undefined` error when running the browser; without altering the node behaviour.
- https://github.com/graphql/graphql-js/pull/2277/commits/d27e528c8912df80c941f7744f30225d65634b9f Vendor'd the `iterall` package to eradicate the only bare specifier import and making the whole codebase dependency free!
- https://github.com/graphql/graphql-js/pull/2277/commits/d7210ecc5cbb361b233f30aeabbc5233242fdcc8 Makes paths in `/index.js` resolvable by extension only (due to `/depencendy.js` and `/dependency/index.js` inconsistency usually handled by nodes resolution algorithm)

### 📦 Output

This PR itself is a POC that I made for myself, I'm not precious about the implementation details but I am quite optimistic about the results.

Here is a published version of this PR: https://www.runpkg.com/?es-graphql@1.0.3/dist/index.es.js
Here is a demo of this build working with urql: https://es-urql.lukejacksonn.now.sh

Running `yarn build && yarn testonly` currently succeeds with no errors but type checking fails due to no definitions for the functions within `iterall.js`. If anyone has suggestions of how to achieve the same result with a different approach to any of the steps outlined by specific commits here, then I am all ears (I am no babel aficionado)!

### 🤔 Conclusion

This library required very few changes in order to make it _just work_ in the browser. I'm almost certain that this would trivialise and encourage gql client library authors to generate es module builds themselves which will lower the barrier to entry for people getting started with GraphQL in general (example projects could now just be an index.html with no install or build process to explain).

### DIFF
--- a/src/jsutils/instanceOf.js
+++ b/src/jsutils/instanceOf.js
@@ -11,7 +11,8 @@ declare function instanceOf(
 
 // See: https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production
 // See: https://webpack.js.org/guides/production/
-export default process.env.NODE_ENV === 'production'
+export default typeof process !== 'undefined' &&
+process.env.NODE_ENV === 'production'
   ? // eslint-disable-next-line no-shadow
     function instanceOf(value: mixed, constructor: mixed) {
       return value instanceof constructor;
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -15,5 +15,27 @@ module.exports = {
     mjs: {
       presets: [['@babel/preset-env', { modules: false }]],
     },
+    esm: {
+      presets: [['@babel/preset-env', { modules: false }]],
+      plugins: [
+        function({ types }) {
+          return {
+            visitor: {
+              ImportDeclaration: function(path, state) {
+                var source = path.node.source.value;
+                if (source.match(/^\.{0,2}\//) && !source.endsWith('.es.js')) {
+                  path.replaceWith(
+                    types.importDeclaration(
+                      path.node.specifiers,
+                      types.stringLiteral(source + '.es.js'),
+                    ),
+                  );
+                }
+              },
+            },
+          };
+        },
+      ],
+    },
   },
 };
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -22,10 +22,24 @@ module.exports = {
           return {
             visitor: {
               ImportDeclaration: function(path, state) {
+                if (!path.node.source) return;
                 var source = path.node.source.value;
                 if (source.match(/^\.{0,2}\//) && !source.endsWith('.es.js')) {
                   path.replaceWith(
                     types.importDeclaration(
+                      path.node.specifiers,
+                      types.stringLiteral(source + '.es.js'),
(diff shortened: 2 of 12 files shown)


src/jsutils/instanceOf.js
```diff
@@ -11,7 +11,8 @@ declare function instanceOf(
 
 // See: https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production
 // See: https://webpack.js.org/guides/production/
-export default process.env.NODE_ENV === 'production'
+export default typeof process !== 'undefined' &&
+process.env.NODE_ENV === 'production'
```
IvanGoncharov: @lukejacksonn Can you please extract it into separate PR?
Also, please test that Webpack is able to recognize this pattern and switch to the production code.

Is it the last change we need to support ESM?

lukejacksonn: Yeh I can do that.. this is the last source modification required for ESM support (assuming that the mjs build now adds extensions to imports paths).

IvanGoncharov: > Yeh I can do that..

Great 👍 Thanks a lot!

> assuming that the mjs build now adds extensions to imports paths

You can see preview build here:
https://github.com/graphql/graphql-js/blob/npm/graphql.mjs
If you have time to try it here is the instruction:

IvanGoncharov: Hi @lukejacksonn,
Did you have a time to test this change with webpack?
I planning to release new RC so it would be great to also include this change.

lukejacksonn: Hey @IvanGoncharov sorry this task completely slipped my mind. I have made a PR https://github.com/graphql/graphql-js/pull/2409 but am not so sure about what webpack specific config needs to be checked

IvanGoncharov:
Superseded by #2409

PR #2279, opened 2019-12-06
Update TravisCi to run full CI suite on Node 12

m14t:
Update TravisCi to run the full `test:ci` commands on Node 12, which was promoted to LTS on 2019-10-21 ([source](https://nodejs.org/en/about/releases/)).


IvanGoncharov:
@m14t Thanks 👍

PR #2285, opened 2019-12-14
Remove forgotten internal error message functions (#2159)

Cito:
In #2519, the internal functions for validation error messages have been removed, but some have been forgotten.


IvanGoncharov:
@Cito Big thanks as always! It's really great to have second eye on all changes 👍

PR #2301, opened 2019-12-20
Remove support for positional args in graphql/execute/subscribe func

IvanGoncharov:
Note: It's a big breaking change and is intended to be released only in v16.x.x line


riginoommen:
Can you please update a note about this change? For the apollo server subscriptions, this is a required dependencies. Is there any solution for overriding this

saihaj:
@riginoommen I don’t think Apollo supports v16 yet they are working on it https://github.com/apollographql/apollo-server/pull/5857 this is something they need to update in docs nothing we can do here.

PR #2318, opened 2019-12-30
Support @defer directive

morrys:
Hello to all,
this PR allows to support the @defer directive. Although the implementation is mostly complete, the PR is in draft to discuss it together and evaluate its integration.

In the PR I also created tests and I performed the integration tests with Relay, express-graphql & [fetch-multipart-graphql](https://github.com/relay-tools/fetch-multipart-graphql) with positive results.

In summary how I implemented defer:
  * @defer has two arguments:
     * if (optional, default true) the fragment is defer when true.
     * label (required) used to distinguish deferred fragments
  * @defer currently provides only FRAGMENT_SPREAD & INLINE_FRAGMENT (Relay) as location, but support for FIELD (Apollo) is commented in the PR https://github.com/graphql/graphql-js/blob/4226f231f74ad30c5e20a3ca3289417af8058ba1/src/execution/execute.js#L663-L668
  * I added `resolverResult: ResultResolver` new property in ExecutionContext which allows the management of the results to be returned at the end of the execution
  * modified execute function now its return type is `AsyncIterable<Promise<ExecutionResult> | PromiseOrValue<ExecutionResult>; `. Using `isAsyncIterable (result)` it is possible to distinguish if there are deferred results.
  * fields that are marked as deferred are resolved after the resolution of the previous result. I also implemented an optimization to avoid solving the same field multiple times.

this is a slow motion gif of how it works.
deferred

Let me know if more information is needed.

Thank you
Lorenzo


wtrocki:
Some additional context from Apollo Defer: https://github.com/apollographql/apollo-server/issues/2700

morrys:
Looking at the Relay test schema
https://github.com/facebook/relay/blob/5da3be070283c6dcd42774ba33c1590db65fe3c7/packages/relay-test-utils-internal/testschema.graphql#L13-L16 i noticed that defer is also supported for inline_fragment, the last commit support it and add some tests

morrys:
In this branch I created support for @stream:
https://github.com/morrys/graphql-js/tree/support-stream-directive

Some checks and optimizations are missing. But the first tests with Relay were positive:
defer-and-stream-directive

IvanGoncharov:
@morrys Thanks for your effort 👍 
I didn't have enough time to read through actual code but on the surface, it looks like both #2318 and #2319 add very similar functionality especially since both of them Relay compatible.
Or I'm missing something?

About the next steps for merging this PR please see my answer in #2319:
https://github.com/graphql/graphql-js/pull/2319

@morrys @lilianammmatos Can you please review each other PRs and figure out how to collaborate?

morrys:
@IvanGoncharov yes the two PR add the same functionality.
Tomorrow I integrate the branch where I implemented the stream and then I dedicate myself to making a review of @lilianammmatos 's PR

Issue #2144, opened 2019-08-31
Combine `subscription` with `execute`?


ericls:
Right now, subscribe and execute are two functions. And there are some issues with that.
1. Semantically, subscription is a type of execution. In the spec, Query, Mutation and Subscription are on the same level. Subscription is not an outlier.
2. When I get a Document, I need to go through validations to get the Operation, and then call `execute` or `subscribe` based on Operation's type, and they go through a similar validation, resulting in duplicated work.

IvanGoncharov:
@ericls Main difference is in the return value since `subscribe` returns promise to async iterator:
https://github.com/graphql/graphql-js/blob/0d2220f0a64238473f4e1c9aa8f73f891d0fd3e0/src/subscription/subscribe.js#L77

Another big difference is that `execute` can return the result without wrapping it in a promise:
https://github.com/graphql/graphql-js/blob/0d2220f0a64238473f4e1c9aa8f73f891d0fd3e0/src/execution/execute.js#L157

> and then call execute or subscribe based on Operation's type, and they go through a similar validation,

Both `execute` and `subscribe` doesn't do the validation.

> resulting in duplicated work.

Since both functions accept args as objects you can share most of the code:
```js
const args = { schema, document, /* ... */ };
switch(getOperationRootType(document, operationName).operation) {
  case 'query':
  case 'mutation':
    return execute(args);
  case 'subscribe':
    const asyncIter = await subscribe(args);
    // send event's using asyncIter
}
```

ericls:
@IvanGoncharov, Thank you for you reply!

But both `execute` and `subscribe` calls `buildExecutionContext`, which validates and selects an operation from Document. And the signature of `getOperationRootType` is `(s: Schema, o: Operation)` which requires operation to be known.

IvanGoncharov:
> And the signature of getOperationRootType is (s: Schema, o: Operation) which requires operation to be known.

@ericls Sorry for the confusion I confused it with:
https://github.com/graphql/graphql-js/blob/0d2220f0a64238473f4e1c9aa8f73f891d0fd3e0/src/utilities/getOperationAST.js#L14-L17

> But both execute and subscribe calls buildExecutionContext, which validates and selects an operation from Document.

It's pretty basic validation (I thought you mean full query validation made by `validate`) and as you correctly pointed out code is shared code is already extracted into `buildExecutionContext`.

-----------------------------------

[...]

Issue #2141, opened 2019-08-28
Flow failing with graphql@^14.5.3:


kedarguy:
I am getting the following errors when running flow:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/graphql/jsutils/memoize3.js.flow:18:29

Cannot call cache0.get with a1 bound to key because read-only array type [1] is incompatible with object type [2].

     node_modules/graphql/jsutils/memoize3.js.flow
 [1]  11│ >(fn: (A1, A2, A3) => R): (A1, A2, A3) => R {
      12│   let cache0;
      13│
      14│   function memoized(a1, a2, a3) {
      15│     if (!cache0) {
      16│       cache0 = new WeakMap();
      17│     }
      18│     let cache1 = cache0.get(a1);
      19│     let cache2;
      20│     if (cache1) {
      21│       cache2 = cache1.get(a2);

     /tmp/flow/flowlib_3fce4660/core.js
 [2] 564│ declare class WeakMap<K: {}, V> extends $ReadOnlyWeakMap<K, V> {


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/graphql/jsutils/memoize3.js.flow:21:27

Cannot call cache1.get with a2 bound to key because read-only array type [1] is incompatible with object type [2].

     node_modules/graphql/jsutils/memoize3.js.flow
 [1]  11│ >(fn: (A1, A2, A3) => R): (A1, A2, A3) => R {
        :
      18│     let cache1 = cache0.get(a1);
      19│     let cache2;
      20│     if (cache1) {
      21│       cache2 = cache1.get(a2);
      22│       if (cache2) {
      23│         const cachedValue = cache2.get(a3);
      24│         if (cachedValue !== undefined) {

     /tmp/flow/flowlib_3fce4660/core.js
 [2] 564│ declare class WeakMap<K: {}, V> extends $ReadOnlyWeakMap<K, V> {


[...]

PR #2319, opened 2019-12-31
add support for defer and stream directives

lilianammmatos:
The implementation in this pull request is a reference implementation of the specification proposal outlined in https://github.com/graphql/graphql-spec/pull/742. It is no longer compatible with with the `@defer` and `@stream` implementation in Relay.

----

At [1stdibs](https://www.1stdibs.com/), our GraphQL server is built on top of [Apollo Server](https://github.com/apollographql/apollo-server). We were first introduced to the concept of the `@defer` directive through an [Apollo blog post](https://blog.apollographql.com/introducing-defer-in-apollo-server-f6797c4e9d6e) detailing an experimental implementation. We attempted to resurrect this initiative by creating an [up-to-date pull request](https://github.com/apollographql/apollo-server/pull/2181) (in hopes of that getting merged); however, it became clear to us over time that support for a `@defer` directive would occur only if it would make its way into the official GraphQL spec. There has been discussion on [this issue](https://github.com/graphql/graphql-spec/issues/269) to add `@defer` to the spec, but it seems it's still in its infancy!

Since then, Relay has added support for `@defer` and `@stream`, but there are no open source GraphQL servers that support its implementation.

This pull request is a proposal for a Relay compatible `@defer` and `@stream` directive. The main distinction between the Relay implementation of `@defer` and that of Apollo's is that the `@defer` directive is supported on a fragment spread and inline fragment instead of on an individual field. 

Furthermore, this pull request is implemented according to our understanding of Relay’s support for `@defer` and `@stream`. We outlined a few examples here: https://gist.github.com/robrichard/f563fd272f65bdbe8742764f1a149b2b

This is a feature we are very passionate about at 1stdibs and would be interested in working with contributors in any way possible!

Sketch RFC: https://github.com/graphql/graphql-spec/blob/master/rfcs/DeferStream.md

### DIFF
--- a/src/execution/execute.d.ts
+++ b/src/execution/execute.d.ts
@@ -44,6 +44,7 @@ export interface ExecutionContext {
  *
  *   - `errors` is included when any errors occurred as a non-empty array.
  *   - `data` is the result of a successful execution of the query.
+ *   - `hasNext` is true if a future payload is expected.
  *   - `extensions` is reserved for adding non-standard properties.
  */
 export interface ExecutionResult<
@@ -53,6 +54,7 @@ export interface ExecutionResult<
   errors?: ReadonlyArray<GraphQLError>;
   // TS_SPECIFIC: TData. Motivation: https://github.com/graphql/graphql-js/pull/2490#issuecomment-639154229
   data?: TData | null;
+  hasNext?: boolean;
   extensions?: TExtensions;
 }
 
@@ -65,6 +67,42 @@ export interface FormattedExecutionResult<
   data?: TData | null;
   extensions?: TExtensions;
 }
+
+/**
+ * The result of an asynchronous GraphQL patch.
+ *
+ *   - `errors` is included when any errors occurred as a non-empty array.
+ *   - `data` is the result of the additional asynchronous data.
+ *   - `path` is the location of data.
+ *   - `hasNext` is true if a future payload is expected.
+ *   - `label` is the label provided to @defer or @stream.
+ *   - `extensions` is reserved for adding non-standard properties.
+ */
+export interface ExecutionPatchResult<
+  TData = { [key: string]: any },
+  TExtensions = { [key: string]: any }
+> {
+  errors?: ReadonlyArray<GraphQLError>;
+  data?: TData | null;
+  path?: ReadonlyArray<string | number>;
+  label?: string;
+  hasNext: boolean;
+  extensions?: TExtensions;
+}
+
+export interface FormattedExecutionPatchResult<
+  TData = { [key: string]: any },
+  TExtensions = { [key: string]: any }
+> {
+  errors?: ReadonlyArray<GraphQLFormattedError>;
+  data?: TData | null;
+  path?: ReadonlyArray<string | number>;
+  label?: string;
+  hasNext: boolean;
+  extensions?: TExtensions;
+}
+
+export type AsyncExecutionResult = ExecutionResult | ExecutionPatchResult;
 
 export interface ExecutionArgs {
   schema: GraphQLSchema;
@@ -89,7 +127,11 @@ export interface ExecutionArgs {
  *
  * Accepts either an object with named arguments, or individual arguments.
  */
-export function execute(args: ExecutionArgs): PromiseOrValue<ExecutionResult>;
+export function execute(
+  args: ExecutionArgs,
+): PromiseOrValue<
+  ExecutionResult | AsyncIterableIterator<AsyncExecutionResult>
+>;
 export function execute(
   schema: GraphQLSchema,
   document: DocumentNode,
@@ -99,7 +141,9 @@ export function execute(
   operationName?: Maybe<string>,
   fieldResolver?: Maybe<GraphQLFieldResolver<any, any>>,
   typeResolver?: Maybe<GraphQLTypeResolver<any, any>>,
-): PromiseOrValue<ExecutionResult>;
+): PromiseOrValue<
+  ExecutionResult | AsyncIterableIterator<AsyncExecutionResult>
+>;
 
 /**
  * Also implements the "Evaluating requests" section of the GraphQL specification.
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -1,4 +1,5 @@
 import arrayFrom from '../polyfills/arrayFrom';
+import { SYMBOL_ASYNC_ITERATOR } from '../polyfills/symbols';
 
 import type { Path } from '../jsutils/Path';
 import type { ObjMap } from '../jsutils/ObjMap';
@@ -8,6 +9,7 @@ import memoize3 from '../jsutils/memoize3';
 import invariant from '../jsutils/invariant';
 import devAssert from '../jsutils/devAssert';
 import isPromise from '../jsutils/isPromise';
+import isAsyncIterable from '../jsutils/isAsyncIterable';
 import isObjectLike from '../jsutils/isObjectLike';
 import isCollection from '../jsutils/isCollection';
 import promiseReduce from '../jsutils/promiseReduce';
@@ -856,6 +858,74 @@ function completeValue(
 }
 
 /**
+ * Complete a async iterator value by completing the result and calling
+ * recursively until all the results are completed.
+ */
+function completeAsyncIteratorValue(
+  exeContext: ExecutionContext,
+  itemType: GraphQLOutputType,
+  fieldNodes: $ReadOnlyArray<FieldNode>,
+  info: GraphQLResolveInfo,
+  path: Path,
+  iterator: AsyncIterator<mixed>,
+): Promise<$ReadOnlyArray<mixed>> {
+  let containsPromise = false;
+  return new Promise((resolve) => {
+    function next(index, completedResults) {
+      const fieldPath = addPath(path, index, undefined);
+      iterator.next().then(
+        ({ value, done }) => {
+          if (done) {
+            resolve(completedResults);
+            return;
+          }
+          // TODO can the error checking logic be consolidated with completeListValue?
+          try {
+            const completedItem = completeValue(
+              exeContext,
+              itemType,
+              fieldNodes,
+              info,
+              fieldPath,
+              value,
+            );
+            if (isPromise(completedItem)) {
+              containsPromise = true;
+            }
+            completedResults.push(completedItem);
+          } catch (rawError) {
+            completedResults.push(null);
+            const error = locatedError(
+              rawError,
+              fieldNodes,
+              pathToArray(fieldPath),
+            );
+            handleFieldError(error, itemType, exeContext);
+            resolve(completedResults);
+            return;
+          }
+
+          next(index + 1, completedResults);
+        },
+        (rawError) => {
+          completedResults.push(null);
+          const error = locatedError(
+            rawError,
+            fieldNodes,
+            pathToArray(fieldPath),
+          );
+          handleFieldError(error, itemType, exeContext);
+          resolve(completedResults);
+        },
+      );
+    }
+    next(0, []);
+  }).then((completedResults) =>
+    containsPromise ? Promise.all(completedResults) : completedResults,
+  );
+}
+
+/**
  * Complete a list value by completing each item in the list with the
  * inner type
  */
@@ -867,6 +937,21 @@ function completeListValue(
   path: Path,
   result: mixed,
 ): PromiseOrValue<$ReadOnlyArray<mixed>> {
+  const itemType = returnType.ofType;
+
+  if (isAsyncIterable(result)) {
+    const iterator = result[SYMBOL_ASYNC_ITERATOR]();
+
+    return completeAsyncIteratorValue(
+      exeContext,
+      itemType,
+      fieldNodes,
+      info,
+      path,
+      iterator,
+    );
+  }
+
   if (!isCollection(result)) {
     throw new GraphQLError(
       `Expected Iterable, but did not find one for field "${info.parentType.name}.${info.fieldName}".`,
@@ -875,7 +960,6 @@ function completeListValue(
 
   // This is specified as a simple map, however we're optimizing the path
   // where the list contains no Promises by avoiding creating another Promise.
-  const itemType = returnType.ofType;
   let containsPromise = false;
   const completedResults = arrayFrom(result, (item, index) => {
     // No need to modify the info object containing the path,
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -53,6 +53,7 @@ import {
   GraphQLIncludeDirective,
   GraphQLSkipDirective,
   GraphQLDeferDirective,
+  GraphQLStreamDirective,
 } from '../type/directives';
 import {
   isNamedType,
@@ -749,6 +750,42 @@ function getDeferValues(
 }
 
 /**
+ * Returns an object containing the @stream arguments if a field should be
+ * streamed based on the experimental flag, stream directive present and
+ * not disabled by the "if" argument.
+ */
+function getStreamValues(
+  exeContext: ExecutionContext,
+  fieldNodes: $ReadOnlyArray<FieldNode>,
+): void | {|
+  initialCount?: number,
+  label?: string,
+|} {
+  // validation only allows equivalent streams on multiple fields, so it is
+  // safe to only check the first fieldNode for the stream directive
+  const stream = getDirectiveValues(
+    GraphQLStreamDirective,
+    fieldNodes[0],
+    exeContext.variableValues,
+  );
+
+  if (!stream) {
+    return;
+  }
+
+  if (stream.if === false) {
+    return;
+  }
+
+  return {
+    initialCount:
+      // istanbul ignore next (initialCount is required number argument)
+      typeof stream.initialCount === 'number' ? stream.initialCount : undefined,
+    label: typeof stream.label === 'string' ? stream.label : undefined,
+  };
+}
+
+/**
  * Determines if a fragment is applicable to the given type.
  */
 function doesFragmentConditionMatch(
@@ -1040,6 +1077,7 @@ function completeAsyncIteratorValue(
   errors: Array<GraphQLError>,
 ): Promise<$ReadOnlyArray<mixed>> {
   let containsPromise = false;
+  const stream = getStreamValues(exeContext, fieldNodes);
   return new Promise((resolve) => {
     function next(index, completedResults) {
       const fieldPath = addPath(path, index, undefined);
@@ -1076,7 +1114,26 @@ function completeAsyncIteratorValue(
             return;
           }
 
-          next(index + 1, completedResults);
+          const newIndex = index + 1;
+          if (
+            stream &&
+            typeof stream.initialCount === 'number' &&
+            newIndex >= stream.initialCount
+          ) {
+            exeContext.dispatcher.addAsyncIteratorValue(
+              stream.label,
+              newIndex,
+              path,
+              iterator,
+              exeContext,
+              fieldNodes,
+              info,
+              itemType,
+            );
+            resolve(completedResults);
+            return;
+          }
+          next(newIndex, completedResults);
         },
         (rawError) => {
           completedResults.push(null);
@@ -1131,6 +1188,8 @@ function completeListValue(
     );
   }
 
+  const stream = getStreamValues(exeContext, fieldNodes);
+
   // This is specified as a simple map, however we're optimizing the path
   // where the list contains no Promises by avoiding creating another Promise.
   let containsPromise = false;
@@ -1140,6 +1199,23 @@ function completeListValue(
     const itemPath = addPath(path, index, undefined);
     try {
       let completedItem;
+
+      if (
+        stream &&
+        typeof stream.initialCount === 'number' &&
+        index >= stream.initialCount
+      ) {
+        exeContext.dispatcher.addValue(
+          stream.label,
+          itemPath,
+          item,
+          exeContext,
+          fieldNodes,
+          info,
+          itemType,
+        );
+        return;
+      }
       if (isPromise(item)) {
         completedItem = item.then((resolved) =>
           completeValue(
@@ -1182,7 +1258,7 @@ function completeListValue(
       const error = locatedError(rawError, fieldNodes, pathToArray(itemPath));
       return handleFieldError(error, itemType, errors);
     }
-  });
+  }).filter((val) => val !== undefined);
 
   return containsPromise ? Promise.all(completedResults) : completedResults;
 }
@@ -1595,6 +1671,129 @@ export class Dispatcher {
     );
   }
 
+  addValue(
+    label?: string,
+    path: Path,
+    promiseOrData: PromiseOrValue<ObjMap<mixed> | mixed>,
+    exeContext: ExecutionContext,
+    fieldNodes: $ReadOnlyArray<FieldNode>,
+    info: GraphQLResolveInfo,
+    itemType: GraphQLOutputType,
+  ): void {
+    const errors = [];
+    this._subsequentPayloads.push(
+      Promise.resolve(promiseOrData)
+        .then((resolved) =>
+          completeValue(
+            exeContext,
+            itemType,
+            fieldNodes,
+            info,
+            path,
+            resolved,
+            errors,
+          ),
+        )
+        // Note: we don't rely on a `catch` method, but we do expect "thenable"
+        // to take a second callback for the error case.
+        .then(undefined, (rawError) => {
+          const error = locatedError(rawError, fieldNodes, pathToArray(path));
+          return handleFieldError(error, itemType, errors);
+        })
+        .then((data) => ({
+          value: createPatchResult(data, label, path, errors),
+          done: false,
+        })),
+    );
+  }
+
+  addAsyncIteratorValue(
+    label?: string,
+    initialIndex: number,
+    path?: Path,
+    iterator: AsyncIterator<mixed>,
+    exeContext: ExecutionContext,
+    fieldNodes: $ReadOnlyArray<FieldNode>,
+    info: GraphQLResolveInfo,
+    itemType: GraphQLOutputType,
+  ): void {
+    const subsequentPayloads = this._subsequentPayloads;
+    function next(index) {
+      const fieldPath = addPath(path, index);
+      const patchErrors = [];
+      subsequentPayloads.push(
+        iterator.next().then(
+          ({ value: data, done }) => {
+            if (done) {
+              return { value: undefined, done: true };
+            }
+
+            // eslint-disable-next-line node/callback-return
+            next(index + 1);
+
+            try {
+              const completedItem = completeValue(
+                exeContext,
+                itemType,
+                fieldNodes,
+                info,
+                fieldPath,
+                data,
+                patchErrors,
+              );
+
+              if (isPromise(completedItem)) {
+                return completedItem.then((resolveItem) => ({
+                  value: createPatchResult(
+                    resolveItem,
+                    label,
+                    fieldPath,
+                    patchErrors,
+                  ),
+                  done: false,
+                }));
+              }
+
+              return {
+                value: createPatchResult(
+                  completedItem,
+                  label,
+                  fieldPath,
+                  patchErrors,
+                ),
+                done: false,
+              };
+            } catch (rawError) {
+              const error = locatedError(
+                rawError,
+                fieldNodes,
+                pathToArray(fieldPath),
+              );
+              handleFieldError(error, itemType, patchErrors);
+              return {
+                value: createPatchResult(null, label, fieldPath, patchErrors),
+                done: false,
+              };
+            }
+          },
+          (rawError) => {
+            const error = locatedError(
+              rawError,
+              fieldNodes,
+              pathToArray(fieldPath),
+            );
+            handleFieldError(error, itemType, patchErrors);
+            return {
+              value: createPatchResult(null, label, fieldPath, patchErrors),
+              done: false,
+            };
+          },
+        ),
+      );
+    }
+    next(initialIndex);
+  }
+
   _race(): Promise<IteratorResult<ExecutionPatchResult, void>> {
     return new Promise((resolve) => {
       this._subsequentPayloads.forEach((promise) => {
@@ -1611,7 +1810,20 @@ export class Dispatcher {
         );
         return promise;
       })
-      .then(({ value }) => {
+      .then(({ value, done }) => {
+        if (done && this._subsequentPayloads.length === 0) {
+          // async iterable resolver just finished and no more pending payloads
+          return {
+            value: {
+              hasNext: false,
+            },
+            done: false,
+          };
+        } else if (done) {
+          // async iterable resolver just finished but there are pending payloads
+          // return the next one
+          return this._race();
+        }
         const returnValue: ExecutionPatchResult = {
           ...value,
           hasNext: this._subsequentPayloads.length > 0,
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -52,6 +52,7 @@ import {
 import {
   GraphQLIncludeDirective,
   GraphQLSkipDirective,
+  GraphQLDeferDirective,
 } from '../type/directives';
 import {
   isNamedType,
@@ -107,6 +108,7 @@ export type ExecutionContext = {|
   fieldResolver: GraphQLFieldResolver<any, any>,
   typeResolver: GraphQLTypeResolver<any, any>,
   errors: Array<GraphQLError>,
+  dispatcher: Dispatcher,
 |};
 
 /**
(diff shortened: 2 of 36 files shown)


src/execution/execute.js
```diff
@@ -114,6 +118,7 @@ export type ExecutionContext = {|
 export type ExecutionResult = {|
   errors?: $ReadOnlyArray<GraphQLError>,
   data?: ObjMap<mixed> | null,
+  patches?: AsyncIterable<ExecutionPatchResult>,
```
lilianammmatos: The result of `execute` (and also the main `graphql` function) has been modified to return a `patches` field, which is an `AsyncIterable`. We chose not to change the result of `execute` to something like `AsyncIterator<AsyncExecutionResult> | Promise<ExecutionResult>` as that would cause type errors for existing consumers of these functions, even if they do not enable these features in their schema.

lilianammmatos:
If anyone wants to try this out locally, I've published these changes to npm under the package `graphql-experimental`: https://www.npmjs.com/package/graphql-experimental.

lilianammmatos:
If you already have `graphql` installed in a project, but would like to use `graphql-experimental` without updating your import/require statements, you can use:
```
yarn add graphql@npm:graphql-experimental
```
to alias `graphql-experimental` under `graphql`.

[Reference](https://twitter.com/sebmck/status/873958247304232961?lang=en)

src/__tests__/starWarsStreamQuery-test.js
```diff
@@ -0,0 +1,200 @@
...
+          hero {
+            friends {
+              id
+            }
+            ...FriendsName
+            ...FriendsAppearsIn
+          }
+        }
+        fragment FriendsName on Character {
+          friends @stream(label: "nameLabel", initial_count: 1) {
```
Keweiqu: At Facebook, we actually invalidate a query that looks like this. In practice, this is more often a mistake developers make rather than an intentional decision.

To allow developers to create multiple streams on the same field, or to both stream and not stream, we take advantage of the "alias" feature in GraphQL and do something like this:

```
friendsWithName: friends @stream(label: "namedLabel", initial_count: 1)
```

IvanGoncharov:
@lilianammmatos Thanks for pushing this functionality forward 👍 

> however, it became clear to us over time that support for a @defer directive would occur only if it would make it's way into the official GraphQL spec. 

Correct, here is document describing how new features are accepted into spec:
https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md#rfc-contribution-stages

> There has been discussion on this issue to add @defer to the spec, but it seems it's still in it's infancy!

So the biggest blocker here is the absence of champion for this feature:
> An RFC at the proposal stage is a solution to a problem with enough fidelity to be discussed in detail. It must be backed by a willing champion. 

Looking at the amount of work done in this PR would be great to have you as a champion for this feature.
Do you have time for it?

If so, good next step would be to write a sketch of RFC document similar to the one that we had for Subscriptions: https://github.com/graphql/graphql-spec/blob/master/rfcs/Subscriptions.md
You can write the first iteration of this document (`Introduction` and `Background` should be enough) and submit it as PR against `rfcs` folder: 
https://github.com/graphql/graphql-spec/blob/master/rfcs/Subscriptions.md

And we can discuss it on the Working Group meeting if you would be able to join the call on Jan 9th 5 PM UTC:
https://github.com/graphql/graphql-wg/blob/master/agendas/2020-01-09.md
You can add yourself and agenda item in PR similar to this one:
https://github.com/graphql/graphql-wg/pull/311/files

@Keweiqu Big thanks for sharing your expertise and reviewing this PR.
It would be great to have your input on the RFC document and have you at the Working Group call.

morrys:
@lilianammmatos @robrichard @IvanGoncharov 
The most important difference between this PR and my PR proposed here:
https://github.com/graphql/graphql-js/pull/2318
this is the mode in which "patches" are queued.

this mode of patch resolution causes a performance deterioration in the execution of the queries because if the same field is requested in 'n' fragment it is resolved 'n' times.

You can easily test it using the test case I created:
[Directives-perf-test.js](https://github.com/graphql/graphql-js/blob/6f28a7b11a4c9845c78ebc48326d93fec44d4f1d/src/execution/__tests__/directives-perf-test.js)

This problem is also present in the @stream directive, and it is for this reason that I have not proposed its integration.

lilianammmatos:
@IvanGoncharov I do have time to champion this! I'll start working on a sketch of an RFC document and create a pull request against `rfcs` folder ahead of the Working Group meeting on Jan 9th, 5 PM UTC.

I'll add myself and my colleague @robrichard to the agenda.

@Keweiqu thank you very much for your input!

src/type/dispatcher.js
```diff
@@ -0,0 +1,102 @@
...
+    }: any);
+  }
+}
+
+export type ExecutionPatchResult = {
+  errors?: $ReadOnlyArray<GraphQLError>,
+  data?: mixed | null,
+  path: $ReadOnlyArray<string | number>,
+  label: string,
+  ...
```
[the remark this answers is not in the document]
robrichard: @Keweiqu thanks for the feedback! This is very helpful information. Do you think `is_final` in the extensions field is something that needs to be part of the spec for this feature, or is that more of a network implementation detail? In our testing, the patches were sent over a streaming HTTP connection. We were able to signal to Relay that there are no more patches by detecting the HTTP connection has been closed.

I will add some more tests to make sure the `AsyncIterator` is closed correctly for queries that contain multiple `@stream`s and `@defer`s.

src/__tests__/starWarsStreamQuery-test.js
robrichard: @Keweiqu I updated the PR to add a validation rule that forbids streaming on the same field unless the label and initial_count are the same. Does this more closely match the implementation at Facebook?

https://github.com/graphql/graphql-js/pull/2319/commits/c29e1e327e796e53047a7d224eac4ea4eb9d48ed#diff-a49ecb51ef73494dfb5db4effdf536a6R601

Keweiqu: @robrichard Thanks for adding the validation rule. I do think this validation is universally beneficial no matter which infrastructure developers are working with.

src/type/dispatcher.js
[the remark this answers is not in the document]
Keweiqu: @robrichard good question. For the purpose of this PR, my inclination is to aim for only the absolutely essential features, because it's easier to add than to remove. 

If we can signal to Relay through detecting a closed HTTP connection without relying on the "extension" field, and @stream/@defer proves to work fine. Let's start with it as the initial step.

We plan to initiate the discussion of evaluating whether to formalize @defer and @stream as a spec and how to best go about it in the next graphql working group meeting. From there on, we will have more information on whether `is_final` will be beneficial to the open source community at large. It will also help inform whether it's a good idea to include it in the reference implementation of graphql-js.

coco98:
Would it be worth talking about how we expect **@defer** and **@stream** to work with a websocket client? Over the https://github.com/apollographql/subscriptions-transport-ws spec might be a good place to start.

I couldn't find anything online based on a preliminary search. I can take a stab at putting some notes together if there isn't something already :)

robrichard:
@coco98 while `@defer` and `@stream` can be used with a websocket in theory, it is more typically used with a standard HTTP connection with `Transfer-Encoding: chunked`. You can see how this works in this express-graphql PR: https://github.com/graphql/express-graphql/pull/583/files#diff-1fdf421c05c1140f6d71444ea2b27638R536

coco98:
@robrichard Thanks! 

I just meant that it would be nice to have support over websockets as well. I think this would be super convenient if one is using something like apollo client exclusively via the websocket transport. Considering that the current subscriptions-transport-ws supports queries/subscriptions/mutations it might be nice to add some information on what support for defer/stream would look like too. Also, this would kind of play well with the idea that GraphQL is a transport independent spec :)

lilianammmatos:
Hi @coco98! Agreed that this should remain transport independent. As far as I know, there hasn’t been any discussion of how to adopt `@defer` and `@stream` responses over websockets, but I can't see why that wouldn't be possible.

sibelius:
isFinal relay implementation https://github.com/facebook/relay/commit/78d20cc8cdeb8328d2922d840dc76aa51eb1ddd6

mike-marcacci:
Out of curiosity, has there been any discussion around accepting an optional "priority" in `@defer`? I have a use-case with data requirements at more than two "tiers" of importance. Making a selection on a page requires:

1. some data for immediate display (priority=1);
2. supplemental "display-as-available" data (priority=2);
3. backgrounded data that is essentially "pre-loading" data the most-common next actions (priority=3)

In our current scenario, I don't actually distinguish between priority 1 and 2 data, since doing so hasn't been ergonomic before `@prefer`, and priority 3 data is made as a separate request with heuristics that delay it until the higher-priority request is likely already in-flight. This has been working, but it feels like defer and stream provide almost all the machinery to solve this far more elegantly, except for the ability to distinguish between priority 2 and 3.

Has this kind of functionality been considered internally?

robrichard:
Hi @mike-marcacci - https://github.com/GraphQL/graphql-spec/issues/691 has been opened to discuss optimistic preloading.

sibelius:
can we resync this with graphql v15 https://github.com/graphql/graphql-js/releases/tag/v15.0.0?

src/execution/dispatcher.js
```diff
@@ -84,5 +94,6 @@ export type ExecutionPatchResult = {
   data?: mixed | null,
   path: $ReadOnlyArray<string | number>,
   label: string,
+  isFinal?: boolean,
```
twhid, line 97: Why aren't we ditching the flow?

lilianammmatos:
@Keweiqu @jhusain @josephsavona @mjmahone @IvanGoncharov we've pushed up the last of our updates for this `@defer` and `@stream` reference implementation! We would appreciate any feedback. Thanks!

src/execution/execute.d.ts
```diff
@@ -54,6 +55,7 @@ export interface ExecutionResult<
   // TS_SPECIFIC: TData. Motivation: https://github.com/graphql/graphql-js/pull/2490#issuecomment-639154229
   data?: TData | null;
   extensions?: TExtensions;
```
IvanGoncharov, line 57: extensions should always be the last field in all types.

src/execution/execute.d.ts
```diff
@@ -66,6 +68,29 @@ export interface FormattedExecutionResult<
...
 
+/**
+ * The result of an asynchronous GraphQL patch.
+ *
+ *   - `errors` is included when any errors occurred as a non-empty array.
+ *   - `data` is the result of the additional asynchronous data.
+ *   - `path` is the location of data.
+ *   - `label` is the label provided to @defer or @stream.
+ *   - `hasNext` is true if a future payload is expected.
+ */
```
IvanGoncharov, line 79: missing `extensions`

src/execution/execute.js
```diff
@@ -106,6 +110,13 @@ export type ExecutionContext = {|
   fieldResolver: GraphQLFieldResolver<any, any>,
   typeResolver: GraphQLTypeResolver<any, any>,
   errors: Array<GraphQLError>,
+  dispatcher: Dispatcher,
+|};
+
+export type FormattedExecutionResult = {|
+  errors?: $ReadOnlyArray<GraphQLFormattedError>,
+  data?: ObjMap<mixed> | null,
+  extensions?: ObjMap<mixed>,
```
IvanGoncharov, line 119: Merge conflict?
It also missing `hasNext`

src/subscription/subscribe.js
```diff
@@ -156,16 +155,18 @@ function subscribeImpl(
 
   // Resolve the Source Stream, then map every source value to a
   // ExecutionResult value as described above.
-  return sourcePromise.then((resultOrStream) =>
+  return sourcePromise.then((resultOrStream) => {
     // Note: Flow can't refine isAsyncIterable, so explicit casts are used.
```
IvanGoncharov, line 159: Now that you have `boolean %checks(value instanceof AsyncIterable)` in `isAsyncIterable` is it still true?

src/jsutils/isAsyncIterable.js
```diff
@@ -0,0 +1,19 @@
...
+  AsyncIterable);
+
+// eslint-disable-next-line no-redeclare
+export default function isAsyncIterable(maybeAsyncIterable) {
+  if (maybeAsyncIterable == null || typeof maybeAsyncIterable !== 'object') {
+    return false;
+  }
+
+  return typeof maybeAsyncIterable[SYMBOL_ASYNC_ITERATOR] === 'function';
+}
```
IvanGoncharov, line 19: Can you please split this change into separate PR to make this PR smaller?

src/type/directives.js
```diff
@@ -169,6 +169,53 @@ export const GraphQLSkipDirective = new GraphQLDirective({
...
+      type: GraphQLString,
+      description: 'Unique name',
+    },
+  },
+});
+
+/**
+ * Used to conditionally defer fragments.
+ */
+export const GraphQLStreamDirective = new GraphQLDirective({
```
IvanGoncharov, line 198: Can you please add them into `specifiedDirectives`?

IvanGoncharov:
> we've pushed up the last of our updates for this @defer and @stream reference implementation! We would appreciate any feedback. Thanks!

Thanks, 👍 I gave it a quick look and left a few minor comments.
The biggest obstacle is that this PR changes return value of `execute` so it produces a huge number of type checking errors for TS & Flow users so we can merge it only in next breaking version (hopefully we will switch master to `16.0.0` in the next few weeks).
Also, since we changing execute to return non-trivial type anyway we can also consider #2141 
I need little bit more time to think about `execute`s return values since I still has some concerns similar to this one: https://github.com/graphql/graphql-js/issues/2144#issuecomment-527076443

So I Propose to put reviewing/merging this PR on a temporary hold until we start 16.0.0-alpha line and switch to TS (ETA ~1month from now). 
In the meantime, I can merge it in a branch (to make it more official and allow other devs to submit PRs against it). Also, I can set up a special branch with NPM package in it similarly to what we currently have for master:
@lilianammmatos What do you think about these steps?
Also, if you want we can publish your blog post about this ongoing effort on GraphQL Foundation blog or promote any other media that will motivate more people to try this feature and provide feedback.

src/jsutils/isAsyncIterable.js
robrichard: @IvanGoncharov I moved support for returning AsyncIterables from resolvers to a new PR: https://github.com/graphql/graphql-js/pull/2712

I don't think there are any breaking changes there, and it will cut down a significant part of this PR. This includes the change of moving `isAsyncIterable` to a new file. Let me know if it should be broken down further.

src/type/directives.js
robrichard: @IvanGoncharov we didn't put these into `specifiedDirectives` because they would be included in schemas by default. We only wanted the directives to be in the schema when the `experimentalDefer`/`experimentalStream` flags are passed.

src/subscription/subscribe.js
robrichard: @IvanGoncharov I left a comment addressing this here: https://github.com/graphql/graphql-js/pull/2712#discussion_r455113122

lilianammmatos:
@IvanGoncharov thanks for taking a first pass at these changes! @robrichard addressed the comments you left, but let us know if you have any additional feedback.

>  I Propose to put reviewing/merging this PR on a temporary hold until we start 16.0.0-alpha line and switch to TS (ETA ~1month from now).

That sounds reasonable! However, would it be at all possible to [merge the PR that @robrichard opened](com/graphql/graphql-js/pull/2712) that supports returning async iterables from resolver functions into `master`? (It's a subset of the changes in this PR that doesn't contain any breaking changes.)

> In the meantime, I can merge it in a branch (to make it more official and allow other devs to submit PRs against it). Also, I can set up a special branch with NPM package in it similarly to what we currently have for master:

Yes, that sounds great! If you could let me know when that's set up I can create another PR against the special branch.

> Also, if you want we can publish your blog post about this ongoing effort on the GraphQL Foundation blog or promote any other media that will motivate more people to try this feature and provide feedback.

We don't have anything ready at the moment, but we are working on writing a blog post to help promote early adoption of `@defer` and `@stream`. I'll keep you updated!

IvanGoncharov:
> Yes, that sounds great! If you could let me know when that's set up I can create another PR against the special branch.

@lilianammmatos I will create branch once #2712 is merged. Do you have an idea for the branch name?
How about `defer-stream` since it what you are using in your fork atm?

lilianammmatos:
`defer-stream` works for me! 👍

yaacovCR:
Hi!

I am curious about response to https://github.com/graphql/graphql-js/pull/2319#issuecomment-570901787 as well as in general the tradeoffs in applying defer to fragments rather than fields. I would imagine that expected behavior -- as defer is a hint -- would be to collect all fields, combining as necessary, and if any fields do not have defer, then defer would not be honored, with defer on fragment being transitive to all fields within a fragment.

@morrys I think I understand the comment btw with respect to defer but not necessarily with respect to stream. Are you able to elaborate?

yaacovCR:
Am I correct in assuming that the need for a label stems from defer on identical fields within different fragments not getting combined, and that thereby serve to identify the why of the individual payloads more than the where, which is actually fully explained by the path?

yaacovCR:
The last comment with regard to label seems to apply especially to stream as well, as test cases document that when path is the same (no alias), label must be the same. I would also suggest that initial batch size be allowed to conflict, as use of these directives as hints only would suggest to honor stream only if identical fields in all selection sets have directive and to use the largest initial batch size...

yaacovCR:
@lilianammmatos @robrichard @IvanGoncharov 

Is it more appropriate to add these comments to the spec RFC issue rather than this implementing PR. Please also let me know if there is a discussion available online explaining these issues. I took a look at the working group notes, but found references to the RFC details but not tradeoffs/alternatives

robrichard:
@yaacovCR thanks for your questions.

> as defer is a hint

The RFC was updated to explain that `@defer` and `@stream` should not be treated as a hint. It's important for product developers to be able to rely on `@defer` and `@stream`. If they was unreliable, the developer may not be able to use them as the performance hit of the directives being ignored would be unacceptable. The spec will only allow advanced use-cases where the server can determine that it is more performant to not defer/stream, but in general it should be expected that all deferred data will actually be deferred. More here: https://github.com/graphql/graphql-spec/issues/704#issuecomment-613438677

> tradeoffs in applying defer to fragments rather than fields

If there is a UI component that renders many fields which are deferred, it would be cumbersome to coordinate the loading state of all of those fields. By deferring all of the fields on a fragment, the component can render its fallback loading state until the fragment is loaded. If only a single field is needed to be deferred it can easily be wrapped in an inline fragment so we didn't think it was necessary to support `@defer` on a field.

> response to #2319 

This was discussed in one of the working group meetings. The consensus was that if a field is present in both a deferred and non-deferred fragment, it's reasonable to expect the resolver to be re-executed, because it could be complex for the server to determine if fields are equivalent when they contain many deeply nested fields.

> label

Label is there because it's required for clients to understand which payloads correspond to which `@defer` or `@stream` directive. This cannot be identified by path alone, e.g. two deferred fragments under the same field. It's necessary for clients that generate types based on fragments to understand which fragments have been loaded and which are still pending.

> suggest that initial batch size be allowed to conflict

This was discussed here: https://github.com/graphql/graphql-js/pull/2319#discussion_r362990726 We're keeping the validation rule strict since its more likely to be a mistake than intentional, and looser validation doesn't seem to enable any additional use cases.

> Is it more appropriate to add these comments to the spec RFC issue

Might be better to discuss these types of issues there since they are not specific to the JS implementation.

yaacovCR:
Thanks for clarification!

yaacovCR:
Thanks again for clarification above and further clarification on the linked spec issue.

I am working on adding support for defer and stream to graphql-tools delegation, schema stitching, etc.

I believe it would be very helpful for the label for deferred fragments to be passed through to executeFields, resolveField, and buildResolveInfo.

This is because schema proxying in GraphQL Tools essentially works by building a request from info and then returning the data within the parent resolver and providing a defaultMergedResolver for child resolvers as necessary for all proxied type field that takes care of aliases, errors, etc, etc.

My plan for implementing defer support was to pass through the defer directive, and if you get an async iterable back, await the initial result, and patch that result with the additional patches as they come in, and enhance the defaultMergedResolver to be smart enough to await the patches it needs to await, which would seem to be identified by the path/label combination.

The path of course is already within the info object, but the label isn't. It is probably easily extracted from the original operation, but it should be simple to pass it through, which seems to make logical sense, as the path is no longer unique, it would seem to make sense to supplement info with the label.

Granted, the label is optional, and if not set to a unique value for that path, it may be difficult to make this work...

I am also voting for label names to not be allowed to conflict for that reason, but either way, I think it might be helpful to pass it through.

yaacovCR:
Actually, I would need the nested list of labels to be passed through, as each defer makes a new branch point, right?

Would you be open to PR adding this to this PR?

lilianammmatos:
Hi @yaacovCR! Thanks again for your input. The aim of this pull request is to provide a reference implementation for the specification proposal and I think the changes you are requesting might be outside of its scope. The specification proposal for these directives is not final and while we hope the proposal does not significantly change, it might complicate things if we start adding additional functionality before it is finalized.

I believe it would be better to open a separate pull request once this pull request is merged that would provide support for defer and stream for schema stitching and the other functionality graphql-tools provides!

yaacovCR:
Hi @lilianammmatos! Thanks so much to you and everyone at 1stdibs pushing ahead support for defer and stream. I agree with the sentiments expressed here and elsewhere that it is a very exciting step forward.

While I think it does not matter too much in terms of pull request mechanics as to whether the change I am requesting becomes part of a different PR or part of this PR, or a PR to this PR, I I disagree in terms of whether the change is out of scope.

The [spec](http://spec.graphql.org/draft/#sec-Executing-Fields) does not seem to say anything about either the context or info arguments to resolvers, but the reference implementation provides support for both. Presumably this is because graphql-js as a reference implementation also represents a minimal real-world implementation, and these arguments represent low-hanging fruit in terms of good ideas --albeit not within the spec.

Whatever standard this repository has for defining the exact level of non-`spec`ified functionality, I guess I am saying, it should not be changed.

Said differently, if the intent of the info argument is to provide a certain level of info, this PR subtracts somewhat from that level, by no longer providing individual resolver functions/methods with their individual "spots" within the overall execution.

That subtraction inhibits certain use cases, including schema/query proxying. If this implementation of defer/stream makes it impossible to provide that info, I don't think it should be merged at all, in favor of a spec-compliant implementation that also provides the full info. If that cannot be done, the spec change itself might need to be reconsidered -- for example, the part about labels being options for defer.

I find that outcome doubtful -- as I think this would be trivial to add. One never really knows unless one tries -- I imagine that is the point of the reference implementation!

If you've made it this far, thanks for giving my request the time!

josephsavona:
> The aim of this pull request is to provide a reference implementation for the specification proposal and I think the changes you are requesting might be outside of its scope. The specification proposal for these directives is not final and while we hope the proposal does not significantly change, it might complicate things if we start adding additional functionality before it is finalized.

> I believe it would be better to open a separate pull request once this pull request is merged that would provide support for defer and stream for schema stitching and the other functionality graphql-tools provides!

Just chiming in to echo @lilianammmatos's answer here. Given that this is the first pass of an experimental feature that won't be enabled by default (presumably, i haven't read through all the code yet), it should absolutely not be required that we support every possible integration on this very first PR. If further changes are needed to support graphql-tools or some other library, that should absolutely be considered but definitely not block landing the core implementation here.

yaacovCR:
>  it should absolutely not be required that we support every possible integration on this very first PR

I agree with this statement as written, but the statement does not accurately describe my request.

I am suggesting that the experimental defer/stream enabled branch of the reference repository for graphql-js retain throughout the experimentation period the ability of the `info` argument to accurately provide resolvers with their position within the execution tree when possible.

As a proof of concept -- or possibly a simple resolution pleasing to all parties - I have created a PR to this PR that does just including a single successful test: https://github.com/lilianammmatos/graphql-js/pull/5

CI shows a problem with code coverage check. I would be happy to add tests for the missing lines, but I could not immediately see from the logs what they were, just the count: 453, not sure whether they are all due to the change I made.

CI also shows a problem initiating the benchmark tests, which should be extraneous to this change, as the logs seem to indicate that they are not kicked off at all.

Again, thanks to you all for your work pushing defer and stream forward.

yaacovCR:
>  it should absolutely not be required that we support every possible integration on this very first PR

I agree with this statement as written, but the statement does not accurately describe my request.

I am suggesting that the experimental defer/stream enabled branch of the reference repository for graphql-js retain 
 -- throughout the experimentation period -- the ability of the `info` argument to accurately provide resolvers with their position within the execution tree when possible.

As a proof of concept -- or possibly a simple resolution pleasing to all parties - I have created a PR to this PR that does so: https://github.com/lilianammmatos/graphql-js/pull/5.

The most recent changes to that PR restore 100% code coverage.

Again, thanks to you all for your work pushing defer and stream forward. I hope this small PR helps in that goal and do not see any obstacles to its inclusion within the experimental branch.

src/execution/execute.js
```diff
@@ -164,7 +199,7 @@ declare function execute(
   operationName?: ?string,
   fieldResolver?: ?GraphQLFieldResolver<any, any>,
   typeResolver?: ?GraphQLTypeResolver<any, any>,
-): PromiseOrValue<ExecutionResult>;
+): PromiseOrValue<ExecutionResult | AsyncIterable<AsyncExecutionResult>>;
```
yaacovCR, line 202: Looks like this change (possibly others?) did not make it over to execute.d.ts

yaacovCR:
See https://github.com/lilianammmatos/graphql-js/pull/6

In terms of https://github.com/lilianammmatos/graphql-js/pull/5 it turns out this is neither sufficient nor necessary for my own use case, but I still believe should be merged for the reasons outlined above. Not sure I understand the resistance. :)

robrichard:
@IvanGoncharov I set up the `experimental` subdirectory containing the new versions of `execute` and `graphql`. There should be no breaking changes in this PR now.

Reminder that this is rebased on top of https://github.com/graphql/graphql-js/pull/2712 which should be reviewed first.

yaacovCR:
Out of curiosity why does the new experimental execute return an AsyncIterable and existing subscribe return an AsyncIterableIterator...

Which is "correct"?

src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.js
```diff
@@ -96,6 +96,137 @@ describe('Validate: Overlapping fields can be merged', () => {
     `);
   });
 
+  it('Same stream directives supported', () => {
+    // @stream is allowed on overlapping fields when they have the same label and
+    // initial count.
```
IvanGoncharov, line 101: nit: It's ok for comments to be longer than 80 characters and overwrap.

AndreasHald:
What is the status of this PR? can't wait for this functionality since i saw the talk at graphql summit.

acao:
so excited about this! let me know if there’s anything i need to do from the LSP/graphiql side. there is a bit of logic for existing built-in directives

n1ru4l:
> Out of curiosity why does the new experimental execute return an AsyncIterable and existing subscribe return an AsyncIterableIterator...
> 
> 
> 
> Which is "correct"?

You probably always want to use AsyncIterableIterator as it allows iterating over it using a async generator function (`async * () { for await (const value of asyncIterableIterator) }`). As as AsyncIterableIterator has a property under the Symbol.asyncIterator key.

There is currently an issue with TypeScript typings and Symbol polyfills (https://github.com/leebyron/iterall/issues/49). Because of that people type their `AsyncIterableIterator` as `AsyncIterable` instead. (e.g. https://github.com/apollographql/graphql-subscriptions/pull/232)

I am also very excited about this as returning `AsyncIterableIterators` from execute would make implementing live queries easier 🤩 (https://github.com/n1ru4l/graphql-live-queries/blob/main/packages/graphql-live-query/src/executeTypes.ts).

sibelius:
can we merge and release this as graphql@experimental

as react@experimental, relay@experimental ?

robrichard:
@acao We have not had a chance to look at graphiql support for defer/stream yet. It would be great if you have some time to see what's needed get it working. 

@sibelius we are working with @IvanGoncharov to publish this as an experimental tag on graphql-js after the next version of graphql-js is published.

src/execution/execute.d.ts
```diff
@@ -89,7 +127,9 @@ export interface ExecutionArgs {
  *
  * Accepts either an object with named arguments, or individual arguments.
  */
-export function execute(args: ExecutionArgs): PromiseOrValue<ExecutionResult>;
+export function execute(
+  args: ExecutionArgs,
+): PromiseOrValue<ExecutionResult | AsyncIterable<AsyncExecutionResult>>;
```
n1ru4l, line 132: AsyncIterableIterator might be the better return type as it can easily be looped over with a for await of loop

robrichard: Updated to `AsyncIterableIterator`

acao:
so I yarn linked this and the `express-graphql` PR up with `graphiql` monorepo real quick and it looks great so far, amazing work everyone! The only thing that doesn't work out of the box on the graphiql end of things is parsing and rendering the multipart responses.

**what works transparently**

- `codemirror-graphql` fragment and `@defer`/`@stream` completion are working
- inline codemirror-graphql fragment validation (placement, invalid types) and field argument validation are working as to spec, with the messages and all
- hover and completion documentation work as to spec

as most of us know, because we're using graphql@15, the introspection schema gives us plenty of information about directives and their arguments.  previously, in graphql@14 era, the language service interface and parser had some baked-in spec directives logic because introspection didn't deal with fragments then

**what doesn't work out of the box?**

as expected, I'm getting raw multipart responses as per the relay spec, so we might need to implement [fetch-multipart-graphql](https://github.com/relay-tools/fetch-multipart-graphql) to parse the responses.

also, pinging @Urigo because we may have to consider this for the subscriptions fetcher, and supporting deferred schema

1. completion
image

2. validation from graphql-js
image

3. raw multipart response
image

graphiql PR forthcoming when I next get a chance, unless someone can beat me to it, haha!

n1ru4l:
Is it actually any recommendation or discussion ongoing about where to implement the initial data + patch merging on the consumer? E.g. on fetcher/network (apollo link) or client layer (apollo-cache)?

Currently, I think guess the network layer is the way to go?

acao:
@n1ru4l looks like [this implementation](https://github.com/relay-tools/fetch-multipart-graphql/blob/master/src/PatchResolver.js) by the authors has a prescribed pattern for consumers as such, and it works with both apollo server and express-graphql (as per @robrichard 's notes [in the express-graphql PR](https://github.com/graphql/express-graphql/pull/583)). Thanks @lilianammmatos and rob for building out such a helpful ecosystem and excellent PRs and documentation for this effort! Was very easy to get caught up on the latest details, and to get everything working.

I'm thinking that, as with subscriptions, we will make this an opt-in feature via the `fetcher` prop in 1.0.x. We are going to adopt @Urigo's wss subscriptions fetcher, and can reccomend an implementation with `fetch-multipart-graphql` so that folks can opt-in to either or both features.

if `graphiql` ends up introducing a higher-level service layer, where we generate the fetcher function (in primitive form, not template replacement) using user input, there can be an `httpMultipartDeferStream` boolean option that enables this for the generated fetcher.

that way, if people want to implement `@defer` or `@stream` using a non-http implementation, they are free to handle that however using the `fetcher` prop

creating a discusson in graphiql for this, so I don't get things off track... will update this comment with the link shortly

robrichard:
This branch is published to npm as version `graphql@15.4.0-experimental-stream-defer.1` under the tag `graphql@experimental-stream-defer`. We encourage anyone interested in this feature to try out this version and report any feedback, issues, or potential improvements to the API. This feedback will help the working group gain confidence to move forward with standardization.

n1ru4l:
I also created https://github.com/n1ru4l/graphql-bleeding-edge-playground for those that want to have a jump start without having to do all the boilerplate themselves

AndreasHald:
Anyone have working examples of the experimental branch using apollo graphql server?

n1ru4l:
@AndreasHald I think it is not possible without manual patching the apollo-server-core library.

AndreasHald:
@n1ru4l dang :( probably got a while to go then. thanks for the quick reply :)

johnculviner:
Looking forward to this one! I'm seeing it on GQL libs for Java, Scala and Ruby now too.

This solves one of my biggest gripes with GQL - your response performance is limited to the performance characteristics of the slowest resolver dependency unless you issue multiple queries which really limits the ability of GQL to serve as a near perfect abstraction over legacy APIs etc.

josephsavona:
> Is there any consensus or discussion ongoing about where to implement the initial data + patch merging on the GraphQL API consumer? Should the fetcher/network (e.g. apollo-link/relay-fetcher) layer always resolve with a ExecutionResult that is built from the initial result and the applied patches that is then consumed by the cache/store/layer or should the client/cache layer (e.g apollo-cache) be able to apply the sent patches to the cache directly?

Relay implements support for `@defer` and `@stream` directly in the core of the runtime. During the processing of an operation we maintain some state about which payloads have been received and which payloads are pending, and incrementally process each new payload unit. This approach allows us to avoid processing previously received data multiple times, as would happen if we were to apply patches at the network layer and push full payloads to the runtime repeatedly. I imagine other clients with a consistent cache will likely use a similar approach, while simpler clients without a cache can patch payloads and re-vend the entire updated response object.

glasser:
Just to make sure I understand this PR: if you are calling `execute` now, it can always be the case that it will return an async iterator of patches? You don't have to opt in to supporting defer/stream by passing some sort of parameter that says "I am going to be serving the result of this `execute` call over a transport that supports incremental delivery, please honor `@defer`/`@stream`"? The defer/stream spec says it's OK to ignore the directives if you don't support them and that clients need to support that, so it seems odd that a version of `graphql-js` with this patch would be unusable in contexts that don't support incremental delivery. Am I missing something?

acao:
@glasse I think this detail may end up being transport specific. in the http multipart [`IncrementalDelivery’ spec rfc](https://github.com/graphql/graphql-over-http/blob/main/rfcs/IncrementalDelivery.md), the client knows to a response is incremental when the server response includes the `multipart/mixed` content type header.

PR #2344, opened 2020-01-13
fix: 🐛#1819 ESM output imports, add file extensions

Banou26:


Banou26:
Any ideas on why only the Node v12 failed tests ?

Seems like the prettier threw, and after that  the ci tests threw.

If we direct me towards the problem i could try to fix it, or if it's just a CI problem and it's not important, you can just review and merge it.

IvanGoncharov:
> Any ideas on why only the Node v12 failed tests ?

Please run "npm test" or yarn test" locally and fix all errors.
Also can you please explain what you are trying to do and what is your use case for these changes?

Banou26:
There were no errors during the tests, and for good reasons, this PR isn't to modify the behavior of graphql.

Its only purpose is to make the ESM output of the build ESM compliant by having import and export paths use file extensions, using extension-less import paths result in errors such as this one when using your package with node in module mode(`"type"="module"`).
```
Error: Cannot find module C:\dev\oz\packages\api\node_modules\graphql\language\source imported from C:\dev\oz\packages\api\node_modules\graphql\language\index.mjs
```

IvanGoncharov:
@Banou26 Thanks for PR 👍 
I decided to implement it slight different in #2379 
It's now merged and will be released as `15.0.0-rc.2`

PR #2362, opened 2020-01-19
feat: Ignore sorting for queries and mutations when printing schema

wtrocki:
## Motivation

When using `printSchema` it's implementation will automatically sort all types by names. 
While for user types scalars and inputs this makes sense, for very large schemas finding mutation, query etc. can be challenging without using find command. Working with that schema layout is also hard as it is usually better to have mutations and queries together. 

I heard opinions from the community that would expect schema to be organized the same way as it was before printing was applied. Print schema is often used by some CLI tools where we just want to append some specific object to the schema while operating on ast objects rather than strings. It will be amazing to see print schema being used more for those use cases.

# How this could be done

## Always put queries/mutations on the bottom of printed schema

This is implemented in this PR as I thought that it will be the least controversial change, but it looks like it is causing failures to some tests that rely on print schema order. 
I wanted to make sure that the community will get that behavior out of the box without using some advanced functions etc. IMHO Scalars and directives could be also reorganized

## Allow developers to apply their own sorting function 

Currently, the order of the types in schema is not preserved as they are stored in map. 
What it means is if users organized schema to their liking in DSL and then we parsed that into object original order is lost (not 100% sure about this but this is what I see happening)

To be able to get some organization when printing schema an separate sorting function could be applied. For example, first list scalars scalars and directives. Then user types in alphabetical order. Then Queries, Mutations. Subscriptions. 

This can be done by adding additional parameter to options or exposing `printFilteredSchema` for advanced use cases


cmwylie19:
Excellent enhancement 👍

wtrocki:
The build is failing due to not updated unit tests snapshots. I will hold of with updates to not obfuscate the change that was done

pratiklodha:
Yes. This is a very good enhancement.

wtrocki:
I totally forgot that schema will contain `loc` fields that can be used to fully reconstruct order of items. Using loc would be much better than putting queries on the bottom etc.

IvanGoncharov:
@wtrocki Thanks for PR 👍 
Totally agree that it's something we need to address.

> I totally forgot that schema will contain loc fields that can be used to fully reconstruct order of items. Using loc would be much better than putting queries on the bottom etc.

Only schemas built from SDL has `loc` 😞 
But in many cases people try to print remote schema s that are built with `buildClientSchema` that totally doesn't have AST nodes at all.
Both SDL and introspection (because types are sent as an array) have an order we just need to preserve it during `GraphQLSchema` construction.

I experimented with a few different approaches and found one that keeps the order intact without introducing the performance penalty and I'm working on PR.
So my proposal is to remove type sorting completely from `printSchema` but if working with the non-stable schema you can always do `printSchema(lexicographicSortSchema(schema))` instead.
What do you think?

wtrocki:
Thank you so much for reaching out. I have done some and found out that order will be preserved when iterating using Object.keys

```
Note: Since ECMAScript 2015, objects do preserve creation order for string and Symbol keys. In JavaScript engines that comply with the ECMEScript 2015 spec, iterating over an object with only string keys will yield the keys in order of insertion.
```
Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

> So my proposal is to remove type sorting completely from printSchema.

Basing on what I see that could be the best way to get this.
There is simply no need to filter things out.
A very simplified version of printing could look as follows:

https://gist.github.com/wtrocki/df0f9cdca8f3c41d4d607a57a211e68f

> printSchema(lexicographicSortSchema(schema)) instead.

That will be also a good alternative

IvanGoncharov:
@wtrocki Thank you for taking time to write a detailed description and PoC.
Fixed in #2411 and will be released as `15.0.0-rc.2`

PR #2364, opened 2020-01-20
Remove 'iterall' as dependency. 'graphql' becomes zero dependency lib 🎉

IvanGoncharov:
We used only couple functions from 'iterall' and it's the only
dependency we have at the moment.
Having no dependencies significantly simplify ESM builds and Deno
support in future
Partially resolves: #2277

Bonus: All supported Node versions already natively support Array.from
so it should improve perfomance

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -7,7 +7,6 @@
   "words": [
     "jsutils",
     "tsutils",
-    "iterall",
     "noflow",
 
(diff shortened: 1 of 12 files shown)


lukejacksonn:
This is awesome @IvanGoncharov 👏 one step closer to a browser compatible ESM build!

PR #2368, opened 2020-01-20
Makes explicit imports that rely on index file inference

lukejacksonn:
Partially resolves https://github.com/graphql/graphql-js/pull/2277 in combination with https://github.com/graphql/graphql-js/pull/2364

There are some import paths that rely on the bundler to resolve `./somefolder` to `./somefolder/index.js` at build time. This doesn't play well with the es module spec which demands import paths be complete.

Adding `/index` to these import statements (which only exist in the root `index.js` file) is minimally invasive and has no adverse effects on any of the existing builds but opens the door to a trivial transformation which adds `.es.js` extensions to import paths in order to generate a fully esm compatible build from source.

Since the project became dependency free the only manual edit to the source after this PR (in order to unlock browser esm compatibility) is to remove the [unprotected use of `process.env`](https://github.com/graphql/graphql-js/pull/2277/files#diff-8ae9c199e45f5279dddc0db0fe063462). I could pull in the changes from that commit to this PR or make a separate PR.

After that.. the babel transform ([which exists here](https://github.com/graphql/graphql-js/pull/2277/files#diff-8363ab4678bef8ef4d374bd410e88532)) would be all that is required.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -125,7 +125,7 @@ export {
   // Validate GraphQL schema.
   validateSchema,
   assertValidSchema,
-} from './type';
+} from './type/index';
 
 export type {
   GraphQLType,
@@ -168,7 +168,7 @@ export type {
   GraphQLScalarSerializer,
   GraphQLScalarValueParser,
   GraphQLScalarLiteralParser,
-} from './type';
+} from './type/index';
 
 // Parse and operate on GraphQL language source files.
 export {
@@ -203,7 +203,7 @@ export {
   isTypeDefinitionNode,
   isTypeSystemExtensionNode,
   isTypeExtensionNode,
-} from './language';
+} from './language/index';
 
 export type {
   ParseOptions,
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@lukejacksonn Thank 👍 
The amount of work you put into #2277 motivated me to try to squeeze this feature into `15.0.0` release.
So I slowly going through changes you made in #2277 and trying to figure out how to merge them. For example, I don't want to merge `/index` without ESLint rule that verifies that we will not have regressions in the future.
Also, I want to keep `*.d.ts` files in sync so I opened #2377

lukejacksonn:
That is good to hear! So what you are saying is that this PR needs to be accompanied by one which adds an ES Lint rule which demands that developers always provide full paths (i.e. `./something/index` rather than `./something`)?

IvanGoncharov:
Fixed in #2377

PR #2379, opened 2020-01-20
build: mjs files import other files using full paths with extensions

IvanGoncharov:
Motivation #2277

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -19,7 +19,8 @@
           "presets": [["@babel/preset-env", { "modules": "commonjs" }]]
         },
         "mjs": {
-          "presets": [["@babel/preset-env", { "modules": false }]]
+          "presets": [["@babel/preset-env", { "modules": false }]],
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@lukejacksonn Big difference for this one and #2277 is that it adds extensions to all imports inside `*.mjs` files.
I don't really want to add separate `*.es.js` files since we already generating 4 files for every source file:
image
Do you see any potential problems with publishing both MJS and ESM builds as `*.mjs` files?

lukejacksonn:
Well, that sure simplifies things! I like the idea.. wasn't too sure that importing `.mjs` still worked in the browser but I just checked and it seems like it does so I can't see there being a problem there.

IvanGoncharov:
@lukejacksonn Thanks for trying it out 👍 
There is still a chance that extensions will break some setups so treat it as an experimental feature and in the worst-case scenario, we will switch to `*.es.js` files.

PR #2382, opened 2020-01-22
Change serialization for Int, Float and Boolean to work with objects

alex-knyazev:
It is good to have opportunity to pass objects which have valueOf() method in resolvers of Int and Float types.

For example, I have such objects as "Money", "Price" and so on in my app, and they have valueOf() method which return value. Now I need to call valueOf() method in every resolver of such field or write custom scalar or directive.

I think changes I did  are implied in native type behavior of Int, Float and Boolean fields.

### DIFF
--- a/src/type/scalars.js
+++ b/src/type/scalars.js
@@ -196,14 +196,16 @@ export const GraphQLString = new GraphQLScalarType({
 });
 
 function serializeBoolean(outputValue: mixed): boolean {
-  if (typeof outputValue === 'boolean') {
-    return outputValue;
+  const coercedValue = serializeObject(outputValue);
+
+  if (typeof coercedValue === 'boolean') {
+    return coercedValue;
   }
-  if (isFinite(outputValue)) {
-    return outputValue !== 0;
+  if (isFinite(coercedValue)) {
+    return coercedValue !== 0;
   }
   throw new GraphQLError(
-    `Boolean cannot represent a non boolean value: ${inspect(outputValue)}`,
+    `Boolean cannot represent a non boolean value: ${inspect(coercedValue)}`,
   );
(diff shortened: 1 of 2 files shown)


src/type/scalars.js
```diff
@@ -31,6 +31,15 @@ function serializeInt(outputValue: mixed): number {
...
   }
 
+  if (
+    outputValue !== null &&
+    typeof outputValue === 'object' &&
+    typeof outputValue.valueOf === 'function' &&
+    !Array.isArray(outputValue)
+  ) {
+    num = Number(outputValue.valueOf());
+  }
```
IvanGoncharov: Please use `serializeObject` instead and add it `Int`, `Float` and `Boolean`.
Also please update tests in https://github.com/graphql/graphql-js/blob/master/src/type/__tests__/serialization-test.js

So basically, please replicate how this is done for ID and String.

alex-knyazev: Don't you think than using of serializeObject method is incorrect in this situation (it contains checking that object has toJSON() method

alex-knyazev: @IvanGoncharov  I did it

IvanGoncharov: > Is it good idea to squash my 4 commits to one?

@alex-knyazev Both work, since it's small commit I can just "squash and merge" in GitHub UI.
Please don't forget to make the same changes and tests for `GraphQLBoolean` type.

alex-knyazev: @IvanGoncharov  I've added boolean type

IvanGoncharov:
@alex-knyazev Merged 🎉 
I plan to release `15.0.0-rc.2` in the next few days.

PR #2402, opened 2020-01-27
Add 'Rule' suffix to all validation rules

IvanGoncharov:
Technically it's not a breaking change since only stuff exported from
`graphql` and `graphql/validation` are part of the public API.
However, many tools/libraries depended on this "internal" imports to
blacklist certain rules so in #2399 I added missing exports so now
you can import them like so:
```
import { ExecutableDefinitionsRule } from 'graphql/validation';
```

Also, I backported missing exports to the `14.6.0` release.
If this change broke your code, please try to migrate it as shown above
and please feel free to open an issue if you need assistance with migration.


IvanGoncharov:
For additional context about this long-awaited please see this discussion:
https://github.com/graphql/graphql-js/pull/808#issuecomment-294237716

src/index.d.ts
```diff
@@ -327,12 +327,12 @@ export {
...
   VariablesInAllowedPositionRule,
   // SDL-specific validation rules
-  LoneSchemaDefinitionRule,
-  UniqueOperationTypesRule,
-  UniqueTypeNamesRule,
-  UniqueEnumValueNamesRule,
+  LoneSchemaDefinitionRuleRule,
+  UniqueOperationTypesRuleRule,
+  UniqueTypeNamesRuleRule,
+  UniqueEnumValueNamesRuleRule,
```
trevor-scheer: @IvanGoncharov I'm worried there may have been a mistake via find and replace here and in other places, can you take another look please?

IvanGoncharov: @trevor-scheer Thanks a lot for reporting 👍 
Fixed it here: https://github.com/graphql/graphql-js/pull/2407

Issue #2203, opened 2019-09-26
Test all functions (with AST input) that they handle `undefined` instead of array properites


IvanGoncharov:
Based on @mjmahone comment https://github.com/graphql/graphql-js/pull/2084#discussion_r313964490
> I know we ran into problems in the past when hand-modeling the AST where, unless you explicitly set directives: [], you'd end up getting an undefined reference, even though the flow type allowed it.

Currently `parse` always return empty arrays (and never `undefined`) but AST definitions explicitly support `undefined`. 

This is done to make AST forward-compatible meaning old code that generate/transformate AST and doesn't know about new properties (e.g. `interfaces` on `InterfaceTypeDefinitionNode` node) is still compatible with newer versions of `graphql-js`.

It also means it's very easy to manually create nodes, for example creating `FieldNode` you can safely omit `arguments`, `directives` and `selectionSet` properties so minimal field will look like:
```js
{
  kind: 'Field', 
  name: { kind: 'Name', value: 'foo' },
}
```

At the same time we need to be sure that functions like `buildSchema`, `extendSchema`, `print`, `execute`, etc. are actually ready to handle such functions and don't crash when they get `undefined` instead of an empty array.

yaacovCR:
Closed by #4206

PR #2405, opened 2020-01-27
Make variableDefinitions not nullable

brandon-leapyear:
:sparkles: _**This is an old work account. Please reference @brandonchinn178 for all future communication**_ :sparkles:

---

It seems like `variableDefinitions` will always be a non-null array.

https://github.com/graphql/graphql-js/blob/f07ec241d51d7c3fb0263bb3ade1975dc87babc3/src/language/parser.js#L258-L285
https://github.com/graphql/graphql-js/blob/f07ec241d51d7c3fb0263bb3ade1975dc87babc3/src/language/parser.js#L307

### DIFF
--- a/src/language/ast.d.ts
+++ b/src/language/ast.d.ts
@@ -220,7 +220,7 @@ export interface OperationDefinitionNode {
   readonly loc?: Location;
   readonly operation: OperationTypeNode;
   readonly name?: NameNode;
-  readonly variableDefinitions?: ReadonlyArray<VariableDefinitionNode>;
+  readonly variableDefinitions: ReadonlyArray<VariableDefinitionNode>;
   readonly directives?: ReadonlyArray<DirectiveNode>;
   readonly selectionSet: SelectionSetNode;
 }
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
Hi @brandon-leapyear 

Thanks for PR 👍 
AST format should have a healthy balance between the ability to easily process but at the same to easily construct AST nodes.
For example, at the moment you can easily extract part of the query into separate query:
```js
return {
  kind: Kind.OPERATION_DEFINITION, 
  selectionSet: someSelectionSet,
};
```
And you don't need to add a bunch of empty arrays, moreover, it makes code less fragile when we add new properties to AST nodes.

But I agree it looks weird and creates complications, for example with testing see #2203 
Personally I think a good long term solution would be to always return `undefined` instead of empty arrays since it's will improve performance and memory consumption.
Also, it would be great if someone researches what other JS parsers do e.g. Babel's parser, ESLint parser, etc.

yaacovCR:
closing this PR in favor of https://github.com/graphql/graphql-js/pull/4206 based on suggestion above in https://github.com/graphql/graphql-js/pull/2405#issuecomment-579077133

PR #2413, opened 2020-01-29
validation: Export rules missing in <14.6.0 using legacy names

IvanGoncharov:
It's recommended to update to 14.6.0 and start using correct exports
added in #2400 but if you stuck on the earlier version you could use these
exports as a temporary workaround.
Note: these exports are deprecated and will be removed in v16


IvanGoncharov:
@trevor-scheer Can you please review it?

PR #2419, opened 2020-01-30
Add support for adding description to schema

IvanGoncharov:
Implements https://github.com/graphql/graphql-spec/pull/466

### DIFF
--- a/src/utilities/__tests__/getIntrospectionQuery-test.js
+++ b/src/utilities/__tests__/getIntrospectionQuery-test.js
@@ -18,7 +18,7 @@ describe('getIntrospectionQuery', () => {
     );
   });
 
-  it('include "isRepeatable" field', () => {
+  it('include "isRepeatable" field on directives', () => {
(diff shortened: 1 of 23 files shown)


src/utilities/__tests__/getIntrospectionQuery-test.js
```diff
@@ -18,7 +18,19 @@ describe('getIntrospectionQuery', () => {
     );
   });
 
-  it('include "isRepeatable" field', () => {
+  it('include "isRepeatable" field on directives', () => {
```
leebyron: Is this related to the isRepeatable change? Or just piggybacking in more tests?

IvanGoncharov: @leebyron Thanks for the review 👍 
Just copy-pasted tests for `isRepeatable` as starting point for testing that schema's description is present but forgot to change them.
I wrote a new test and committed them.

PR #2422, opened 2020-02-03
fix(validation): unnecessary constraint forcing root query type

jjangga0214:
Currently, there is a constraint that forces people to provide root query type.

However, IMHO, it's unnecessary.

Mutation-and-subscription-only GraphQL is also possible.

Therefore, I patched it, to allow queryless schema.

More fundamentally speaking, why should you restrict people's intentions?

You cannot predict every use case, as the world is a quite complex place for a limited personal experience and a single philosophy to foresee.

Give people freedom.

JUST LET PEOPLE BUILD WHAT THEY WANT.

Thanks.


Cito:
I hate to be a naysayer, but GraphQL.js has been created as a reference implementation for GraphQL, so it needs to follow the specification [which says](http://spec.graphql.org/draft/#sel-FANNDABAB0DnpX) that the type system *must* contain a query type.

IvanGoncharov:
@Cito Thanks for the support 👍 

@jjangga0214 I totally understand the frustration of being forced to add dummy field to Query type just to satisfy GraphQL specification.
That said as @Cito correctly pointed out we need to follow the spec so a way forward would be to first solve it in the spec.
Moreover, the proposal for this change is discussed here: 
https://github.com/graphql/graphql-spec/pull/631

PR #2444, opened 2020-02-08
TypeInfo-test: use proper test name

Cito:
test name has been duplicated and not adapted


IvanGoncharov:
@Cito Was really concentrated on achieving 100% coverage so I lost quality in the process.
Thanks a lot 👍

Cito:
It's always great to have complete test coverage, thanks for your effort to reach that.

PR #2449, opened 2020-02-08
add full unicode support

andimarek:
This is the complementary implementation for the proposed spec change https://github.com/graphql/graphql-spec/issues/687.

Most notably full unicode is already supported today, albeit without having explicit tests for it. (It is fully supported because the lexer never checks that the actual unicode code point is <=0xFFFF as required by the spec)

The only new code that is added is the verification of the surrogate pairs. The current implementation allows for invalid surrogate pairs.


andimarek:
@IvanGoncharov could you have look at the failing coverage tests and give me hint how to fix them? thanks

acao:
@andimarek I think it wants you to add a test case for this exception:

https://codecov.io/gh/graphql/graphql-js/compare/f662f95931153db2c0d36659f2a4da4bfbf70c94...b437de925a0729f24e995d5ee5a84ae334ae7dd5/diff#D1-569

andimarek:
thanks a lot @acao for the hint  👍 ... now the build is green.

leebyron:
I just updated this to reflect the latest changes in https://github.com/graphql/graphql-spec/pull/849/

  * Requires surrogate pairs, regardless of if they are escaped
  * Support braced unicode escapes
  * Improved error messages with more tests

src/language/__tests__/lexer-test.js
```diff
@@ -262,6 +262,95 @@ describe('Lexer', () => {
...
+      },
+    );
+
+    expect(
+      lexOne('"string with unicode escape outside BMP \\u{1F600}"'),
+    ).to.contain({
+      kind: TokenKind.STRING,
+      start: 0,
+      end: 50,
+      value: 'string with unicode escape outside BMP 😀',
```
IvanGoncharov, line 281: I don't trust Unicode symbols in tests, because you can't visually inspect that codes are matching.
Can we use JS escape sequences instead?

IvanGoncharov: Actually, I wrote a simple ESLint rule to find non-ASCII characters in #3053
And discovered that my editor (NeoVim) use strange encoding for ⏱️ emoji:
image

PR #2451, opened 2020-02-10
fix: use babel-plugin-transform-require-extensions

arlac77:
use _babel-plugin-transform-require-extensions_ to generate fully qualified (with file name extension) imports


IvanGoncharov:
@arlac77 Thanks for PR 👍 
But it's already fixed in #2379 and I working to release it today.

PR #2459, opened 2020-02-17
suggestionList: Speed up 

IvanGoncharov:
It becomes almost 2.5 times faster:
image


IvanGoncharov:
Update: after I cleaned it a bit it became almost 3x faster:
image

acao:
very nice! codecov just wants the other condition tested on line 126 it looks like:
https://codecov.io/gh/graphql/graphql-js/compare/8152fb61d13d5a66e81f74a2fa942a2df77d61a6...d9ce66c40a9360b5ff5266fb8462f0aa19cc56c6/diff#D1-126

acao:
much easier to comprehend now, too

Issue #2104, opened 2019-08-22
Road to Typescript


IvanGoncharov:
This is an issue is to discuss our plan to convert this library to TypeScript.
Note: We will still continue to provide full Flow typings in the same package.

This effort is in line with the ongoing conversion of GraphQL IDE packages (GraphiQL, GraphQL LSP, etc.): https://github.com/graphql/graphiql/tree/convert-to-ts

Note this decision is not related to the technical aspect of Flow vs TS.
GraphQL.js has complete Flow coverage and it is excellent for ongoing maintenance. But our goal is to increase the number of active contributors and by switching codebase to TypeScript we hope to make the codebase feel more familiar and easy to contribute to.

Big thanks to @JacksonKearl from Apollo team for making PoC for such conversion and exploring different conversion strategies.

Before we start discussing the action plan here is a few important points:
* `14.x.x` is stable release so we shouldn't do any potentially breaking things in this release.
* We need to do `15.0.0` anyway since it becomes hard to maintain support for Node6 since ESLint and other projects stop supporting it. Plus it allows us simply remove all deprecated stuff that was scheduled for removal in this release.
* Even after we release `15.0.0` we will still need to support `14.x.x` as a branch and backport all critical or security-related bugfixes until `16.0.0`.

Giving these constraints here is a plan I want to propose:
- [x] 1. Freeze all non-critical merges for the duration of this conversion
- [x] 2. I copied all `*.d.ts` from `@types/graphql` into [tstypes](https://github.com/graphql/graphql-js/tree/master/tstypes) folder and working on [removing it from DefinetlyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package). I included this typings in `14.5.0` release as the baseline for our future work.
- [x] 3. Synchronize `*.d.ts` files with Flow types (some stuff is missing and a lot of types are based on `14.0.0` release) and identify all changes that were done in `@types/graphql` but weren't reflected in Flow. We will release these changes as `14.5.1` release. 
- [x] 4. After we are sure that `*.d.ts` files are complete and don't break existing TS libraries we can branch of `14.x.x` and switch master to `15.0.0-alpha.1`. Remove all stuff marked as deprecated and was scheduled for removal in v15.
[...]

PR #2465, opened 2020-02-28
Change type of extensions from anonymous Record to named interfaces

benjie:
Following on from https://github.com/graphql/graphql-js/issues/2104#issuecomment-592420333 I thought it best to demonstrate with a PR.

I'm concerned about losing type safety of the `extensions` field due to the `any`'s here. I consulted @orta on the best way to type this in TypeScript, and he confirmed that using an empty interface and [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) is the best practice here. He provided the following example:

https://www.typescriptlang.org/play/?ssl=12&ssc=1&pln=16&pc=1#code/JYOwLgpgTgZghgYwgAgKIA9IgM7APY7IDeAsAFDkC+55E6ADnlGMggDZzbbIDiUc9ABYBFADIB5AEYArCAjAAVAJ70UpMsmR0suAtgBcaTBBz4cNDaz1goAV3lMAFAEpiycps1hBwbADptE11CAF5iaksIiPJQSFhEFAwdM251TQBbJQA1OCgAfkNsG1AAcypyIA

He noted that the fields added to the interface by merged declarations should contain `?` to avoid issues at the internal callsite, but I don't think that's an issue for us since all extension attributes would be optional; accessed via something like: `myGraphQLObjectType?.extensions?.myAttribute`.

~~This would be a breaking change to the types (`any` -> `unknown`), so I think it's important to include it before GraphQL.js v15 is released.~~


IvanGoncharov:
@benjie Thanks for PR 👍 
The big problem here is that schemas from different sources can have different `extensions` so we can't pollute global types.
Also, it will make it impossible to generate `extensions` from introspection in the future.

IvanGoncharov:
@benjie Can you please describe your use case in more detail so I can think about alternative solutions?

benjie:
> The big problem here is that schemas from different sources can have different extensions so we can't pollute global types.

That's what declaration merging solves; everyone's type completion would be relevant to their project solely based on the dependencies they pull in; e.g.

In MyType.ts:

```ts
import { GraphQLObjectType } from "graphql";

declare module "graphql" {
  interface GraphQLObjectTypeExtensions {
    /*
     * This attribute declaration gets merged in with all the other
     * attribute declarations from every other module _that you have
     * imported_ that declares this interface.
     */
    myNumber?: number;
  }
}

export default new GraphQLObjectType({
  name: "MyType",
  fields: {...},
  extensions: {
    myNumber: 42,
  },
});
```

In SomeOtherFile.ts:

```ts
import { GraphQLSchema } from "graphql";
import MyType from "./MyType";
import DifferentType from "some-other-module";

console.log(MyType.extensions?.myNumber); // 42
console.log(DifferentType.extensions?.myNumber); // undefined
```

^ The above would not throw typing errors, and would know what `myNumber` was `number | undefined` and would know what attributes are likely to be available in `extensions` of a `GraphQLObjectType`.

The main issue this would open is that once an attribute is defined on extensions, all attributes with that name on that type of extension (e.g. GraphQLObjectTypeExtensions) would have to have the same type. This is probably a good thing; I'd suggest using namespacing to help avoid conflicts from different modules. If you want to introspect over the properties of `extensions` then having the attributes have consistent types is probably important anyway.

> Also, it will make it impossible to generate `extensions` from introspection in the future.

I'd need to understand more what you have in mind here to comment; I assume it relates to https://github.com/graphql/graphql-spec/issues/300 ? I don't currently see any issues here that using `Record<string, any>` would solve - you can just cast the type to `Record<string, any>` internally.

IvanGoncharov:
> I'd need to understand more what you have in mind here to comment;

@benjie I mean that when you generate `extensions` dynamically you can't guarantee that they will match any type other than `Record<string, any>` and it creates false sense of security.
A long term plan is to actually switch to `Record<string, unknown>` so it will match current Flow types:
https://github.com/graphql/graphql-js/blob/3f859d4e5e6088deea97193d5106382a901a9c21/src/type/definition.js#L574

> Can you please describe your use case in more detail so I can think about alternative solutions?

benjie:
> I mean that when you generate extensions dynamically you can't guarantee that they will match any type other than Record<string, any> and it creates false sense of security.

You can still add the loss-of-type-safety necessary for that with declaration merging; e.g.:

https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=25&pc=1#code/JYOwLgpgTgZghgYwgAgKIA9IgM7APY7IDeAsAFDkC+55E6ADnlGMggDZzbbIDiUc9ABYBFADIB5AEYArCAjAAVAJ70UpMsmR0suAtgBcaTBBz4cNDaz1goAV3lMAFAEpiycps1hBwbADptE11CAF5iaksIiPJQSFhEFAwdM251TQBbJQA1OCgAfkNsG1AAcyoLWOh4JCNkvWIPZABtAGsIJULikBKAXUM4ECUAbnKKMgRrZDxkMJAIAHdefiExKVl5ZVUXEbG8AONTPTy-TJyoHb3Aw5xjmDw8HaA

```ts
interface Extensions {

}

export class GraphQLObjectType {
  extensions: Extensions

  constructor() { 
    this.extensions = {}
  }
}

interface Extensions {
  myVar?: string
}

interface Extensions {
  [key: string]: any;
}

const o = new GraphQLObjectType();

o.extensions?.myVar;
o.extensions?.foo;
```

> Can you please describe your use case in more detail so I can think about alternative solutions?

The main motivation here is to avoid loss of type safety - currently as soon as you read an attribute from `thing.extensions` it's an `any`. In PostGraphile we maintain internal maps from GraphQL type instances to values we use (e.g. "look-ahead" data for efficient SQL execution, etc); it would be great to bake these into the schema directly on the fields/arguments/types/etc so that they can be accessed directly from GraphQL resolvers without requiring pulling data from a closure or demanding that context or rootValue has all this information passed in. This'd allow us to make use of `.toConfig()` -> save -> load and make user's lives easier.

benjie:
@IvanGoncharov Would you be open to me changing these to:

```ts
export interface GraphQLObjectTypeExtensions {
  [key: string]: unknown;
}
```

This way you still get basically the same as you have currently (except a named interface rather than anonymous Record), but TypeScript can tab-complete and understand the declared attributes if people opt-into that.

MichalLytek:
>  Can you please describe your use case in more details

The simplest use case would be query complexity:
https://github.com/slicknode/graphql-query-complexity

When we put some metadata to type's `extensions` property that are read in runtime and used to calculate query complexity.

So if you install `graphql-query-complexity`, thanks to the declarations merging it would add type info to `GraphQLObjectTypeExtensions` so it will provide intellisense that `extensions` has `complexity` property and prevent passing there wrong values like string:

```ts
  name: 'Post',
  fields: () => ({
    title: { type: GraphQLString },
    text: {
      type: GraphQLString,
      extensions: {
        complexity: 5
      },
    },
  }),
});
```

Or provide intellisense (type-inference) for the callback calculator parameters:
```ts
  name: 'Query',
  fields: () => ({
    posts: {
      type: new GraphQLList(Post),
      args: {
        count: {
          type: GraphQLInt,
          defaultValue: 10
        }
      },
      extensions: {
        complexity: ({args, childComplexity}) => childComplexity * args.count,
      },
    },
  }),
});
```

benjie:
I've added this change. I think this is now a fairly non-controversion change:

- We've given names to the previously anonymous record types (basically equivalent)
- We're using an interface with index signature rather than a record (basically equivalent)
- We've changed "any" to "unknown" so that types must be checked

benjie:
To make this even more uncontroversial I've converted the `unknown`s back to `any`s. So now we're just changing anonymous record types to named and exported interface types. This code is 100% backwards compatible.

src/type/index.d.ts
```diff
@@ -75,28 +75,38 @@ export {
   Thunk,
   GraphQLArgument,
   GraphQLArgumentConfig,
+  GraphQLArgumentExtensions,
```
IvanGoncharov, line 78: you need to also expose it in `src/index.d.ts`

src/type/definition.d.ts
```diff
@@ -271,6 +271,10 @@ export function getNamedType(type: GraphQLType): GraphQLNamedType;
  */
 export type Thunk<T> = (() => T) | T;
 
+export interface GraphQLScalarTypeExtensions {
```
IvanGoncharov, line 274: For a time being we need to maintain parity between TS and Flow so you should mirror these types into `definition.js`

IvanGoncharov:
@benjie You also need to apply the same changes to `GraphQLSchema` and `GraphQLDirective`. 
Other than that and review comments that left I don't have any objections.

That said I still think extending types globally is a good idea but I don't want to block this PR since I don't object changes themself I just object how they would be used.

IvanGoncharov:
@benjie Just to capture note from our chat:
Descriptions should mentioning that it's highly recommended to create wrapper object for stuff that you put inside extension to prevent clashes:
```
new GraphQLObjectType({
  name: 'Foo',
  // ...
  extensions: {
    apolloFederation: {
      keys: [/* <Federations specific keys> */],
    },
    postgraphile: {
      keys: [/* <Postrgress specific keys> */],
    }
  }
}
```

src/type/definition.d.ts
benjie: A type for Flow is 

```js
  extensions?: ?ReadOnlyObjMapLike<mixed>,
```

where

```js
export type ReadOnlyObjMap<T> = { +[key: string]: T, __proto__: null, ... };
export type ReadOnlyObjMapLike<T> =
  | ReadOnlyObjMap<T>
  | { +[key: string]: T, ... };
```

What's the resulting Flow type you'd like?

Maybe:

```js
export type GraphQLFieldConfigExtensions = ReadOnlyObjMapLike<mixed>;

// ...

  extensions: ?GraphQLFieldConfigExtensions;
```

Or maybe the more verbose:

```js
export type GraphQLFieldConfigExtensions =
  | {
      +[key: string]: mixed,
      __proto__: null,
      ... 
    }
  | {
      +[key: string]: mixed,
      ... 
    };

// ...

  extensions: ?GraphQLFieldConfigExtensions;
```

I assume the former.

benjie: Done as the former.

benjie:
I've extended to GraphQLSchema and GraphQLDirective; exposed through `src/index.d.ts`, mirrored the changes to Flow, and added the following comment above the extensions in the TypeScript typings:

```js
/**
 * It's strongly recommended that you prefix each added attribute's name with an
 * identifier unique to your project to avoid conflicts with other projects.
 */
```

This allows for using a namespaced wrapper object as you suggest, or just having a mess of attributes that shouldn't conflict (which may be more convenient for other projects).

There's a small issue with the Flow typings: previously you were using `ReadOnlyObjMapLike` for the extensions input, and then converting that to `ReadOnlyObjMap` on the constructed types. Since we're now sharing a single named type between these places, it's not clear how or if you want me to make this distinction. Maybe we should just use `ReadOnlyObjMapLike` everywhere? Or just drop the `null` prototype requirement?

airhorns:
Would you folks be open to making the Extensions types generic over the same things the containing type is generic over so that the properties of our strongly typed extensions can be generic over those as well? If you're attaching a function as an extension to a  `GraphQLObjectType`, it seems reasonably to want to pass that function the context some times, so to strongly type the function the extension needs to be generic over the `TContext` that the `GraphQLObjectType` is as well. 

Prior to extensions being a thing at all, one library I've worked with was using TypeScript declaration merging to extend the definition `GraphQLObjectType` to facilitate the same thing extensions does like so: https://github.com/join-monster/join-monster/blob/master/src/index.d.ts . The optional properties it adds to those types can be thunked such that they can be functions that get the context and return a dynamic value, and that's strongly typed by extending the root type and getting a handle on the generics again. If the empty `*Extensions` types aren't generic, is there a way to still get a handle on the generic types?

IvanGoncharov:
> Would you folks be open to making the Extensions types generic over the same things the containing type is generic over so that the properties of our strongly typed extensions can be generic over those as well? 

@airhorns Sounds reasonable the only downside I see is that unused template arguments will look strange so it we should extend their description saying they are open types and it's possible to extend them so we provide template arguments for this reason.

@benjie Can you please add simple test extending at least a couple of them in our TS integration tests:
https://github.com/graphql/graphql-js/blob/master/integrationTests/ts/index.ts
We have `GraphQLObjectType` there so you can extend object type, field, and argument.

IvanGoncharov:
> There's a small issue with the Flow typings: previously you were using ReadOnlyObjMapLike for the extensions input, and then converting that to ReadOnlyObjMap on the constructed types. Since we're now sharing a single named type between these places, it's not clear how or if you want me to make this distinction. Maybe we should just use ReadOnlyObjMapLike everywhere? Or just drop the null prototype requirement?

@benjie My initial concern was that some developers will generate extensions from some 3rd-party sources or even user-provided data in case platforms like CMS. Since the end goal is to allow exposing `extensions` through introspection so `buildClientSchema` can contract schema with `extensions`.
So we should be able to handle stuff like `toString`, `toValue` without constant calls to `hasOwnProperty`, especially since `hasOwnProperty` is also valid GraphQL name :(

Thinking about it just remove those types from Flow since we migrating TS in next release anyway.

IvanGoncharov:
> This allows for using a namespaced wrapper object as you suggest, or just having a mess of attributes that shouldn't conflict (which may be more convenient for other projects).

@benjie Can you please explain user cases that you have in mind where it's more convenient?

Because I actually prefer to standardize on one approach to not confuse library authors and their users.
Also, people like to create extremely short prefixes so for example instead of `postgraphile` wrapper most of the developers will try to use `pg` prefix which has non zero chance to conflict with any other PostgreSQL related code.

IvanGoncharov:
@benjie Looking at @airhorns example it sims that join monster is currently using `jm` prefix that is proving my point of people get tendency for extremely short prefixes. Since the end-goal is to eventually allow to expose `extensions` through introspection (disabled for all fields by default) it would make sense to recommend wrapping extensions in longer names from the beginning.

benjie:
@IvanGoncharov would you like to provide alternative wording for the comment?

integrationTests/ts/index.ts
```diff
@@ -2,17 +2,58 @@ import { GraphQLString, GraphQLSchema, GraphQLObjectType } from 'graphql/type';
...
+    TContext,
+    TArgs = { [argName: string]: any }
+  > {
+    someFieldExtension?: SomeExtension;
+  }
+  interface GraphQLArgumentExtensions {
+    someArgumentExtension?: SomeExtension;
+  }
+}
+*/
```
benjie, line 33: I don't know how best to selectively enable this for TypeScript 3.2 and above only.

Further, the types themselves are not concretely tested, pointing to an example in the GraphQL source where this takes place would be helpful.

benjie:
@IvanGoncharov This is ready for re-review. All comments were addressed, however I was not able to figure out what testing could be added for the types that supports back to TypeScript 2.6. If you uncomment the block I commented, it works on TS>=3.2, but fails on 3.1:


```
Testing on typescript-3.9 ...
Testing on typescript-3.8 ...
Testing on typescript-3.7 ...
Testing on typescript-3.6 ...
Testing on typescript-3.5 ...
Testing on typescript-3.4 ...
Testing on typescript-3.3 ...
Testing on typescript-3.2 ...
Testing on typescript-3.1 ...
../../npmDist/index.d.ts:147:3 - error TS2484: Export declaration conflicts with exported declaration of 'GraphQLArgumentExtensions'.
147   GraphQLArgumentExtensions,
      ~~~~~~~~~~~~~~~~~~~~~~~~~
../../npmDist/index.d.ts:156:3 - error TS2484: Export declaration conflicts with exported declaration of 'GraphQLFieldExtensions'.
156   GraphQLFieldExtensions,
      ~~~~~~~~~~~~~~~~~~~~~~
../../npmDist/index.d.ts:172:3 - error TS2484: Export declaration conflicts with exported declaration of 'GraphQLObjectTypeExtensions'.
172   GraphQLObjectTypeExtensions,
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Since this is user code, this should not be a concern (we just shouldn't add this block of code in TypeScript 3.1 or below). I considered changing `index.ts` to `index.ts.template` and then re-writing it before each test, but that seems a bad approach. I considered moving from running `tsc` to loading the TypeScript library programatically, but that seems expensive (loading so many versions, never unloading any). I considered doing this but in forked node processes, but this started to look complex. Perhaps you could demonstrate how you'd like this tested?

integrationTests/ts/index.ts
```diff
@@ -1,30 +0,0 @@
-import { GraphQLString, GraphQLSchema, GraphQLObjectType } from 'graphql/type';
```
benjie, line 1: This entire file was renamed to `integrationTests.ts.template`

integrationTests/ts/index.ts.template
```diff
@@ -0,0 +1,70 @@
...
+  string: string;
+}
+
+const example: SomeExtension = {
+  number: 42,
+  string: 'Meaning of life',
+};
+
+// The following code block requires a newer version of TypeScript
+/*! >=3.2 !*/
```
benjie, line 16: This "magic comment" syntax is documented in `test.js`

integrationTests/ts/test.js
```diff
@@ -4,16 +4,56 @@
...
+  return { writeFileSync };
+});
+
 for (const version of tsVersions) {
   console.log(`Testing on ${version} ...`);
 
+  for (const template of templates) {
+    template.writeFileSync(version);
+  }
+
```
benjie, line 56: @IvanGoncharov I seek your guidance on this approach to allowing a block of code to be omitted when testing older TypeScript versions. It works; but it does mean the tests write to the filesystem, which makes me uncomfortable. It's a fairly simple implementation though, basically powered by a single regexp and the `semver` module.

IvanGoncharov: @benjie Thanks for trying this solution.
How about just have `index-3.2.ts` and `index.ts` cycle starts with `index.ts` since we start from newest versions first but if `fs.fileExistSync(`index-${version.replace('typescript-', '')}.ts`)` exist we will switch to this file for this and latter versions.

benjie: That was my first thought; but I realised that keeping these files in sync would be challenging, and it increases the amount of code in the tests without offering much in the way of value. It also doesn't allow for more complex ranges like `>=3.2 <5` which may be required in future. "Turning off" a block of code seemed like the simplest solution from the point of view of writing the tests, at a slight cost of the test infrastructure being more complex.

benjie: I suppose the more complex ranges are enabled by having a `4.9` file and another `3.1` file; but this still seems more complex/cumbersome than just removing a block.

IvanGoncharov: @benjie couple of points:
1. don't like having `glob` and `semver` as dependencies it means we need to update them constantly to prevent GitHub marking this repo as affected by yet another security vulnerability.
2. Version after next will drop all TS up to 3.7 (to support recursive types) so this is the temporary workaround to keep maintaining `15.x.x` for some time and possibly backport critical fixes.
3. I don't expect this file to be changed frequently since we will migrate our codebase to TS we will need it just as a sanity check and for very rare occasions where we can't test stuff in our codebase like this one.

So I will likely remove this mechanism in `16.0.0` anyway so no need to make it too complex or future proof just make it to require as low maintenance as possible.

IvanGoncharov: @benjie Let's not block this PR and keep it KISS just add test as a comment with `FIXME` and I will uncomment it in `16.0.0`.
How does it sounds?

benjie: 1. You already have a lot of dependencies on both of these modules according to `package-lock.json`, but I totally get where you're coming from
2. Great :+1: 
3. Okay.

I've reverted the commit that adds this testing support.

airhorns:
Bless both of your hearts for making this happen and sweating the details! Thanks!!

PR #2468, opened 2020-03-02
docs: fixed couple of typos & lang

pranshuchittora:


README.md
```diff
@@ -30,16 +30,16 @@ With yarn:
...
-or alternatively using npm:
+alternatively using npm:
 
 ```sh
 npm install --save graphql
 ```
 
 GraphQL.js provides two important capabilities: building a type schema, and
-serving queries against that type schema.
+serving queries against that type of schema.
```
IvanGoncharov: @pranshuchittora I'm not sure about some of the changes in this PR, especially this one (`type schema` is a single entity here).
So I think this PR needs to be proof-read by a native speaker.

pranshuchittora: Sounds good, looking forward for further review.

rmcteggart-r7:
It looks like some of this has been put through a grammar tool without double checking it? 

I only noticed it after opening up a similar PR https://github.com/graphql/graphql-js/pull/2680

There may be some from this PR that needs added to mine but there are some unusual changes in this one that don't make sense.

IvanGoncharov:
@rmcteggart-r7 Thanks for the review 👍 
We would gladly accept any grammar fixes.

PR #2471, opened 2020-03-02
Feature/add comment in ast

### LINKED ISSUE #2241: Maintain comments when parsing schema
```graphql
type GlobalCinemaVenue implements Node @preLoad {
  # @todo Refactor into a dynamic set of identifiers, i.e.
  """
  https://webediamoviespro.com identifier.
  """
  foreignWebediaId: String
}

```

Currently when using `parse`, the `@todo Refactor into a dynamic set of identifiers, i.e.` is not made available in AST. As a result, this data is lost when implementing utilities for formatting SDL, e.g. https://github.com/gajus/sort-graphql-sdl.

It would be nice if there was a way to parse/ print SDL while maintaining comments and descriptions.

nveenjain:
Fixes #2241 
Added support to add comments in lexer itself.
Modified the tests as the root structure of the response is changed.
Added some tests. Please comment if more test cases are required.


src/language/__tests__/parser-test.js
```diff
@@ -147,6 +147,39 @@ describe('Parser', () => {
     );
   });
 
+  it('Add comments from type in AST', () => {
+    const ast = parse(`
+      #This comment has a \u0A0A multi-byte character.
+      type alpha{ field(arg: string):string }
+    `);
+
+    expect(ast).to.have.nested.property(
```
IvanGoncharov: It's better to just do `expect(toJSONDeep(ast.comments)).to.deep.equal([`

src/language/ast.d.ts
```diff
@@ -204,6 +204,16 @@ export interface DocumentNode {
   readonly kind: 'Document';
   readonly loc?: Location;
   readonly definitions: ReadonlyArray<DefinitionNode>;
+  readonly comments: ReadonlyArray<CommentNode>;
+}
+
+export interface CommentNode {
+  readonly kind: 'Comment';
```
IvanGoncharov: Please use the same `readonly loc?: Location;` as all other nodes

src/language/ast.d.ts
```diff
@@ -204,6 +204,16 @@ export interface DocumentNode {
...
+  readonly start: number;
+  readonly end: number;
+  readonly column: number;
+  readonly line: number;
+  readonly value: string | void;
```
IvanGoncharov: Value can be empty but it should be always present.

src/language/ast.d.ts
```diff
@@ -204,6 +204,16 @@ export interface DocumentNode {
   readonly kind: 'Document';
   readonly loc?: Location;
   readonly definitions: ReadonlyArray<DefinitionNode>;
+  readonly comments: ReadonlyArray<CommentNode>;
```
IvanGoncharov: Please `comments` optional: `comments?: ReadonlyArray<CommentNode>` and remove empty arrays from manually created document nodes in unrelated tests (e.g. `buildASTSchema-test.js`)

src/language/ast.d.ts
```diff
@@ -204,6 +204,16 @@ export interface DocumentNode {
...
+}
+
+export interface CommentNode {
```
IvanGoncharov: Also add `CommentNode` to `ASTNode` union.

src/language/lexer.js
```diff
@@ -28,6 +28,10 @@ export class Lexer {
    */
   token: Token;
 
+  /**
+   * The list of comments token encountered so far
+   */
+  commentsList: Array<CommentNode>;
```
IvanGoncharov: It makes lexer stateful and also doesn't fit the goals of the lexer.
Lexer should produce tokens and parser should produce AST nodes.

nveenjain: Okay, should I then iterate over the tokens produced by lexer which will include the comment tokens too?
It would add additional O(N) complexity then...

IvanGoncharov: No just move comment skipping functionality into the parser.

src/language/parser.js
```diff
@@ -206,6 +206,11 @@ class Parser {
       ),
       loc: this.loc(start),
     };
+    const comments = this._lexer.commentsList;
+    if (comments.length >= 1) {
```
IvanGoncharov: This check is not needed just return an empty array.
Maybe it's not ideal but this is how all other arrays inside AST are handled.

src/language/visitor.js
```diff
@@ -58,6 +58,7 @@ export const QueryDocumentKeys: VisitorKeyMap<ASTKindToNode> = {
     'directives',
     'selectionSet',
   ],
+  Comment: [],
```
IvanGoncharov: Please add visitor tests

src/language/ast.d.ts
```diff
@@ -139,7 +139,8 @@ export type ASTNode =
   | InterfaceTypeExtensionNode
   | UnionTypeExtensionNode
   | EnumTypeExtensionNode
-  | InputObjectTypeExtensionNode;
+  | InputObjectTypeExtensionNode
+  | CommentNode;
```
IvanGoncharov: Please move it after `DocumentNode` type in all lists

src/language/parser.js
nveenjain: Yes, but then we would have to modify those tests having where comments is an empty array, right? such as `buildASTSchema-test.js`

IvanGoncharov: @nveenjain No, please look at other optional arrays in AST.
We return empty arrays in parser but allow users to create nodes where such arrays can be omitted.

src/language/visitor.js
IvanGoncharov: Also, please add `comments` into Document's array.

nveenjain: Sure!

nveenjain:
@IvanGoncharov while migrating the logic of comment skipping to the parser, I noticed that we are using `this._lexer.token` to get the location for many nodes... This might cause an issue when called after ```this._lexer.advance()```/```this._lexer.lookahead()```(through parser) it would mean that we would need to change the logic to skip comments at all the places where `this._lexer.advance()`/```this._lexer.lookahead()``` is called in parser, since next token might be a comment...  (which means lots of changes, and lots of tests to satisfy coverage check)

Is it desired, or should I follow some other approach towards moving skipping functionality to parser like creating lexer.lookahead in the parser itself and renaming `lexer.lookahead` function to `lexer._lookahed` or something else?

Reference:-
I added skipping functionality to `peek`, `expectToken`, `expectKeyword`, `expectOptionalKeyword` and `expectOptionalToken` but 
```parses type with description multi-line string``` failed because we are using `this._lexer.lookahead()` and checking whether ```keywordToken.kind === TokenKind.NAME``` where keywordToken is lookahead, it fails because keywordToken is comment token...

File: src/language/__tests__/visitor-test.js [hunk 10]
```diff
@@ -115,6 +116,69 @@ describe('Visitor', () => {
     });
   });
 
+  it('allows visiting single comments', () => {
+    const ast = parse(
+      dedent`
+      #This is a comment
+    { a }
+    `,
```
IvanGoncharov: Please fix indenting

src/language/__tests__/visitor-test.js
```diff
@@ -115,6 +116,69 @@ describe('Visitor', () => {
...
+      },
+    });
+
+    expect(visited).to.deep.equal([
+      ['enter', 'Comment'],
+      ['leave', 'Comment'],
+    ]);
+  });
+
+  it('allows visiting multiple comments including comments in definitions', () => {
```
IvanGoncharov: I think one test with multiple comments would be enough

src/language/__tests__/visitor-test.js
```diff
@@ -115,6 +116,69 @@ describe('Visitor', () => {
...
+          visited.push(['leave', node.kind]);
+        },
+      },
+    });
+
+    expect(visited).to.deep.equal([
+      ['enter', 'Comment'],
+      ['leave', 'Comment'],
+      ['enter', 'Comment'],
+      ['leave', 'Comment'],
```
IvanGoncharov: Please add comment values here.

src/utilities/stripIgnoredCharacters.js
```diff
@@ -76,6 +76,9 @@ export function stripIgnoredCharacters(source: string | Source): string {
 
   let wasLastAddedTokenNonPunctuator = false;
   while (lexer.advance().kind !== TokenKind.EOF) {
+    if (lexer.token.kind === TokenKind.COMMENT) {
+      continue;
+    }
```
IvanGoncharov: nit: empty line will improve readability

src/language/parser.js
```diff
@@ -1404,6 +1417,22 @@ class Parser {
     }
   }
 
+  // Add the comment token to Parser
```
IvanGoncharov: Developers are capable of reading function names no need to duplicate it inside comments :)

src/language/parser.js
```diff
@@ -1404,6 +1417,22 @@ class Parser {
...
+  addCommentToParser() {
```
IvanGoncharov: Please inline it inside `advance` and also explain why you have nested `while (token.kind === TokenKind.COMMENT) {`

IvanGoncharov:
> Is it desired, or should I follow some other approach towards moving skipping functionality to parser like creating lexer.lookahead in the parser itself and renaming `lexer.lookahead` function to `lexer._lookahed` or something else?

@nveenjain If comments now included in AST as first-class nodes it means comments should be handled by the parser. So adding `lookahead` into parser is an appropriate solution for this problem.

But please keep comments inside the double-linked list.

P.S. Based on our Slack conversation please don't forget to investigate how comments are handled inside Babel, TS, etc. parsers. I'm especially interested in having some mechanism to allow comments to survive at least some AST modification.

nveenjain:
>  Based on our Slack conversation please don't forget to investigate how comments are handled inside Babel, TS, etc. parsers

@IvanGoncharov, For acorn, comments are not parsed, i.e. skipped.
For babel, comments are parsed, added to top-level as well as to internal nodes ( each node has leadingComments, and trailingComments depending on how comments are present ).

> But please keep comments inside the double-linked list.

Sure will add the comments to the linked list.

Also, Do we need to add Support for leading and trailing comments too?

src/language/parser.js
nveenjain: Sorry! Nested `while` were by mistake, will remove it!

nveenjain:
But if we keep nodes in linked list, there might be some edge cases:-
Like for finding location, we rely on lastNode, consider the test case present in repo:-
```graphql
      """
      Description
      """
      # Even with comments between them
      type Hello {
        world: String
      }
```
Here the description string literal will has endPosition=19 according to current implementation now, but if we add comments to linked list endPosition would become 53 (since we check the last token and it will include the comment token)...
What should I do in such case?

src/utilities/stripIgnoredCharacters.js
nveenjain: I tried to do it, but prettier won't allow me, all the files are automatically linted by VSCode... When i disabled auto format on save ```yarn test``` came up with a `prettier error`

IvanGoncharov: @nveenjain Strange prettier should always keep single empty line:
https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AE8AzjABQCU+wAOlPvgO4AWOANnPqewB5wBOAOgCGAEwBuQrHAoCA1jigj8AQgC8q-ABUIshAGkFIgQFEA8gDFKNOvXw5UnHvwEwdCOYfzqN23VAOKAgDCpgCyocYAcppWtLa2kLAK6HAA3HG2AL4Z9Dn4AEasQoVC6TbZUBUgADQgEAAOuNBEyKBCfHwQDAAK7QgtKEJiEDgiNSD5fEJgujAAyvXTCgDmyDB8KbVMMAC2rADqLMSLYHBz-Ti4YpcAnsjgRC21CkT8MN1TyztCyKhCrK9agArIjcABCUxmcHmQh2cAAMgo4L9-oCQCDuHMVuwAIroCDwFEAuC1RZ8V58e75YpwVjjep8BQwfajGBMZAADgADKTOq99lN6vcGXAKWJkbUAI74+AfBoDEBCIgAWigcDgIg14z4cGlOB1HyEXx+SD+xNqrx2ODWGxJICI2LgeIJyNNqLtMGKLJEbOQACZaushGwVkEIDtvvcoNAJSB0K9NMUBma0WIUgBJRQIeZgRmNACCijmMBu7CJr0ymSAA

nveenjain: Oh sorry I thought you meant:
```js
if (lexer.token.kind === TokenKind.COMMENT) {

      continue;
}
```

Vincz:
Ping @nveenjain, @IvanGoncharov 
Hey guys! Any update about this feature? Having the comments in the AST would be awesome to parse/dump. 
It seems this PR was almost finish, no?

adammiller:
@nveenjain @IvanGoncharov I'd also love this feature - are you open to someone continuing this work? 

I'm working on some codemod tooling to handle schema deprecations in a fleet of UI apps and a lot of the apps leverage comments in their `.graphql` files. This would really help with helping preserve the original document structure after transforming the AST.

PR #2472, opened 2020-03-04
Improve error reporting for union member types

Cito:
First make sure that the included type is an object type, because that would be the more severe problem. Only then we can be sure that the type has a name and check for duplicates. Also, create only one error per different non-object type.


src/type/validate.js
```diff
@@ -438,7 +438,20 @@ function validateUnionMembers(
   }
 
   const includedTypeNames = Object.create(null);
+  const includedInvalidTypes = Object.create(null);
```
IvanGoncharov: I think it strange behavior where after fixing one error you would get the same error in the next location.
We should definitely prevent implicit cascading errors but I think we should never mask explicit errors made by the user.
Also if we add this error reduction trick we need to add it in many other places (e.g. if you implement non-interface types) for consistency.

So I think moving check to beginning of the loop and adding `continue` would be enough.

Cito: Note that this does not hide any errors - there is only one message, but all locations are reported and the same message applies to them all.

src/type/validate.ts
```diff
@@ -480,7 +480,20 @@ function validateUnionMembers(
...
+      const typeString = inspect(memberType);
+      if (!includedInvalidTypes.has(typeString)) {
+        context.reportError(
+          `Union type ${union.name} can only include Object types, ` +
+            `it cannot include ${inspect(memberType)}.`,
+          getUnionMemberTypeNodes(union, String(memberType)),
+        );
+        includedInvalidTypes.add(typeString);
+      }
+      continue;
```
yaacovCR, line 495: this reminds me of https://github.com/graphql/graphql-js/pull/4181 => what do you think about changing the error message string to something that references both errors, at least if there are repetitions?

'Something like Union type can only include unique Object types, ....'

What do you think?

yaacovCR (commented):
Seems like a good idea, left some comments/questions!

PR #2475, opened 2020-03-04
schema: fix error message when provided type has no name

Cito:
Small fix to show an error message when one provided type has no name and to avoid error messages like "multiple types named undefined" when multiple provided types have no name.

### DIFF
--- a/src/type/schema.js
+++ b/src/type/schema.js
@@ -213,6 +213,11 @@ export class GraphQLSchema {
       }
 
       const typeName = namedType.name;
+      if (!typeName) {
+        throw new Error(
+          'One of the provided types for building the Schema is missing a name.',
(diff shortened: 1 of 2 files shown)


src/type/schema.js
```diff
@@ -213,6 +213,11 @@ export class GraphQLSchema {
       }
 
       const typeName = namedType.name;
+      if (!typeName) {
+        throw new Error(
+          'One of the provided types for building the Schema is missing a name.',
+        );
+      }
```
IvanGoncharov: Since according to TS & Flow typings `namedType` always have `name` it should be `devCheck`.

Cito: Ok, changed that.

PR #2482, opened 2020-03-09
chore(jsutils): improved didYouMean to handle more edge cases

pranshuchittora:
### Description
Better edge cases handling for the `jsutils/didYouMean`. 

- Handling duplication
- Handling broken values like (`undefined`, `null`, `NaN`)

### Current behaviour
As of now, the module is not able to handle any duplicated values in the `Array`.
Similarly, if it encounters `undefined`, `null` etc then it converts it to the string "undefined", "null".

### Approach
I tried creating a new `Set` from the given `Array`, then finally filtering it in order to eliminate the broken or garbage values.

- [x] Tests Added

### DIFF
--- a/src/jsutils/__tests__/didYouMean-test.js
+++ b/src/jsutils/__tests__/didYouMean-test.js
@@ -35,4 +35,27 @@ describe('didYouMean', () => {
       ' Did you mean the letter "A"?',
     );
   });
+  // Test for handling Duplications
+  it('Handle duplicates, resulting in single unique element', () => {
+    expect(didYouMean(['A', 'A'])).to.equal(' Did you mean "A"?');
+  });
+  it('Handle duplicates, resulting in two elements', () => {
+    expect(didYouMean(['A', 'B', 'A', 'B'])).to.equal(
+      ' Did you mean "A" or "B"?',
+    );
+  });
+  it('Handle duplicates, resulting in more thentwo elements', () => {
+    expect(didYouMean(['A', 'B', 'C', 'C', 'B', 'A'])).to.equal(
+      ' Did you mean "A", "B", or "C"?',
+    );
+  });
+  // Test for filtering null, undefined, NaN...
+  it('should filter the broken values', () => {
+    expect(didYouMean(['A', 'B', 'C', 'C', undefined, null])).to.equal(
+      ' Did you mean "A", "B", or "C"?',
+    );
+  });
+  it('Does accept an empty list', () => {
+    expect(didYouMean([null, undefined, NaN])).to.equal('');
+  });
 });
--- a/src/jsutils/didYouMean.js
+++ b/src/jsutils/didYouMean.js
@@ -19,12 +19,16 @@ export default function didYouMean(firstArg, secondArg) {
       ? [firstArg, secondArg]
       : [undefined, firstArg];
 
+  const suggestionsArgFromSet = Array.from(new Set(suggestionsArg)).filter(
+    x => x,
+  );
+  console.log({ suggestionsArgFromSet });
   let message = ' Did you mean ';
   if (subMessage) {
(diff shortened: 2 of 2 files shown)


src/jsutils/__tests__/didYouMean-test.js
```diff
@@ -35,4 +35,27 @@ describe('didYouMean', () => {
...
+      ' Did you mean "A", "B", or "C"?',
+    );
+  });
+  // Test for filtering null, undefined, NaN...
+  it('should filter the broken values', () => {
+    expect(didYouMean(['A', 'B', 'C', 'C', undefined, null])).to.equal(
+      ' Did you mean "A", "B", or "C"?',
+    );
+  });
+  it('Does accept an empty list', () => {
```
nveenjain: @pranshuchittora I don't think ```didYouMean``` function is defined for Numbers, hence NaN can't be passed... It only accepts string array, and all strings are required, and hence null/undefined can't be passed...

src/jsutils/didYouMean.js
```diff
@@ -19,12 +19,16 @@ export default function didYouMean(firstArg, secondArg) {
       ? [firstArg, secondArg]
       : [undefined, firstArg];
 
+  const suggestionsArgFromSet = Array.from(new Set(suggestionsArg)).filter(
+    x => x,
+  );
+  console.log({ suggestionsArgFromSet });
```
IvanGoncharov: It's an internal function that has Flow typings to accept only strings so this check is done by Flow linter.
Also, please review your PRs before submission and try to remove stuff like `console.log`.

src/jsutils/__tests__/didYouMean-test.js
```diff
@@ -35,4 +35,27 @@ describe('didYouMean', () => {
       ' Did you mean the letter "A"?',
     );
   });
+  // Test for handling Duplications
+  it('Handle duplicates, resulting in single unique element', () => {
+    expect(didYouMean(['A', 'A'])).to.equal(' Did you mean "A"?');
+  });
+  it('Handle duplicates, resulting in two elements', () => {
```
IvanGoncharov: For future PRs: Please spend time and try to understand the code style of the file that you changing.
For example, other tests doesn't have comments before them but have an empty line between tests.

IvanGoncharov:
@pranshuchittora Also please run `npm test` before submitting PRs.

src/jsutils/__tests__/didYouMean-test.js
```diff
@@ -35,4 +35,27 @@ describe('didYouMean', () => {
...
+    expect(didYouMean(['A', 'B', 'A', 'B'])).to.equal(
+      ' Did you mean "A" or "B"?',
+    );
+  });
+  it('Handle duplicates, resulting in more thentwo elements', () => {
+    expect(didYouMean(['A', 'B', 'C', 'C', 'B', 'A'])).to.equal(
+      ' Did you mean "A", "B", or "C"?',
+    );
+  });
```
IvanGoncharov: This single test actually make sense so please extract it into separate PR

Cito: there is a typo: "more thentwo" -> "more than two"

IvanGoncharov: On second thought: It's the responsibility of `suggestionList` to remove duplicates so `didYouMean` should show duplicates as is.

src/jsutils/__tests__/didYouMean-test.js
IvanGoncharov: @nveenjain Thanks for review 👍

src/jsutils/didYouMean.js
pranshuchittora: Pardon

PR #2485, opened 2020-03-11
Fix #894, add perEventContextResolver for subscribe

### LINKED ISSUE #894: Context lifecycle in subscriptions
After reading through the codebase, my understanding of the subscription implementation is as follow :
* initial call `subscribe()` when a subscription operation hits the server, 
* `subscribe()` returns an `AsyncIterator`
* This `AsyncIterator` contains a stream of results of subsequent calls to `execute()`

Again, if my understanding is correct, `contextValue` is passed to `subscribe` on the initial subscription operation and then passed down to `execute` whenever an event triggers a reexecution.

In short, `contextValue` is set once a subscription operation happens and will remain the same until the client unsubscribes.

A common graphql-js implementation is to use DataLoader as a batching/caching library. For queries and mutations, a set of fresh dataloader instances is usually bound to `contextValue` on a per request basis. When the query/mutation is done, those DataLoader instances are left for garbage collection.

Now, subscription context has a much longer life span, making DataLoader caching behaviour unwarranted in most cases.

Ideally, I guess providing a way to set `contextValue` per `execute()` call would allow a behaviour that matches what one could except when setting `contextValue` on a per request basis for queries and mutations.

Is there any way to do so in current implementation ? Is this a behaviour that could be implemented in graphql-js or should I settle for one context per subscription operation and work my way around that ?

tgriesser:
Fixes #894 

Adds ~`contextValueExecution`~ `perEventContextResolver`, a new option for `subscribe` which allows you to derive a new context value for each independent execution of a subscription. 

It takes the original `context` value as an argument, so you can derive the new context from the original.


IvanGoncharov:
@tgriesser I like this solution and I think it is generic enough to be added into lib.
However, I think `contextValueExecution` is a confusing name but I can't suggest anything better 🤷‍♂ 
It's the only blocking issue I have.

Also please update JSDoc comments for relevant functions.

tgriesser:
Nice! The idea with the name was to take `contextValue`, which is the name for the original context, combined with the `execute`, called in `mapSourceToResponse`.

Maybe one of:

- `executeContext`
- `executeContextValue`
- `contextValueExecute`
- `contextValueForExecute`

> Also please update JSDoc comments for relevant functions.

Sounds good, will do once we land on the name.

IvanGoncharov:
> combined with the execute, called in mapSourceToResponse.

@tgriesser The fact that we call `execute` under the hood is an implementation detail.
Also since it's a function it should be reflected in the name.

Spec describes "stream of events" so I think it should be something like `perEventContextResolver`.
http://spec.graphql.org/June2018/#sec-Subscription

tgriesser:
> The fact that we call execute under the hood is an implementation detail.
Also since it's a function it should be reflected in the name.

Ah right, that makes sense.

> Spec describes "stream of events" so I think it should be something like `perEventContextResolver`.

Sounds great.

tgriesser:
@IvanGoncharov docs should be updated, let me know if you see anything that needs correction.

src/subscription/subscribe.d.ts
```diff
@@ -36,6 +37,9 @@ export interface SubscriptionArgs {
  * If the operation succeeded, the promise resolves to an AsyncIterator, which
  * yields a stream of ExecutionResults representing the response stream.
  *
+ * If a `perEventContextResolver` argument is provided, it will be invoked for
+ * each event and return a new context value specific to that event's execution.
```
IvanGoncharov: Please copy it inside `subscribe.js`

IvanGoncharov: Forgot to update PR 🤦‍♂

src/subscription/subscribe.js
```diff
@@ -143,16 +151,23 @@ function subscribeImpl(
...
-  const mapSourceToResponse = payload =>
-    execute({
+  // If `perEventContextResolver` is provided, it is invoked with the original
+  // `contextValue` to return a new context unique to this `execute`.
+  const mapSourceToResponse = payload => {
+    const executeContextValue =
+      typeof perEventContextResolver === 'function'
+        ? perEventContextResolver(contextValue)
+        : contextValue;
+    return execute({
```
IvanGoncharov: It's better to do a single check per stream:
```js
const perEventContextResolverFn = perEventContextResolver ?? (ctx) => ctx
```

tgriesser: Updated to a single check per-stream, kept with the `typeof ` check rather than the `??` just because it's more explicit.

IvanGoncharov:
@tgriesser Thanks for the update, now it looks perfect 👍 
But I don't want to increase scope of `15.0.0` so will merge after release.

andreialecu:
Just ran into the issue at #2289 

Seems that the milestone was missed. Anything holding this back?

97amarnathk:
I have a similar use case while implementing graphql subscriptions with dataloaders. Is there any reason why this isn't merged yet?

benjie:
Is there a way to "release" the context? Say, for example, that the context were to contain an authenticated client connected to a datastore.

tgriesser:
@benjie that's a really good point... the current change wouldn't support that case well.

In thinking about it more, and looking through the `subscribe` / `createSourceEventStream` code, I feel a better option is to add a `perEventExecute`, so you can tap into / modify anything you'd need about the execution. That way you can do something like:

```ts
const iterator = await subscribe({
  schema,
  document,
  variableValues,
  contextValue: new DataContext(),
  operationName: obj.request.name,
  perEventExecute: async (execArgs) => {
    const { contextValue, release } = getContext(...);
    try {
      return await execute({...execArgs, contextValue})
    } finally {
      release();
    }
  }
})
```

@IvanGoncharov thoughts on that? Something like:

```diff
export type SubscriptionArgs = {|
  schema: GraphQLSchema,
  document: DocumentNode,
  rootValue?: mixed,
  contextValue?: mixed,
  variableValues?: ?{ +[variable: string]: mixed, ... },
  operationName?: ?string,
  fieldResolver?: ?GraphQLFieldResolver<any, any>,
  subscribeFieldResolver?: ?GraphQLFieldResolver<any, any>,
+ perEventExecute?: (args: ExecutionArgs) => PromiseOrValue<ExecutionResult>
|};
```

```diff
function subscribeImpl(
  args: SubscriptionArgs,
): Promise<AsyncGenerator<ExecutionResult, void, void> | ExecutionResult> {
  const {
    schema,
    document,
    rootValue,
    contextValue,
    variableValues,
    operationName,
    fieldResolver,
    subscribeFieldResolver,
+   perEventExecute = execute
  } = args;
```

```diff
  const mapSourceToResponse = (payload) =>
-   execute({
+   perEventExecute({
      schema,
```

97amarnathk:
Are there any plans to merge this  ? If not, then what is the best way to generate a context for each execution ?

97amarnathk:
PR which solves this in another way:
https://github.com/graphql/graphql-js/pull/3071

slumericanBx (approved):
send it

yaacovCR:
Closed by #4211

PR #2487, opened 2020-03-11
Make it work with Node ESModules

kamilkisiela:
A missing piece to run `graphql` with Node's experimental ecmascript modules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,10 @@
   "module": "index.mjs",
   "types": "index.d.ts",
   "sideEffects": false,
+  "exports": {
+    "require": "index.js",
+    "default": "index.mjs"
+  },
   "homepage": "https://github.com/graphql/graphql-js",
   "bugs": {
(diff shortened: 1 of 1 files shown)


Urigo:
@IvanGoncharov Is there something we can do to help progress this one?

IvanGoncharov:
@kamilkisiela We need to support imports like `graphql/language` (officially supported) or `graphql/language/parser` (non-supported but used by a lot in the wild).
Also can you please post a link to a document that describes `exports`?

package.json
```diff
@@ -8,6 +8,10 @@
   "module": "index.mjs",
   "types": "index.d.ts",
   "sideEffects": false,
+  "exports": {
+    "require": "index.js",
+    "default": "index.mjs"
+  },
```
ooflorent, line 14: This is a breaking change since it does not allow:

```js
require("graphql/package.json");
require("graphql/language");
// …
```

Issue #2510, opened 2020-04-09
There is no way to provider custom logic of field extracting for GraphQLObjectType


alex-knyazev:
At business level of applications people often work with instances of classes, not just objects.

Common pattern is to put data in separate field of class, for example: 

```
type DataT = {
   name: string
}

class Cat {
  data: DataT

  methods: {
      ....
  }
}
```

Such data can contain other classes instances, and it is very comfortable to work with business entities in application in this way.

But if then you want to return info to API level (to graphql in our case), every resolver should extract data property from classes instances.  It couldn't be problem if types have resolver functions, but they haven't.

Also it could be done via directives, but directives isn't supported in GraphQLObjectType.

So my proposal is to add additional property to GraphQLObjectType constructor, like: "fieldsExtractor", "getFieldsPath", or something like this, which will be called in completeObjectValue if provided. 

Example: 
```
  const Cat = new GraphQLObjectType({
    name: "Cat",
    fields: {
      name: {
        type: GraphQLString,
      },
    },
    fieldsExtractor: (source) {
      return source.data
    }
  });
```

What is your opinion on my proposal? 
I can do it and make pull request if you don't mind my idea.

IvanGoncharov:
@alex-knyazev Sorry for the late reply 😞 

We trying to keep this lib as simple as possible.
So you can solve your use case by customizing defaultFieldResolver:
```ts
import { defaultFieldResolver } from 'graphql'

execute({
  // ...
  fieldResolver(source, ...rest) {
    const result = defaultFieldResolver(source, ...rest);
    if (result === undefined && typeof source === 'object' && source != null) {
      return defaultFieldResolver(source.data, ...rest);
    }
    return result;
  }
})
```

PR #2514, opened 2020-04-19
add opportunity to set path to object fields value

alex-knyazev:
Related issue: https://github.com/graphql/graphql-js/issues/2510

I've added simple string config fields which optionally can be passed to graphql object type.

It this field is provided - in process of object value competing  fields values will be taken not from provided to graphql value, but from internal object field.

P.S.
I wanted to add test for this behaviour, but I didn' found corresponding section in tests (If you want tests - give me advise, please)


IvanGoncharov:
Answered here: https://github.com/graphql/graphql-js/issues/2510#issuecomment-897449404

PR #2517, opened 2020-04-23
Test on Node 14 and don't deploy on forks

MichaelDeBoey:


.travis.yml
```diff
@@ -5,10 +5,10 @@ cache: yarn
...
 # https://github.com/nodejs/Release
 node_js:
-  - '14'
-  - '13'
-  - '12'
-  - '10'
+  - 10
+  - 12
+  - 14
+  - node
```
IvanGoncharov, line 11: AFAIK Travis will run duplicate versions so we will have duplicating runs of v14.

.travis.yml
```diff
@@ -21,10 +21,10 @@ script: |
 jobs:
   include:
     - stage: deploy
-      if: type = push AND branch = master
+      if: type = push AND branch = master AND fork = false
       script: npm run gitpublish
       skip_cleanup: true
-      node_js: '12'
+      node_js: 14
```
IvanGoncharov, line 27: We use LTS for running scripts so it's too early to switch to v14

.travis.yml
```diff
@@ -21,10 +21,10 @@ script: |
 jobs:
   include:
     - stage: deploy
-      if: type = push AND branch = master
+      if: type = push AND branch = master AND fork = false
```
IvanGoncharov, line 24: Do you maintain your own fork of `graphql-js` and trying to use Travis for it?
If not can you please describe your use case?

.travis.yml
MichaelDeBoey: At this moment it's indeed duplicate, but whenever v15 comes out, it's automatically tested too

.travis.yml
MichaelDeBoey: When I have a fork of the repo AND I have Travis myself, this script will be run too.
This will fail my own pipeline because of the failure of deploy.

Doing this PR, the pipeline won't fail anymore

.travis.yml
IvanGoncharov: Make sense however I do all commits only through separate PRs so CI time is critical for this workflow.

.travis.yml
IvanGoncharov: Can you please extract it into separate PR?

MichaelDeBoey: https://github.com/graphql/graphql-js/pull/2530

IvanGoncharov:
@MichaelDeBoey Node14 added in #2529 and fork change in #2530 
If you want to delete extra single quotes please open a new PR.

PR #2535, opened 2020-05-09
Make 'toJSON' definition explicit

IvanGoncharov:


lifeiscontent:
@IvanGoncharov this commit broke some of the tooling I use. I'm expecting `source` when calling `toJSON` to be available on the Location type but this commit removed it.

PR #2541, opened 2020-05-12
docs:Added GraphQL link to title in README.md

chaytanyasinha:
Added GraphQL organisation link to the title so that if new users even click on the title they can reach to organisation website.


chaytanyasinha:
Okay @iamdarshshah

IvanGoncharov:
@chaytanyasinha Thanks for PR but we already have both links at the top of README:
image

Issue #2522, opened 2020-04-23
Unsupported URL Type "link:": link:./resources/eslint-rules


AaronNGray:
Getting the following on :-
```
git clone https://github.com/graphql/graphql-js.git
cd graphql-js
npm install
```
on both Windows and Mac OSX with latest NPM.

```
C:\Users\aaron\GitHub\graphql-js>npm install
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "link:": link:./resources/eslint-rules

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\aaron\AppData\Roaming\npm-cache\_logs\2020-04-23T17_49_59_928Z-debug.log
```
C:\Users\aaron\AppData\Roaming\npm-cache\_logs\2020-04-23T17_49_59_928Z-debug.log
```
0 info it worked if it ends with ok
1 verbose cli [
1 verbose cli   'C:\\Program Files\\nodejs\\node.exe',
1 verbose cli   'C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js',
1 verbose cli   'install'
1 verbose cli ]
2 info using npm@6.14.4
3 info using node@v14.0.0
4 verbose npm-session 0c652f051e4fd10e
5 silly install runPreinstallTopLevelLifecycles
6 silly preinstall graphql@15.0.0
7 info lifecycle graphql@15.0.0~preinstall: graphql@15.0.0
8 silly install loadCurrentTree
9 silly install readLocalPackageData
10 timing stage:loadCurrentTree Completed in 10ms
11 silly install loadIdealTree
12 silly install cloneCurrentTreeToIdealTree
13 timing stage:loadIdealTree:cloneCurrentTree Completed in 0ms
14 silly install loadShrinkwrap
15 timing stage:loadIdealTree:loadShrinkwrap Completed in 1ms
16 silly install loadAllDepsIntoIdealTree
17 timing stage:rollbackFailedOptional Completed in 1ms
18 timing stage:runTopLevelLifecycles Completed in 39ms
19 silly saveTree graphql@15.0.0
20 verbose stack Error: Unsupported URL Type "link:": link:./resources/eslint-rules
20 verbose stack     at unsupportedURLType (C:\Program Files\nodejs\node_modules\npm\node_modules\npm-package-arg\npa.js:200:15)
[...]

PR #2552, opened 2020-05-14
Switch to 'package-lock.json'

IvanGoncharov:


jgcmarins:
@IvanGoncharov any special reason for this decision?

IvanGoncharov:
@jgcmarins It will help to prevent issues like #2522 and make this repo friendlier for contributors who never used Yarn.
Also, my personal philosophy says: "The fewer dependencies the better" and since we can't drop `npm` it means we should drop `yarn`.
Finally, since `npm` was bought by GitHub they intensify their development cycle and published a very promising roadmap.

jgcmarins:
Thanks

Issue #2512, opened 2020-04-16
`printSchema` produces incorrect AST for descriptions ending in \ (backslash)


pcarrier:
```typescript
import { buildASTSchema, parse, printSchema } from 'graphql';
console.log(printSchema(buildASTSchema(parse(`"Example\\\\" scalar Foo`))));
```
=>
```graphql
"""Example\"""
scalar Foo
```

Which `parse` does not accept.

IvanGoncharov:
@pcarrier Thanks for reporting 👍 
Definitely something we need to address.

hereisnaman:
@IvanGoncharov what should be the correct result?

IvanGoncharov:
@hereisnaman It should be reproducible even with:
```
const sdl = `
"""Example\\"""
scalar Foo
`;
expect(print(parse(sdl))).to.equal(sdl);
```
So it's not an issue with `printSchema` it's an issue with how we are printing block strings in general.

pcarrier:
Oh yeah sorry I didn't mean to imply this was specific to schemas, absolutely a general printer issue, it's just where we reproduced it using our wealth of customer data and that's the easiest way to make GraphQL make those strings I know.

IvanGoncharov:
@pcarrier No worries.
I'm mentoring @hereisnaman as part of "Google Summer of Code" and this issue is an excellent opportunity for him to learn how parser/printer works.
So my comments intended to guide him in the right direction so he could figure how to fix this issue.

PR #2560, opened 2020-05-17
BlockString: print multi line for trailing backslash

hereisnaman:
Fixes: #2512

`\"""` is not allowed inside a `BlockString`, so handled it by putting a `\n` between the backslash and triple quotes. The trailing '\n' are already handled.

Thanks for the help @IvanGoncharov .


hereisnaman:
> @hereisnaman Looks great 👍
> The next step is to write tests.

Added a test.

src/language/__tests__/blockString-test.js
```diff
@@ -174,4 +174,15 @@ describe('printBlockString', () => {
...
+  it('correctly prints single-line with trailing backslash', () => {
+    const str = 'backslash \\';
+
+    expect(printBlockString(str)).to.equal(
+      joinLines('"""', 'backslash \\', '"""'),
+    );
+    expect(printBlockString(str, '', true)).to.equal(
+      joinLines('"""', 'backslash \\', '"""'),
+    );
+  });
```
IvanGoncharov, line 187: It makes more sense to move it one test up since has the same logic.
Also `joinLines` make sense only in tests related to "indent".

hereisnaman: done.

IvanGoncharov:
@hereisnaman Merged 🎉

Issue #2524, opened 2020-04-27
Add max int to exports


LeoVS09:
# Feature request

GraphQL has a limit to use 32-bit max and min Integer, but js, not built-in 32-bit max Integer.
I prefer not to set constants in the project, if they already defined.
https://github.com/graphql/graphql-js/blob/f31f5050644c3d2925557be3b285032345e91920/src/type/scalars.js#L21-L22

Please, add exports to contents

PR #2582, opened 2020-05-23
export constants MAX_INT and MIN_INT

overhead525:
Resolving issue [#2524](https://github.com/graphql/graphql-js/issues/2524)

### DIFF
--- a/src/type/scalars.js
+++ b/src/type/scalars.js
@@ -18,8 +18,8 @@ import { type GraphQLNamedType, GraphQLScalarType } from './definition';
 //
 // n.b. JavaScript's integers are safe between -(2^53 - 1) and 2^53 - 1 because
 // they are internally represented as IEEE 754 doubles.
-const MAX_INT = 2147483647;
-const MIN_INT = -2147483648;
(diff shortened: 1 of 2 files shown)


src/type/scalars.js
```diff
@@ -18,8 +18,8 @@ import { type GraphQLNamedType, GraphQLScalarType } from './definition';
 //
 // n.b. JavaScript's integers are safe between -(2^53 - 1) and 2^53 - 1 because
 // they are internally represented as IEEE 754 doubles.
-const MAX_INT = 2147483647;
-const MIN_INT = -2147483648;
+export const MAX_INT = 2147483647;
```
danielrearden, line 21: Please also add these constants to `graphql-js/src/type/scalars.d.ts`.

danielrearden:
@overhead525 Thanks for your PR!

It looks like your changes inadvertently also changed some of the existing formatting, which is causing this PR to fail the formatting check. The two constants should also be added as exports to `scalars.d.ts` (i.e. `export declare const MAX_INT: number;`) to keep the TypeScript type declarations inline with the Flow ones.

Issue #2577, opened 2020-05-22
printSchema: creates invalid block strings for descriptions


IvanGoncharov:
Not every string can be represented as block string but currently `printSchema` uses block strings for all descriptions.
Example:
```js
const { buildSchema, parse, printSchema } = require('./dist');
const schema = buildSchema('" " scalar Foo');
const printedSDL = printSchema(schema);
console.log(printedSDL);
// """ """
// scalar Foo

console.log(printSchema(buildSchema(printedSDL)));
// """"""
// scalar Foo
```

We should add a new utility function (in `blockString.js`) that checks if the string can be represented as a block string and use it inside `printSchema.js`.
Based on our previous experience with block strings it would be hard to cover all edge cases so I strongly suggest writing fuzzing test for this new function, similar to this one:
https://github.com/graphql/graphql-js/blob/47ac276f21598db5690f287f2f0ea41bbdeef4ca/src/language/__tests__/blockString-fuzz.js#L24

hereisnaman:
I want to work on this, please assign this to me.

jay-deshmukh:
in `printSchema.js` shouldn't `printDescription` method return `''` if `description === ''`, I see that happens when `description` is `null` but for empty string shouldn't that be same ? 

eg 
```js
function printDescription(def, indentation = '', firstInBlock = true) {
  const { description } = def;

  if (description == null || description == '') {  // change 
    return '';
  }

  const blockString = print({
    kind: Kind.STRING,
    value: description,
    block: isPrintableAsBlockString(description),
  });
  const prefix =
    indentation && !firstInBlock ? '\n' + indentation : indentation;
  return prefix + blockString.replace(/\n/g, '\n' + indentation) + '\n';
}

```

for me `printSchema(buildClientSchema(obj))` returns the schema with block-string for all descriptions 


`Upstream. => https://countries.trevorblades.com/`

```graphql
""""""
directive @cacheControl(maxAge: Int, scope: CacheControlScope) on FIELD_DEFINITION | OBJECT | INTERFACE

""""""
type Query {
  """"""
  continents(filter: ContinentFilterInput): [Continent!]!

  """"""
  continent(code: ID!): Continent

  """"""
  countries(filter: CountryFilterInput): [Country!]!

  """"""
  country(code: ID!): Country

  """"""
  languages(filter: LanguageFilterInput): [Language!]!

  """"""
  language(code: ID!): Language
}

""""""
input ContinentFilterInput {
  """"""
[...]

PR #2589, opened 2020-05-27
Print only required BlockStrings

thenamankumar:
fixes: https://github.com/graphql/graphql-js/issues/2577

Currently `printSchema` uses block strings for all descriptions. This PR add a check if the given string is printable block string, otherwise don't print it.

### DIFF
--- a/src/language/__tests__/blockString-fuzz.js
+++ b/src/language/__tests__/blockString-fuzz.js
@@ -10,7 +10,11 @@ import invariant from '../../jsutils/invariant';
 
 import { Lexer } from '../lexer';
 import { Source } from '../source';
-import { printBlockString } from '../blockString';
+import {
+  printBlockString,
+  isPrintableBlockString,
+  isBlank,
+} from '../blockString';
 
 function lexValue(str) {
   const lexer = new Lexer(new Source(str));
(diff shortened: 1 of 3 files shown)


src/language/__tests__/blockString-fuzz.js
```diff
@@ -66,3 +70,77 @@ describe('printBlockString', () => {
...
+    })) {
+      if (isPrintableBlockString(fuzzStr)) {
+        const testStr = '"""' + fuzzStr + '"""';
+
+        let testValue;
+        try {
+          testValue = lexValue(testStr);
+        } catch (e) {
+          continue; // skip invalid values
+        }
```
IvanGoncharov, line 90: You don't need to skip these invalid values this is the purpose of this fuzz test to detect them.

src/language/__tests__/blockString-fuzz.js
```diff
@@ -66,3 +70,77 @@ describe('printBlockString', () => {
...
+describe('isPrintableBlockString', () => {
+  it('correctly tells if random strings are printable block string', () => {
+    // Testing with length >7 is taking exponentially more time. However it is
+    // highly recommended to test with increased limit if you make any change.
+    for (const fuzzStr of genFuzzStrings({
+      allowedChars: ['\n', '\t', ' ', '"', 'a', '\\'],
+      maxLength: 7,
+    })) {
+      if (isPrintableBlockString(fuzzStr)) {
+        const testStr = '"""' + fuzzStr + '"""';
```
IvanGoncharov, line 83: Use `printBlockString`

src/language/__tests__/blockString-fuzz.js
```diff
@@ -66,3 +70,77 @@ describe('printBlockString', () => {
...
+          testValue = lexValue(testStr);
+        } catch (e) {
+          continue; // skip invalid values
+        }
+        invariant(typeof testValue === 'string');
+
+        const printedValue = lexValue(printBlockString(testValue));
+
+        invariant(
+          testValue === printedValue,
```
IvanGoncharov, line 96: It should be:
```js
lexValue(printBlockString(fuzzStr)) === fuzzStr
```

IvanGoncharov: and you can remove all previous code

src/language/__tests__/blockString-fuzz.js
```diff
@@ -66,3 +70,77 @@ describe('printBlockString', () => {
...
+            but got  ${inspectStr(printedValue)}
+        `,
+        );
+
+        const printedMultilineString = lexValue(
+          printBlockString(testValue, ' ', true),
+        );
+
+        invariant(
+          testValue === printedMultilineString,
```
IvanGoncharov, line 109: same as the previous `invariant`

src/language/__tests__/blockString-fuzz.js
```diff
@@ -66,3 +70,77 @@ describe('printBlockString', () => {
...
+            testValue,
+          )}, ' ', true))
+            to equal ${inspectStr(testValue)}
+            but got  ${inspectStr(printedMultilineString)}
+        `,
+        );
+      } else {
+        if (isBlank(fuzzStr)) {
+          continue;
+        }
```
IvanGoncharov, line 121: Why you are skipping blank lines?

src/language/__tests__/blockString-fuzz.js
```diff
@@ -66,3 +70,77 @@ describe('printBlockString', () => {
...
+
+        invariant(
+          fuzzStr !== printedValue,
+          dedent`
+          Expected lexValue(printBlockString(${inspectStr(fuzzStr)}))
+            not to equal ${inspectStr(fuzzStr)}
+        `,
+        );
+
+        const printedMultilineString = lexValue(
```
IvanGoncharov, line 132: In this test no need to test `printBlockString(fuzzStr, ' ', true)` it will slow down fuzzing
`printBlockString` fuzz tests already ensure that `lexValue(printBlockString(fuzzStr)) === lexValue( printBlockString(fuzzStr, ' ', true))` so no need to test it twice.

src/language/blockString.js
```diff
@@ -66,10 +66,14 @@ function leadingWhitespace(str) {
   return i;
 }
 
-function isBlank(str) {
+export function isBlank(str: string): boolean {
   return leadingWhitespace(str) === str.length;
 }
 
+export function isPrintableBlockString(str: string): boolean {
+  return Boolean(dedentBlockStringValue(str));
```
IvanGoncharov, line 74: please explicitly compare with empty string `=== ''`

src/language/printer.js
```diff
@@ -84,7 +84,9 @@ const printDocASTReducer: any = {
   FloatValue: ({ value }) => value,
   StringValue: ({ value, block: isBlockString }, key) =>
     isBlockString
-      ? printBlockString(value, key === 'description' ? '' : '  ')
+      ? isPrintableBlockString(value)
+        ? printBlockString(value, key === 'description' ? '' : '  ')
+        : ''
```
IvanGoncharov, line 89: You don't need to change `print` if AST node marked as block string we should print it only as block string.
You need to change only `printSchema` please use test case from original issue.

src/language/blockString.js
```diff
@@ -66,10 +66,14 @@ function leadingWhitespace(str) {
   return i;
 }
 
-function isBlank(str) {
+export function isBlank(str: string): boolean {
```
IvanGoncharov, line 69: No need to export this function for tests.
In tests, you can use new `trimStart` function.

In general, we never export functions to use them in tests, since tests should be independent of implementation.

src/language/blockString.js
IvanGoncharov: `dedentBlockStringValue` is an expensive function since split strings and does a lot of other manipulations. So it totally ok to use for the draft stage but it should be replaced with a simpler function before merge.

src/language/__tests__/blockString-fuzz.js
hereisnaman: fuzzStr can have lines with only whitespaces or trailing or leading multiple new lines which will be removed on print. Can't compare the 2.

IvanGoncharov: Yes, this is exactly the purpose of `isPrintableBlockString` to detect such strings.
Any string can be printed as standard strings but only subset of strings can be safely printed as block strings so `isPrintableBlockString` should be predicate function to detect them.

PR #2600, opened 2020-06-01
Add NoSchemaIntrospectionCustomRule

danielrearden:
#2597

### DIFF
--- a/src/validation/__tests__/NoSchemaIntrospectionCustomRule-test.js
+++ b/src/validation/__tests__/NoSchemaIntrospectionCustomRule-test.js
@@ -1,0 +1,142 @@
+// @flow strict
+
+import { describe, it } from 'mocha';
+
+import { buildSchema } from '../../utilities/buildASTSchema';
+
+import { NoSchemaIntrospectionCustomRule } from '../rules/custom/NoSchemaIntrospectionCustomRule';
+
+import { expectValidationErrorsWithSchema } from './harness';
+
+function expectErrors(queryStr) {
+  return expectValidationErrorsWithSchema(
+    schema,
+    NoSchemaIntrospectionCustomRule,
+    queryStr,
+  );
+}
+
+function expectValid(queryStr) {
+  expectErrors(queryStr).to.deep.equal([]);
+}
+
+const schema = buildSchema(`
+  type Query {
+    someQuery: SomeType
+  }
+
+  type SomeType {
+    someField: String
+    introspectionField: __EnumValue
+  }
+`);
+
+describe('Validate: Prohibit introspection queries', () => {
+  it('ignores valid fields including __typename', () => {
+    expectValid(`
+      {
+        someQuery {
+          __typename
+          someField
+        }
+      }
+    `);
+  });
+
+  it('ignores fields not in the schema', () => {
+    expectValid(`
+      {
+        __introspect
+      }
+    `);
+  });
+
+  it('reports error when a field with an introspection type is requested', () => {
+    expectErrors(`
+      {
+        __schema {
+          queryType {
+            name
+          }
+        }
+      }
+    `).to.deep.equal([
+      {
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "__schema".',
+        locations: [{ line: 3, column: 9 }],
+      },
+      {
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "queryType".',
+        locations: [{ line: 4, column: 11 }],
+      },
+    ]);
+  });
+
+  it('reports error when a field with an introspection type is requested and aliased', () => {
+    expectErrors(`
+      {
+        s: __schema {
+          queryType {
+            name
+          }
+        }
+      }
+      `).to.deep.equal([
+      {
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "__schema".',
+        locations: [{ line: 3, column: 9 }],
+      },
+      {
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "queryType".',
+        locations: [{ line: 4, column: 11 }],
+      },
+    ]);
+  });
+
+  it('reports error when using a fragment with a field with an introspection type', () => {
+    expectErrors(`
+      {
+        ...QueryFragment
+      }
+
+      fragment QueryFragment on Query {
+        __schema {
+          queryType {
+            name
+          }
+        }
+      }
+    `).to.deep.equal([
+      {
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "__schema".',
+        locations: [{ line: 7, column: 9 }],
+      },
+      {
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "queryType".',
+        locations: [{ line: 8, column: 11 }],
+      },
+    ]);
+  });
+
+  it('reports error for non-standard introspection fields', () => {
+    expectErrors(`
+      {
+        someQuery {
+          introspectionField
+        }
+      }
+    `).to.deep.equal([
+      {
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "introspectionField".',
+        locations: [{ line: 4, column: 11 }],
+      },
+    ]);
+  });
+});
--- a/src/validation/index.js
+++ b/src/validation/index.js
@@ -94,3 +94,6 @@ export { UniqueEnumValueNamesRule } from './rules/UniqueEnumValueNamesRule';
 export { UniqueFieldDefinitionNamesRule } from './rules/UniqueFieldDefinitionNamesRule';
 export { UniqueDirectiveNamesRule } from './rules/UniqueDirectiveNamesRule';
 export { PossibleTypeExtensionsRule } from './rules/PossibleTypeExtensionsRule';
+
+// Optional rules not defined by the GraphQL Specification
+export { NoSchemaIntrospectionCustomRule } from './rules/custom/NoSchemaIntrospectionCustomRule';
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -335,6 +335,8 @@ export {
   UniqueFieldDefinitionNamesRule,
   UniqueDirectiveNamesRule,
   PossibleTypeExtensionsRule,
+  // Custom validation rules
+  NoSchemaIntrospectionCustomRule,
   ValidationRule,
 } from './validation/index';
 
--- a/src/index.js
+++ b/src/index.js
@@ -333,6 +333,8 @@ export {
   UniqueFieldDefinitionNamesRule,
   UniqueDirectiveNamesRule,
   PossibleTypeExtensionsRule,
+  // Custom validation rules
+  NoSchemaIntrospectionCustomRule,
 } from './validation/index';
 
 export type { ValidationRule } from './validation/index';
--- a/src/validation/index.d.ts
+++ b/src/validation/index.d.ts
@@ -90,3 +90,6 @@ export { UniqueEnumValueNamesRule } from './rules/UniqueEnumValueNamesRule';
 export { UniqueFieldDefinitionNamesRule } from './rules/UniqueFieldDefinitionNamesRule';
 export { UniqueDirectiveNamesRule } from './rules/UniqueDirectiveNamesRule';
 export { PossibleTypeExtensionsRule } from './rules/PossibleTypeExtensionsRule';
+
+// Optional rules not defined by the GraphQL Specification
+export { NoSchemaIntrospectionCustomRule } from './rules/custom/NoSchemaIntrospectionCustomRule';
--- a/src/validation/rules/custom/NoSchemaIntrospectionCustomRule.d.ts
+++ b/src/validation/rules/custom/NoSchemaIntrospectionCustomRule.d.ts
@@ -1,0 +1,16 @@
+import { ASTVisitor } from '../../../language/visitor';
+import { ValidationContext } from '../../ValidationContext';
+
+/**
+ * Prohibit introspection queries
+ *
+ * A GraphQL document is only valid if all fields selected are not fields that
+ * return an introspection type.
+ *
+ * Note: This rule is optional and is not part of the Validation section of the
+ * GraphQL Specification. This rule effectively disables introspection, which
+ * does not reflect best practices and should only be done if absolutely necessary.
+ */
+export function NoSchemaIntrospectionCustomRule(
+  context: ValidationContext,
+): ASTVisitor;
--- a/src/validation/rules/custom/NoSchemaIntrospectionCustomRule.js
+++ b/src/validation/rules/custom/NoSchemaIntrospectionCustomRule.js
@@ -1,0 +1,39 @@
+// @flow strict
+
+import { GraphQLError } from '../../../error/GraphQLError';
+
+import type { FieldNode } from '../../../language/ast';
+import type { ASTVisitor } from '../../../language/visitor';
+
+import { getNamedType } from '../../../type/definition';
+import { isIntrospectionType } from '../../../type/introspection';
+
+import type { ValidationContext } from '../../ValidationContext';
+
+/**
+ * Prohibit introspection queries
+ *
+ * A GraphQL document is only valid if all fields selected are not fields that
+ * return an introspection type.
+ *
+ * Note: This rule is optional and is not part of the Validation section of the
+ * GraphQL Specification. This rule effectively disables introspection, which
+ * does not reflect best practices and should only be done if absolutely necessary.
+ */
+export function NoSchemaIntrospectionCustomRule(
+  context: ValidationContext,
+): ASTVisitor {
+  return {
+    Field(node: FieldNode) {
+      const type = getNamedType(context.getType());
+      if (type && isIntrospectionType(type)) {
+        context.reportError(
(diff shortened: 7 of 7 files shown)


src/validation/rules/NoIntrospectionFieldsRule.js
```diff
@@ -0,0 +1,32 @@
...
+ * A GraphQL document is only valid if all fields selected are not
+ * fields that return an introspection type. Note: This rule is not
+ * part of the Validation section of the GraphQL Specification.
+ */
+export function NoIntrospectionFieldsRule(
+  context: ValidationContext,
+): ASTVisitor {
+  return {
+    Field(node: FieldNode) {
+      if (node.name.value === '__schema' || node.name.value === '__type') {
```
IvanGoncharov, line 22: Please check that you are in root type, `context` has functions for that.

danielrearden: @IvanGoncharov Fixed!

IvanGoncharov:
@danielrearden Thanks, looks very tidy and match the style of other rules 👍 

I give it more though and technically `__typename` is also introspection field so we can't call it `NoIntrospectionFields` 😞 
Also, we have other fields that can leak similar info for example Apollo federation has `_service` that returns SDL.
So I more incline towards more generic rules like: 
* https://eslint.org/docs/rules/no-restricted-properties
* https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md
* etc.

Also, I think in this case being able to customize error messages would be very helpful.

But I'm not sure if I'm overcomplicating it or not so I need more time to think.
Meanwhile, if you have time you can work on #2601 which is very similar to this one.

IvanGoncharov:
@danielrearden Actually scratch it this rule is very short so config for the more generic rule would be more longer that rule itself.
So back the initial question, `__typename` is also introspection field so we need to change rule name.
How about `BanIntrospectionQueriesRule`?

src/validation/rules/NoIntrospectionFieldsRule.js
```diff
@@ -0,0 +1,35 @@
...
+ * part of the Validation section of the GraphQL Specification.
+ */
+export function NoIntrospectionFieldsRule(
+  context: ValidationContext,
+): ASTVisitor {
+  return {
+    Field(node: FieldNode) {
+      if (
+        context.getSchema().getQueryType() === context.getParentType() &&
+        (node.name.value === '__schema' || node.name.value === '__type')
```
IvanGoncharov, line 24: It's better to use `getFieldDef` and compare it against constants from `src/type/instospection`.

IvanGoncharov: that way you will have single check

src/validation/rules/NoIntrospectionFieldsRule.js
```diff
@@ -0,0 +1,35 @@
+// @flow strict
```
IvanGoncharov, line 1: better to move it to `rules/extra` directory.

IvanGoncharov: or if you can come up with a better name for this directory.

src/validation/rules/NoIntrospectionFieldsRule.js
danielrearden: > How about `BanIntrospectionQueriesRule`?

That makes sense. Would `NoIntrospectionQueriesRule` be more inline with the naming of the existing rules?

danielrearden:
> How about `BanIntrospectionQueriesRule`?

@IvanGoncharov  That makes sense. Would `NoIntrospectionQueriesRule` be more inline with the existing rule names?

src/validation/__tests__/NoIntrospectionFieldsRule-test.js
```diff
@@ -0,0 +1,100 @@
+// @flow strict
+
+import { describe, it } from 'mocha';
+
+import { NoIntrospectionFieldsRule } from '../rules/NoIntrospectionFieldsRule';
+
+import { expectValidationErrors } from './harness';
+
+function expectErrors(queryStr) {
+  return expectValidationErrors(NoIntrospectionFieldsRule, queryStr);
```
IvanGoncharov, line 10: it's better to use `expectValidationRulesWithSchema` and have a short SDL snippet in this file than to depend on a huge schema.
I was planning to break "harness" into multiple small schemas and put them directly into test files so for new rules let's do it from the start.
BTW. Are you interested in working on breaking "harness" schema?

src/validation/rules/NoIntrospectionFieldsRule.js
danielrearden: So a name like `NoFieldsWithIntrospectionTypesRule` would maybe be more accurate.

src/validation/__tests__/NoIntrospectionFieldsRule-test.js
danielrearden: I can do that. Maybe I could knock that out as a separate PR first and then rebase this.

IvanGoncharov:
> Would NoIntrospectionQueriesRule be more inline with the existing rule names?

@danielrearden Generally, yes. But in this case, we forbid something that is totally valid and you ban it because of your own preference/requirements not even because it represents best practice.
Speaking of this can you please expand description with some strong wording asking not to use this rule unless it's absolutely necessary.

src/validation/__tests__/NoIntrospectionFieldsRule-test.js
IvanGoncharov: Actually, if you are interested in doing that work, we can merge this first and your later change all the rules in one decisive PR.

danielrearden: That works too!

src/validation/rules/NoIntrospectionFieldsRule.js
IvanGoncharov: Technically you right. But it's better to use names that improve discoverability/readability than a name 100% correct. 
For example, `SDL` is an incorrect abbreviation since specification defines `Type System Definition Language` but we use `SDL` anyway since it what community settled on.

I think the spirit of this rule is that we forbidding "introspection queries" so it's better to reflect it a name even through `{ __typename }` is also introspection query.

danielrearden:
@IvanGoncharov 

* Changed name to `ProhibitIntrospectionQueriesRule`
* Updated rule logic to just check the field type
* Moved rule to `/rules/optional`
* Updated comment to include usage warning
* Added test for non-standard introspection fields -- ended up using `expectValidationRulesWithSchema` anyway so I wouldn't have to touch the harness schema 
* Had to build schema programmatically since `buildSchema` doesn't support using introspection types for fields

The use of introspection types with user-defined fields is unconventional but I don't believe it's prohibited by the spec. Since the rule just tests the field's type, it's possible it would also throw in the event a non-standard introspection field like that was requested. I think that behavior lines up with the intent behind the rule.

src/validation/rules/optional/ProhibitIntrospectionQueriesRule.js
```diff
@@ -0,0 +1,47 @@
...
+    Field(node: FieldNode) {
+      const fieldDef = context.getFieldDef();
+      if (fieldDef) {
+        // Unwrap the field's type in case it is a NonNull or List type
+        let fieldType = fieldDef.type;
+        while (isWrappingType(fieldType)) {
+          fieldType = fieldType.ofType;
+        }
+
+        if (introspectionTypes.includes(fieldType)) {
```
IvanGoncharov, line 36: Why not simply call `isIntrospectionType`?

src/validation/rules/optional/ProhibitIntrospectionQueriesRule.js
```diff
@@ -0,0 +1,47 @@
...
+): ASTVisitor {
+  return {
+    Field(node: FieldNode) {
+      const fieldDef = context.getFieldDef();
+      if (fieldDef) {
+        // Unwrap the field's type in case it is a NonNull or List type
+        let fieldType = fieldDef.type;
+        while (isWrappingType(fieldType)) {
+          fieldType = fieldType.ofType;
+        }
```
IvanGoncharov, line 34: We have `getNamedType` for that

src/validation/rules/optional/ProhibitIntrospectionQueriesRule.js
```diff
@@ -0,0 +1,47 @@
...
+ * Note: This rule is optional and is not part of the Validation section of the
+ * GraphQL Specification. This rule effectively disables introspection, which
+ * does not reflect best practices and should only be done if absolutely necessary.
+ */
+export function ProhibitIntrospectionQueriesRule(
+  context: ValidationContext,
+): ASTVisitor {
+  return {
+    Field(node: FieldNode) {
+      const fieldDef = context.getFieldDef();
```
IvanGoncharov, line 28: If you want to just get resulting type please use `getType`

danielrearden:
@IvanGoncharov Ok, that should be all of it 😓

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "javascript.validate.enable": false
+}
```
IvanGoncharov, line 3: Please remove this file.

src/validation/__tests__/ProhibitIntrospectionQueriesRule-test.js
```diff
@@ -0,0 +1,146 @@
...
+    ProhibitIntrospectionQueriesRule,
+    queryStr,
+  );
+}
+
+function expectValid(queryStr) {
+  expectErrors(queryStr).to.deep.equal([]);
+}
+
+describe('Validate: Prohibit introspection queries', () => {
```
IvanGoncharov, line 45: Please add one test with introspection field being reported in fragment.

src/validation/__tests__/ProhibitIntrospectionQueriesRule-test.js
```diff
@@ -0,0 +1,146 @@
...
+        __schema {
+          queryType {
+            name
+          }
+        }
+      }
+    `).to.deep.equal([
+      {
+        locations: [
+          {
```
IvanGoncharov, line 77: Please format locations as in the other test: single line with `line` followed by `column`

src/validation/__tests__/ProhibitIntrospectionQueriesRule-test.js
```diff
@@ -0,0 +1,146 @@
...
+            column: 9,
+            line: 3,
+          },
+        ],
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "__schema".',
```
IvanGoncharov, line 83: Also, `message` should be first property followed by `locations`

src/validation/__tests__/ProhibitIntrospectionQueriesRule-test.js
```diff
@@ -0,0 +1,146 @@
...
+      {
+        someQuery {
+          introspectionField
+        }
+      }
+    `).to.deep.equal([
+      {
+        message:
+          'GraphQL introspection has been disabled, but the requested query contained the field "introspectionField".',
+        locations: [{ line: 4, column: 11 }],
```
IvanGoncharov, line 142: Here is good example of correct formatting.

src/validation/rules/optional/ProhibitIntrospectionQueriesRule.js
```diff
@@ -0,0 +1,41 @@
...
+ * Note: This rule is optional and is not part of the Validation section of the
+ * GraphQL Specification. This rule effectively disables introspection, which
+ * does not reflect best practices and should only be done if absolutely necessary.
+ */
+export function ProhibitIntrospectionQueriesRule(
+  context: ValidationContext,
+): ASTVisitor {
+  return {
+    Field(node: FieldNode) {
+      const fieldType = context.getType();
```
IvanGoncharov, line 28: You can write it like this:
```js
const type = getNamedType(context.getType());
if (type != null && isIntrospectionType(type) {
```

src/validation/rules/optional/ProhibitIntrospectionQueriesRule.js
```diff
@@ -0,0 +1,41 @@
...
+
+import { GraphQLError } from '../../../error/GraphQLError';
+
+import { type FieldNode } from '../../../language/ast';
+import { type ASTVisitor } from '../../../language/visitor';
+
+import { type ValidationContext } from '../../ValidationContext';
+
+import { getNamedType } from '../../../type/definition';
+import { isIntrospectionType } from '../../../type/introspection';
```
IvanGoncharov, line 11: You should order imports in hierarchical so `../../../type` should be before `../../ValidationContext`.
So first ordering criteria.

We put type import value imports in the same import statements so it doesn't make sense to order by them by type/non-type criteria since latter you would mix them in the same statement anyway.

src/validation/__tests__/ProhibitIntrospectionQueriesRule-test.js
```diff
@@ -0,0 +1,146 @@
...
+const schema = new GraphQLSchema({
+  query: new GraphQLObjectType({
+    name: 'Query',
+    fields: {
+      someQuery: {
+        type: new GraphQLObjectType({
+          name: 'SomeType',
+          fields: {
+            someField: { type: GraphQLString },
+            introspectionField: { type: __Schema },
```
IvanGoncharov, line 25: If you like to test a non-standard introspection field it's better to use some type that is not exposed directly by GraphQL itself, for example, type for enum value.

src/validation/__tests__/ProhibitIntrospectionQueriesRule-test.js
```diff
@@ -0,0 +1,146 @@
...
+import { describe, it } from 'mocha';
+
+import { ProhibitIntrospectionQueriesRule } from '../rules/optional/ProhibitIntrospectionQueriesRule';
+
+import {
+  GraphQLSchema,
+  GraphQLObjectType,
+  GraphQLString,
+  __Schema,
+} from '../../type/index';
```
IvanGoncharov, line 12: Same note about order of import, more "external" files should go first.

src/validation/index.js
```diff
@@ -94,3 +94,6 @@ export { UniqueEnumValueNamesRule } from './rules/UniqueEnumValueNamesRule';
 export { UniqueFieldDefinitionNamesRule } from './rules/UniqueFieldDefinitionNamesRule';
 export { UniqueDirectiveNamesRule } from './rules/UniqueDirectiveNamesRule';
 export { PossibleTypeExtensionsRule } from './rules/PossibleTypeExtensionsRule';
+
+// Optional rules not defined by the GraphQL Specification
+export { ProhibitIntrospectionQueriesRule } from './rules/optional/ProhibitIntrospectionQueriesRule';
```
IvanGoncharov, line 99: Please duplicate it in `src/validation/index.d.ts`, `src/index.{js,d.ts}`

.vscode/settings.json
danielrearden: Dangit

src/validation/rules/optional/ProhibitIntrospectionQueriesRule.js
danielrearden: Totally missed that's the code style being enforced. Would it make sense to update the [import/order](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md) ESLint rule to help enforce that?

IvanGoncharov:
@danielrearden Forget about this one.
But it too late in my timezone so I will go to 🛌 will check it tomorrow.

P.S. It's totally okay to ping in comments after you addressed all review comments. 
Moreover actually makes my life easier since I don't need to check all current PRs to see their statuses 😄

src/validation/rules/optional/ProhibitIntrospectionQueriesRule.js
IvanGoncharov: If you can do that would be great at the same time I don't think we can enforce everything with their current config.
For example, I order imports in order (it matches order of chapters in GraphQL spec):
* polyfills
* \__testUtils__
* jsutils
* error
* **\<same order as chapters in graphql spec>**
* language
* type
* validation
* execution
* subscription
* **\<end of the same order>**
* utilities
* `src/graphql.js`
* files in upper dir
* files in the same dir
* \__fixtures__

If we can enforce something like this with `import/order` I'm totally for it if not I think we need to write our own custom rule.
I'm also ok to change the order if it would be easier to enforce it but for folders that present spec chapters like to keep them in the same order.

@danielrearden Can you please open an issue to track this?

IvanGoncharov: Also, it definitely not compete with TS migration but actually help with reviewing such PRs.

src/validation/__tests__/NoSchemaIntrospectionCustomRule-test.js
```diff
@@ -0,0 +1,153 @@
...
+          name: 'SomeType',
+          fields: {
+            someField: { type: GraphQLString },
+            introspectionField: { type: __EnumValue },
+          },
+        }),
+      },
+    },
+  }),
+});
```
IvanGoncharov, line 43: @danielrearden I merged #2608 so now `buildSchema` should work fine can you please use it instead.

Issue #2601, opened 2020-06-02
Extract validation rule for deprecated fields


IvanGoncharov:
We already have a rule for detecting deprecated fields but the community constantly reimplement it because we don't expose it as rule but as a function:
https://github.com/apollographql/eslint-plugin-graphql/blob/c465fedc8fea239ee1731ad4ec3ee1183a3cdddf/src/customGraphQLValidationRules.js#L152-L153

So we need to add it under `src/validation/rules/extra` and deprecate `findDeprecatedUsages` :)

PR #2605, opened 2020-06-03
Add NoDeprecatedCustomRule and deprecate findDeprecatedUsages

danielrearden:
#2601


src/validation/rules/optional/ProhibitDeprecatedFieldsRule.js
```diff
@@ -0,0 +1,52 @@
...
+ * GraphQL Specification.
+ */
+export function ProhibitDeprecatedFieldsRule(
+  context: ValidationContext,
+): ASTVisitor {
+  return {
+    Field(node: FieldNode) {
+      const fieldDef = context.getFieldDef();
+      const parentType = context.getParentType();
+      if (parentType && fieldDef && fieldDef.deprecationReason != null) {
```
danielrearden, line 28: I modified the original logic here to account for the fact that a deprecated field will always have a deprecation reason (since `isDeprecated` is derived from the truthiness of the `deprecationReason` value). The original function had code that was effectively unreachable because it tried to take into account a scenario where `isDeprecated` is true but no `deprecationReason` is defined.

I'd rather test for `isDeprecated` here, but then we'd have to cast `fieldDef.deprecationReason` to `any` and again to `string`, and this seemed like the lesser evil.

Ideally, GraphQLField would be a union type like `{ isDeprecated: true, deprecationReason: string } | { isDeprecated: false, deprecationReason: null | void }` -- maybe that's something we could change?

src/utilities/findDeprecatedUsages.js
```diff
@@ -11,7 +11,7 @@ import { type GraphQLSchema } from '../type/schema';
 import { TypeInfo, visitWithTypeInfo } from './TypeInfo';
 
 /**
- * A validation rule which reports deprecated usages.
+ * @deprecated: A validation rule which reports deprecated usages.
  *
  * Returns a list of GraphQLError instances describing each deprecated use.
  */
```
IvanGoncharov, line 17: Can you switch old implementation to use the new validation rule?

src/utilities/__tests__/findDeprecatedUsages-test.js
```diff
@@ -1,5 +1,7 @@
 // @flow strict
 
+/* eslint-disable import/no-deprecated */
+
```
IvanGoncharov, line 4: If you implement an old function on top of a new validation rule you can simply remove all tests and disable coverage on `findDeprecatedUsages.js`

src/utilities/findDeprecatedUsages.js
```diff
@@ -11,7 +11,7 @@ import { type GraphQLSchema } from '../type/schema';
 import { TypeInfo, visitWithTypeInfo } from './TypeInfo';
 
 /**
- * A validation rule which reports deprecated usages.
+ * @deprecated: A validation rule which reports deprecated usages.
```
IvanGoncharov, line 14: Please provide a code sample on how to migrate to the new rule + validate.

src/validation/rules/optional/ProhibitDeprecatedFieldsRule.d.ts
```diff
@@ -0,0 +1,15 @@
+import { ASTVisitor } from '../../../language/visitor';
+import { ValidationContext } from '../../ValidationContext';
+
+/**
+ * Prohibit deprecated fields
+ *
+ * A GraphQL document is only valid if all selected fields and all used enum
+ * values have not been deprecated.
```
IvanGoncharov, line 8: For comment/docs/strings you can use more than 80 columns especially if it's just a few words.

src/validation/rules/optional/ProhibitDeprecatedFieldsRule.d.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Note: This rule is optional and is not part of the Validation section of the
+ * GraphQL Specification.
+ */
+export function ProhibitDeprecatedFieldsRule(
```
IvanGoncharov, line 13: I wouldn't use "fields" since we also forbid enum values and plan to support deprecation on arguments so we need a more generic name.

src/validation/rules/optional/ProhibitDeprecatedFieldsRule.js
```diff
@@ -0,0 +1,52 @@
...
+import { type ValidationContext } from '../../ValidationContext';
+
+/**
+ * Prohibit deprecated fields
+ *
+ * A GraphQL document is only valid if all selected fields and all used enum
+ * values have not been deprecated.
+ *
+ * Note: This rule is optional and is not part of the Validation section of the
+ * GraphQL Specification.
```
IvanGoncharov, line 19: I would add something that specifies that its main purpose is to detect deprecation usages not necessary forbid.

src/validation/rules/optional/ProhibitDeprecatedFieldsRule.js
IvanGoncharov: I actually think `isDeprecated` itself was a mistake and we should just use `deprecationReason` everywhere.
BTW. You can create and issue for that mark `isDeprecated` as deprecated :)
and update all our code to use `deprecationReason` instead.

src/validation/rules/optional/ProhibitDeprecatedFieldsRule.js
```diff
@@ -0,0 +1,52 @@
...
+              fieldDef.deprecationReason,
+            [node],
+          ),
+        );
+      }
+    },
+    EnumValue(node: EnumValueNode) {
+      const type = getNamedType(context.getInputType());
+      const enumValue = context.getEnumValue();
+      if (type && enumValue?.deprecationReason != null) {
```
IvanGoncharov, line 41: Why you can't use the same trick with optional chaining for fields?

src/validation/rules/optional/ProhibitDeprecatedFieldsRule.js
```diff
@@ -0,0 +1,52 @@
...
+  return {
+    Field(node: FieldNode) {
+      const fieldDef = context.getFieldDef();
+      const parentType = context.getParentType();
+      if (parentType && fieldDef && fieldDef.deprecationReason != null) {
+        context.reportError(
+          new GraphQLError(
+            `The field ${parentType.name}.${fieldDef.name} is deprecated. ` +
+              fieldDef.deprecationReason,
+            [node],
```
IvanGoncharov, line 33: You don't need to wrap node in the array, you actually did it bellow with enum error.

src/validation/rules/optional/ProhibitDeprecatedFieldsRule.d.ts
danielrearden: Hmm. Is there a name for "things in the schema"? I think the closest I've seen in the spec is "things" and "portions of the schema".

`ProhibitDeprecatedUsages`?

src/validation/__tests__/ProhibitDeprecatedFieldsRule-test.js
```diff
@@ -0,0 +1,135 @@
...
+      }
+    `);
+  });
+
+  it('ignores unknown fields and enum values', () => {
+    expectValid(`
+      fragment UnknownFragment on UnknownType {
+        unknownField(unknownArg: UNKNOWN_VALUE)
+      }
+    `);
```
IvanGoncharov, line 51: You not testing unknown field on known type combination.
That's why coverage showed that branch as uncovered in optional chaining.

IvanGoncharov: the same problem with enum value you always check the only left side of `&&`.

src/validation/__tests__/ProhibitDeprecatedFieldsRule-test.js
```diff
@@ -0,0 +1,135 @@
...
+        locations: [{ line: 4, column: 9 }],
+      },
+    ]);
+  });
+
+  it('reports error when a deprecated field with default reason is selected', () => {
+    expectErrors(`
+      {
+        normalField
+        deprecatedFieldWithNoReason
```
IvanGoncharov, line 73: Since this test doesn't actually test anything new (since default for deprecation reason was added during `buildSchema`) it's better to merge it with above test.
It's normal to have two errors in expected result.
Same for enum

src/validation/__tests__/ProhibitDeprecatedFieldsRule-test.js
```diff
@@ -0,0 +1,135 @@
...
+      },
+    ]);
+  });
+
+  it('reports error when a deprecated field is selected or an enum value is used inside a fragment', () => {
+    expectErrors(`
+      {
+        ...QueryFragment
+      }
+
```
IvanGoncharov, line 117: You don't need to write full query here just having fragment would be enough.

src/validation/__tests__/ProhibitDeprecatedFieldsRule-test.js
```diff
@@ -0,0 +1,135 @@
...
+  return expectValidationErrorsWithSchema(
+    schema,
+    ProhibitDeprecatedFieldsRule,
+    queryStr,
+  );
+}
+
+function expectValid(queryStr) {
+  expectErrors(queryStr).to.deep.equal([]);
+}
```
IvanGoncharov, line 35: nit: these two functions kind of boring since you see them in every validation tests so it's better to put `schema` closer to actual tests so you do less scrolling.
I always put boring and repairing parts closer to imports and more specific stuff closer to tests

src/validation/rules/optional/ProhibitDeprecatedFieldsRule.d.ts
IvanGoncharov: `ProhibitDeprecatedUsage` sounds reasonable.

@benjie You are really good at coming up with good names. Maybe you can advise?
We creating a folder with optional rules and two first rules would be `ProhibitIntrospectionQueriesRule` and `ProhibitDeprecatedUsages`.
Does this naming scheme make sense for these two rules and similar rules in the future?

benjie: All the current rules describe the desired outcome, what the valid query has, rather than the action of the rule itself. And they end in “Rule”. So in this we want a valid query to have no deprecated values, fields, etc... unless we can think of a good collective noun; I think calling it `NoDeprecatedRule` would be acceptable. I’d rather add a word between deprecated and rule, but I can’t think of a good one: usages, things, entities, constructs, fields, components, productions; none of these are quite right. Maybe “elements”?

benjie: I’d go with `NoIntrospectionRule` for the other; but presumably it doesn’t block __typename.

IvanGoncharov: @benje Thanks for review 👍 

> I’d go with NoIntrospectionRule for the other; but presumably it doesn’t block __typename.

Yes, that's why we added `Queries`.

>  I think calling it NoDeprecatedRule would be acceptable.

I kind of wanted to separate it from other rules since other `No*` mean it's forbidden by the spec idea was to show that you forbid something that is legal but it's just your decision to do so.
Do you think it's worth having this separation?

benjie: > Yes, that's why we added Queries.

I don't feel adding `Queries` really adds any difference between `__typename` vs `__type` here. E.g. `{ currentUser { __typename id friends { name } teams { name } } __type(name: "UserType") { enumValues { name } } }` is not an "introspection query" exactly, but it does contain introspection. I think `NoIntrospectionRule` is simpler and more direct, and it's fairly obvious to most GraphQL users that `__typename` (though an introspection field technically) would not be forbidden.

If you want a more precise name, `NoSchemaIntrospectionRule` would be a better choice, IMO.

> I kind of wanted to separate it from other rules since other No* mean it's forbidden by the spec idea was to show that you forbid something that is legal but it's just your decision to do so.

I don't think having a subtle language difference like this is a good way of demarking this separation. I'd stick to the existing naming conventions; but potentially prefix the rule with `Custom` or similar text*, i.e. `CustomNoSchemaIntrospectionRule`.

Strictly the `NoSchemaIntrospectionRule` actually breaks the GraphQL spec (which requires introspection be available), so I'm not really happy with something like "Optional" as a description for it, hence `Custom`.

* `Custom`: `NonStandard`, `OptIn`, `Optional`, `Bespoke`, `Unspecified`, `Extra`, `Unsanctioned`, ...

> Do you think it's worth having this separation?

It seems like a good idea if it can be executed well.

IvanGoncharov: @benjie Thanks `NoSchemaIntrospectionRule` looks great 👍 

>  so I'm not really happy with something like "Optional" as a description for it, hence Custom.

Great logic 👍  Actually refusing deprecated rules is also breaking the spec.
How about moving `Custom` to the end so `NoSchemaIntrospectionCustomRule` and `NoDeprecatedCustomRule`?

benjie: Sure, sounds good :+1:

IvanGoncharov: @danielrearden What do you think about the proposed naming schema?
So `NoSchemaIntrospectionCustomRule` and `NoDeprecatedCustomRule` will live in `src/validation/rules/custom`.

danielrearden: @IvanGoncharov Sounds great. I'll get both PRs updated and address the rest of the comments in this one ASAP. Thanks for the help @benjie !

IvanGoncharov: @danielrearden Don't hurry we still blocked by TS conversion of tests. Don't want to prolong it so we in freeze for everything unrelated that doesn't touch tests probably for next couple days.

danielrearden: I figured, also why I'm holding off on breaking up the harness schema like we discussed

IvanGoncharov: @danielrearden If you have time for this it would be great if you can convert tests for `express-graphql` into TS meanwhile. should be way simpler but try to separate changes into PRs (.d.ts updates (if needed), missing types (you can start without `noImplicitAny` and enable it in the next PR) so basically try to separate syntax changes from something serious.

src/utilities/findDeprecatedUsages.js
```diff
@@ -2,57 +2,29 @@
...
-              node,
-            ),
-          );
-        }
-      },
-    }),
-  );
-
-  return errors;
+): $ReadOnlyArray<GraphQLError> {
```
danielrearden, line 28: Changed the type here and in the declaration file to readonly to match validate

PR #2609, opened 2020-06-04
Migrate all tests code to TypeScript

dotansimha:
This PR is based on the TypeScript migration path from https://github.com/graphql/graphql-js/issues/2104 

This PR is also inspired by https://github.com/graphql/graphql-js/pull/2139 , but it uses the most recent codebase, and migrates all tests files to TS.

The changes in this PR includes the following:

### TS Migration
- Migrate all tests code to use TypeScript, and added some missing `.d.ts` files for test utils and other `.js` files that are in use by both Flow code and TypeScript tests.
- Added root `tsconfig.json`, this is needed because we are going to compile TypeScript files, and the `tsconfig.json` that located under `src` is different, and used by `dtslint` only.
- Added `ts-node` and use `ts-node/register` in Mocha configuration to support `.ts` files.

### Bug fixes (non-breaking)
- `GraphQLError.d.ts` - use `Maybe` for `nodes` field.
- `lexer.d.ts` - Added missing declaration for `isPunctuatorTokenKind`.
- `definition.d.ts` - changed `serialize` to be optional in `GraphQLScalarTypeConfig`.
- `schema.d.ts` - changed `query` in `GraphQLSchemaConfig` to be optional.
- `getIntrospectionQuery.d.ts` - Changed `descriptions` to be optional in `IntrospectionOptions`.
- `getOperationAST.d.ts ` - changed `getOperationAST` signature to allow 2nd arg to be optional.
- `KnownTypeNamesRule.d.ts` - updated signature to match the implementing type.

### Lint
- Turn off `@typescript-eslint/ban-ts-comment` rule, because we need it in test files (replaces `$DisableFlowOnNegativeTest`).
- Updated TSLint config - currently, TSLint only used by `dtslint` in order to lint the `.d.ts` files, so it needs to ignore.
- Changed `@typescript-eslint/consistent-type-assertions` to allow `{} as Something` in test files.

### Next steps in later PRs?
- Replace TSLint with ESLint ([deprecated](https://medium.com/palantir/tslint-in-2019-1a144c2317a9)) and use a single configuration file for linting all files. Also, use a single linter for all files (just eslint, and remove dtslist and tslint) 
- Replace Mocha and Chai with Jest?

PR #2616, opened 2020-06-05
Fixes for `d.ts` files

dotansimha:
Originally from: https://github.com/graphql/graphql-js/pull/2609


src/__fixtures__/index.d.ts
```diff
@@ -0,0 +1,4 @@
+export declare const bigSchemaSDL: string;
+export declare const bigSchemaIntrospectionResult: Record<string, any>;
```
IvanGoncharov, line 2:
```suggestion
export declare const bigSchemaIntrospectionResult: {};
```

IvanGoncharov: You will still need to do cast inside tests but at least you will be able to get `data` without cast so it will be one cast instead of two.

src/jsutils/didYouMean.d.ts
```diff
@@ -0,0 +1,4 @@
+export default function didYouMean(
+  firstArg: string | ReadonlyArray<string>,
+  secondArg?: ReadonlyArray<string>,
+): string;
```
IvanGoncharov, line 4: Typescript supports normal overloading so try to match flow here:
```
declare function didYouMean(suggestions: $ReadOnlyArray<string>): string;
// eslint-disable-next-line no-redeclare
declare function didYouMean(
  subMessage: string,
  suggestions: $ReadOnlyArray<string>,
): string;
```

src/jsutils/didYouMean.d.ts
```diff
@@ -0,0 +1,4 @@
+export default function didYouMean(
```
IvanGoncharov, line 1: Please copy comments jsdoc comments for all methods of new stuff if they present in Flow sources.

src/jsutils/toObjMap.d.ts
```diff
@@ -0,0 +1,13 @@
...
+  ObjMap,
+  ObjMapLike,
+  ReadOnlyObjMap,
+  ReadOnlyObjMapLike,
+} from './ObjMap';
+
+// eslint-disable-next-line import/export
+export default function toObjMap<T>(obj: ObjMapLike<T>): ObjMap<T>;
+// eslint-disable-next-line import/export
+export default function toObjMap<T>(
```
IvanGoncharov, line 11: I think in TS you can first overload function and then expose it by name?

src/language/lexer.d.ts
```diff
@@ -50,3 +51,5 @@ export class Lexer {
  * @internal
  */
 export function isPunctuatorToken(token: Token): boolean;
+
+export function isPunctuatorTokenKind(kind: TokenKindEnum): boolean;
```
IvanGoncharov, line 55: please copy description

IvanGoncharov:
@dotansimha All the fixes in existing `*.d.ts` files look good but new typing requires an extra review.
So can you please extract fixes to existing files into separate PR?

dotansimha:
@IvanGoncharov I saw you already moved the fixes to a new PR and merged, so that's great! Thank you. 
I rebased and applied the fixes.
 
Do you think I should rebase https://github.com/graphql/graphql-js/pull/2609 to this PR? because those changes are needed in order to support the TS tests.

yaacovCR:
This can be closed considering that the code base has been converted to TS.

Thank you @dotansimha for your work on this, which has been integrated into other PRs as well as for your energy and initiative in pushing forward the migration.

PR #2621, opened 2020-06-06
Moved `Maybe` to `jsutils` folder and remove `tsutils`

IvanGoncharov:


wodka:
This change breaks Apollo Server

```
node_modules/@apollo/federation/dist/composition/utils.d.ts:3:19
  error TS2307: Cannot find module 'graphql/tsutils/Maybe' or its corresponding type declarations.

import Maybe from 'graphql/tsutils/Maybe';
```

IvanGoncharov:
@wodka Thanks for reporting 👍 
It is an internal type not intended to be used outside of our package that is why it's not exported through `intex.d.ts`.
Can you please report it to the Apollo team?

PR #2623, opened 2020-06-06
TS: sync TS typings with Flow typings

IvanGoncharov:
extracted from #2609


dotansimha:
@IvanGoncharov I think you missed the `GraphQLError` fix? here: https://github.com/graphql/graphql-js/pull/2616/files#diff-7dbaf2fce9ebeb528b2cae60c208ee06

PR #2669, opened 2020-06-19
Add parentType to path to avoid path ambiguity

benjie:
Imagine you have the schema:

```graphql
type A {
  int: Int
  union: C
}
type B {
  int: Int
  union: C
}
union C = A | B
type Query {
  root: C
}
```

You could write the query:

```graphql
{
  root {
    ... on A {
      union {
        ... on A {
          alias1: union { # EXAMPLE1
            int
          }
          alias2: union {
            int
          }
        }
        ... on B {
          alias1: union {
            int
          }
          alias2: union {
            int
          }
        }
      }
    }
    ... on B {
      union {
        ... on A {
          alias1: union { # EXAMPLE2
            int
          }
          alias2: union {
            int
          }
        }
        ... on B {
          alias1: union {
            int
          }
          alias2: union {
            int
          }
        }
      }
    }
  }
}
```

In the resolver for `A.union`, in `resolveInfo` (the fourth argument to resolve) you would not be able to tell if you are in position `EXAMPLE1` or `EXAMPLE2` without looking into the `loc` data in `fieldNodes`. This makes it challenging to write certain classes of lookahead engine.

Namely in both EXAMPLE1 and EXAMPLE2, you have:

- `parentType: 'A'`
- `path: {key: 'alias1', prev: { key: 'union', prev: { key: 'root' } } }`

So the `path` is ambiguous. This PR adds `parentType` to the `Path` type, so these paths would become:

- EXAMPLE1: `path: {key: 'alias1', parentTypeName: 'A', prev: { key: 'union', parentTypeName: 'A', prev: { key: 'root', parentTypeName: 'Query' } } }`
- EXAMPLE2: `path: {key: 'alias1', parentTypeName: 'A', prev: { key: 'union', parentTypeName: 'B', prev: { key: 'root', parentTypeName: 'Query' } } }`

Note the subtle difference between the A/B parentTypeName in the middle of the path.

### DIFF
--- a/src/execution/__tests__/executor-test.js
+++ b/src/execution/__tests__/executor-test.js
@@ -329,7 +329,7 @@ describe('Execute: Handles basic execution tasks', () => {
       name: 'TestUnion',
       types: [a],
       resolveType() {
-        return a;
+        return 'A';
       },
     });
     const testType = new GraphQLObjectType({
--- a/src/execution/__tests__/executor-test.js
+++ b/src/execution/__tests__/executor-test.js
@@ -314,8 +314,8 @@ describe('Execute: Handles basic execution tasks', () => {
 
   it('populates path correctly with complex types', () => {
     let path;
-    const a = new GraphQLObjectType({
-      name: 'A',
+    const someObject = new GraphQLObjectType({
+      name: 'SomeObject',
       fields: {
         test: {
           type: GraphQLString,
@@ -325,41 +325,46 @@ describe('Execute: Handles basic execution tasks', () => {
         },
       },
     });
-    const testUnion = new GraphQLUnionType({
-      name: 'TestUnion',
-      types: [a],
+    const someUnion = new GraphQLUnionType({
+      name: 'SomeUnion',
+      types: [someObject],
       resolveType() {
-        return 'A';
+        return 'SomeObject';
       },
     });
     const testType = new GraphQLObjectType({
-      name: 'Test',
+      name: 'SomeQuery',
       fields: {
         test: {
           type: new GraphQLNonNull(
-            new GraphQLList(new GraphQLNonNull(testUnion)),
+            new GraphQLList(new GraphQLNonNull(someUnion)),
           ),
-          resolve(_val, _args, _ctx, _info) {
-            return [{ type: 'A' }];
-          },
         },
       },
     });
     const schema = new GraphQLSchema({ query: testType });
+    const rootValue = { test: [{}] };
+    const document = parse(`
+      query {
+        l1: test {
+          ... on SomeObject {
+            l2: test
+          }
+        }
+      }
+    `);
 
-    const document = parse('query { l1: test { ... on A { l2: test } } }');
-
-    execute({ schema, document });
+    execute({ schema, document, rootValue });
 
     expect(path).to.deep.equal({
       key: 'l2',
-      parentType: 'A',
+      parentType: 'SomeObject',
       prev: {
         key: 0,
         parentType: undefined,
         prev: {
           key: 'l1',
-          parentType: 'Test',
+          parentType: 'SomeQuery',
           prev: undefined,
         },
       },
--- a/src/execution/__tests__/executor-test.js
+++ b/src/execution/__tests__/executor-test.js
@@ -17,6 +17,7 @@ import {
   GraphQLScalarType,
   GraphQLInterfaceType,
   GraphQLObjectType,
+  GraphQLUnionType,
 } from '../../type/definition';
 
 import { execute } from '../execute';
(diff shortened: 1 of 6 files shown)


src/utilities/coerceInputValue.js
```diff
@@ -126,7 +126,7 @@ function coerceInputValueImpl(
         fieldValue,
         field.type,
         onError,
-        addPath(path, field.name),
+        addPath(path, field.name, undefined),
```
IvanGoncharov, line 129: Actually here you have access to `type.name` so you can put it in the `Path`.

benjie: I considered that, but I'm not sure that inputs should set `parentType`: since we don't know what the polymorphic input solutions are going to be yet, letting people rely on `parentType` existing is risky. Besides, the path is only used in errors and that doesn't currently use `parentType`, so adding it offers little or no benefit, IMO.

If you're sure, I'm happy to add it, but I was deliberate in not adding it to input types.

benjie: I've added it as the final commit so you can pop it or add it as you see fit.

benjie:
Ready for review; I've added a test that tests a union/list too.

src/utilities/coerceInputValue.js
IvanGoncharov: @benjie Any input union solution will result in exactly one `parentType` if it's not like that we would have a bigger breaking change anyway.
Leaving `parentType` unset will be confusing for people accessing path.

IvanGoncharov: And it would be mean we will need two separate types for execution path and for path inside input values.

IvanGoncharov: Scratch it, it's not worth to add functionality that no one asked for just to unify types.
It actually makes sense to have separate `ExecutionPath` type just for execution.

src/jsutils/Path.d.ts
```diff
@@ -1,12 +1,17 @@
 export interface Path {
   prev: Path | undefined;
   key: string | number;
+  parentType: string | undefined;
```
IvanGoncharov, line 4: I'm not sure about `parentType` name since it's too tied `info` object.
I would like something that you can easily understand even if you never looked into `info`.
I actually think you touch a bigger topic here since you can't identify resolver that was triggered just from `path` so I feel like this structure would be useful for logging, debugging, and other similar purposes and in a future figure how to add it inside `GraphQLError` instead of array path.

How about `typename` (we can't use underscores since it will look like private field)?
It will also solve the confusion for JS users since context `parentType` is a `GraphQLOutputType` and here it's a string and I don't really like `parentTypeName` as an alternative.

IvanGoncharov: it will also explain why we don't have it on an array segment of the path since you will get `__typename` only on this levels in your response.

src/execution/__tests__/executor-test.js
```diff
@@ -306,11 +307,65 @@ describe('Execute: Handles basic execution tasks', () => {
...
+            path = info.path;
+          },
+        },
+      },
+    });
+    const testUnion = new GraphQLUnionType({
+      name: 'TestUnion',
+      types: [a],
+      resolveType() {
+        return a;
```
IvanGoncharov, line 332: Just do 
```suggestion
        return 'A';
```

src/execution/__tests__/executor-test.js
```diff
@@ -306,11 +307,65 @@ describe('Execute: Handles basic execution tasks', () => {
...
+      },
+    });
+    const testType = new GraphQLObjectType({
+      name: 'Test',
+      fields: {
+        test: {
+          type: new GraphQLNonNull(
+            new GraphQLList(new GraphQLNonNull(testUnion)),
+          ),
+          resolve(_val, _args, _ctx, _info) {
```
IvanGoncharov, line 342: Since you have `resolveType` you don't really need this resolver.
Let's keep tests minimal.

src/execution/__tests__/executor-test.js
```diff
@@ -306,11 +307,65 @@ describe('Execute: Handles basic execution tasks', () => {
...
+            return [{ type: 'A' }];
+          },
+        },
+      },
+    });
+    const schema = new GraphQLSchema({ query: testType });
+
+    const document = parse('query { l1: test { ... on A { l2: test } } }');
```
IvanGoncharov, line 350: this is pretty complex query so it deserves to be written as multi-line similar to some of the tests below in this file.

src/execution/__tests__/executor-test.js
```diff
@@ -306,11 +307,65 @@ describe('Execute: Handles basic execution tasks', () => {
...
-      path: { prev: undefined, key: 'result' },
+      path: { prev: undefined, key: 'result', parentType: 'Test' },
       variableValues: { var: 'abc' },
     });
   });
 
+  it('populates path correctly with complex types', () => {
+    let path;
+    const a = new GraphQLObjectType({
+      name: 'A',
```
IvanGoncharov, line 318:
```suggestion
      name: 'SomeType',
```

src/execution/__tests__/executor-test.js
```diff
@@ -306,11 +307,65 @@ describe('Execute: Handles basic execution tasks', () => {
...
+        test: {
+          type: GraphQLString,
+          resolve(_val, _args, _ctx, info) {
+            path = info.path;
+          },
+        },
+      },
+    });
+    const testUnion = new GraphQLUnionType({
+      name: 'TestUnion',
```
IvanGoncharov, line 329:
```suggestion
      name: 'SomeUnion',
```
We slowly migrating all the tests to use same naming convention for test types: `SomeObject`, `AnotherObject`, `SomeUnion`, `SomeScalar`, etc.

IvanGoncharov: and the same goes for variable names `someUnion`.

src/utilities/coerceInputValue.js
IvanGoncharov: Thinking more about it, it's actually useful to be able to reference input field definition as part of the error message. So yes let's keep it.
Sorry for all the back and forth.

benjie: No worries; it's valuable to see your thought process. I'll keep it :+1:

src/execution/__tests__/executor-test.js
benjie: I will, but could you explain why? To me it seems that it'd be more expensive, because now you have to look up the type by name?

src/execution/__tests__/executor-test.js
benjie: If we don't have a resolver here, then GraphQL will surely throw an error since the `test` field will not return anything? I could keep the resolver but change it to `[{}]` if you like?

benjie: Oh; we could use rootValue. I'll do that :+1:

src/jsutils/Path.d.ts
benjie: Done

benjie:
@IvanGoncharov Just a ping in case the "request review" above didn't hit your radar.

src/execution/__tests__/executor-test.js
```diff
@@ -306,11 +307,70 @@ describe('Execute: Handles basic execution tasks', () => {
...
+      query {
+        l1: test {
+          ... on SomeObject {
+            l2: test
+          }
+        }
+      }
+    `);
+
+    execute({ schema, document, rootValue });
```
IvanGoncharov, line 357:
```suggestion
    executeSync({ schema, document, rootValue });
```

benjie:
Suggestion merged; I have not corrected all the other places in the file that use `execute` rather than `executeSync`.

PR #2680, opened 2020-06-30
Docs: some language cleanup in readme and API reference

rmcteggart-r7:
Noticed a small grammar mistake. I went to fix it and then noticed some more so I'm including them also. Cleaner language hopefully makes for easier understanding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Looking for help? Find resources [from the community](https://graphql.org/commun
 
 ## Getting Started
 
-An overview of GraphQL, in general, is available in the
+A general overview of GraphQL is available in the
 [README](https://github.com/graphql/graphql-spec/blob/master/README.md) for the
 [Specification for GraphQL](https://github.com/graphql/graphql-spec). That overview
 describes a simple set of GraphQL examples that exist as [tests](src/__tests__)
@@ -141,7 +141,7 @@ GraphQL.js is [MIT-licensed](./LICENSE).
 
 ### Credits
 
-The `*.d.ts` files in this project are based on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/54712a7e28090c5b1253b746d1878003c954f3ff/types/graphql). Definitions written by:
+The `*.d.ts` files in this project are based on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/54712a7e28090c5b1253b746d1878003c954f3ff/types/graphql) definitions written by:
 
 <!--- spell-checker:disable -->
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ Looking for help? Find resources [from the community](https://graphql.org/commun
 
 ## Getting Started
 
-An overview of GraphQL in general is available in the
+An overview of GraphQL, in general, is available in the
```
danielrearden, line 16: issue: This changes the meaning of the original sentence.

"an overview of GraphQL in general" is equivalent to "a general overview of GraphQL". By adding the commas, we are changing the meaning to "an overview of GraphQL, in most situations".

The original sentence is syntactically valid, although changing it to "A general overview of GraphQL" would be more succinct and clearer.

README.md
```diff
@@ -142,7 +141,7 @@ GraphQL.js is [MIT-licensed](./LICENSE).
 
 ### Credits
 
-The `*.d.ts` files in this project are based on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/54712a7e28090c5b1253b746d1878003c954f3ff/types/graphql) definitions written by:
+The `*.d.ts` files in this project are based on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/54712a7e28090c5b1253b746d1878003c954f3ff/types/graphql). Definitions written by:
```
danielrearden, line 144: issue: This changes the original meaning of the sentence.

"DefinitelyTyped" describes what definitions the files are based on.

docs/APIReference-Language.md
```diff
@@ -106,9 +106,9 @@ export class Source {
 ```
 
 A representation of source input to GraphQL. The name is optional,
-but is mostly useful for clients who store GraphQL documents in
+but is most useful for clients who store GraphQL documents in
```
danielrearden, line 109: suggestion: Either word would be correct, but they convey different meanings ("generally" vs "to the greatest degree").

I think dropping the word altogether would make the sentence clearer and make the use of "but" justifiable. We should also either drop the comma before "but" or add "it" after "but". A comma is only used if joining two independent clauses. I think this would be fine:

> The name is optional, but it is useful...

docs/APIReference-Language.md
```diff
@@ -233,7 +233,7 @@ instead provide functions named the same as the kinds of AST nodes, or
 enter/leave visitors at a named key, leading to four permutations of
 visitor API:
```
danielrearden, line 234: suggestion: "the visitor API"

README.md
rmcteggart-r7: Agreed. Updated!

README.md
rmcteggart-r7: Whoops read this differently first time. Fixed

docs/APIReference-Language.md
rmcteggart-r7: Agreed on omitting 👍 . Also updated

PR #2684, opened 2020-07-02
fix: Permit `isDeprecated` field in presence of `deprecationReason`

trevor-scheer:
Currently, the typings for a `GraphQLField` and `GraphQLEnumValue` are at odds with their runtime expectations.

The typings require the existence of `isDeprecated`.
https://github.com/graphql/graphql-js/blob/13ab792895c7056186fddee552d0db9776c735a3/src/type/definition.js#L1002
https://github.com/graphql/graphql-js/blob/13ab792895c7056186fddee552d0db9776c735a3/src/type/definition.js#L1470

However, at runtime, we throw an error if `isDeprecated` exists.
https://github.com/graphql/graphql-js/blob/13ab792895c7056186fddee552d0db9776c735a3/src/type/definition.js#L838-L841
https://github.com/graphql/graphql-js/blob/13ab792895c7056186fddee552d0db9776c735a3/src/type/definition.js#L1431-L1434

This PR proposes that we _permit_ the presence of that field as long as `deprecationReason` is also provided. Alternatively, we could update the types and make `isDeprecated` optional.

I'm currently running into this error in a PR to the `apollo-tooling` repo to migrate to v15:
https://github.com/apollographql/apollo-tooling/pull/1743

### DIFF
--- a/src/type/__tests__/definition-test.js
+++ b/src/type/__tests__/definition-test.js
@@ -399,6 +399,22 @@ describe('Type System: Objects', () => {
     );
   });
 
+  it('allows an Object type with an isDeprecated field if a deprecationReason field is present', () => {
+    const OldObject = new GraphQLObjectType({
+      name: 'OldObject',
+      // $DisableFlowOnNegativeTest
+      fields: {
+        field: {
+          type: ScalarType,
+          isDeprecated: false,
+          deprecationReason: undefined,
+        },
+      },
+    });
+
+    expect(() => OldObject.getFields()).not.to.throw();
+  });
+
   it('rejects an Object type with incorrectly typed interfaces', () => {
     const objType = new GraphQLObjectType({
       name: 'SomeObject',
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
https://github.com/graphql/graphql-js/blob/13ab792895c7056186fddee552d0db9776c735a3/src/type/definition.js#L1458-L1474

@trevor-scheer For creating types you should use only `GraphQ*Config` types and they deliberately missing `isDeprecated`. We plan to remove `isDeprecated` completely in v16 (except of introspection) so removing it from config types is part of our deprecation plan and was done in `15.0.0` as a breaking change.

trevor-scheer:
For anyone else that may run into this issue, here is my commit which resolved the issue. Exactly as @IvanGoncharov explained. Thank you!

[`e9db99d` (#1743)](https://github.com/apollographql/apollo-tooling/pull/1743/commits/e9db99de5b5a4e128aee7409c5c58d7f4f5833ed)

PR #2688, opened 2020-07-05
Added new 'FormattedExecutionResult' type

IvanGoncharov:


src/execution/index.js
```diff
@@ -8,6 +8,11 @@ export {
   defaultFieldResolver,
   defaultTypeResolver,
 } from './execute';
-export type { ExecutionArgs, ExecutionResult } from './execute';
+
+export type {
```
GreenGremlin, line 12: I'm curious, why do these types need to be exported from `index.js` if they are already exported from `index.d.ts`?

PR #2711, opened 2020-07-15
Add failing test in findBreakingChanges-test.js

DrewML:
Not sure if I'm overlooking some cases, so I'm starting with a failing test to help demonstrate the proposed change.

The output of the test in this PR is:

```sh
1) findBreakingChanges
    should not flag args changing from String or Int to ID as breaking:

    AssertionError: expected [ Array(2) ] to deeply equal []
    + expected - actual

    -[
    -  {
    -    "description": "Type1.field1 arg arg1 has changed type from String to ID."
    -    "type": "ARG_CHANGED_KIND"
    -  }
    -  {
    -    "description": "Type1.field2 arg arg1 has changed type from Int to ID."
    -    "type": "ARG_CHANGED_KIND"
    -  }
    -]
    +[]
```

Looking through the commit history of `findBreakingChanges`, it seems like the feature was designed to follow the spirit of "breaking changes" as mentioned in the spec:

> As GraphQL type system schema evolve over time by adding new types and new fields, it is possible that a request which was previously valid could later become invalid. Any change that can cause a previously valid request to become invalid is considered a breaking change

If an argument's type is changed from either a `String` or `Int` to `ID`, a query providing either of those input types should still pass validation, and be capable of executing successfully (assuming the field resolve functions adjusts for input coercion).

### DIFF
--- a/src/utilities/__tests__/findBreakingChanges-test.js
+++ b/src/utilities/__tests__/findBreakingChanges-test.js
@@ -562,6 +562,24 @@ describe('findBreakingChanges', () => {
     expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([]);
   });
 
+  it('should not flag args changing from String or Int to ID as breaking', () => {
+    const oldSchema = buildSchema(`
+      type Type1 {
+        field1(arg1: String): Int
+        field2(arg1: Int): Int
+      }
+    `);
+
+    const newSchema = buildSchema(`
+      type Type1 {
+        field1(arg1: ID): Int
+        field2(arg1: ID): Int
+      }
+    `);
+
+    expect(findBreakingChanges(oldSchema, newSchema)).to.deep.equal([]);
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@DrewML It's breaking for the case where your clients pass it as query variable since it's explicitly marked as `String`:
```graphql
query ($id: String) {
  field1(arg1: $id)
}
```
changing it to string will break this rule
http://spec.graphql.org/draft/#sel-IALbLDHCJCAACIBohU
> Validation failures occur when variables are used in the context of types that are complete mismatches, or if a nullable type in a variable is passed to a non‐null argument type.

For context it was discussion about changing the spec to allow this use case:
https://github.com/graphql/graphql-spec/issues/526

DrewML:
Thanks @IvanGoncharov! I definitely overlooked that 😄. Makes total sense though.

PR #2712, opened 2020-07-15
Support returning async iterables from resolver functions

robrichard:
Returning AsyncIterables in resolver methods is useful when `@stream` is supported and can be added without any breaking changes. This is split out from https://github.com/graphql/graphql-js/pull/2319

### DIFF
--- a/src/execution/execute.js
+++ b/src/execution/execute.js
@@ -1,4 +1,5 @@
 import arrayFrom from '../polyfills/arrayFrom';
+import { SYMBOL_ASYNC_ITERATOR } from '../polyfills/symbols';
 
 import type { Path } from '../jsutils/Path';
 import type { ObjMap } from '../jsutils/ObjMap';
@@ -8,6 +9,7 @@ import memoize3 from '../jsutils/memoize3';
 import invariant from '../jsutils/invariant';
 import devAssert from '../jsutils/devAssert';
 import isPromise from '../jsutils/isPromise';
+import isAsyncIterable from '../jsutils/isAsyncIterable';
 import isObjectLike from '../jsutils/isObjectLike';
 import isCollection from '../jsutils/isCollection';
 import promiseReduce from '../jsutils/promiseReduce';
@@ -851,6 +853,48 @@ function completeValue(
 }
 
 /**
+ * Complete a async iterator value by completing the result and calling
+ * recursively until all the results are completed.
+ */
+function completeAsyncIteratorValue(
+  exeContext: ExecutionContext,
+  itemType: GraphQLOutputType,
+  fieldNodes: $ReadOnlyArray<FieldNode>,
+  info: GraphQLResolveInfo,
+  path: Path,
+  index: number,
+  completedResults: Array<mixed>,
+  iterator: AsyncIterator<mixed>,
+): Promise<$ReadOnlyArray<mixed>> {
+  const fieldPath = addPath(path, index, undefined);
+  return iterator.next().then(
+    ({ value, done }) => {
+      if (done) {
+        return completedResults;
+      }
+      completedResults.push(
+        completeValue(exeContext, itemType, fieldNodes, info, fieldPath, value),
+      );
+      return completeAsyncIteratorValue(
+        exeContext,
+        itemType,
+        fieldNodes,
+        info,
+        path,
+        index + 1,
+        completedResults,
+        iterator,
+      );
+    },
+    (error) => {
+      completedResults.push(null);
+      handleFieldError(error, fieldNodes, fieldPath, itemType, exeContext);
+      return completedResults;
+    },
+  );
+}
+
+/**
  * Complete a list value by completing each item in the list with the
  * inner type
  */
@@ -862,6 +906,23 @@ function completeListValue(
   path: Path,
   result: mixed,
 ): PromiseOrValue<$ReadOnlyArray<mixed>> {
+  const itemType = returnType.ofType;
+
+  if (isAsyncIterable(result)) {
+    const iterator = result[SYMBOL_ASYNC_ITERATOR]();
+
+    return completeAsyncIteratorValue(
+      exeContext,
+      itemType,
+      fieldNodes,
+      info,
+      path,
+      0,
+      [],
+      iterator,
+    );
+  }
+
   if (!isCollection(result)) {
     throw new GraphQLError(
       `Expected Iterable, but did not find one for field "${info.parentType.name}.${info.fieldName}".`,
@@ -870,7 +931,6 @@ function completeListValue(
 
   // This is specified as a simple map, however we're optimizing the path
   // where the list contains no Promises by avoiding creating another Promise.
-  const itemType = returnType.ofType;
   let containsPromise = false;
   const completedResults = arrayFrom(result, (item, index) => {
     // No need to modify the info object containing the path,
--- a/src/subscription/subscribe.js
+++ b/src/subscription/subscribe.js
@@ -1,6 +1,5 @@
-import { SYMBOL_ASYNC_ITERATOR } from '../polyfills/symbols';
-
 import inspect from '../jsutils/inspect';
+import isAsyncIterable from '../jsutils/isAsyncIterable';
 import { addPath, pathToArray } from '../jsutils/Path';
 
 import { GraphQLError } from '../error/GraphQLError';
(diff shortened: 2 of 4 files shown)


src/subscription/subscribe.js
```diff
@@ -160,7 +159,7 @@ function subscribeImpl(
     // Note: Flow can't refine isAsyncIterable, so explicit casts are used.
     isAsyncIterable(resultOrStream)
       ? mapAsyncIterator(
-          ((resultOrStream: any): AsyncIterable<mixed>),
+          resultOrStream,
```
robrichard, line 162: @IvanGoncharov since adding `boolean %checks(value instanceof AsyncIterable)`  to `isAsyncIterable` I was able to remove the cast here, but I was not able to remove it from the other side of the ternary.

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/subscription/subscribe.js:166:9

Cannot call sourcePromise.then with function bound to onFulfill because mixed [1] is incompatible
with object type [2] in type argument Yield [3] of the return value of property @@asyncIterator of
the return value. [incompatible-call]

     src/subscription/subscribe.js
 [2] 117│ ): Promise<AsyncIterator<ExecutionResult> | ExecutionResult> {
        :
     163│           mapSourceToResponse,
     164│           reportGraphQLError,
     165│         )
     166│       : resultOrStream,
     167│   );
     168│ }
     169│
        :
 [1] 206│ ): Promise<AsyncIterable<mixed> | ExecutionResult> {

     /private/tmp/flow/flowlib_5bfb8b1/core.js
 [3] 562│ interface $AsyncIterator<+Yield,+Return,-Next> {
```

src/__tests__/starWarsSchema.js
```diff
@@ -170,6 +170,29 @@ const humanType = new GraphQLObjectType({
...
+        'The friends of the droid, or an empty list if they have none. Returns an AsyncIterable',
+      args: {
+        errorIndex: { type: GraphQLInt },
+      },
+      async *resolve(droid, { errorIndex }) {
+        const friends = getFriends(droid);
+        let i = 0;
+        for (const friend of friends) {
+          // eslint-disable-next-line no-await-in-loop
+          await new Promise((r) => setTimeout(r, 1));
```
IvanGoncharov, line 185: Why do you need `setTimeout`?
A promise is always processed on a next tick even if it already resolved.

src/__tests__/starWarsSchema.js
```diff
@@ -170,6 +170,29 @@ const humanType = new GraphQLObjectType({
...
+          if (i === errorIndex) {
+            throw new Error('uh oh');
+          }
+          yield friend;
+          i++;
+        }
+        // close iterator asynchronously
+        await new Promise((r) => setTimeout(r, 1));
```
IvanGoncharov, line 193: same here.

src/__tests__/starWarsSchema.js
robrichard: Leftover from when I was trying to time results for tests combining `@stream` and `@defer`. Not needed any more and removed from this PR

src/__tests__/starWarsSchema.js
robrichard: removed

src/__tests__/starWarsSchema.js
```diff
@@ -170,6 +170,29 @@ const humanType = new GraphQLObjectType({
         'The friends of the human, or an empty list if they have none.',
       resolve: (human) => getFriends(human),
     },
+    friendsAsync: {
+      type: GraphQLList(characterInterface),
+      description:
+        'The friends of the droid, or an empty list if they have none. Returns an AsyncIterable',
+      args: {
+        errorIndex: { type: GraphQLInt },
```
IvanGoncharov, line 178: It's better to move non-trivial tests here: https://github.com/graphql/graphql-js/blob/master/src/execution/__tests__/lists-test.js

Starwars schema is intended as just a sample schema to quickly demonstrate all basic features.

src/execution/execute.js
```diff
@@ -916,6 +918,58 @@ function completeValue(
   );
 }
 
+/**
+ * Complete a async iterable value by completing each item in the list with
+ * the inner type
+ */
+
```
IvanGoncharov, line 925: Extra line

src/execution/execute.js
```diff
@@ -916,6 +918,58 @@ function completeValue(
...
+  let index = 0;
+
+  const itemType = returnType.ofType;
+
+  const handleNext = () => {
+    const fieldPath = addPath(path, index);
+    return iterator.next().then(
+      ({ value, done }) => {
+        if (done) {
+          return;
```
IvanGoncharov, line 948: You can return `completedResults` here right?
And save one tick.

src/execution/execute.js
```diff
@@ -916,6 +918,58 @@ function completeValue(
...
+  // $FlowFixMe
+  const iteratorMethod = result[SYMBOL_ASYNC_ITERATOR];
+  const iterator = iteratorMethod.call(result);
+
+  const completedResults = [];
+  let index = 0;
+
+  const itemType = returnType.ofType;
+
+  const handleNext = () => {
```
IvanGoncharov, line 943: Usually, if we don't need `this` we use standard `function` syntax and put functions after return of the main function.

src/execution/execute.js
```diff
@@ -916,6 +918,58 @@ function completeValue(
...
+    const fieldPath = addPath(path, index);
```
IvanGoncharov, line 944: Can you please check if you need to add third argument.

src/__tests__/starWarsSchema.js
robrichard: Removed these tests and added new tests to `lists-test.js`

src/execution/execute.js
robrichard: Yes, updated to return `completedResults` here and in the error handler.

src/execution/execute.js
robrichard: I refactored this a little bit so the `handleNext` function is no longer required. We do not need to create inline functions which relying on closures any more.

src/execution/execute.js
robrichard: I'm not sure I understand what the third argument is for and when it should be passed. I updated it to pass `undefined` as that's what's done currently for list values: https://github.com/graphql/graphql-js/blob/6012d28b3b1fe8e002eefed9a3f013d4b352f08d/src/execution/execute.js#L944

robrichard:
@IvanGoncharov this is ready for another review whenever you have some time.

IvanGoncharov:
@robrichard Had a surgery so was offline for the last few weeks.
Now returning back to open-source so it will take me a few days to get back to this PR.

robrichard:
@IvanGoncharov no rush, please take your time

IvanGoncharov:
@robrichard I merged some changes to `subscribe` can you please rebase it?
It should pretty simple.

robrichard:
@IvanGoncharov this is up to date now

src/execution/execute.js
```diff
@@ -862,6 +906,24 @@ function completeListValue(
   path: Path,
   result: mixed,
 ): PromiseOrValue<$ReadOnlyArray<mixed>> {
+  const itemType = returnType.ofType;
+
+  if (isAsyncIterable(result)) {
+    const iteratorMethod = result[SYMBOL_ASYNC_ITERATOR];
+    const iterator = iteratorMethod.call(result);
```
IvanGoncharov, line 913: @robrichard Can we do?
```suggestion
    const iterator = result[SYMBOL_ASYNC_ITERATOR](result);
```

robrichard: Yes, updated.

IvanGoncharov:
@robrichard Thanks for your patience and sorry for such long delay.
I reviewed this PR in more detail and suspect we not handling a few cases like: `completeValue` throwing or returning a promise, etc.
Also, I feel like we need to figure out a way to unify more code between iterating async iterable and other types of collections.
In the next few days, I will try to add as much as possible missing test cases and we can proceed from there.

IvanGoncharov:
@robrichard I rebased this PR against `master` and added trivial commit on top of it.
To allow both of us to submit and review each other PRs I created branch `asynciteratable` and merge it there.

Issue #2717, opened 2020-07-23
introspectionQuery removed from library but not documentation


dobesv:
This constant was removed in https://github.com/graphql/graphql-js/commit/6b550192e0ce3b458abaec92e4c65d36eeafb7d0 but it still documented in https://github.com/graphql/graphql-js/blob/master/docs/APIReference-Utilities.md#introspectionquery

The documentation should be updated to describe the new way of doing this - using `getIntrospectionQuery` I suppose?

Issue #2503, opened 2020-04-02
Error: Must provide Source. Received: undefined after upgrade to graphql@15.0


Fi1osof:
Title.

Downgrade to graphql@14.6.0 solve problem.

ssypi:
Same here, not sure if the problem is in graphql-tools though as it seems to begin from there.

`
Error: Must provide Source. Received: undefined.
    at devAssert (node_modules/graphql/jsutils/devAssert.js:13:11)
    at new Parser (node_modules/graphql/language/parser.js:82:67)
    at Object.parse (node_modules/graphql/language/parser.js:35:16)
    at Object.<anonymous> (node_modules/graphql-tools/dist/stitching/introspectSchema.js:39:42)

`
at `graphql-tools/dist/stitching/introspectSchema.js:39:42`

var graphql_1 = require("graphql");
var linkToFetcher_1 = require("./linkToFetcher");
var parsedIntrospectionQuery = graphql_1.parse(**graphql_1.introspectionQuery**);

It seems that the bolded part graphql_1.introspectionQuery is undefined which then causes the error at node_modules/graphql/language/parser.js:82:67)

yaacovCR:
Should be fixed in graphql-tools v4.0.7. (https://github.com/apollographql/graphql-tools/blob/master/CHANGELOG.md)

Which version are you using?

ssypi:
I


> Should be fixed in graphql-tools v4.0.7. (https://github.com/apollographql/graphql-tools/blob/master/CHANGELOG.md)
> 
> Which version are you using?

I was on latest v5 alpha, I’ll try latest stable tomorrow.

mike-marcacci:
This is almost certainly due to multiple different versions of `graphql` in your dependency tree. You can look for this by running either:

```bash
# using yarn
yarn why graphql

# using npm
npm ls graphql
```

While GraphQL 15 _is_ a major change, you may find that the libraries you use that have a stated dependency of 14 will still work with the new version. If this is the case, you can force the use of version 15 everywhere using [yarn's resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) or [npm-force-resolutions](https://www.npmjs.com/package/npm-force-resolutions). However, the real solution is to make sure all libraries are updated to use the newest version of `graphql`.

ssypi:
> I
> 
[...]

PR #2718, opened 2020-07-23
introspectionQuery -> getIntrospectionQuery

dobesv:
`introspectionQuery` has been deleted and replaced with `getIntrospectionQuery`

Fixes https://github.com/graphql/graphql-js/issues/2717

See also https://github.com/graphql/graphql-js/issues/2503

### DIFF
--- a/docs/APIReference-Utilities.md
+++ b/docs/APIReference-Utilities.md
@@ -21,9 +21,9 @@ _Introspection_
 
 <ul class="apiIndex">
   <li>
-    <a href="#introspectionquery">
-      <pre>var introspectionQuery</pre>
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@dobesv Thanks for the PR 👍

JasonKleban:
Still at https://graphql.org/graphql-js/utilities/

image

dobesv:
@JasonKleban maybe you can open another PR to fix that one?

saihaj:
I think the docs are not automatically synced with [`graphql/graphql.github.io`](
graphql.github.io) so these should be updated in there https://github.com/graphql/graphql.github.io/tree/source/src/content/graphql-js

PR #2720, opened 2020-07-25
NoDeprecatedCustomRule-test: simplify test cases

IvanGoncharov:


src/validation/__tests__/NoDeprecatedCustomRule-test.js
```diff
@@ -8,112 +8,127 @@ import { NoDeprecatedCustomRule } from '../rules/custom/NoDeprecatedCustomRule';
...
-}
+function buildAssertion(sdlStr: string) {
+  const schema = buildSchema(sdlStr);
+  return { expectErrors, expectValid };
 
-const schema = buildSchema(`
-  enum EnumType {
-    NORMAL_VALUE
-    DEPRECATED_VALUE @deprecated(reason: "Some enum reason.")
-    DEPRECATED_VALUE_WITH_NO_REASON @deprecated
```
IvanGoncharov, line 27: Setting up default `deprecationReason` is tested in `buildASTSchema-test.js` so no need to duplicate it here.

PR #2726, opened 2020-08-02
Fix typo README.md

gejimayu:
Revision:

Please don't forget to set `NODE_ENV=production` if you are running a production server. It will disable some checks that can be useful during development but will significantly ~~improve~~ **impact** performance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ graphql(schema, query).then((result) => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@gejimayu Thanks for the PR 👍 
But why is `improve` is problematic?

gejimayu:
@IvanGoncharov oops sorry to waste your time. it seems like i misunderstood it. i thought `it` refers to not setting `NODE_ENV=production`

Issue #2727, opened 2020-08-03
Allow result type and variables type to be inferred from `DocumentNode`/`Source` using TypeScript type system


dotansimha:
Hi!

We recently implemented an improved version of `DocumentNode`, called `TypedDocumentNode`, that allow type generics for result type and variables type:

```ts
export interface TypedDocumentNode<Result = { [key: string]: any }, Variables = { [key: string]: any }> extends DocumentNode {}
```

This improved version allows TypeScript to infer the types automatically when an object that matches the signature is used, for example:

```ts
const typedDocument: TypedDocumentNode<{ test: string }, { v: string }> = parse(`query($v: String) { test(v: $v) }`);

// Now `result` will be automatically of type `{ test: string }`
const result = await execute({
  // ...
  document: typedDocument,
  // variableValues will be typed to `{ v: string }` automatically
  variableValues: { v: 'test' },
});
```

This will allow to have the types set within the `DocumentNode` and not externally as used today, and avoid incorrect TypeScript types used. 
This made possible because TypeScript does [type inference](https://www.typescriptlang.org/docs/handbook/type-inference.html) when possible. 

The existing solution is implemented in this repo: https://github.com/dotansimha/graphql-typed-document-node and it's currently implemented by patching `graphql` library and add support to it. I think this concept could be merged into this PR, just like we allow `extensions` or other to be typed with generics.

Developers who chose to automate the typings creation, can use [GraphQL Code Generator](https://graphql-code-generator.com/) to generate a pre-compiled `DocumentNode<T, R>` from their operations, the output file looks like that: 

https://github.com/dotansimha/graphql-typed-document-node/blob/master/examples/graphql/typed-document-nodes.ts#L45

Adding this to the core implementation of `graphql` won't break anything, since it effects only TypeScript types, and the generics are mostly there today, so it's just a few non-breaking adjustments to support this. 

To have it fully supported across `graphql` and `execute` method, it needs to be added to `Source` and `DocumentNode`. I created a PR here: https://github.com/graphql/graphql-js/pull/2728

Related:
- https://github.com/graphql/graphql-js/pull/2490#issuecomment-639154229
- https://github.com/graphql/graphql-js/issues/2104 

@IvanGoncharov

benjamn:
[...]

PR #2728, opened 2020-08-03
Allow to pass generics to `Source` and `DocumentNode` to allow TS type inference

dotansimha:
Resolves https://github.com/graphql/graphql-js/issues/2727


dotansimha:
Hi @IvanGoncharov , @leebyron  how can we progress with this one?  thanks!

dotansimha:
Thank you @IvanGoncharov  for the review! 

@ForbesLindesay suggested (https://github.com/dotansimha/graphql-typed-document-node/pull/34) to add internal types to `DocumentNode` to make sure TypeScript treats it as structural type and not removing the generics because they are unused. 
I'm fine with having another type for that (`TypedDocumentNode extends DocumentNode`) but to be honest, it will be kind of useless, especially because we can type the generics with `any` and it won't effect anything else. 

Regarding `execute`/`graphql` functions:
1. Type inference won't work without adding the generics to the entire chain of calls. If `execute` won't have the generics, it won't get inferred later based on `DocumentNode` passed to it. 
2. I do understand how `@defer` could effect it, but just like `@skip` and `@include` - this effect types only, and probably just make it optional. 
3. `graphql-codegen` will handle the `@defer` when it will be available in `graphql-js` and will allow developers to reflect the effect of it on generated types. But please note that `graphql-codegen` isn't the only way to use generate types. Developers can manually specify the types.
4. No one expects GraphQL to be responsible for the result and match the generic types. If you allow to pass generics - developers knows that he's in charge of those types, and it's not part of the package itself. 
5. We don't have to change `execute` / `graphql` at all if you wish, but then it will mean that only tools that uses `DocumentNode` will benefit from the inferred types. Also, `graphql-js` is used by many projects locally, and `execute` sometimes is being executed directly with an operation that has been declared locally.

@IvanGoncharov what do you think about the following: I can remove all the changes from `execute`, `graphql` and even from `Source` and limit those changes only for `DocumentNode` to allow generics in it. These will be defaulted to any / record value in order to make sure it's not breaking anything.
That means that types could still be inferred while `DocumentNode` is being used by other libraries / graphql clients.

dotansimha:
@IvanGoncharov I updated the PR with the minimal changes possible, in order to support that. What do you think?

IvanGoncharov:
@dotansimha Thanks 👍 
As I wrote above can we add separate `TypedQueryDocumentNode` type with template parameters inherited from `DocumentNode`?

dotansimha:
Thanks @IvanGoncharov . 
As It wrote before, I think making it a new variable might make it more difficult to adopt in GraphQL client libraries. Since it means they need to modify the codebase to use `TypedDocumentNode` instead of the original `DocumentNode`. 

Are you open to keep it as part of the original `DocumentNode`? it doesn't break anything, and it fully compatible.

IvanGoncharov:
> As It wrote before, I think making it a new variable might make it more difficult to adopt in GraphQL client libraries.

@dotansimha Sorry missed that in you previous comment.

> I don't think adding a new type called TypedDocumentNode will be the best solution, because then it means we do need to change execute and others to allow DocumentNode | TypedDocumentNode.

Here is a snippet that `execute` should work perfectly fine with a new type and it also shows how to implement wrapper function around `execute` that utilize full potential of this new type:
```
interface DocumentNode {
  someValue: number
}


interface TypedDocumentNode<TArg> extends DocumentNode {
}

function execute(document: DocumentNode): unknown { return null; }

const stringTypedDocume...
```

[Playground Link](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCIHsEFcC2FwBy6AJigN4BQyyAzungGpwA2WEAXMiLgEbQUBfCsNCRYiFABUAngAcIxDNjyESEADySAglADmAPmQQAHpBDEaaTLnxgipZJSEUYWEAjDB0II8YjZIAApiaxUwTiUbVVIASk43AGsQdAB3HzJkKAgwLCgfbmZmAG5kZwRvGjBaMChQXRl5RVDbTgaFSLD7DUrakANkAF5HWnoIJlYOZABWUqKKE38sIJ66tqblWxjhV3dPb2QUqDhZRoBRPwCNSX1g5vBWuXa7uzVNfRjHKkzs3J8Fy9uG3AHzglkkczKFSqNAAFugsMxiAAhCAAZRqdU4Kz6gwORxOCnOi2WGL6aw6myKQA)

PR #2731, opened 2020-08-05
execute/subscribe: simplify to improve debugging experience

IvanGoncharov:
Also move handling for error that not inherit from Error to `locatedError`


IvanGoncharov:
Also gives small perfomance boost:
image

PR #2757, opened 2020-08-25
Support returning async iterables from resolver functions

IvanGoncharov:
Continuation of #2712

### DIFF
--- a/src/execution/__tests__/lists-test.js
+++ b/src/execution/__tests__/lists-test.js
@@ -64,6 +64,47 @@ describe('Execute: Accepts any iterable as list value', () => {
   });
 });
 
+describe('Execute: Accepts async iterables as list value', () => {
+  function complete(rootValue: mixed) {
+    return execute({
+      schema: buildSchema('type Query { listField: [String] }'),
+      document: parse('{ listField }'),
+      rootValue,
+    });
+  }
+
+  it('Accepts an AsyncGenerator function as a List value', async () => {
+    async function* listField() {
+      yield await 'two';
+      yield await 4;
+      yield await false;
+    }
+
+    expect(await complete({ listField })).to.deep.equal({
+      data: { listField: ['two', '4', 'false'] },
+    });
+  });
+
+  it('Handles an AsyncGenerator function that throws', async () => {
+    async function* listField() {
+      yield await 'two';
+      yield await 4;
+      throw new Error('bad');
+    }
+
+    expect(await complete({ listField })).to.deep.equal({
+      data: { listField: ['two', '4', null] },
+      errors: [
+        {
+          message: 'bad',
+          locations: [{ line: 1, column: 3 }],
+          path: ['listField', 2],
+        },
+      ],
+    });
+  });
+});
+
 describe('Execute: Handles list nullability', () => {
   async function complete({ listField, as }) {
     const schema = buildSchema(`type Query { listField: ${as} }`);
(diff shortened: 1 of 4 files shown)


IvanGoncharov:
@robrichard Since ATM we are using a very similar algorithm I'm worried that we will also have memory leak:
https://github.com/leebyron/iterall/issues/24
So working on adding performance tests against it.

src/execution/__tests__/lists-test.ts
```diff
@@ -64,6 +71,146 @@ describe('Execute: Accepts any iterable as list value', () => {
...
+    });
+    return execute({
+      schema,
+      document: parse('{ listField { index } }'),
+    });
+  }
+
+  it('Accepts an AsyncGenerator function as a List value', async () => {
+    async function* listField() {
+      yield await Promise.resolve('two');
```
dnalborczyk, line 119: just a nit, could we remove `await Promise.resolve()` as an `async [generator] function` always returns/yields a Promise. it also does not schedule additional microtasks.

```js
// yields the same 
async function* foo() {
	yield await Promise.resolve(1)
	yield Promise.resolve(1)
	yield 1
}
```

robrichard: I added them because of the lint rule.

```
Async generator function 'listField' has no 'await' expression.eslint@typescript-eslint/require-await
```

I don't mind changing it to ignore the lint rule if that's preferred.

yaacovCR:
@robrichard @IvanGoncharov 

As a first step towards keeping the design space open to accommodate streaming in chunks (discussed during [recent wg](https://github.com/graphql/graphql-wg/blob/main/agendas/2022/2022-02-03.md)), do we want to change the allowed signature for returning async iterables from resolver functions?

Currently, this PR allows you to asynchronously return individual list items, but we could change that to allow asynchronously returning __portions__ (chunks) of the list, with the naive case being chunks of size 1.

Specifically,

### if the field has a return type of: `[Int]`
async resolver return type should be: `AsyncIterable<Iterable<number>>` instead of `AsyncIterable<number>`

### if the field has a return type of: `[[Int]]`
async resolver return type should be: `AsyncIterable<Iterable<Iterable<number>>>` instead of `AsyncIterable<Iterable<number>>`

Thoughts?

yaacovCR:
cf https://github.com/yaacovCR/graphql-executor/pull/154

yaacovCR (changes requested):
My review (requested by @saihaj ) is basically that we should allow resolvers to return async iterables of iterables, not of individual items. See https://github.com/robrichard/defer-stream-wg/discussions/38

src/execution/execute.ts
```diff
@@ -697,6 +698,65 @@ function completeValue(
...
+      }
+    } catch (rawError) {
+      completedResults.push(null);
+      const error = locatedError(rawError, fieldNodes, pathToArray(fieldPath));
+      handleFieldError(error, itemType, exeContext);
+      break;
+    }
+    index += 1;
+  }
+  return containsPromise ? Promise.all(completedResults) : completedResults;
```
yaacovCR, line 757: To my eye, this code does NOT suffer from the subtle issue raised by #3528 (addressed by #3529), i.e. even with a sync error, all promises will be awaited. Perhaps the loop format here can be used to fix #3528 as #3529 seems a bit stuck CLA-wise).

yaacovCR (approved):

yaacovCR:
We decided at last js-wg to give this a few days, but it seems to have slipped. It looks good to me. @IvanGoncharov @saihaj ?

saihaj (approved):

PR #2779, opened 2020-08-31
Replace resolved types in lexicographic schema sort

mattleff:
When using `lexicographicSortSchema()` with a schema containing `resolveType` for either interfaces or unions the resolved type wasn't being replaced with the new `GraphQLObjectType` created during sorting. This PR wraps any `resolveType` functions provided and substitutes the sorted named type for the type returned from the `resolveType` function.

Downstream issue: https://github.com/nestjs/graphql/issues/1107

### DIFF
--- a/src/utilities/lexicographicSortSchema.js
+++ b/src/utilities/lexicographicSortSchema.js
@@ -11,7 +11,6 @@ import type {
   GraphQLFieldConfigMap,
   GraphQLFieldConfigArgumentMap,
   GraphQLInputFieldConfigMap,
-  GraphQLTypeResolver,
 } from '../type/definition';
 import { GraphQLSchema } from '../type/schema';
 import { GraphQLDirective } from '../type/directives';
@@ -75,20 +74,6 @@ export function lexicographicSortSchema(schema: GraphQLSchema): GraphQLSchema {
     return maybeType && replaceNamedType(maybeType);
   }
 
-  function replaceResolvedType(
-    maybeTypeResolver: ?GraphQLTypeResolver<mixed, mixed>,
-  ) {
-    return (
-      maybeTypeResolver &&
-      ((...args) =>
-        Promise.resolve(maybeTypeResolver(...args)).then((resolvedType) =>
-          isObjectType(resolvedType)
-            ? replaceNamedType(resolvedType)
-            : resolvedType,
-        ))
-    );
-  }
-
   function sortDirective(directive: GraphQLDirective) {
     const config = directive.toConfig();
     return new GraphQLDirective({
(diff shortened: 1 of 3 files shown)


src/utilities/lexicographicSortSchema.js
```diff
@@ -74,6 +75,20 @@ export function lexicographicSortSchema(schema: GraphQLSchema): GraphQLSchema {
     return maybeType && replaceNamedType(maybeType);
   }
 
+  function replaceResolvedType(
+    maybeTypeResolver: ?GraphQLTypeResolver<mixed, mixed>,
+  ) {
+    return (
+      maybeTypeResolver &&
+      ((...args) =>
+        Promise.resolve(maybeTypeResolver(...args)).then((resolvedType) =>
```
IvanGoncharov, line 84: An important property of `execute` is that it returns only if at least one resolver return promise.
This change will convert sync schema into async.
Please use `isPromise`

src/utilities/lexicographicSortSchema.js
```diff
@@ -74,6 +75,20 @@ export function lexicographicSortSchema(schema: GraphQLSchema): GraphQLSchema {
...
+          isObjectType(resolvedType)
```
IvanGoncharov, line 85: Just return `resolvedType.name` is should be simpler and faster.
Also by using `replaceNamedType` you making reference to the previous schema so it can't be garbage collected.

IvanGoncharov:
@mattleff Very interesting bug, once review comments are addressed I will merge it as a temporary fix.
But this uncovered generic issue since you can also reproduce with `extendSchema` or any other function that modify schema.
So the correct solution would be to just forbid returning `GraphQLObjectType` instance in the upcoming `16.0.0` so names of types would be the only option.
Do you see any potential problems with these approach?

IvanGoncharov:
@mattleff Thinking more about it it's better to have this temporary workaround directly inside `execute`:
https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js#L1006-L1009
That way it's smaller and fix this problem for any schema transformation instead of just fixing `lexicographicSortSchema`

mattleff:
@IvanGoncharov Ok, I'll see if I can implement it in `execute()`.

mattleff:
@IvanGoncharov I've pushed that change. Are you good with the current tests or is there a better place or method for testing this?

IvanGoncharov:
@mattleff Midnight in my location 🛏️ 
Will look into it tomorrow morning 🌅

IvanGoncharov:
@mattleff Took more time than I expected and resulted in #2790 #2791 #2793
So now we prepared for dropping `GraphQLObject` as a result of `resolveType` in the upcoming 16.0.0
Can you please test it with Next.js?

IvanGoncharov:
@mattleff You can use our NPM branch for tests:

mattleff:
@IvanGoncharov That works great. Thanks for all your work to square this away!

PR #2793, opened 2020-09-05
resolveType: add workaround for transformed schemas

IvanGoncharov:
closes #2779


adrianmxb:
Is there some ETA on when this will be tagged/released?

PR #2804, opened 2020-09-18
Added check for specific symbols in polyfills/symbols

dionisnote:
Added check for specific symbols in polyfills/symbols because in nodejs 8 Symbol exists, but Symbol.asyncIterator is not.
This cause an error 
`
message: "Subscription field must return Async Iterable. Received: { next: [function next], return: [function return], throw: [function throw], @@asyncIterator: [function] }."
`
just because `isAsyncIterable` returns false
Checking for specific symbols in Symbol will fix it


Junoburger (approved):

IvanGoncharov:
@dionisnote Thanks for PR 👍 
Merged but please note that we support only current versions of Node.js: https://nodejs.org/en/about/releases/
And this list doesn't include Node8.

PR #2806, opened 2020-09-24
Update outdated documentation

Tapped:
This method is not named `parseSchemaIntoAST` anymore, but is instead called `parse`.

I'm not sure how the process is, but the error can also be found here https://github.com/graphql/graphql.github.io/blob/9a6e78630448d9235664209a0026c7a19912ce5e/site/graphql-js/APIReference-Utilities.md


IvanGoncharov:
@Tapped Thanks you 👍 
Merged 🎉

PR #2807, opened 2020-09-24
Don't Expose Mabye<T> in typescript def

ctopherhague:
[Pull 2621](https://github.com/graphql/graphql-js/pull/2621#issuecomment-641927660) Removed Maybe<T> from tsutils export, however it still remains in the def for GraphQLError.originalError.

This broke our use case where we have leverage [customFormatErrorFn](https://github.com/graphql/express-graphql/blob/master/src/index.ts#L103) in express, in which we : 

``` Typescript
export function errorHandler(error?: Error) {
	//
}

errorHandler(error.originalError); 
```

Results in:
``` 
Argument of type 'Maybe<Error>' is not assignable to parameter of type 'Error | undefined'.
  Type 'null' is not assignable to type 'Error | undefined'.ts(2345)
```

We could go the route of the [Apollo Fix](https://github.com/apollographql/apollo-server/pull/4230) and re-define maybe in our own package, but if this is doable, would like to simply not expose it in `GraphQLError.d.ts`

### DIFF
--- a/src/error/GraphQLError.d.ts
+++ b/src/error/GraphQLError.d.ts
@@ -72,7 +72,7 @@ export class GraphQLError extends Error {
   /**
    * The original error thrown from a field resolver during execution.
    */
-  readonly originalError: Maybe<Error>;
+  readonly originalError?: Error;
 
   /**
    * Extension fields to add to the formatted error.


IvanGoncharov:
@ctopherhague Maybe is still exported just in different folder `jsutils`.
Your issue is that you need to write `export function errorHandler(error?: Error | null)`.
Note: In 16.0.0 we are planning to do breaking change and remove all `Maybe` wrappers.

PR #2809, opened 2020-09-25
fix(docs): Update readme

iamdarshshah:
### Changes
- [x] Added logo 
- [x] aligned header elements to the center

> This will make it look readable.

#### _Before:_
before

#### _After:_
after


IvanGoncharov:
@iamdarshshah Thanks for the effort but we like to keep our README in pure markdown.

Issue #2110, opened 2019-08-23
Error with flowlint in graphql@14.5.0


HsuTing:
I got the error with flow in `graphql@14.5.0`. I do not get any error in `graphql@14.4.2`. Can someone help me? Or maybe you can remove this rule of `flowlint`?

```
$ /Users/hsuting/Desktop/work/core/node_modules/.bin/flow
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/graphql/type/definition.js.flow:4:13

Redundant argument. This argument doesn't change any lint settings.

     1│ // @flow strict
     2│
     3│ // FIXME
     4│ // flowlint deprecated-type: off
     5│ import objectEntries from '../polyfills/objectEntries';
     6│
     7│ import inspect from '../jsutils/inspect';
```

gajus:
Is there a way to suppress this error?

IvanGoncharov:
@HsuTing @gajus I'm working on `14.5.1` to fix this issue ASAP. ETA < 1hr.

IvanGoncharov:
@HsuTing @gajus Sorry for the delay.
Had unexpected meeting :(
Fixed in `14.5.1` 📦

HsuTing:
@IvanGoncharov 
That is ok. Thank you. 👍

magicmark:
Looks like this has come back?

```
╚═ ♪ yarn flow status
yarn run v1.13.0                                                                                                                                                        
warning package.json: No license field                                                                                                                                  
$ /nail/home/markl/pg/dlcg/examples/swapi/node_modules/.bin/flow status                                                                                                 
Launching Flow server for /nail/home/markl/pg/dlcg/examples/swapi
Spawned flow server (pid=22550)
Logs will go to /tmp/flow/zSnailzShomezSmarklzSpgzSdlcgzSexampleszSswapi.log
Monitor logs will go to /tmp/flow/zSnailzShomezSmarklzSpgzSdlcgzSexampleszSswapi.monitor_log
[...]

PR #2854, opened 2020-11-23
Remove "uninitialized-instance-property" flowlint rule from GraphQLError.js

rheaditi:
Hey Folks!

I ran into the same issue as described [in this comment here](https://github.com/graphql/graphql-js/issues/2110#issuecomment-608111262). Basically, `flow check` throws an error in the `GraphQLError.js.flow` file:

```sh
 $ flow check
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/graphql/error/GraphQLError.js.flow:3:13

Redundant argument. This argument doesn't change any lint settings. [lint-setting]

     1│ // @flow strict
     2│ // FIXME:
     3│ // flowlint uninitialized-instance-property:off
     4│
     5│ import isObjectLike from '../jsutils/isObjectLike';
     6│ import { SYMBOL_TO_STRING_TAG } from '../polyfills/symbols';
```

As suggested in [another comment](https://github.com/graphql/graphql-js/issues/2110#issuecomment-644205632), I tried adding the file to both the `[untyped]`/`[ignore]` sections of our `.flowconfig` but it still seems to throw an error in our project.

My best guess is that this rule (`uninitialized-instance-property`) [does not exist in flowlint anymore](https://flow.org/en/docs/linting/rule-reference/) (?) and hence doesn't allow for overriding either.

Going by the above assumption we removed the following two lines from the file manually & this seems to make `flow check` pass for our project:
```js
// FIXME:	
// flowlint uninitialized-instance-property:off
```

Is this lint rule safe to remove from `graphql-js`? 

Thanks :)

### DIFF
--- a/src/error/GraphQLError.js
+++ b/src/error/GraphQLError.js
@@ -1,6 +1,3 @@
-// FIXME:
-// flowlint uninitialized-instance-property:off
-
 import isObjectLike from '../jsutils/isObjectLike';
 import { SYMBOL_TO_STRING_TAG } from '../polyfills/symbols';
 


rheaditi:
Seems like it is required 😅 

Is there a way around this error? Or, with some guidance I could attempt to fix the root-cause instead of just removing the lint rule override comment 🙂

IvanGoncharov:
@rheaditi You need to specify `node_modules` under `declarations` here is example:
https://github.com/graphql/graphql-js/blob/8c6e7c7289882dd6b5bca8feaa435e7e6d18dba3/integrationTests/flow/.flowconfig#L4-L5

PR #2855, opened 2020-11-24
introspection: Add missing support for deprecated input values

### LINKED ISSUE #2834: Breaking change with deprecated fields not being included by default
In [15.4.0](https://github.com/graphql/graphql-js/releases/tag/v15.4.0) release, [this change](https://github.com/graphql/graphql-js/pull/2733/files#diff-ca360cc3700b4c9414922e5f3cff0ee44a71e65bd31862fa5a7b3ac1beac9aaeR341) was merged which prevents deprecated field from being returned when fetching the schema by default. Because these fields were returned before, this is a breaking change for projects that use this to update a local file for their schema.

In our case, the schema is used by a linter and our CI is now blocked because some fields are not in the schema anymore, yet we still use them in some places.

Proposed solutions:
- Change `includeDeprecated`'s `defaultValue` to `true`
- Or mark this change as breaking in the release notes

Happy to contribute to one of these options if the maintainers think it's the way to go. 
Ping @IvanGoncharov since you worked on this recently.

IvanGoncharov:
Fixes #2834

### DIFF
--- a/src/utilities/__tests__/buildClientSchema-test.js
+++ b/src/utilities/__tests__/buildClientSchema-test.js
@@ -486,6 +486,14 @@ describe('Type System: build schema from introspection', () => {
 
   it('builds a schema aware of deprecation', () => {
     const sdl = dedent`
+      directive @someDirective(
+        """This is a shiny new argument"""
+        shinyArg: SomeInputObject
+
+        """This was our design mistake :("""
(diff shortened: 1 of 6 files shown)


dmateiu:
@IvanGoncharov Is there an estimated time of arrival for this fix to be included in a new package release?

As many tools are dependent on the introspection query, I think this is a rather important fix.

Issue #2715, opened 2020-07-22
Single field subscriptions rule has conflicts with the spec


nikis05:
The rule: [https://github.com/graphql/graphql-js/blob/master/src/validation/rules/OverlappingFieldsCanBeMergedRule.js]()

The spec [states](https://spec.graphql.org/June2018/#sec-Single-root-field) that subscriptions should have a single root field. Current implementation only checks that there is one selection on selection set of the subscription, not taking fragments into account. If a single selection is a fragment spread or an inline fragment (with multiple fields) this rule will wrongfully pass.

kshitij86:
I would like to look into this issue and work on a fix. @IvanGoncharov can I please get assigned to solve this bug?

benjie:
I noticed this too; whoever fixes this, please can you ping me on #2861 as I'll need to incorporate the fix into that PR.

benjie:
I may have resolved this bug as part of #2861

yaacovCR:
@benjie can this be closed?

PR #2861, opened 2020-11-25
RFC: Assert subscription field is not introspection.

benjie:
This PR implements the spec changes in the following GraphQL Spec RFC:

https://github.com/graphql/graphql-spec/pull/776

### DIFF
--- a/src/validation/rules/SingleFieldSubscriptionsRule.ts
+++ b/src/validation/rules/SingleFieldSubscriptionsRule.ts
@@ -1,9 +1,79 @@
 import { GraphQLError } from '../../error/GraphQLError';
 
 import type { ASTVisitor } from '../../language/visitor';
-import type { OperationDefinitionNode } from '../../language/ast';
+import type {
+  OperationDefinitionNode,
+  SelectionSetNode,
+} from '../../language/ast';
+import { Kind } from '../../language/kinds';
 
 import type { ASTValidationContext } from '../ValidationContext';
+
+/**
+ * Walks the selection set and returns a list of selections where extra fields
+ * were selected, and selections where introspection fields were selected.
+ */
+function walkSubscriptionSelectionSet(
+  context: ASTValidationContext,
+  selectionSet: SelectionSetNode,
+  responseKeys,
+  fieldNames = new Set(),
+  visitedFragmentNames = new Set(),
+  extraFieldSelections = [],
+  introspectionFieldSelections = [],
+) {
+  for (const selection of selectionSet.selections) {
+    switch (selection.kind) {
+      case Kind.FIELD: {
+        const fieldName = selection.name.value;
+        fieldNames.add(fieldName);
+        const responseName = selection.alias
+          ? selection.alias.value
+          : fieldName;
+        responseKeys.add(responseName);
+        if (fieldName[0] === '_' && fieldName[1] === '_') {
+          // fieldName represents an introspection field if it starts with `__`
+          introspectionFieldSelections.push(selection);
+        } else if (fieldNames.size > 1 || responseKeys.size > 1) {
+          extraFieldSelections.push(selection);
+        }
+        break;
+      }
+      case Kind.FRAGMENT_SPREAD: {
+        const fragmentName = selection.name.value;
+        if (!visitedFragmentNames.has(fragmentName)) {
+          visitedFragmentNames.add(fragmentName);
+          const fragment = context.getFragment(fragmentName);
+          if (fragment) {
+            walkSubscriptionSelectionSet(
+              context,
+              fragment.selectionSet,
+              responseKeys,
+              fieldNames,
+              visitedFragmentNames,
+              extraFieldSelections,
+              introspectionFieldSelections,
+            );
+          }
+        }
+        break;
+      }
+      case Kind.INLINE_FRAGMENT: {
+        walkSubscriptionSelectionSet(
+          context,
+          selection.selectionSet,
+          responseKeys,
+          fieldNames,
+          visitedFragmentNames,
+          extraFieldSelections,
+          introspectionFieldSelections,
+        );
+        break;
+      }
+    }
+  }
+  return [extraFieldSelections, introspectionFieldSelections];
+}
 
 /**
  * Subscriptions must only include a non-introspection field.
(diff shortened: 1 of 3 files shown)


src/validation/rules/SingleFieldSubscriptionsRule.js
```diff
@@ -6,9 +6,10 @@ import type { OperationDefinitionNode } from '../../language/ast';
 import type { ASTValidationContext } from '../ValidationContext';
 
 /**
- * Subscriptions must only include one field.
+ * Subscriptions must only include a non-introspection field.
```
benjie, line 9: This is very precise wording; if we were to use `one non-introspection field` or `a single non-introspection field` here then it wouldn't place limits on the _introspection_ fields we added, only the non-introspection fields.

benjie:
I've updated the PR to overhaul the subscription validation rule to walk fragments too (see #2715).

IvanGoncharov:
@benjie You need to handle recursive fragments:
```graphql
subscription {
  ...A
}

fragment A {
  ...A
}
```
Even thought it's invalid, validation shouldn't be stuck on it

benjie:
@IvanGoncharov I've handled that by tracking `visitedFragmentNames`; but I've added a test with your operation to show this is the case for completeness :+1:

IvanGoncharov (approved):
Tests look great 👍 
Implementation is also looking correct.
I just like to play with the source code of the rule to see if it can be simplified.
Will try to do this on weekends.

src/validation/rules/SingleFieldSubscriptionsRule.js
```diff
@@ -1,28 +1,119 @@
...
+ * A GraphQL subscription is valid only if it contains a single root field and
+ * that root field is not an introspection field.
  */
 export function SingleFieldSubscriptionsRule(
   context: ASTValidationContext,
 ): ASTVisitor {
   return {
     OperationDefinition(node: OperationDefinitionNode) {
       if (node.operation === 'subscription') {
-        if (node.selectionSet.selections.length !== 1) {
```
leebyron, line 19: Happy to also see this validation bug being fixed - fields are collected before asserting a single selection

src/validation/rules/SingleFieldSubscriptionsRule.js
```diff
@@ -1,28 +1,119 @@
...
+import { Kind } from '../../language/kinds';
 
 import type { ASTValidationContext } from '../ValidationContext';
 
 /**
- * Subscriptions must only include one field.
+ * Walks the selection set and returns a list of selections where extra fields
+ * were selected, and selections where introspection fields were selected.
+ */
+function walkSubscriptionSelectionSet(
```
leebyron, line 16: Any reason why this can't be a more typical `CollectFields()` implementation? I'm surprised there isn't something to be reused from another validation rule.

src/validation/rules/SingleFieldSubscriptionsRule.js
```diff
@@ -1,28 +1,119 @@
...
+        const responseName = selection.alias
+          ? selection.alias.value
+          : fieldName;
+        responseKeys.add(responseName);
+        if (fieldName[0] === '_' && fieldName[1] === '_') {
+          // fieldName represents an introspection field if it starts with `__`
+          introspectionFieldSelections.push(selection);
+        } else if (fieldNames.size > 1 || responseKeys.size > 1) {
+          extraFieldSelections.push(selection);
+        }
```
leebyron, line 39: I think this is the only bit that's unique from `CollectFields()` - I'd suggest moving this bit of logic down to the Operation visit function after the grouped fields are collected.

src/validation/rules/SingleFieldSubscriptionsRule.js
benjie: The closest I could find was `_collectFieldsAndFragmentNames` from `src/validation/rules/OverlappingFieldsCanBeMergedRule.js` but it doesn't recurse into named fragments. We also can't use collectFields from execution because it requires an execution context.

My algorithm is more efficient than recursing over calls to `_collectFieldsAndFragmentNames` because it has no need to reference the underlying types - there's no need for calls to `parentType.getFields()`/`typeFromAST`/`isObjectType`/etc since all we want to know is:

1. is there any `__introspection` fields?
2. is there more than one field?

The actual type they're defined on (or even whether it exists or not!) is irrelevant.

benjie: I suppose that could potentially be untrue; e.g.

```graphql
type Query implements Node {...}
type Subscription implements Node {...}

subscription ShouldBeFine($bool: Boolean!) {
  ... on Node {
    ... on Query { __typename id firstName }
    ... on Subscription {
      mySubscriptionField @include(if: $bool)
      myOtherSubscriptionField @skip(if: $bool)
    }
  }
}
```

in this case a runtime `collectFields` would return only `mySubscriptionField` or `myOtherSubscription` field (depending on the value of `$bool`) because the Query selections would be scoped out as incompatible types and the `@include`/`@skip` would ensure only one field was valid. This is problematic though, since in validate we don't have access to the runtime variables so we cannot determine which of the fields will/won't be included.

In the [spec edit](https://github.com/graphql/graphql-spec/pull/776/files) I wrote this leveraging the `CollectFields` call which was already present, which already relies on `varibleValues`; but GraphQL.js allows validation without variables so... is this a fundamental issue in the way GraphQL.js is built? Personally I think we should edit the spec to reflect GraphQL.js' behaviour - validating the query without access to the variables seems desirable and so having the validation rule for "single root field" depend on CollectFields which depends on variableValues is undesirable.

benjie: Reading more carefully the spec explicitly makes variableValues an empty object, so in this case the operation would be valid; however so would the following:

```graphql
type Query implements Node {...}
type Subscription implements Node {...}

subscription ShouldBeFine($bool: Boolean!) {
  ... on Node {
    ... on Query { __typename id firstName }
    ... on Subscription {
      mySubscriptionField1 @include(if: $bool)
      mySubscriptionField2 @include(if: $bool)
      mySubscriptionField3 @skip(if: $bool)
    }
  }
}
```

In this case the CollectFields algorithm would result in only `mySubscriptionField3` being considered; the operation would be valid, but then if `$bool` was set to `true` at runtime there'd be two selections and the operation would be invalid.

This is definitely a spec bug. I'm going to write up a proper fix.

benjie:
To address @leebyron's feedback I've rewritten the implementation to more closely follow the spec - it calls into `collectFields` (and to do so it sets up a fake execution context, similar to how the spec creates the empty variableValues map). This allows the factoring in of `@skip` / `@include` and fragment traversal as the spec currently states (previously this implementation did not fully consider these concerns, being overly strict and rejecting more operations than necessary).

This is the only place in Section 5 of the GraphQL spec that utilises `CollectFields`, so it's not unexpected that the code in GraphQL.js to set this up is a little gross with a bunch of `fake` things. I think this is prefereable to re-implementing the algorithms locally.

src/validation/rules/SingleFieldSubscriptionsRule.js
benjie: (After more thought, I'm not sure it's a bug so much as an "unexpected solution"; but I've filed a potential spec edit anyway: https://github.com/graphql/graphql-spec/pull/860)

Given the above discussion I've rewritten the solution to use CollectFields so that it can take into account the fragment compatibility and the `@include`/`@skip` directives in the same way that the GraphQL spec would.

src/validation/rules/SingleFieldSubscriptionsRule.js
benjie: Have instead delegated to `collectFields`

src/validation/rules/SingleFieldSubscriptionsRule.js
```diff
@@ -1,30 +1,95 @@
...
+            fieldResolver: defaultFieldResolver,
+            typeResolver: defaultTypeResolver,
+            errors: [],
+          };
+          const fields = collectFields(
+            fakeExecutionContext,
+            subscriptionType,
+            node.selectionSet,
+            {},
+            {},
```
IvanGoncharov, line 62: With objects as maps they would count `fragment hasOwnProperty` as visited and exhibit other weird behavior.
In other places we used `Object.create(null)` but that is confusing as your PR shows, so I changed `collectFields` to accept `Map` and `Set`.
@benjie Can you rebase your PR?

benjie: Done :+1:

src/validation/rules/SingleFieldSubscriptionsRule.ts
```diff
@@ -1,30 +1,92 @@
...
+            [variable: string]: any;
+          } = Object.create(null);
+          const document = context.getDocument();
+          const fragments: ObjMap<FragmentDefinitionNode> = Object.create(null);
+          for (const definition of document.definitions) {
+            if (definition.kind === Kind.FRAGMENT_DEFINITION) {
+              fragments[definition.name.value] = definition;
+            }
+          }
+          const fakeExecutionContext: ExecutionContext = {
```
IvanGoncharov, line 45:
```suggestion
          // FIXME: refactor out `collectFields` into utility function that doesn't need fake context.
          const fakeExecutionContext: ExecutionContext = {
```

IvanGoncharov (approved):
Generally looks good 👍 
Not really happy with the fact that we create fake execution context but it's the only way to use `collectFields` at the moment.
Added FIXME to show that we plan to change that in the future since `collectFields` doesn't really need the whole execution context.

PR #2864, opened 2020-12-02
Fix JSON Serialization for source

lifeiscontent:
The Type definitions say that source should be exposed, but when its transformed into JSON its not there.

### DIFF
--- a/src/language/ast.d.ts
+++ b/src/language/ast.d.ts
@@ -33,7 +33,7 @@ export class Location {
 
   constructor(startToken: Token, endToken: Token, source: Source);
 
-  toJSON(): { start: number; end: number };
+  toJSON(): { start: number; end: number; source: Source };
 }
 
 /**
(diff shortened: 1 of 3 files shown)


lifeiscontent:
@IvanGoncharov is it possible to backport this fix to v14?

IvanGoncharov:
@lifeiscontent Sadly this change will make AST dumps unreadable.
Since every node will include the source.
That's why tests are falling on your PR.

IvanGoncharov:
I need to think more about this one.
Also, it would help if someone do research on how other tools store filenames in AST.

lifeiscontent:
@IvanGoncharov I've been working on https://github.com/lifeiscontent/storybook-addon-apollo-client

The only way I've managed to get the formatter to display code consistently is by dynamically injecting babel-plugin-graphql-tag into the storybook runtime to then parse all the code which seems to be including the source as I need in order to display it to the user in the panel.

If you have a better suggestion on how to achieve this, please let me know.

lifeiscontent:
Ping @IvanGoncharov

PR #2870, opened 2020-12-14
Feature/custom resolve field

emaciel10:


IvanGoncharov:
@emaciel10 Thanks for the PR 👍 
From your code example, it looks like you need to cache only `doc` fields.
Can you just do something like this? (not tested)
```js
for (const type of Object.values(schema.getTypeMap())) {
  if (isObjectType(type)) {
    for (const [name, field] of Object.entries(type.getFields())) {
      if (name === 'doc') {
        const oldResolve = field.resolve;
        field.resolve = (...args) => {
          const source = args[0];
          // ...
          return oldResolve(...args);
        }
      }
    }
  }
}
```

emaciel10:
Hey Ivan, thanks for the suggestion, I'll take a look and see if we can override the resolve to drop in our custom caching at that level instead!

PR #2876, opened 2021-01-05
Replace 'localeCompare' with function independent from locale

### LINKED ISSUE #2869: Allow specifying locale in `lexicographicSortSchema`
I use `lexicographicSortSchema()` to get a stable output when generating TS types from GraphQL schema and recently I discovered a problem when a field called `chemicalName` is misplaced and causing a diff.

The reason is that one of my machines has non-English, Czech locale and in Czech alphabet, the [digraph _Ch_](https://en.wikipedia.org/wiki/Ch_(digraph)) is handled as a single letter and sorted between _H_ and _I_.

<details>
<summary>Example</summary>

```js
const items = ['xx', 'aa', 'dd', 'cx', 'ch', 'cc'];

items.sort((a, b) => a.localeCompare(b, 'en'));
// ["aa", "cc", "ch", "cx", "dd", "xx"]

items.sort((a, b) => a.localeCompare(b, 'cs'));
// ["aa", "cc", "cx", "dd", "ch", "xx"]
```

</details>

Would be great if `lexicographicSortSchema()` could accept `locales` and possibly also `options` parameters for [`localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) to be able to force a particular (English) locale for the schema sorting independently on the machine's locale.

IvanGoncharov:
Fixes #2869

### DIFF
--- a/src/jsutils/__tests__/naturalCompare-test.js
+++ b/src/jsutils/__tests__/naturalCompare-test.js
@@ -1,0 +1,72 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import naturalCompare from '../naturalCompare';
+
+describe('naturalCompare', () => {
+  it('Handles empty strings', () => {
+    expect(naturalCompare('', '')).to.equal(0);
+
+    expect(naturalCompare('', 'a')).to.equal(-1);
+    expect(naturalCompare('', '1')).to.equal(-1);
+
+    expect(naturalCompare('a', '')).to.equal(1);
+    expect(naturalCompare('1', '')).to.equal(1);
+  });
+
+  it('Handles strings of different length', () => {
+    expect(naturalCompare('A', 'A')).to.equal(0);
+    expect(naturalCompare('A1', 'A1')).to.equal(0);
+
+    expect(naturalCompare('A', 'AA')).to.equal(-1);
+    expect(naturalCompare('A1', 'A1A')).to.equal(-1);
+
+    expect(naturalCompare('AA', 'A')).to.equal(1);
+    expect(naturalCompare('A1A', 'A1')).to.equal(1);
+  });
+
+  it('Handles numbers', () => {
+    expect(naturalCompare('0', '0')).to.equal(0);
+    expect(naturalCompare('1', '1')).to.equal(0);
+
+    expect(naturalCompare('1', '2')).to.equal(-1);
+    expect(naturalCompare('2', '1')).to.equal(1);
+
+    expect(naturalCompare('2', '11')).to.equal(-1);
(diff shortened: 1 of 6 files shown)


backbone87:
naturalCompare is case sensitive, while localeCompare isnt. this leads to changes in schema dumps. its a breaking change, but more an annoyance than causing real issues. one could argue that sorting case sensitive is not as natural as one expected.

IvanGoncharov:
@backbone87 Thanks for the feedback 👍 

> Intl.Collator.prototype.compare.bind(new Intl.Comparator('en-US'))

Since we support a bunch of platforms Node/Deno/Different browser we don't want to depend on anything beyond core JS APIs.
Also, the main purpose of `graphql-js` is to be reference implementation so libraries in other languages can easily replicate our code by just converting to their language (e.g. whole GraphQL in Python ecosystem is using line by line port of this lib https://github.com/graphql-python/graphql-core#graphql-core-3) so we try not to use stuff that can't be easily ported.

>  i think it would be better to accept a comparator function for string comparison which defaults

Happy to do that if we discover a few popular use cases that can't be solved by the single compare function.


>  one could argue that sorting case sensitive is not as natural as one expects.

I'm open to this change if you think it provides better DX in popular cases.

> naturalCompare is case sensitive, while localeCompare isn't. this leads to changes in schema dumps. its a breaking change, but more an annoyance than causing real issues.

Partly agree 👍  If it causes a major pain point for someone I can revert it on `15.x.x` line.
One downside, in that case, is that `15.x.x` and upcoming `16.x.x` will produce different dumps so it would be hard for library authors that need to support both `15.x.x` and `16.x.x` through peer dependencies.

backbone87:
> Partly agree 👍 If it causes a major pain point for someone I can revert it on 15.x.x line.

it depends for what you use schema dumps for. in our case we use type-graphql and commit a dumped schema snapshot for easier review (which is checked by CI to be "clean"). updating to latest graphql-js@^15 changed the snapshot in non-semantical ways, which would appear as an "unrelated" change during review, for appearently unknown reason. 

thats why i commented here for reference that maybe safes some others a few minutes of research.

> I'm open to this change if you think it provides better DX in popular cases.

as a reader of schema dumps: i would say yes its more intuitive to have case insensitive order. as a programer looking at the complexity of naturalCompare: do we really want to reimplement collation rules?

so i would change my proposal: introduce a comparator argument to `lexicographicSortSchema` which defaults to `naturalCompare`, but can then be just changed to `Intl.Collator.prototype.compare.bind(new Intl.Comparator('en-US'))`

PR #2828, opened 2020-10-27
Convert flow types to TS

saihaj:
### Summary
Convert all the `.js` to `.ts` and fix all syntax related issues that exist because of Flow and TypeScript.

* [x] Fix `%checks` and remove `eslint-disable` comments (https://github.com/saihaj/graphql-js/commit/5d966bd052b692844e729cc83d84ee7ddea06205)
* [x] Change `_` https://github.com/saihaj/graphql-js/commit/fbbb7a557db0f5cd921a0d8b434bdad2c512c40c#r43625283
* [x] use `object` since it is a `WeakMap` (https://github.com/saihaj/graphql-js/commit/ae7c3f406e948ea7411acf02a2f88dc058428213#r43625385)
* [x] Restrict to `integrationTests/*/**` (https://github.com/saihaj/graphql-js/commit/28023e6bf172c63ed5ce00a102fb3c42684cb501)
* [x] Convert  `?` to `Maybe` (few were missed in 41b6975)
* [x] Convert `mixed` to `unknown` (few were missed in cd55ba9)
* [x] Convert to `$Keys` to `keyof`
* [x] Convert `$ReadOnly` to `Readonly`
* [x] Fixup caa8334 into c244b37
* [x] Complete comparison https://github.com/saihaj/graphql-js/pull/1 compare changes made manually with changes made by a tool https://github.com/Khan/flow-to-ts 
* [X] Replace `$ObjMap` to `ObjMap` utility function (need help on this. See discussion here https://github.com/graphql/graphql-js/pull/2828#discussion_r520238458)
* [X] Update `.babelrc-deno` to use TypeScript

Action items Based on https://github.com/graphql/graphql-js/pull/2828#issuecomment-761102728
For @IvanGoncharov 
* [x] Research if we can simplify `visit` typings in `16.x.x`
* [x] Remove polyfills
* [x] Fix test coverage on `16.x.x`
* [ ] https://github.com/graphql/graphql-js/pull/2828#discussion_r554121151
* [ ] Implement `GraphQLAgregatedError`
* [ ] Help @saihaj to generate declarations

For @saihaj 
* [x] Compare and remove .d.ts
* [x] https://github.com/graphql/graphql-js/pull/2828#discussion_r556184614
* [x] Remove Flow infrastructure, even integration test (remove in one commit and split if requested)
* [x] Convert type to interfaces
* [x] Enable as much as possible ESLint rules

~Fix no redeclare `src/polyfills/objectValues.ts`~ (these will be removed so no need)

PR #2883, opened 2021-01-16
build: use `@tsconfig/node10`

saihaj:
Instead of adding specific `lib` targets for example in here 487377a5cae97adcc52d7b9dac54bb49391f9928 (#2828) use [`@tsconfig/node10`](https://www.npmjs.com/package/@tsconfig/node10) since that config has everything we need to support targeted node versions. Suggested by @IvanGoncharov


IvanGoncharov:
@saihaj Sorry for the confusion.
I meant that we can copy and paste the setting from node10 config.
Saving few lines doesn't forth introducing a new dev dependency.

IvanGoncharov:
@saihaj I incorporated `lib` and `target` into https://github.com/graphql/graphql-js/commit/1951bce42092123e844763b6a8e985a8a3327511#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0

PR #2887, opened 2021-01-20
fix: change OperationTypeNode to enum

lekoaf:
Changes OperationTypeNode to enum instead of a union type. This allows developers to use the type as `OperationTypeNode.QUERY` instead of risking to msispell the string 'query'.

Example:

Instead of having to write

`operation === 'query'`

you can instead write

`operation === OperationTypeNode.QUERY`

This should not be a breaking change, since the string version should still work.


IvanGoncharov:
@lekoaf Thanks for the PR 👍 
We definitely need to merge it.
One thing that we currently in process of migrating our codebase to TS with the goal of minimal changes in mind, see #2828 
Let's merge it after we switch `master` to TS.

saihaj (approved):
Thanks @lekoaf! @IvanGoncharov we can merge this now

yaacovCR:
Now that TS migration is done, this can perhaps be included in v16? Is there a reason this was held up?

lekoaf:
> Now that TS migration is done, this can perhaps be included in v16? Is there a reason this was held up?

Yupp, I missed the rebase comment. I'll do that tomorrow. 😃

PR #2894, opened 2021-01-26
Avoid relying on constructor.name for instanceOf error check.

benjamn:
PR #1174 disabled [`instanceOf`](https://github.com/graphql/graphql-js/blob/master/src/jsutils/instanceOf.js) error checking when `process.env.NODE_ENV === 'production'`, which eliminated some false positives due to collisions of minified class names in production. This PR aims to eliminate the same kind of false positive errors in non-production environments that enable minification, such as `NODE_ENV='staging'`. Please see https://github.com/apollographql/apollo-client/issues/7446#issuecomment-767660631 for motivating use cases and background explanation.

Fortunately, the `graphql` package goes to the trouble of providing a [`Symbol.toStringTag`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) property for most of the classes it exports ([one of many examples](https://github.com/graphql/graphql-js/blob/3bce13f652e4043d940df8eec29f79eef24b6215/src/type/definition.js#L793-L795)), and that string is immune to minification (unlike `constructor.name`). This PR proposes using that tag string for error checking in `instanceOf(value, constructor)` whenever the `constructor` provides a `Symbol.toStringTag` property, falling back to `constructor.name` and `value.constructor.name` for constructors that do not define `Symbol.toStringTag` (as before).

If this is controversial in any way, I would be happy to discuss alternative solutions, but I believe this change will resolve many of the remaining issues linked at the bottom of the #1174 issue thread.

Remaining to-dos:
- [x] Add more [`instanceOf` tests](https://github.com/graphql/graphql-js/blob/master/src/jsutils/__tests__/instanceOf-test.js) to restore full coverage.

### DIFF
--- a/src/jsutils/__tests__/instanceOf-test.js
+++ b/src/jsutils/__tests__/instanceOf-test.js
@@ -2,6 +2,15 @@ import { expect } from 'chai';
 import { describe, it } from 'mocha';
 
 import { instanceOf } from '../instanceOf';
+import { SYMBOL_TO_STRING_TAG } from '../../polyfills/symbols';
+import {
+  GraphQLScalarType,
+  GraphQLObjectType,
+  GraphQLInterfaceType,
+  GraphQLUnionType,
+  GraphQLEnumType,
+  GraphQLInputObjectType,
+} from '../../type/definition';
 
 describe('instanceOf', () => {
   it('fails with descriptive error message', () => {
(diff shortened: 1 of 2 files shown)


benjamn:
That code coverage failure is reasonable! I will add some more [`instanceOf` tests](https://github.com/graphql/graphql-js/blob/master/src/jsutils/__tests__/instanceOf-test.js) to restore full coverage.

IvanGoncharov:
@benjamn Thanks for PR.
We moved the `15.x.x` line into the maintenance but want to address this issue in the upcoming `16.0.0`.
And we need to drop our reliance on `NODE_ENV` completely to support Deno natively.
Current idea is to use an approach from React repo and adapt it to `graphql-js`.

Currently, we are focused on TS migration so let's keep this PR open until we figure out a future-proof solution.

benjamn:
@IvanGoncharov Dropping `NODE_ENV` sounds like a good idea, but this problem occurs whenever class names are minified and `instanceOf` uses `constructor.name` to determine the class name at runtime. You can use `NODE_ENV` to hide the problem (for better or worse), but dropping `NODE_ENV` won't make this problem go away.

Would you consider making an update to the 15.x.x line, given that no (pre)release of 16.x is available yet? This is a real problem for Apollo Client developers, and I don't know of any good workarounds. Here's [one](https://github.com/apollographql/apollo-client/issues/7446#issuecomment-767867590) that I suggested recently, if you want to see what I mean.

I would also caution against following the lead of the React repo, since they are still using [a conditional `require` approach](https://github.com/facebook/react/blob/master/packages/react/npm/index.js) that only works with CommonJS exports, not ESM (since you can't nest `import` declarations inside conditional blocks).

benjamn:
@IvanGoncharov Do you think you could take another look at this PR, in light of my previous comment? Thanks!

IvanGoncharov:
@benjamn Will take a look into this.
Resolving this issue is planned for `16.0.0-alpha.1` so it's on top of my list.

benjamn:
@IvanGoncharov Just rebased against `main`, though the old branch is [here](https://github.com/apollographql/graphql-js/tree/old-avoid-constructor.name-in-instanceOf) in case you need it.

IvanGoncharov:
@benjamn Did another quick review and agree that it should be merged.
I Will try to merge in the next few days and release it as part of `16.0.0-alpha.1`.

hwillson:
👋 @IvanGoncharov - do you think this will make it into the next alpha?

IvanGoncharov:
@benjamn @hwillson Sorry for the delay, I'm back on it.
Want to do some preparation work first like require that every exported class have `toStringTag`.
ETA today or tomorrow.

hwillson:
Awesome @IvanGoncharov - thanks very much!

IvanGoncharov:
@benjamn @hwillson Release as https://github.com/graphql/graphql-js/releases/tag/v16.0.0-alpha.4
Your PR helped a lot I simplified the resulting code using TS to ensure that all our classes have `Symbol.toStringTag` and also added a custom ESLint rule to be extra safe.
Please report if it solved your issues, also if you are interested in backporting #3172 to https://github.com/graphql/graphql-js/tree/15.x.x I can merge and release it.

tubbo:
@IvanGoncharov I just tried this code and I get an error:

```
"export 'TypeKind' was not found in './type/index.mjs'
```

I'm running node v16.3.0 as I had to upgrade to take this update.

tubbo:
I backported this in #3186 so we can get this fix on v15.x. Not much of the tooling has caught up to v16 yet so this should allow the change to work for a lot more people.

PR #2900, opened 2021-01-27
Drop support for long deprecated comments as descriptions

IvanGoncharov:


jonathanstanley:
Just to clarify in case others come here trying to resolve their lost comments: GraphQL differentiates between "[descriptions](https://spec.graphql.org/October2021/#sec-Descriptions)" and "[comments](https://spec.graphql.org/October2021/#sec-Comments)"

Here is a good resource on the topic: https://github.com/graphql/graphql-spec/issues/420

PR #2901, opened 2021-01-30
Clarify comment about mutually exclusive parents

andimarek:


yaacovCR:
While it's true that we can currently never find a conflict when the parent type is a union, if we chose to allow a union to be constrained to a type that implement an interface, we might eventually allow fields on unions, and the current code/spec is future-proof in that regard.

See RFC: https://github.com/graphql/graphql-wg/blob/main/rfcs/ExpandingSubtyping.md

For now, I am going to close this old PR, really just for housekeeping. Feel free to reopen if you still think this change should go in, if you would like to continue the conversation, if for any other reason.

PR #2905, opened 2021-02-01
resolveType: remove support for returning GraphQLObjectType

IvanGoncharov:


boetersep:
The documentation was not updated

saihaj:
@boetersep yeah we are running very behind on apps. We have a new site which should be landing soon which provides TS Docs api #3269

PR #2910, opened 2021-02-09
Allow Buffer objects to be seamlessly treated as string and ID types

pospi:
This results in an ergonomic improvement when GraphQL is used with `msgpack` API calls in resolvers.


IvanGoncharov:
@pospi Sorry for the late response.
`Buffer` is untyped data even if we assume it can represented as a string there are no way we can guess encoding.
Assuming `utf-8` will work in many cases but it not something we can make a default.

PR #2920, opened 2021-02-15
Deprecate fragments with variables and reflect that in naming

IvanGoncharov:


drewbaker:
Very curious what this means in real world? Any examples of what this changed? I'm getting some AST warnings with v16 and suspect it's from this, but not sure what it does.

PR #2926, opened 2021-02-18
build: add types for node, chai and mocha

saihaj:
- build: add @types/node
- build: commit package-lock
- build: add mocha types
- build: add chai types


saihaj:
should go in #2828

PR #2946, opened 2021-03-07
docs: add more exported utilities to APIReference

zcei:
Hej all,

I was playing around a bit with some GraphQL utilities and ended up looking at the source and was surprised to find so many more useful methods in there than in documentation.

Are these fine to be documented? They're ready to be imported on the public package interfaces via `graphql` and `graphql/utilities` already.

### DIFF
--- a/docs/APIReference-Utilities.md
+++ b/docs/APIReference-Utilities.md
@@ -3,7 +3,7 @@ title: graphql/utilities
 layout: ../_core/GraphQLJSLayout
 category: API Reference
 permalink: /graphql-js/utilities/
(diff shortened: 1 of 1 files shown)


zcei:
(Wow that's a whole lot to read in the CLA just to get some docs changes out 🤣

zcei:
How can I link properly to another function (`buildClientSchema`) without CSpell complaining?

IvanGoncharov:
@zcei Thanks for the help and effort you spend on this PR 👍 
Currently, our docs are severely outdated and broken.
It's unrealistic to support such a big API surface with manual API references so our current plan is to generate docs after we switch to TS in #2828 

If you want to improve our current docs hosted under https://graphql.org/graphql-js/ please make these against `graphql.org` repo: https://github.com/graphql/graphql.github.io/tree/source/src/content/graphql-js

PR #2949, opened 2021-03-08
introspection: Allow interfaces to be undefined for interfaces

Tapped:
The old code was using strict equality check for null to check if the introspected field `interfaces` was not defined, which means that it would fail in case `interfaces` was set to undefined. Older GraphQL implementations will not send the `interfaces` field when introspecting an interface, and thus would fail with `Introspection result missing interfaces`.


IvanGoncharov:
@Tapped Sorry for delay response.
But according to the spec all types should have `interfaces` as `null` even for old servers.
From previous spec release:
image

PR #2975, opened 2021-03-19
fix(race): concurrent next calls with defer/stream

yaacovCR:
This seems to work with me for my project.

not quite sure what to do about isDone which seems to be used just when return is  called (parenthetically, should that be renamed to isReturned for clarity/distinction between `isDone` and `done` attributes?)

namely, should the isDone check go up top prior to checking for the next resolved promise, within the next resolved promise check, or maybe both.

in general, not sure how we are really handling secure stopping issues, seems like -- for example -- there could be condition in which we await a promise which never resolves, but that is ok, because return has been called?

but still quite the novice with all of this async stuff

---From my opinionated perspective, I realize that graphql-js is a reference lib, and has a zero-dependency footprint, but seems like handling all these async issues (in JS in general and within `graphql-js`) would be easier with a Repeater-like construct (see https://repeater.js.org/, @brainkim) that could be even theoretically inlined in downsized version into `js-utils`.

### DIFF
--- a/src/execution/__tests__/stream-test.ts
+++ b/src/execution/__tests__/stream-test.ts
@@ -133,16 +133,18 @@ async function complete(document: DocumentNode) {
   return result;
 }
 
-async function completeAsync(document, numCalls) {
+async function completeAsync(document: DocumentNode, numCalls: number) {
   const schema = new GraphQLSchema({ query });
 
   const result = await execute({ schema, document, rootValue: {} });
 
   invariant(isAsyncIterable(result));
 
+  const iterator = result[Symbol.asyncIterator]();
+
   const promises = [];
   for (let i = 0; i < numCalls; i++) {
-    promises.push(result.next());
+    promises.push(iterator.next());
   }
   return Promise.all(promises);
 }
--- a/src/execution/__tests__/stream-test.ts
+++ b/src/execution/__tests__/stream-test.ts
@@ -437,6 +437,72 @@ describe('Execute: stream directive', () => {
       },
     ]);
   });
+  it('Can stream a field that returns an async iterable', async () => {
+    const document = parse(`
+      query { 
+        asyncIterableList @stream(initialCount: 2) {
+          name
+          id
+        }
+      }
+    `);
+    const schema = new GraphQLSchema({ query });
+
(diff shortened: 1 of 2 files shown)


yaacovCR:
@robrichard hope this helps @IvanGoncharov et al can’t seem to read test coverage failure results as now on mobile, let me know if real and maybe what lines need help

yaacovCR:
@robrichard @IvanGoncharov ready for review

yaacovCR:
Although this change suggests that _subsequentPayloads should be a set rather than an array

yaacovCR:
This PR fixes by modifying the general approach as little as possible, we still create Promise reactions to all the pending promises on each call to `next`.

See https://github.com/graphql/graphql-js/issues/2848#issuecomment-799032492 where I link to https://github.com/repeaterjs/repeater/blob/219a0c8faf2c2768d234ecfe8dd21d455a4a98fe/packages/repeater/src/repeater.ts#L687 that uses a different approach to convert all pending promises to async iterables (which each yield one value) and then combining those iterables in settle order.

If you would be more open to that type of approach, I can try to see if I can work to extract it (maybe with the help of @brainkim)

robrichard:
@yaacovCR thanks again for your help on this. I'm not tied to the current approach. I think a generic well-tested jsutil that can combine and race promises and async iterables makes sense.

yaacovCR:
@robrichard @IvanGoncharov I think the best path forward so as to not lose steam/forward momentum with any experimenters, would be to merge this bug fix that keeps the existing approach, and then as a separate PR, see if myself or @brainkim is able to work something else out in a refactor.

yaacovCR:
Would this be accepted if rebased? graphql tools stitching support uses patch package to apply this fix....

robrichard:
@yaacovCR I'm planning to spend some time refactoring the dispatcher to a generic jsutil that can be more easily tested and handle all the edge cases

yaacovCR:
Sure but until then can we merge it so the bug is fixed

yaacovCR:
Meaning prior to next experimental release it would be nice if fixed somehow with this PR if not a refactor

PR #2980, opened 2021-03-22
fix(stream): continue on iterator error values

yaacovCR:
if the iterator errors when attempting to get the next value, we can assume subsequent calls to next will also error, and abort, but if we successfully get the next value, but it ends up triggering an error, we can continue, optimistic that the next value will not do so.

### DIFF
--- a/benchmark/benchmark.js
+++ b/benchmark/benchmark.js
@@ -346,9 +346,9 @@ function sampleModule(modulePath) {
 
     clock(7, module.measure); // warm up
     global.gc();
-    process.nextTick(() => {
+    process.nextTick(async () => {
       const memBaseline = process.memoryUsage().heapUsed;
-      const clocked = clock(module.count, module.measure);
+      const clocked = await clock(module.count, module.measure);
       process.send({
         name: module.name,
         clocked: clocked / module.count,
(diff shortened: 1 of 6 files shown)


yaacovCR:
@robrichard @IvanGoncharov this is more of a question than a PR -- what is the desired behavior?

robrichard:
I don't think it's expected to receive more values after an async iterable throws an error. How is this handled with subscriptions?

yaacovCR:
That’s just it, this code is not about the iterator returning an error, that’s further down. This code section is an error encountered in completion of a list item. With iterables, the next item is allowed to be processed, shouldn’t it be the same with async iterables?

yaacovCR:
Added tests to explain

robrichard:
@yaacovCR should this target the `asynciteratable` branch (https://github.com/graphql/graphql-js/pull/2757)? It doesn't seem to be bug in defer/stream, but rather the related feature of allowing async iterable return values from resolvers.

yaacovCR:
Yes! Will rework when I have a moment to target that branch

yaacovCR:
@robrichard @IvanGoncharov 

ready for review

yaacovCR:
Bump, would this be accepted if rebased?

robrichard:
@yaacovCR I pulled this in while rebasing. Thanks!

PR #2985, opened 2021-03-24
TS: switch to Record utility type

saihaj:
- feat: switch to Record utility type

### DIFF
--- a/src/error/GraphQLError.d.ts
+++ b/src/error/GraphQLError.d.ts
@@ -18,7 +18,7 @@ export class GraphQLError extends Error {
     positions?: Maybe<ReadonlyArray<number>>,
     path?: Maybe<ReadonlyArray<string | number>>,
     originalError?: Maybe<Error>,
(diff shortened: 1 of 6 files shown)


saihaj:
@IvanGoncharov as per #2860 we are dropping support for Typescript v3.7 so that is the only integration test that is failing.

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "exclude": ["integrationTests/ts/**/*"],
   "compilerOptions": {
     "module": "commonjs",
-    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string"],
+    "lib": ["es2019", "es2020.string"],
```
IvanGoncharov, line 5: No, keep this as is.
Not all code published as an NPM package and nothing preventing us from using these types there.

IvanGoncharov (changes requested):

IvanGoncharov:
@saihaj Can you please extract the `AsyncGenerator` change into separate PR?

PR #3000, opened 2021-03-29
feat: add `.d.ts` for jsutils

saihaj:


src/jsutils/instanceOf.d.ts
```diff
@@ -0,0 +1,3 @@
+/* eslint-disable import/no-default-export */
+declare const _default: (value: unknown, constructor: unknown) => boolean;
+export default _default;
```
IvanGoncharov, line 3: @saihaj It's changed in `main` please remove default export.

src/jsutils/invariant.d.ts
```diff
@@ -0,0 +1 @@
+export function invariant(condition: unknown, message?: string): void;
```
IvanGoncharov, line 1: Can you please update the return type to `: asserts condition`?
Please also do this change in TS migration PR as it should remove a bunch of TS errors.

PR #3012, opened 2021-04-02
mapAsyncIterator: allow 'rejectCallback' to only be synchronous

IvanGoncharov:


yaacovCR:
Just curious, why the restriction?

PR #3044, opened 2021-04-16
Schema Coordinates

leebyron:
Depends on #3115

Implements https://github.com/graphql/graphql-spec/pull/794/

Adds:

* DOT punctuator in lexer
* Improvements to lexer errors around misuse of `.`
* Minor improvement to parser core which simplified this addition
* `SchemaCoordinate` node and `isSchemaCoodinate()` predicate
* Support in `print()` and `visit()`
* Added function `parseSchemaCoordinate()` since it is a parser entry point.
* Added function `resolveSchemaCoordinate()` and `resolveASTSchemeCoordinate()` which implement the semantics (name mirrored from `buildASTSchema`) as well as the return type `GraphQLSchemaElement`


src/language/lexer.js
```diff
@@ -284,6 +285,38 @@ function unexpectedCharacterMessage(code: number): string {
...
+      start,
+      'Cannot parse the unexpected character "..".',
+    );
+  }
+  if (nextCode >= 48 && nextCode <= 57) {
+    // 0 - 9
+    throw syntaxError(
+      source,
+      start,
+      'Invalid number, expected digit before ".".',
```
magicmark, line 314: maybe i'm just understanding this wrong, but it looks the the issue that we found a number _after_ "."?

Would `Unexpected number after "."` be more appropriate?

leebyron: You're exactly right - this error is from `. [lookahead != Digit]`, so a number coming after the "." - however the message that we provide for the failure should be as informative as possible based on the intent.

The rationale for limiting this is that in many languages, including javascript, `.123` is actually a legal number token, but in GraphQL it is not, and the leading `0.` is required. Many new to GraphQL may write `.123` directly in a query, expecting it to be interpreted as a number and ideally we can give a clear message as to why.

If you wrote a query: `{ doThing(num: .123) }` and get back `'Unexpected number after "."'` you might be confused given the intent - you *meant* to put a number after the `.` it wasn't a typo. Instead we should make it clear that this style of number isn't supported and how to fix it.

I wonder if we should make this message even more specific about needing a `0` before the `.`?

src/language/printer.js
```diff
@@ -301,6 +301,18 @@ const printDocASTReducer: any = {
...
+
+  // Schema Coordinate
+
+  SchemaCoordinate: {
+    leave: ({ isDirective, name, fieldName, argumentName }) =>
+      join([
+        isDirective && '@',
+        name,
+        wrap('.', fieldName),
+        wrap('(', argumentName, ':)'),
```
magicmark, line 313: :)

magicmark (approved):
Amazing!

This is way beyond what I was hacking with. Hadn't even considered reusing the existing parser. (I was fiddling around with regexes...)

Thanks for putting this together - this makes a lot of sense.

src/language/ast.js
```diff
@@ -625,3 +627,14 @@ export type InputObjectTypeExtensionNode = {|
...
 |};
+
+// Schema Coordinates
+
+export type SchemaCoordinateNode = {|
+  +kind: 'SchemaCoordinate',
+  +loc?: Location,
+  +isDirective: boolean,
+  +name: NameNode,
+  +fieldName?: NameNode,
```
IvanGoncharov, line 638: My main concern is enum values are accessed through the property `field` of the AST node.
I still think we need to have separate syntax for enum values.
I also have other arguments for separate syntax for non-fields, I will try to formulate them later.

leebyron: Yeah this is definitely a generalization and a tradeoff to keep this name short. What do you think about `fieldOrValueName` to make it clear that the one AST shape supports both forms?

src/utilities/resolveSchemaCoordinate.js
```diff
@@ -0,0 +1,111 @@
...
+ * A schema coordinate is resolved in the context of a GraphQL schema to
+ * uniquely identifies a schema element. It returns undefined if the schema
+ * coordinate does not resolve to a schema element.
+ *
+ * https://spec.graphql.org/draft/#sec-Schema-Coordinates.Semantics
+ */
+export function resolveSchemaCoordinate(
+  schema: GraphQLSchema,
+  schemaCoordinate: string | Source,
+): GraphQLSchemaElement | void {
```
IvanGoncharov, line 23: From a DX standpoint returning just an object is bad because fields, args, and enums values are plain objects. That means you can easily detect what was returned.

There two solutions wrap the result in object and add `kind` tag field, e.g.:
```
{
  kind: Kind.ARGUMENT,
  name: 'someArg',
  ...
}
```
alternatively, it can be, "one of" pattern:

```
{
  arguments: {
    name: 'someArg',
    // ...
  },
}
```

src/language/ast.js
IvanGoncharov: @leebyron I was thinking about something even more general like `subElementName`.
Since in future we can potentially make even more things addressable.

leebyron:
Great feedback - I think I've resolved both:

* Replaced `fieldName` with `memberName` in the AST to ensure it's generalized between fields, input fields, and enum values. Those are all specific instances of the general pattern of "members of a set", which is pretty common terminology for this generalized dot access pattern.
* Took both of your great suggestions for wrapping the return type of `resolveSchemaCoordinate` - the `GraphQLSchemaElement` type is now a union of wrappers. This had the added benefits of disambiguating between "field argument" and "directive argument" and including the coordinate context in the return value

src/language/ast.ts
```diff
@@ -670,3 +672,14 @@ export interface InputObjectTypeExtensionNode {
   readonly directives?: ReadonlyArray<ConstDirectiveNode>;
   readonly fields?: ReadonlyArray<InputValueDefinitionNode>;
 }
+
+// Schema Coordinates
+
+export interface SchemaCoordinateNode {
+  readonly kind: 'SchemaCoordinate';
+  readonly loc?: Location;
+  readonly isDirective: boolean;
```
IvanGoncharov, line 681: @leebyron `isDirective` sounds good if we reference the directive itself e.g. `@foo` but it confusing if we reference the directive argument e.g. `@foo.arg`
Maybe we can just have two node kinds `TypeCoordinate` and `DirectiveCoordinate`?
Since directives represent another namespace anyway?

leebyron: That's a reasonable suggestion, I tried it out but found it made things a lot more complicated. Created 3 types instead of 1 to represent coordinates, complicated the predicate checker, and then backporting the change to the spec text (assuming the AST should mirror the grammar) made it a little bit more awkward to explain (sets of semantics for each sub-grammar). It wasn't awful, but it's definitely a bit of cognitive cost.

Then IMHO `DirectiveCoordinate` implies something is a directive more or less equivalently as `isDirective` - or at least not more so to offset the cognitive cost of a more complex implementation and spec text.

IvanGoncharov: @leebyron Agree, with the above arguments 👍 
But still `isDirective` doesn't look right for coordinates for directive arguments.
How about renaming it to `ofDirective` meaning it either:
- schema coordinate `of directive`
- schema coordinate `of directive` argument

leebyron: Sure, that makes some sense. Let's do that

src/utilities/index.ts
```diff
@@ -127,3 +127,10 @@ export type {
 
 // Wrapper type that contains DocumentNode and types that can be deduced from it.
 export type { TypedQueryDocumentNode } from './typedQueryDocumentNode.js';
+
+// Schema coordinates
+export {
+  resolveSchemaCoordinate,
+  resolveASTSchemaCoordinate,
```
JoviDeCroock, line 134: We should make a note about documenting this when we start on the v17 documentation

benjie:
I've merged the latest `main` into this branch and got it passing tests/etc again; am adding it to the next GraphQL WG.

JoviDeCroock (approved):

src/language/__tests__/lexer-test.ts
```diff
@@ -965,6 +965,13 @@ describe('Lexer', () => {
       value: undefined,
     });
 
+    expect(lexOne('.')).to.contain({
+      kind: TokenKind.DOT,
+      start: 0,
+      end: 1,
+      value: undefined,
+    });
```
martinbonnin: Can we add tests for the `'Unexpected ".."` and `Invalid number, expected digit before "."` cases?

src/language/ast.ts
```diff
@@ -762,3 +764,14 @@ export interface InputObjectTypeExtensionNode {
...
 }
+
+// Schema Coordinates
+
+export interface SchemaCoordinateNode {
+  readonly kind: 'SchemaCoordinate';
+  readonly loc?: Location;
+  readonly ofDirective: boolean;
+  readonly name: NameNode;
+  readonly memberName?: NameNode | undefined;
```
martinbonnin: Representing things like so makes it possible to do this:

```ts
{
  kind: 'SchemaCoordinate',
  ofDirective: true,
  name: "foo",
  memberName: "bar", // ⚠️ invalid syntax for directives
  argumentName: "arg"
}
```

Is there any other Node that allows this?

The alternative is to use an intermediate rule, like for `Selection`:

```
SchemaCoordinate:

- TypeCoordinate
- MemberCoordinate
- ArgumentCoordinate
- DirectiveCoordinate
- DirectiveArgumentCoordinate
```

Has this been discussed already?

src/utilities/index.ts
martinbonnin, line 134: Those functions are defined in the spec (see https://github.com/graphql/graphql-spec/pull/794#discussion_r1901002157), should we put them in `language` instead of `utlilities`?

src/utilities/resolveSchemaCoordinate.ts
```diff
@@ -0,0 +1,190 @@
...
+      if (!directive) {
+        return;
+      }
+      return { kind: 'Directive', directive };
+    }
+
+    // SchemaCoordinate : @ Name ( Name : )
+    // Assert {directive} must exist.
+    if (!directive) {
+      return;
```
martinbonnin: Should assertions throw instead of returning `undefined`?

src/utilities/resolveSchemaCoordinate.ts
```diff
@@ -0,0 +1,190 @@
...
+    if (!type) {
+      return;
+    }
+    return { kind: 'NamedType', type };
+  }
+
+  if (!argumentName) {
+    // SchemaCoordinate : Name . Name
+    // If {type} is an Enum type:
+    if (isEnumType(type)) {
```
martinbonnin: Depending what we decide about meta fields (see https://github.com/graphql/graphql-spec/pull/794/files#r1900996045), we might want to assert that `memberName.value` does not start with `__` here.

src/language/__tests__/lexer-test.ts
benjie: We have those tests in GraphQL js already I think?

martinbonnin: Indeed, sorry I missed them! ([here](https://github.com/graphql/graphql-js/blob/b0f467d6d00da71d182e50280ac180b694f1cd63/src/language/__tests__/lexer-test.ts#L1034) and [here](https://github.com/graphql/graphql-js/blob/b0f467d6d00da71d182e50280ac180b694f1cd63/src/language/__tests__/lexer-test.ts#L863))

src/language/ast.ts
magicmark: https://github.com/graphql/graphql-js/pull/4422/files

magicmark:
@benjie I don't have write access to this PR since I wasn't the original author -- do you think we could merge in https://github.com/graphql/graphql-js/pull/4422 and then https://github.com/magicmark/graphql-js/pull/1 to collapse the stack for easier review?

(Alternatively, i'm happy to make a yet-another new PR to collapse everything, was trying to avoid making things even more confusing though.)

martinbonnin:
> do you think we could merge in https://github.com/graphql/graphql-js/pull/4422 and then https://github.com/magicmark/graphql-js/pull/1 to collapse the stack for easier review?

+1 on having a single PR for all changes.

benjie:
@magicmark let's merge https://github.com/magicmark/graphql-js/pull/1 into https://github.com/graphql/graphql-js/pull/4422 (you have to do that since you own that repository), then incorporate changes to support https://github.com/magicmark/graphql-spec/pull/1, then I can merge the whole lot into this PR assuming there's no pushback.

src/language/__tests__/printer-test.ts
```diff
@@ -299,4 +299,18 @@ describe('Printer: Query document', () => {
...
+    expect(print(parseSchemaCoordinate('  Name . field '))).to.equal(
+      'Name.field',
+    );
+    expect(print(parseSchemaCoordinate('  Name . field ( arg: )'))).to.equal(
+      'Name.field(arg:)',
+    );
+    expect(print(parseSchemaCoordinate(' @ name  '))).to.equal('@name');
+    expect(print(parseSchemaCoordinate(' @ name (arg:) '))).to.equal(
+      '@name(arg:)',
+    );
```
benjie: This is missing an enum value print.

src/language/tokenKind.ts
```diff
@@ -10,8 +10,10 @@ export const TokenKind = {
   AMP: '&' as const,
   PAREN_L: '(' as const,
   PAREN_R: ')' as const,
+  DOT: '.',
```
benjie:
```suggestion
  DOT: '.' as const,
```

src/language/__tests__/printer-test.ts
magicmark: https://github.com/graphql/graphql-js/pull/4429/files

src/utilities/__tests__/resolveSchemaCoordinate-test.ts
```diff
@@ -0,0 +1,189 @@
...
+      kind: 'NamedType',
+      type: schema.getType('Business'),
+    });
+
+    expect(resolveSchemaCoordinate(schema, 'String')).to.deep.equal({
+      kind: 'NamedType',
+      type: schema.getType('String'),
+    });
+
+    expect(resolveSchemaCoordinate(schema, 'private')).to.deep.equal(undefined);
```
martinbonnin, line 52: should this be `null`?

```suggestion
    expect(resolveSchemaCoordinate(schema, 'private')).to.deep.equal(null);
```

src/language/parser.ts
```diff
@@ -182,6 +188,26 @@ export function parseType(
...
+ * resolveASTSchemaCoordinate(). Or calling resolveSchemaCoordinate() directly
+ * with an unparsed source.
+ */
+export function parseSchemaCoordinate(
+  source: string | Source,
+  options?: ParseOptions,
+): SchemaCoordinateNode {
+  const parser = new Parser(source, options);
+  parser.expectToken(TokenKind.SOF);
+  const type = parser.parseSchemaCoordinate();
```
yaacovCR:
```suggestion
  const coordinate = parser.parseSchemaCoordinate();
```

src/language/parser.ts
```diff
@@ -182,6 +188,26 @@ export function parseType(
...
+  parser.expectToken(TokenKind.EOF);
+  return type;
```
yaacovCR:
```suggestion
  return coordinate;
```

src/utilities/index.ts
yaacovCR: ~~fwiw, I vote for `utilities`~~

yaacovCR (approved):

src/utilities/__tests__/resolveSchemaCoordinate-test.ts
magicmark: per @benjie's comment here https://github.com/magicmark/graphql-js/pull/1#discussion_r2122309237 I believe `undefined` is correct, despite the spec saying "null". I think a note _somewhere_ in the spec text saying something like this:

> "Algorithms for functions in this spec may return "null" (in a non-error state). In the reference graphql-js implementation, this corresponds to a value of `undefined`"

would be a helpful clarification - so far it's caught us both out :P

src/language/parser.ts
magicmark: are you able to hit the "accept suggestion" button? I don't have permissions to do so, but lg2m.

src/utilities/index.ts
yaacovCR: Actually, it could fit better into the `type` section or its own 'coordinate' section.

src/utilities/resolveSchemaCoordinate.ts
yaacovCR: resolving this conversatin, we have decided to allow meta-fields

yaacovCR:
### Update

The 2nd member of this stack by @leebyron was #3145, with that PR adding a coordinate field to all schema elements, allowing library users to move from schema element => to coordinate (as opposed to this PR, which allows the reverse, from coordinate => to schema element. I have now updated #3808 to be rebased on top of this PR with this PR as the base. We can still merge it separately.

### Point for discussion

Now that we have allowed meta-fields in this PR, the other PR #3145/#3808 demonstrates that the schema element <==> coordinate relationship will not allow cycles with meta-fields.

For example, coordinate `SomeType.__typename` will resolve to `TypeNameMetaFieldDef` by this PR, an instance of `GraphQLField`. In #3808, `TypeNameMetaFieldDef` has a "coordinate" property of `<meta>.__typename`, which is not a valid coordinate and cannot be resolved back to `TypeNameMetaFieldDef`. This also affects `SchemaMetaFieldDef` and `TypeMetaFieldDef`. The meta-fields are each defined once by the library and then accessed "magically" by `schema.getField(...)` -- rather than being artificially defined on every type (for `__typename`) or every root query type of every schema (for the `__schema` and `__type`).

To preserve the symmetry we have with non-meta-fields, we would have to make `<meta>.__typename` or something like it valid and resolve to `TypeNameMetaFieldDef` and make `SomeType.__typename` not resolve to `TypeNameMetaFieldDef`.

Other options would be to simply acknowledge this asymmetry:
- perhaps by renaming the `coordinate` property introduced by #3808 to something else
- perhaps by not including a `coordinate` property at all => `schemaElement.toString()` essentially holds that value anyway on the main branch after the work done in #4288)
- perhaps simply by documentation

src/utilities/__tests__/resolveSchemaCoordinate-test.ts
martinbonnin: Right, it'd help a lot 👍 Also, could we include the reason why we're doing this? Is there any advantage of using `undefined` instead of just `null`?

magicmark: `undefined` is what @leebyron's original implementation used. I don't have a strong opinion either way here, there's a valid semantic argument for either. I would be ok to switch. Do you feel strongly that it should be `null`?

magicmark:
a bare `<meta>.__typename` isn't something that would/could exist in a GraphQL operation right? I can't think what use cases there would be to need to support that as something we'd want to parse as a schema coordinate. Therefore, your suggestion of "perhaps simply by documentation" (and possibly a runtime assertion if required) sg2m?

(I don't know if this blocks the spec, or if this is just an edge casey implementation detail)

src/utilities/__tests__/resolveSchemaCoordinate-test.ts
martinbonnin: From an outside vantage point, it sure looks surprising that the spec says `null` but `graphql-js` returns `undefined`.

But also I'm not the biggest `graphql-js` connoisseur out there so I'll leave y'all to decide.

magicmark: @benjie wdyt?

benjie:
```
> [1, 2, 3].find(n => n === 4)
undefined
```

If it's a big deal, don't say `Return {null}` just say `Return the blah, if it exists.` and leave the "if it doesn't exist" up to the implementation. Which for the vast majority of implementations will be `null` but in JS it's more common to use `undefined` for "not found".

benjie:
@yaacovCR thanks for finding out why Lee explicitly forbade meta fields, even he couldn't remember why he did it!

So (pseudocode):

```ts
// Get the Type.field GraphQLField instance:
const Type_field = schema.getType('Type').getFields()['field'];

// Schema coord should point to this field instance:
assert.equal(resolveSchemaCoordinate(schema, `Type.field`), Type_field);
// Field instance should have this field coordinate.
assert.equal(Type_field.getSchemaCoordinate(), 'Type.field');
// Works both ways ✅

// Resolving the schema coordinate for `RootQueryType.__schema` could point to our meta definition:
assert.equal(resolveSchemaCoordinate(schema, `RootQueryType.__schema`), SchemaMetaFieldDef);
// But from this global we don't have access to the schema so we can't determine the root operation type name!
SchemaMetaFieldDef.getSchemaCoordinate() // 💢 ERROR! 
// The `RootQueryType.__schema` field is actually "virtual", so we can only resolve it one way! No cycle!
```

I'm not too worried about the two-way lookup not working; **however** `Foo.__typename` is the exact same "virtual" field as `Bar.__typename` (`resolveSchemaCoordinate('Foo.__typename') === resolveSchemaCoordinate('Bar.__typename')`), which breaks the "every schema element should have exactly one coordinate" rule.

So we have a few options:

1. Forbid all meta fields (as Lee proposed)
2. Forbid `__typename` only, and create new instances of `SchemaMetaFieldDef` and `TypeMetaFieldDef` for each schema
3. Create new instances of `SchemaMetaFieldDef` and `TypeMetaFieldDef` for every schema, and a new instance of `TypeNameMetaFieldDef` for every object type in every schema
4. Break the cycle _only_ for meta fields, and state in the spec that every _non-metafield_ schema element should have exactly one coordinate

I'm leaning towards option 4.

src/utilities/__tests__/resolveSchemaCoordinate-test.ts
benjie: The more I think about it the more this feels like the right answer. We shouldn't mandate `null` e.g. in Elixir you might want to return something like `:not_found` on failure vs `{:field, ...}` on success

magicmark: Addressed https://github.com/graphql/graphql-spec/pull/794/commits/5ac68a005b07b70a581574774cbd58b1971dd78a

magicmark: I'll leave it out for now to avoid more things to bikeshed, but I think it would be worth a note in the spec explaining this context above

magicmark:
> state in the spec that every non-metafield schema element should have exactly one coordinate

Done in https://github.com/graphql/graphql-spec/pull/794/commits/62ce19aafb0e0be105cd248e3e044d3f5103a897

src/language/__tests__/parser-test.ts
```diff
@@ -679,4 +685,118 @@ describe('Parser', () => {
...
+    });
+
+    it('rejects @ Name . Name', () => {
+      expect(() => parseSchemaCoordinate('@myDirective.field'))
+        .to.throw()
+        .to.deep.include({
+          message: 'Syntax Error: Expected <EOF>, found ".".',
+          locations: [{ line: 1, column: 13 }],
+        });
+    });
```
benjie, line 854: Please add parser tests for:

- `__Type`
- `Type.__metafield`
- `Type.__metafield(arg:)`

src/language/__tests__/printer-test.ts
benjie, line 329: Please add printer tests for:

- `__Type`
- `Type.__metafield`
- `Type.__metafield(arg:)`

src/utilities/__tests__/resolveSchemaCoordinate-test.ts
```diff
@@ -0,0 +1,239 @@
...
+    );
+
+    expect(() => resolveSchemaCoordinate(schema, '@unknown(arg:)')).to.throw(
+      'Expected "unknown" to be defined as a directive in the schema.',
+    );
+  });
+
+  it('resolves a meta-field', () => {
+    const type = schema.getType('Business') as GraphQLObjectType;
+    const field = schema.getField(type, '__typename');
```
benjie, line 196: The implementation of `schema.getField()` may change in the future to return undefined for metafields. Let's ensure the test throws in that case:

```suggestion
    const field = schema.getField(type, '__typename');
    assert.ok(field);
```

(Not sure if this is the correct code; but I've deliberately used `assert()` rather than `expect()` here to indicate it's an assumption of the test rather than something actually being tested _for_.)

Hopefully this one assertion means we won't need to add more in the other tests, but feel free to do so if you're so inclined.

src/utilities/__tests__/resolveSchemaCoordinate-test.ts
```diff
@@ -0,0 +1,239 @@
...
+    expect(resolveSchemaCoordinate(schema, '@private(unknown:)')).to.deep.equal(
+      undefined,
+    );
+
+    expect(() => resolveSchemaCoordinate(schema, '@unknown(arg:)')).to.throw(
+      'Expected "unknown" to be defined as a directive in the schema.',
+    );
+  });
+
+  it('resolves a meta-field', () => {
```
benjie, line 194:
```suggestion
});

/*
 * NOTE: the following are not required for spec compliance; resolution
 * of meta-fields is implementation-defined.
 *
 * These tests are here to ensure a change of behavior will only be made
 * in a semver-major release of GraphQL.js.
 */
   
describe('resolveSchemaCoordinate (meta-fields and introspection types)', () => {
  it('resolves a meta-field', () => {
```

src/utilities/resolveSchemaCoordinate.ts
```diff
@@ -0,0 +1,313 @@
...
+ * A resolved schema element may be one of the following kinds:
+ */
+export interface ResolvedNamedType {
+  readonly kind: 'NamedType';
+  readonly type: GraphQLNamedType;
+}
+
+export interface ResolvedField {
+  readonly kind: 'Field';
+  readonly type: GraphQLNamedType;
```
benjie: Is there a reason we're not being more specific? We do have assertions for this, I think?

```suggestion
  readonly type: GraphQLObjectType | GraphQLInterfaceType;
```

(May need `<unknown, unknown>` added.)

src/utilities/resolveSchemaCoordinate.ts
```diff
@@ -0,0 +1,313 @@
...
+  readonly field: GraphQLField<unknown, unknown>;
+}
+
+export interface ResolvedInputField {
+  readonly kind: 'InputField';
+  readonly type: GraphQLNamedType;
```
benjie:
```suggestion
  readonly type: GraphQLInputObjectType;
```

src/utilities/resolveSchemaCoordinate.ts
```diff
@@ -0,0 +1,313 @@
...
+  readonly inputField: GraphQLInputField;
+}
+
+export interface ResolvedEnumValue {
+  readonly kind: 'EnumValue';
+  readonly type: GraphQLNamedType;
```
benjie:
```suggestion
  readonly type: GraphQLEnumType;
```

src/utilities/resolveSchemaCoordinate.ts
```diff
@@ -0,0 +1,313 @@
...
+  readonly enumValue: GraphQLEnumValue;
+}
+
+export interface ResolvedFieldArgument {
+  readonly kind: 'FieldArgument';
+  readonly type: GraphQLNamedType;
```
benjie:
```suggestion
  readonly type: GraphQLObjectType | GraphQLInterfaceType;
```

src/utilities/resolveSchemaCoordinate.ts
```diff
@@ -0,0 +1,313 @@
...
+  | ResolvedField
+  | ResolvedInputField
+  | ResolvedEnumValue
+  | ResolvedFieldArgument
+  | ResolvedDirective
+  | ResolvedDirectiveArgument;
+
+/**
+ * A schema coordinate is resolved in the context of a GraphQL schema to
+ * uniquely identifies a schema element. It returns undefined if the schema
```
benjie:
```suggestion
 * uniquely identify a schema element. It returns undefined if the schema
```

src/utilities/resolveSchemaCoordinate.ts
```diff
@@ -0,0 +1,313 @@
...
+ * coordinate does not resolve to a schema element.
```
benjie:
```suggestion
 * coordinate does not resolve to a schema element, meta-field, or introspection
 * schema element. It will throw if the containing schema element (if applicable)
 * does not exist.
```

benjie (changes requested):
I think this is essentially good to go except for the comments noted, and that I think we should explicitly forbid {Ignored} from schema coordinates.

src/utilities/resolveSchemaCoordinate.ts
magicmark: I suppose this makes sense, `GraphQLUnionType` ∈ `GraphQLNamedType`, but `GraphQLUnionType` is not a valid type here.

benjie (approved):
Excellent work everyone, especially @magicmark!

PR #3059, opened 2021-05-05
Refine parse and AST to represent ConstValue

leebyron:
This adds:

* ConstValueNode - A subtype of ValueNode which recursively excludes Variables
* Improved syntax error when encountering variable in a const value
* parseConstValue(): ConstValueNode
* isConstValue(): ConstValueNode
* Various refinements to AST types to use ConstValueNode (or a type which includes it) to better align to the spec.


src/language/ast.js
```diff
@@ -379,23 +396,26 @@ export type EnumValueNode = {|
   +value: string,
 |};
 
-export type ListValueNode = {|
+export type ConstListValueNode = ListValueNode<ConstValueNode>;
+export type ListValueNode<+V = ValueNode> = {|
```
IvanGoncharov, line 400: @leebyron Can we do it like this:
```js
type _ListValueNode<+V> = {|
  +kind: 'ListValue',
  +loc?: Location,
  +values: $ReadOnlyArray<ValueNode>,
  +values: $ReadOnlyArray<V>,
|};

export type ListValueNode = _ListValueNode<ValueNode>;
export type ConstListValueNode = _ListValueNode<ConstValueNode>;
```

That way it's more clear to the reader that we just reusing type definition.
Also, we not exposing the ability to change this template argument as part of public API.

leebyron: I tried that first actually, but we have a linter that blocks the use of `_` in type names.

leebyron: I instead followed the pattern I used for `ConstArgumentNode` and `ConstDirectiveNode` which I think is easier to follow and addresses your concern

src/language/ast.js
```diff
@@ -385,19 +402,38 @@ export type ListValueNode = {|
   +values: $ReadOnlyArray<ValueNode>,
 |};
 
+export type ConstListValueNode = {|
+  +kind: 'ListValue',
```
IvanGoncharov, line 406: I understand why we use `ListValue` in both types.
But it means we lose the ability to distinguish them in runtime.
Also, it makes `isConstValueNode` the only predicate that recursive check nodes.

Can we do:
```js
export type ConstListValueNode = {|
  +kind: 'ListValue',
  +loc?: Location,
  +values: $ReadOnlyArray<ConstValueNode>,
  +isConst: true,
|};

export type ListValueNode = ConstListValueNode | {|
  +kind: 'ListValue',
  +loc?: Location,
  +values: $ReadOnlyArray<ValueNode>,
  +isConst: false,
|}
```

@leebyron What do you think?

leebyron: I'd rather not add Booleans to encode this since it's already knowable from the structure itself, and this would add two sources of truth that could disagree and would need to be carefully kept in sync during any edits.

For example parsing a value with parseValue() instead of parseConstValue() returns ValueNode instead of ConstValueNode since that's the safe type, but it might be const if it doesn't contain a variable. 

That means the parser would need an inlined version of isConstValue to set the booleans correctly. 

Another example is editing one with visit() to add or remove Variables would need to update these booleans. Right now that's not necessary

IvanGoncharov: @leebyron Argument about changing totally makes sense 👍

PR #3065, opened 2021-05-06
Add valueToLiteral()

leebyron:
Depends on #3077

**Provides the "Value to Literal" methods in this [data flow chart](https://user-images.githubusercontent.com/50130/118379946-51ac5300-b593-11eb-839f-c483ecfbc875.png).**

* Adds `valueToLiteral()` which takes an external input value and translates it to a literal, allowing for custom scalars to define this behavior.
    
**This also adds important changes to Input Coercion, especially for custom scalars:**
    
* The value provided to `parseLiteral` is now `ConstValueNode` and the second `variables` argument has been removed. For all built-in scalars this has no effect, but any custom scalars which use complex literals no longer need to do variable reconciliation manually (in fact most do not -- this has been an easy subtle bug to miss).
    
  This behavior is possible with the addition of `replaceVariables`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -416,6 +416,10 @@ export {
   /** A helper to use within recursive-descent visitors which need to be aware of the GraphQL type system. */
   TypeInfo,
   visitWithTypeInfo,
+  /** Converts a value to a const value by replacing variables. */
+  replaceVariables,
+  /** Create a GraphQL literal (AST) from a JavaScript input value. */
+  valueToLiteral,
   /** Coerces a JavaScript value to a GraphQL type, or produces errors. */
   coerceInputValue,
   /** Coerces a GraphQL literal (AST) to a GraphQL type, or returns undefined. */
--- a/src/type/__tests__/definition-test.ts
+++ b/src/type/__tests__/definition-test.ts
@@ -4,7 +4,7 @@ import { describe, it } from 'mocha';
 import { inspect } from '../../jsutils/inspect';
 import { identityFunc } from '../../jsutils/identityFunc';
 
-import { parseValue } from '../../language/parser';
+import { parseConstValue } from '../../language/parser';
 
 import type { GraphQLType, GraphQLNullableType } from '../definition';
 import {
@@ -83,15 +83,12 @@ describe('Type System: Scalars', () => {
       },
     });
 
-    expect(scalar.parseLiteral(parseValue('null'))).to.equal(
+    expect(scalar.parseLiteral(parseConstValue('null'))).to.equal(
       'parseValue: null',
     );
-    expect(scalar.parseLiteral(parseValue('{ foo: "bar" }'))).to.equal(
+    expect(scalar.parseLiteral(parseConstValue('{ foo: "bar" }'))).to.equal(
       'parseValue: { foo: "bar" }',
     );
-    expect(
-      scalar.parseLiteral(parseValue('{ foo: { bar: $var } }'), { var: 'baz' }),
-    ).to.equal('parseValue: { foo: { bar: "baz" } }');
   });
 
   it('rejects a Scalar type without name', () => {
@@ -137,6 +134,17 @@ describe('Type System: Scalars', () => {
     ).to.throw(
       'SomeScalar must provide both "parseValue" and "parseLiteral" functions.',
     );
+  });
+
+  it('rejects a Scalar type defining valueToLiteral with an incorrect type', () => {
+    expect(
+      () =>
+        new GraphQLScalarType({
+          name: 'SomeScalar',
+          // @ts-expect-error
+          valueToLiteral: {},
+        }),
+    ).to.throw('SomeScalar must provide "valueToLiteral" as a function.');
   });
 
   it('rejects a Scalar type defining specifiedByURL with an incorrect type', () => {
--- a/src/type/__tests__/scalars-test.ts
+++ b/src/type/__tests__/scalars-test.ts
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
 import { describe, it } from 'mocha';
 
-import { parseValue as parseValueToAST } from '../../language/parser';
+import { parseConstValue } from '../../language/parser';
 
 import {
   GraphQLID,
@@ -66,7 +66,7 @@ describe('Type System: Specified scalar types', () => {
 
     it('parseLiteral', () => {
       function parseLiteral(str: string) {
-        return GraphQLInt.parseLiteral(parseValueToAST(str), undefined);
+        return GraphQLInt.parseLiteral(parseConstValue(str));
       }
 
       expect(parseLiteral('1')).to.equal(1);
@@ -103,9 +103,6 @@ describe('Type System: Specified scalar types', () => {
       );
       expect(() => parseLiteral('ENUM_VALUE')).to.throw(
         'Int cannot represent non-integer value: ENUM_VALUE',
-      );
-      expect(() => parseLiteral('$var')).to.throw(
-        'Int cannot represent non-integer value: $var',
       );
     });
 
@@ -231,7 +228,7 @@ describe('Type System: Specified scalar types', () => {
 
     it('parseLiteral', () => {
       function parseLiteral(str: string) {
-        return GraphQLFloat.parseLiteral(parseValueToAST(str), undefined);
+        return GraphQLFloat.parseLiteral(parseConstValue(str));
       }
 
       expect(parseLiteral('1')).to.equal(1);
@@ -263,9 +260,6 @@ describe('Type System: Specified scalar types', () => {
       );
       expect(() => parseLiteral('ENUM_VALUE')).to.throw(
         'Float cannot represent non numeric value: ENUM_VALUE',
-      );
-      expect(() => parseLiteral('$var')).to.throw(
-        'Float cannot represent non numeric value: $var',
       );
     });
 
@@ -344,7 +338,7 @@ describe('Type System: Specified scalar types', () => {
 
     it('parseLiteral', () => {
       function parseLiteral(str: string) {
-        return GraphQLString.parseLiteral(parseValueToAST(str), undefined);
+        return GraphQLString.parseLiteral(parseConstValue(str));
       }
 
       expect(parseLiteral('"foo"')).to.equal('foo');
@@ -370,9 +364,6 @@ describe('Type System: Specified scalar types', () => {
       );
       expect(() => parseLiteral('ENUM_VALUE')).to.throw(
         'String cannot represent a non string value: ENUM_VALUE',
-      );
-      expect(() => parseLiteral('$var')).to.throw(
-        'String cannot represent a non string value: $var',
       );
     });
 
@@ -456,7 +447,7 @@ describe('Type System: Specified scalar types', () => {
 
     it('parseLiteral', () => {
       function parseLiteral(str: string) {
-        return GraphQLBoolean.parseLiteral(parseValueToAST(str), undefined);
+        return GraphQLBoolean.parseLiteral(parseConstValue(str));
       }
 
       expect(parseLiteral('true')).to.equal(true);
@@ -488,9 +479,6 @@ describe('Type System: Specified scalar types', () => {
       );
       expect(() => parseLiteral('ENUM_VALUE')).to.throw(
         'Boolean cannot represent a non boolean value: ENUM_VALUE',
-      );
-      expect(() => parseLiteral('$var')).to.throw(
-        'Boolean cannot represent a non boolean value: $var',
       );
     });
 
@@ -571,7 +559,7 @@ describe('Type System: Specified scalar types', () => {
 
     it('parseLiteral', () => {
       function parseLiteral(str: string) {
-        return GraphQLID.parseLiteral(parseValueToAST(str), undefined);
+        return GraphQLID.parseLiteral(parseConstValue(str));
       }
 
       expect(parseLiteral('""')).to.equal('');
@@ -603,9 +591,6 @@ describe('Type System: Specified scalar types', () => {
       );
       expect(() => parseLiteral('ENUM_VALUE')).to.throw(
         'ID cannot represent a non-string and non-integer value: ENUM_VALUE',
-      );
-      expect(() => parseLiteral('$var')).to.throw(
-        'ID cannot represent a non-string and non-integer value: $var',
       );
     });
 
--- a/src/type/definition.ts
+++ b/src/type/definition.ts
@@ -20,7 +20,6 @@ import { Kind } from '../language/kinds';
 import { print } from '../language/printer';
 import type {
   FieldNode,
-  ValueNode,
   ConstValueNode,
   OperationDefinitionNode,
   FragmentDefinitionNode,
@@ -590,8 +589,38 @@ export interface GraphQLScalarTypeExtensions {
  *         if (value % 2 === 1) {
  *           return value;
  *         }
+ *       },
+ *       parseValue(value) {
+ *         if (value % 2 === 1) {
+ *           return value;
+ *         }
+ *       }
+ *       valueToLiteral(value) {
+ *         if (value % 2 === 1) {
+ *           return parse(`${value}`);
+ *         }
  *       }
  *     });
+ *
+ * Custom scalars behavior is defined via the following functions:
+ *
+ *  - serialize(value): Implements "Result Coercion". Given an internal value,
+ *    produces an external value valid for this type. Returns undefined or
+ *    throws an error to indicate invalid values.
+ *
+ *  - parseValue(value): Implements "Input Coercion" for values. Given an
+ *    external value (for example, variable values), produces an internal value
+ *    valid for this type. Returns undefined or throws an error to indicate
+ *    invalid values.
+ *
+ *  - parseLiteral(ast): Implements "Input Coercion" for literals. Given an
+ *    GraphQL literal (AST) (for example, an argument value), produces an
+ *    internal value valid for this type. Returns undefined or throws an error
+ *    to indicate invalid values.
+ *
+ *  - valueToLiteral(value): Converts an external value to a GraphQL
+ *    literal (AST). Returns undefined or throws an error to indicate
+ *    invalid values.
  *
  */
 export class GraphQLScalarType extends GraphQLSchemaElement {
@@ -601,6 +630,7 @@ export class GraphQLScalarType extends GraphQLSchemaElement {
   serialize: GraphQLScalarSerializer<unknown>;
   parseValue: GraphQLScalarValueParser<unknown>;
   parseLiteral: GraphQLScalarLiteralParser<unknown>;
+  valueToLiteral: Maybe<GraphQLScalarValueToLiteral>;
   extensions: Maybe<Readonly<GraphQLScalarTypeExtensions>>;
   astNode: Maybe<ScalarTypeDefinitionNode>;
   extensionASTNodes: ReadonlyArray<ScalarTypeExtensionNode>;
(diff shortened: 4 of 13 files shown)


leebyron:
@andimarek this is my attempt at introducing the version of "toLiteral" we discussed the other day.

leebyron:
Open for discussion for a future change is whether `replaceASTVariables` should even exist. I'm not thrilled with it, but it's necessary to be a non-breaking change.

We currently allow variables to exist in the middle of a complex custom scalar:

```graphql
query ($var: String) {
  example(arg: { field: $var })
}

type Query {
  example(arg: CustomScalar): String
}

scalar CustomScalar
```

This currently passes validation ["All Variable Usages are Allowed"](https://spec.graphql.org/draft/#sec-All-Variable-Usages-are-Allowed) because if we do not know the expect type of a variable's position then we allow it - and the position in the example above is contained within a scalar - it does not have an expected type.

The validation rule in the spec does not specify this behavior, it just [assumes a variable position has a type](https://spec.graphql.org/draft/#sel-IALbLDFCDCAACKM3gG) which based on this example is not necessarily a safe assumption.

We should probably make this behavior clear. Either by allowing it in the spec (and support this existing behavior) or disallow it in the spec making the above example invalid (which would allow us to assert that anything passed to `parseLiteral` is already `ConstValue` and therefore `replaceASTVariables` could be removed)

andimarek:
> Open for discussion for a future change is whether `replaceASTVariables` should even exist. I'm not thrilled with it, but it's necessary to be a non-breaking change.
> 
> 
> 
> We currently allow variables to exist in the middle of a complex custom scalar:
> 
> 
> 
> ```graphql
> 
> query ($var: String) {
> 
>   example(arg: { field: $var })
> 
> }
> 
> 
> 
> type Query {
> 
>   example(arg: CustomScalar): String
> 
> }
> 
> 
> 
> scalar CustomScalar
> 
> ```
> 
> 
> 
> This currently passes validation ["All Variable Usages are Allowed"](https://spec.graphql.org/draft/#sec-All-Variable-Usages-are-Allowed) because if we do not know the expect type of a variable's position then we allow it - and the position in the example above is contained within a scalar - it does not have an expected type.
> 
> 
> 
> The validation rule in the spec does not specify this behavior, it just [assumes a variable position has a type](https://spec.graphql.org/draft/#sel-IALbLDFCDCAACKM3gG) which based on this example is not necessarily a safe assumption.
> 
> 
> 
> We should probably make this behavior clear. Either by allowing it in the spec (and support this existing behavior) or disallow it in the spec making the above example invalid (which would allow us to assert that anything passed to `parseLiteral` is already `ConstValue` and therefore `replaceASTVariables` could be removed)

The most popular example is the JSON Scalar which can contain variables in its literals. I don't really think we can break that behavior.

leebyron:
That's the main use case I'm aware of, but I'm curious how common it is - it seems weird to allow one well-defined scalar to be embedded within another well-defined scalar. There's almost certainly going to be some weird edge-case behavior in there. From a usefulness point of view - I see the merit. From a clarity of execution and option-value preservation point of view, it seems wise to restrict complex scalars to be always fully formed rather than provided in pieces

We should probably clarify that in the spec either way.

andimarek:
I want to point out a challenge regarding validation here:

This query:
```graphql
query($var: ID){hello(arg: {a: $var})}
```

should be invalid for this schema:
```graphql
  scalar JSON

  type Query {
    hello(arg: JSON): ID
  }
```
at least according to the implementation of `parseLiteral` in the custom scalar (https://github.com/taion/graphql-type-json/blob/master/src/index.js#L45) because `ID` is not an allowed Literal for a JSON value.

But because it is passed in as variable the custom Scalar just accepts the coerced variable.

leebyron:
Yeah you've got this exactly right. What does it mean to have a variable within a leaf value? There is no way to represent that in the type system.

src/utilities/replaceASTVariables.d.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Replaces any Variables found within an AST Value literal with literals
+ * supplied from a map of variable values, returning a constant value.
+ *
+ * Used primarily to ensure only complete constant values are used during input
+ * coercion of custom scalars which accept complex literals.
+ */
+export function replaceASTVariables(
+  valueNode: ValueNode,
+  variables: Maybe<ObjMap<unknown>>,
```
andimarek, line 15: just to be clear here: these are the raw (uncoerced variables) right?

andimarek:
One important detail: 

What should the formal relationship between `valueToLiteral` and `literalToValue` be?

For example for ID: This code does `valueToLiteral(literalToValue(AstInt(123))) == AstInt(123)`
but also `valueToLiteral(literalToValue(AstString("123"))) == AstInt(123)`

This means `valueToLiteral` maps two different values to the same Literal (In math speak: it is not injective) and both functions are not the inverse to each other. 

If we want to make both functions the inverse to each other we would require that different literals always produce different external input values and different external input values produce different literals. I don't think we can do that as this is a constraint we have not had so far and we would break existing code.

This probably means we require that for every valid literal `x` `literalToValue(valueToLiteral(literalToValue(x)))` is possible and for every valid input value `y` `valueToLiteral(literalToValue(valueToLiteral(y)))` is possible, but not more.

src/utilities/replaceASTVariables.js
```diff
@@ -0,0 +1,36 @@
...
+ * Used primarily to ensure only complete constant values are used during input
+ * coercion of custom scalars which accept complex literals.
+ */
+export function replaceASTVariables(
+  valueNode: ValueNode,
+  variables: ?ObjMap<mixed>,
+): ConstValueNode {
+  return visit(valueNode, {
+    Variable(node) {
+      return valueToLiteral(variables?.[node.name.value]);
```
andimarek, line 22: we need to consider non provided variables values: a variables reference ast reference would need to completely removed I think to emulate a non existing value (and not just null)

src/utilities/valueFromAST.js
```diff
@@ -129,12 +132,16 @@ export function valueFromAST(
...
   if (isLeafType(type)) {
+    const constValueNode = replaceASTVariables(valueNode, variables);
+
     // Scalars and Enums fulfill parsing a literal value via parseLiteral().
     // Invalid values represent a failure to parse correctly, in which case
     // no value is returned.
     let result;
     try {
-      result = type.parseLiteral(valueNode, variables);
+      result = type.parseLiteral
```
andimarek, line 142: how do I need to read this? if we have an existing `parseLiteral` we use it? I would expect that it would be: if we have a `literalToValue` we use it, otherwise we fall back

andimarek:
I want to state that I think getting rid of variable references inside complex ast literal is the right way to preserve the current behaviour while making custom scalar implementations easier and more robust. Great work @leebyron

src/utilities/valueToLiteral.js
```diff
@@ -0,0 +1,97 @@
...
+
+  // Does this type (if provided) define `valueToLiteral` which returns a value?
+  if (isLeafType(namedType) && namedType.valueToLiteral != null) {
+    const literal = namedType.valueToLiteral(value);
+    if (literal) {
+      return literal;
+    }
+  }
+
+  // Otherwise, perform a JS-to-Literal default equivalency.
```
andimarek, line 58: we should make clear that this is deprecated, possibility wrong behaviour and Scalars should define their appropriate logic.

src/utilities/valueToLiteral.js
```diff
@@ -0,0 +1,97 @@
...
+
+  // Convert JavaScript array to GraphQL list.
+  if (isIterableObject(value)) {
+    return {
+      kind: Kind.LIST,
+      values: Array.from(value, (item) => valueToLiteral(item, namedType)),
+    };
+  }
+
+  // Does this type (if provided) define `valueToLiteral` which returns a value?
```
andimarek, line 50: what about input object types? Does InputObject define a `valueToLiteral`? It either should or we should define the appropriate logic here.

src/utilities/replaceASTVariables.js
leebyron: Correct, the section on "ObjectValue" below performs that task, the last case in replaceASTVariables-test.js tests the behavior

src/utilities/valueFromAST.js
leebyron: `valueFromAST` is an older function that is deprecated in #3049 by `coerceInputLiteral`. It actually does "input coercion" via `parseLiteral` similar to how `coerceInputValue` does with `parseValue`.

Reading this:

* If `parseLiteral` exists on `type`:
  * Call `parseLiteral` on `type` with `constValueNode`
* Otherwise:
  * Call the fallback implementation of `parseLiteral` on `type`
    * Where that fallback implementation is defined as:
    * Convert `constValueNode` to `value` with `literalToValue` using `type`
    * Call `parseValue` on `type` with `value`

src/utilities/valueToLiteral.js
leebyron: No, only leaf types, similar to coercion. Input Objects should never have overridden behavior they should always use the behavior described here.

src/utilities/valueToLiteral.js
leebyron: This is not deprecated, these are the default conversions. Other languages might implement these defaults differently - or require all scalars to define this if a default is not well defined.

For Javascript, it's natural to work with Javascript objects, numbers, strings, booleans as the primary domain of data.

src/utilities/valueFromAST.js
leebyron: Previously this fallback implementation was provided in definition.js, but doing this creates a circular dependency now that `literalToValue` uses type information.

Here's the replaced logic from that file:

```js
this.parseLiteral =
     config.parseLiteral ??
     ((node, variables) => parseValue(valueFromASTUntyped(node, variables)));
```

You can see how `valueFromASTUntyped(node, variables)` has been replaced by `replaceASTVariables(valueNode, variables)` and `literalToValue(constValueNode, type)`

src/utilities/replaceASTVariables.js
andimarek: 👍

src/utilities/replaceASTVariables.d.ts
andimarek: We are calling `valueToLiteral` [here](https://github.com/graphql/graphql-js/pull/3065/files#diff-39c231e63679a9d86e346550424c6deeba68d4b4685b0cf044ec5d3d0bb5314cR22) on the variables which means from my understanding it must be raw variables, as `valueToLiteral` is external input value (aka raw variables) to Literal.

src/utilities/valueToLiteral.js
```diff
@@ -0,0 +1,97 @@
...
+ * | Object            | Input Object         |
+ * | Array             | List                 |
+ * | Boolean           | Boolean              |
+ * | String            | String               |
+ * | Number            | Int / Float          |
+ * | null / undefined  | Null                 |
+ *
+ * Note: This function does not perform any type validation or coercion.
+ */
+export function valueToLiteral(
```
andimarek, line 31: I think this method tries to be too clever: the type argument should never be undefined and the whole conversion logic should be based on the type.

If it is a List type => take the value apart and perform it per value.
If Object type => iterate over field definitions and call it per field
If Scalar/Enum => call the `valueToLiteral` on its type definition.

If you have an input object as type and the value is a boolean it should raise an error: at the moment it returns a `boolean` literal which is wrong.

src/utilities/literalToValue.js
```diff
@@ -0,0 +1,79 @@
...
+ * | Input Object         | Object           |
+ * | List                 | Array            |
+ * | Boolean              | Boolean          |
+ * | String / Enum        | String           |
+ * | Int / Float          | Number           |
+ * | Null                 | null             |
+ *
+ * Note: This function does not perform any type validation or coercion.
+ */
+export function literalToValue(
```
andimarek, line 31: similar to other `valueToLiteral` this tries too much: we need a version of this which is definitely type based and does not infer logic from the value: an input object value should never have fields for example which are not defined in the type.

src/utilities/valueFromAST.js
andimarek: I think replacing `valueFromASTUntyped` with `literalToValue` is not correct: `literalToValue` and  consequently `replaceASTVariables` needs to always have type informations to act on.

andimarek:
Looking at your flow chat: I fully agree regarding the new well defined behavior defined by `valueToLiteral` and `literalToValue`. 👍 

I commented inline, but want to highlight it here again: I think we are trying to conserve some not well defined behavior which we don't have to: `valueFromAst` (and the revsere)  should be deprecated and left alone imho. It should never be used anymore inside GraphQL.js execution itself as fas as I can tell.

src/utilities/valueFromAST.js
leebyron: What do you propose instead? I'd like to remove all uses of `valueFromASTUntyped` and `literalToValue` is designed as it's replacement

src/utilities/valueToLiteral.js
leebyron: I thought about this, but the problem with this approach is that we're operating on "external" pre-coerced, pre-validated values and it's entirely possible that those values do not match the types. These conversion functions are used before validation takes place in #3049. That may require us to define validation rules again in each of these functions, ideally they can be passed through so that we define validation once.

leebyron: We also need reasonable defaults for when we don't have a type, since that's going to be common for any custom scalars with complex definitions (like the JSON type)

leebyron:
`valueFromAst` is misnamed, as it does "uncoercion" and needs to stay in use until we can fix default values in #3049 - it's deprecated in that PR

src/utilities/valueToLiteral.js
andimarek: Took me a bit to wrap my head around, but I think our misunderstanding comes from a fundamental principles about   `Scalar.literalToValue`/`Scalar.valueToLiteral` does: I think these two methods should follow the same principles as `parseValue/parseLiteral` and should throw errors if something is wrong. I want basically another "coercing" (already a bit overloaded term btw) which converts from Value and Literal and vice verse. 

Looking at the `ID` implementation in this PR I see that exactly this is missing: currently it returns `undefined` instead of throwing `GraphQLError`. Building on this principle I think method is basically the equivalent of `coerceInputValueImpl` but going from `value` to `literal` .

The reason I want this: for example for a JSON `literalToValue` method : you would want to throw a clear error message for an expected Literal deep inside the provided `literal`. What is the alternative? Return `undefined`? Silently ignore?

There is also general value in having a well defined, validated literal or external value for inputs. 

And as you mentioned: this would mean validated twice, yes. But I don't see logic defined twice really: all validation logic resided inside the Scalar except for Input Objects. And I am sure we can extract/generalize this logic if that is a concern.

To be specific: we check for two things when validating input objects: if all required fields are provided and not more fields are provided.

leebyron: I don't necessarily disagree with that. I think being more strict is useful, and I'm sure we could find a way to generalize the validation portion.

The root of the issue is still variables within custom scalars. We have no types there and all bets are off. As long as we decide to preserve that behavior we need untyped transform

andimarek: I don't think I follow completely regarding the variables:

We actually have a type of the variables and we could use that type to convert the External input value into an Ast Literal in a well defined manner. We could then simply replace the VariableReference ast node with the Ast Literal we got. 

Custom Scalar implementation can't type check variables values anyway at the moment: they just get a variable and they basically take the value. 

I think that would preserver the current behavior.

src/utilities/valueFromAST.js
andimarek: There is no type information for the `field` correct, but there is for `$var`: the type of the variable. As discussed if this is matches is currently not guaranteed by the type system. But we can safely and in well defined manner translate `$var` into an literal and then replace `$var` with it

leebyron: That's a great point. I think implementing this will require a bit more surgery since right now we pass around "variables" as a map of name to coerced value, but to do this we need to preserve the original external value and type.

andimarek: I understand that this might not be implemented trivially, but this ties into the other topic regarding `valueToLiteral/literalToValue` operating always on type which I think is really needed to produce a correct behavior for custom Scalars.

leebyron: My comment above is now implemented via #3077

andimarek:
# literal to value needed?
I changed my mind a bit and I am not convinced `literalToValue` is worth adding after going through GraphQL Java and looking at this PR: `literalToValue` is only used by `valueFromAst` as fallback if a Scalar doesn't define `parseLiteral`. And inside `literalToValue` we have a fallback for Scalars which don't  define `literalToValue`. But if a Scalar doesn't define `parseLiteral` it will very likely also not define `literalToValue` as the latter is only added now. Which means `literalToValue` is no really needed for `valueFromAst`. 

The only real usage I can see is that users want to convert the default value from external input values to literals. 

Related to that: In GraphQL Java we plan to offer to set defaultValues either as Literal or external input value. GraphQL.js will only offer the former, correct? If you would offer to set the default value as external input value too, there would be no need anymore for `literalToValue` afaiu. 

# Error handling

When will the default values literals be validated? I think we need a schema validation for that, as we currently only have Query validation level.

leebyron:
>  I am not convinced literalToValue is worth adding. literalToValue is only used by valueFromAst as fallback if a Scalar doesn't define parseLiteral.

Yeah, it's really feels there for completeness. Especially after the stack of diffs before this one, this is the sole use case.

> But if a Scalar doesn't define parseLiteral it will very likely also not define literalToValue as the latter is only added now. Which means literalToValue is no really needed for valueFromAst.

Definitionally true since this PR hasn't landed yet :)

An alternative would be to revert back to the previous behavior of using `valueFromASTUntyped` (which is more or less what we now have defined as `defaultScalarLiteralToValue` - to your point) and require `parseLiteral` to be defined if `literalToValue` is also defined. This would avoid a scenario where a custom scalar defines `parseLiteral` but not `literalToValue` which seems to be the only purpose for this function at the moment.

> Related to that: In GraphQL Java we plan to offer to set defaultValues either as Literal or external input value. GraphQL.js will only offer the former, correct? If you would offer to set the default value as external input value too, there would be no need anymore for literalToValue afaiu.

This is introduced earlier in the PR stack in #3074. Once the entire stack is complete you will be able to provide a default value as a literal, an external input value, or (deprecated existing cases) as an assumed-coerced internal input value.

> When will the default values literals be validated? I think we need a schema validation for that, as we currently only have Query validation level. Reading the code it seems that valuteToLiteral assumes that the external value is valid.

This isn't done in this PR, but in the one stacked above it in #3049. Since existing default values are assumed coerced internal input values, there's nothing to validate just yet. Once we can get them defined properly, then we should validate them as part of schema validation, which is part of that RFC.

`valueToLiteral` does not assume the external value is valid, it returns `undefined` if it encounters a value that cannot be converted to a literal in a well defined way according to the provided type.

> Documenting constraints

Agreed. I'll expand on the documentation for these methods

andimarek:
> This is introduced earlier in the PR stack in #3074. Once the entire stack is complete you will be able to provide a default value as a literal, an external input value, or (deprecated existing cases) as an assumed-coerced internal input value.

Sounds good, but then you need to differentiate between these three states in `getCoercedDefaultValue`. I can't see this happening so far, but maybe I missing something?

leebyron:
Yeah it's a bit hard to follow since defaultValues are still stored as coerced internal values. `getCoercedDefaultValue` checks both cases: first that `.value` is defined (which until #3049 is coerced internal value), then that `.literal` is defined

leebyron:
@andimarek - removed `literalToValue` and improved on documentation. Proper schema validation comes in a later PR

yaacovCR:
Merged as #3812

PR #3071, opened 2021-05-10
feat(subscribe): allow providing custom execute function for subscribe

n1ru4l:
This allows solving https://github.com/graphql/graphql-js/issues/894 and https://github.com/enisdenjo/graphql-ws/pull/180 in user-land by providing a custom execute function for `subscribe`, that creates the `contextValue` per published value (for each `execute` invocation), instead of once for the whole subscription.

### DIFF
--- a/src/subscription/subscribe.d.ts
+++ b/src/subscription/subscribe.d.ts
@@ -1,7 +1,7 @@
 import { Maybe } from '../jsutils/Maybe';
 
 import { DocumentNode } from '../language/ast';
-import { ExecutionResult } from '../execution/execute';
+import { execute, ExecutionResult } from '../execution/execute';
 import { GraphQLSchema } from '../type/schema';
 import { GraphQLFieldResolver } from '../type/definition';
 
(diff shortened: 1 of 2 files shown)


97amarnathk:
I too have a usecase where I need to invalidate dataloader cache after each execution result.  This will be quite helpful.

n1ru4l:
@97amarnathk feel also free to check out https://github.com/dotansimha/envelop/pull/183 which kind of tries to work around this limitation

dotansimha:
Interesting. @IvanGoncharov can you please review this PR? :)

dotansimha (approved):

n1ru4l:
Since this will probably not be reviewed or released for a long-time, I built a solution that can be easily installed as a envelop plugin: https://www.envelop.dev/docs/guides/resolving-subscription-data-loader-caching-issues

pantajoe:
Question: Is there any reason this does not get merged?
It'd be great to be able to use `dataloader`'s cache again when using subscriptions.

yaacovCR:
#4211 solves this slightly differently for v17, avoiding revalidation of arguments/recoercion of variables

PR #3074, opened 2021-05-10
Preserve defaultValue literals

### LINKED ISSUE #3051: Default value declared via SDL may not introspect correctly
> One interesting aspect here is that a default value declared via SDL is coerced twice:
> Once via `valueFromAST` and then again via `coerceInputValueImpl`. Is my reading here correct?
>
> I am not really sure this is a problem, but I wanted to call that out.
>
> _Originally posted by @andimarek in https://github.com/graphql/graphql-js/issues/3049#issuecomment-827036666_

That's a good call out. Coercing more than once should be safe, assuming that `coerce(X) == coerce(serialize(coerce(X)))` 

However I think this is likely exposing a real (but maybe low-impact) bug, even before this diff. Default values are exposed via introspection and are not expected to include nested default values:

```graphql
input A {
  a: B = {}
}

input B {
  c: Int = 0
}
```

When introspecting `A.a`, we should see `"defaultValue": "{}"` but if I read correctly likely will see `"defaultValue": "{c:0}"`

If you were to re-print this SDL after introspection, you'd have:

```graphql
input A {
  a: B = { c: 0 }
}
```

which broke the original intent.

Executed results would be correct, however schema representation would be off.

Notably, constructing the schema programmatically would not encounter this bug - only using AST creation would because of how `valueFromAST` works.

leebyron:
Depends on #3092
Fixes #3051

This change solves the problem of default values defined via SDL not always resolving correctly through introspection by preserving the original GraphQL literal in the schema definition. This changes argument and input field definitions `defaultValue` field from just the "value" to a new `GraphQLDefaultValueUsage` type which contains either or both "value" and "literal" fields.

Since either of these fields may be set, new functions for resolving a value or literal from either have been added - `getLiteralDefaultValue` and `getCoercedDefaultValue` - these replace uses that either assumed a set value or were manually converting a value back to a literal.

Here is the flow for how a default value defined in an SDL would be converted into a functional schema and back to an SDL:

**Before this change:**

```
(SDL) --parse-> (AST) --coerceInputLiteral--> (defaultValue config) --valueToAST--> (AST) --print --> (SDL)
```

`coerceInputLiteral` performs coercion which is a one-way function, and `valueToAST` is unsafe and set to be deprecated in #3049.

**After this change:**

```
(SDL) --parse-> (defaultValue literal config) --print --> (SDL)
```

### DIFF
--- a/src/execution/values.ts
+++ b/src/execution/values.ts
@@ -23,6 +23,7 @@ import { typeFromAST } from '../utilities/typeFromAST';
 import {
   coerceInputValue,
   coerceInputLiteral,
+  coerceDefaultValue,
 } from '../utilities/coerceInputValue';
 
 type CoercedVariableValues =
@@ -177,8 +178,11 @@ export function getArgumentValues(
     const argumentNode = argNodeMap[name];
 
     if (!argumentNode) {
-      if (argDef.defaultValue !== undefined) {
-        coercedValues[name] = argDef.defaultValue;
+      if (argDef.defaultValue) {
+        coercedValues[name] = coerceDefaultValue(
+          argDef.defaultValue,
+          argDef.type,
+        );
       } else if (isNonNullType(argType)) {
         throw new GraphQLError(
           `Argument ${argDef} of required type ${argType} was not provided.`,
@@ -197,8 +201,11 @@ export function getArgumentValues(
         variableValues == null ||
         !hasOwnProperty(variableValues, variableName)
       ) {
-        if (argDef.defaultValue !== undefined) {
-          coercedValues[name] = argDef.defaultValue;
+        if (argDef.defaultValue) {
+          coercedValues[name] = coerceDefaultValue(
+            argDef.defaultValue,
+            argDef.type,
+          );
         } else if (isNonNullType(argType)) {
           throw new GraphQLError(
             `Argument ${argDef} of required type ${argType} ` +
--- a/src/index.ts
+++ b/src/index.ts
@@ -181,6 +181,7 @@ export type {
   GraphQLScalarSerializer,
   GraphQLScalarValueParser,
   GraphQLScalarLiteralParser,
+  GraphQLDefaultValueUsage,
 } from './type/index';
 
 /** Parse and operate on GraphQL language source files. */
--- a/src/type/__tests__/definition-test.ts
+++ b/src/type/__tests__/definition-test.ts
@@ -826,6 +826,69 @@ describe('Type System: Input Objects', () => {
       );
     });
   });
+
+  describe('Input Object fields may have default values', () => {
+    it('accepts an Input Object type with a default value', () => {
+      const inputObjType = new GraphQLInputObjectType({
+        name: 'SomeInputObject',
+        fields: {
+          f: { type: ScalarType, defaultValue: 3 },
+        },
+      });
+      expect(inputObjType.getFields()).to.deep.equal({
+        f: {
+          coordinate: 'SomeInputObject.f',
+          name: 'f',
+          description: undefined,
+          type: ScalarType,
+          defaultValue: { value: 3 },
+          deprecationReason: undefined,
+          extensions: undefined,
+          astNode: undefined,
+        },
+      });
+    });
+
+    it('accepts an Input Object type with a default value literal', () => {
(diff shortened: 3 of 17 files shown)


leebyron:
I'll follow up with additional test coverage - but would love a review on this solution and approach

andimarek:
I think that is the right direction and as far as I understand you want that merged before the other bigger changes.

The more comprehensive end state we are aiming for In GraphQL Java is this:

```java
public enum ValueState {
    NOT_SET,
    LITERAL,
    EXTERNAL_VALUE,
    INTERNAL_VALUE // this is deprecated and should not be used going forward, will be removed in the future
}
```
to capture the different states of the default value:
`NOT_SET` you probably don't need because you JS has `undefined`.

This is needed because we need to differentiate between external input value which will be converted to `literal` via `valueToLiteral` and the legacy behaviour of using a not well defined already coerced internal value, which will continue to use  `astFromValue`.

src/type/defaultValues.js
```diff
@@ -0,0 +1,28 @@
...
+  invariant(literal, 'Value cannot be converted to literal for this type');
+  return literal;
+}
+
+export function getCoercedDefaultValue(
+  usage: GraphQLDefaultValueUsage,
+  type: GraphQLInputType,
+): mixed {
+  return usage.value !== undefined
+    ? usage.value
```
yaacovCR, line 26: Looks to me like valueFromAST gets called during execution, could you call it at schema build time and just save the result as the defaultValue.value.

Just curious also, why do you save the default value node separately as part of the type if the entire ast is also preserved, why not just read from there?

leebyron: Calling `valueFromAST` during schema construction causes a circular dependency that can result in some TDZ issues. Also it's possible that the value is invalid, and schema validation occurs after schema construction, so we need to preserve the pre-coerced values.

We save the default value literal AST separately to simplify the logic - otherwise detecting a defaultValue would require inspecting the configuration as well as the SDL's AST - this moves it to a single source of truth. It also allows a default value literal to be provided outside the context of SDL-based construction which is at least useful for writing unit tests and may also be useful in future cases as well.

yaacovCR: I see your points, but this still does not seem right to me. Taking a different tack, if node/ast SDL is so separate, we should not have separate literal preserved at all and scalars that require uncoercion would have functions defined that allow that or default to serialize

i thought that your other related PRs were headed in that direction, but this seems to suggest not, and then makes ast values canonical for default values, which seems less optimal to me from a purist standpoint

i am not sure I am a purist, so, I could ask this the other way, why can’t we then preserve directive ast nodes on the type as well (or maybe just a subset of them that are marked in some way) one argument was the SDL / code separation.

Anyway, this seems overall inconsistent. I think a refactor with a staged build of types from ast in which default values are added last might help. If the default value is not valid, but you want to preserve the ast, you could save it in the defaultValue as an Error or have another way of storing it within the schema.

it just doesn’t make sense to me to delay until runtime something that can be calculated at schema construction time. It’s a sign that something is off imo

yaacovCR: Maybe a comment “todo: add memoization if/when possible” like elsewhere in code base. Thanks for the discussion, certainly been clarifying.

i meant mostly for schema, not query directives.

as an aside, not sure yet how these PRs affect schema directive default value processing generally

leebyron:
Updated to do proper memoization and error if multiple sources of truth are presented in a config

jjsanchz46 (approved):

src/type/defaultValues.js
yaacovCR: Just a little unnecessary follow-up. I have spent new few hours trying to put my money where my mouth and implementing an alternative solution and was humbled quite quickly by circular references and preserving literals that error on coercion, and definitely now see the benefit of consolidating the literal with the default value separately from the AST considering how often the code has to check for the existence of either, i.e. all the points that @leebyron brought up.

Thanks @leebyron very much for indulging my comments and teaching me along the way about some of the goals of GraphQL, the reference implemationation, etc, etc.

There may be a better way of doing this, but it certainly is not as simple as I thought. The solution here now seems much more elegant than it did to me earlier. Huge apologies.

PR #3077, opened 2021-05-11
Preserve sources of variable values

leebyron:
Depends on #3074

By way of introducing type `VariableValues`, allows `getVariableValues` to return both the coerced values as well as the original sources, which are then made available in `ExecutionContext`.

While variable sources are not used directly here, they're used directly in #3065. This PR is pulled out as a pre-req to aid review


yaacovCR:
Merged as #3811

PR #3086, opened 2021-05-13
Input Value Validation

leebyron:
Depends on #3065

Factors out input validation to reusable functions:

* Introduces `validateInputLiteral` by extracting this behavior from `ValuesOfCorrectTypeRule`.
* Introduces `validateInputValue` by extracting this behavior from `coerceInputValue`
* Simplifies `coerceInputValue` to return early on validation error
* Unifies error reporting between `validateInputValue` and `validateInputLiteral`, causing some error message strings to change, but error data (eg locations) are preserved.

These two parallel functions will be used to validate default values in #3049

Potentially breaking if you rely on the existing behavior of `coerceInputValue` to call a callback function, as the call signature has changed. GraphQL behavior should not change, though error messages are now slightly different.

### DIFF
--- a/src/execution/__tests__/nonnull-test.ts
+++ b/src/execution/__tests__/nonnull-test.ts
@@ -643,7 +643,7 @@ describe('Execute: handles non-nullable types', () => {
         errors: [
           {
             message:
-              'Argument Query.withNonNullArg(cannotBeNull:) of non-null type String! must not be null.',
+              'Argument Query.withNonNullArg(cannotBeNull:) has invalid value: Expected value of non-null type String! not to be null.',
             locations: [{ line: 3, column: 42 }],
             path: ['withNonNullArg'],
           },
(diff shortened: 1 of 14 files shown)


andimarek:
@leebyron question: are you going to validate default argument values of Directive definitions and  arguments of applied Directives too?

For example:

```graphql
# custom Scalar
scalar CacheKey

directive @cached(
  key: CacheKey = {key: "DEFAULT", duration=1000}
) on FIELD_DEFINITION 

type Query {
  hello: String @cached(key: {key: "hello", duration = 500})
}
```

leebyron:
Yes both. Though one additional follow up might be necessary. 

Arguments of applied directives in _operations_ have always been validated in the ValuesOfCorrectType validator. In schema they never translate into something so they haven't been validated. This seems like an easy thing to fix. 

Directive definition argument default value validation is in the RFC spec change and included in the last PR in this stack

yaacovCR:
Merged rebased version of this PR as #3813.

PR #3092, opened 2021-05-15
Add coerceInputLiteral()

leebyron:
Depends on #3145

Removes `valueFromAST()` and adds `coerceInputLiteral()` as an additional export from `coerceInputValue`.

The implementation is almost exactly the same as `valueFromAST()` with a slightly more strict type signature and refactored tests to improve coverage (the file unit test has 100% coverage)

While this does not change any behavior, it could be breaking if you rely directly on the `valueFromAST()` method. Use `coerceInputLiteral()` as a direct replacement.


src/utilities/buildClientSchema.js
```diff
@@ -369,7 +370,10 @@ export function buildClientSchema(
 
     const defaultValue =
       inputValueIntrospection.defaultValue != null
-        ? valueFromAST(parseValue(inputValueIntrospection.defaultValue), type)
+        ? coerceInputLiteral(
+            parseValue(inputValueIntrospection.defaultValue),
```
leebyron, line 374: ParseConstValue?

yaacovCR:
Merged as #3809

PR #3109, opened 2021-05-17
scalars: ID and String support for serializing BigInt

Hkmu:
Add support to serialize BigInt for String and ID scalar.


yaacovCR:
I think it would make sense to use a custom scalar for BigInt a la https://the-guild.dev/graphql/scalars/docs/scalars/big-int

yaacovCR:
I misunderstood, this is for coercion of values from BigInt to string/id, not to int, this is reasonable, see discussion at newer PR https://github.com/graphql/graphql-js/pull/4088

PR #3115, opened 2021-05-19
Refactor Lexer

leebyron:
The lexer needed some cleanup, I found myself doing this as part of a Unicode RFC, but factoring all that out to make the Unicode RFC PR easier to follow.

* Always use hexadecimal form for code values.
* Remove use of `isNaN` for checking source over-reads.
* Defines `isSourceCharacter`
* Add more documentation and comments, also replaces regex with lexical grammar
* Simplifies error messages
* Adds additional tests

### DIFF
--- a/src/language/__tests__/lexer-test.ts
+++ b/src/language/__tests__/lexer-test.ts
@@ -30,12 +30,12 @@ function expectSyntaxError(text: string) {
 describe('Lexer', () => {
   it('disallows uncommon control characters', () => {
     expectSyntaxError('\u0007').to.deep.equal({
-      message: 'Syntax Error: Cannot contain the invalid character "\\u0007".',
+      message: 'Syntax Error: Invalid character: U+0007.',
       locations: [{ line: 1, column: 1 }],
     });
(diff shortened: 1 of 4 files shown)


IvanGoncharov:
@leebyron It's hard to review since a lot of changes happened simultaneously.
I also spotted some minor issues I like to fix.
So if you don't mind I will extract some of the changes in separate PRs?

leebyron:
I'm happy to factor some things out if it helps

IvanGoncharov:
> I'm happy to factor some things out if it helps

@leebyron Thanks, a lot it would be helpful.
I got carried away for a bit so I already extracted 5 commits and create one of my own 😊 
I will push those in a branch for you to review.

IvanGoncharov:
@leebyron Can you please take a look at #3155?

leebyron:
Applied most of your suggested changes and a few other related clarifying changes in https://github.com/graphql/graphql-js/pull/3115/commits/0d6a09673ee813937c614d2914f76245969a3662

PR #3117, opened 2021-05-19
RFC: Support full Unicode in lexer

leebyron:
Depends on #3115

Implements RFC at graphql/graphql-spec#849.

* Replaces `isSourceCharacter` with `isUnicodeScalarValue`
* Adds `isSupplementaryCodePoint`, used in String, BlockStrings, and Comments to ensure correct lexing of JavaScript's UTF-16 source.
* Updates `printCodePointAt` to correctly print supplementary code points.
* Adds variable-width Unicode escape sequences
* Adds explicit support for legacy JSON-style fixed-width Unicode escape sequence surrogate pairs.
* Adds `printString` to no longer rely on `JSON.stringify`. Borrows some implementation details from Node.js internals for string printing.

  Implements:

  > When producing a {StringValue}, implementations should use escape sequences to
  > represent non-printable control characters (U+0000 to U+001F and U+007F to
  > U+009F). Other escape sequences are not necessary, however an implementation may
  > use escape sequences to represent any other range of code points.

Closes #2449

Co-authored-by: Andreas Marek <andimarek@fastmail.fm>

### DIFF
--- a/src/language/__tests__/printString-test.ts
+++ b/src/language/__tests__/printString-test.ts
@@ -1,0 +1,82 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { printString } from '../printString';
+
+describe('printString', () => {
+  it('prints a simple string', () => {
+    expect(printString('hello world')).to.equal('"hello world"');
+  });
+
+  it('escapes quotes', () => {
+    expect(printString('"hello world"')).to.equal('"\\"hello world\\""');
+  });
+
+  it('does not escape single quote', () => {
+    expect(printString("who's test")).to.equal('"who\'s test"');
+  });
+
+  it('escapes backslashes', () => {
+    expect(printString('escape: \\')).to.equal('"escape: \\\\"');
+  });
+
+  it('escapes well-known control chars', () => {
+    expect(printString('\b\f\n\r\t')).to.equal('"\\b\\f\\n\\r\\t"');
+  });
+
+  it('escapes zero byte', () => {
+    expect(printString('\x00')).to.equal('"\\u0000"');
+  });
+
+  it('does not escape space', () => {
+    expect(printString(' ')).to.equal('" "');
+  });
+
+  it('does not escape non-ascii character', () => {
+    expect(printString('\u21BB')).to.equal('"\u21BB"');
+  });
+
+  it('does not escape supplementary character', () => {
+    expect(printString('\u{1f600}')).to.equal('"\u{1f600}"');
+  });
+
+  it('escapes all control chars', () => {
+    /* spellchecker:ignore abcdefghijklmnopqrstuvwxyz */
+    expect(
+      printString(
+        '\u0000\u0001\u0002\u0003\u0004\u0005\u0006\u0007' +
+          '\u0008\u0009\u000A\u000B\u000C\u000D\u000E\u000F' +
+          '\u0010\u0011\u0012\u0013\u0014\u0015\u0016\u0017' +
+          '\u0018\u0019\u001A\u001B\u001C\u001D\u001E\u001F' +
+          '\u0020\u0021\u0022\u0023\u0024\u0025\u0026\u0027' +
+          '\u0028\u0029\u002A\u002B\u002C\u002D\u002E\u002F' +
+          '\u0030\u0031\u0032\u0033\u0034\u0035\u0036\u0037' +
+          '\u0038\u0039\u003A\u003B\u003C\u003D\u003E\u003F' +
+          '\u0040\u0041\u0042\u0043\u0044\u0045\u0046\u0047' +
+          '\u0048\u0049\u004A\u004B\u004C\u004D\u004E\u004F' +
+          '\u0050\u0051\u0052\u0053\u0054\u0055\u0056\u0057' +
+          '\u0058\u0059\u005A\u005B\u005C\u005D\u005E\u005F' +
+          '\u0060\u0061\u0062\u0063\u0064\u0065\u0066\u0067' +
+          '\u0068\u0069\u006A\u006B\u006C\u006D\u006E\u006F' +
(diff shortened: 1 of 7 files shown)


src/language/__tests__/printString-test.js
```diff
@@ -0,0 +1,54 @@
...
+        `"\\u00${i <= 0x000f ? '0' : ''}${i.toString(16).toUpperCase()}"`,
+      );
+    }
+  });
+
+  it('does not escape non-ascii character', () => {
+    expect(printString('\u21BB')).to.equal('"\u21BB"');
+  });
+
+  it('does not escape supplementary character', () => {
```
andimarek, line 51: This is tricky detail where a GraphQL String literal is not compatible with a JSON string.
I could imagine that the default JSON serialization doesn't handle that correctly.

You could escape it as surrogate pair and make it compatible without any additional work.

leebyron: These are all just characters that a JSON string would encode/escape as it defines.

For example the printed GraphQL string literal `"\u{12345}"` would JSON stringify as `"\"\\u{12345}\""`

Do you have a test case in mind that would cover your concern?

leebyron: JSON also supports all supplemental Unicode code points, escaped or unescaped

leebyron: There should be no case where a GraphQL string literal is directly provided to a JSON parser

andimarek: unescaped unicode code points are fine as you said, I am just thinking about the braced escape as this is something not supported by JSON.

I was just thinking if these printed string are every serialized directly.

For example a Field description like  "This is a funny field \u{1F921}".  (1F921 is = 🤡)

If this GraphQL string would be serialized to a JSON String  it would be printed as 
"This is a funny field \\u{12345}" which is not a correct JSON serialization. The correct JSON serialization would have to use surrogate pair.

If these printed String are never directly serialized, then it should be fine.

leebyron: That's right, a printed GraphQL string literal should be and is never represented directly as parseable JSON.

The one case where we would serialize a GraphQL string literal is during [introspection when returning default values](https://spec.graphql.org/draft/#sel-GAJXVHHDAAACDB78R). But this is always a printed string, then serialized as JSON, so `defaultValue: "123"` is the integer value `123` and `defaultValue: "\"123\""` is the string value `"123"`.

andimarek: ok ...so the default value "\u{1F921}"  would be serialized as "\"\\u{1F921}\"" which is a valid GraphQL String value encoded as JSON. My point was that this is not a direct representation of this code point in JSON: "\uD83E\uDD21" would be a better JSON encoded value, but if you print this value as "\uD83E\uDD21" and then serialize it would be "\"\\uD83E\\uDD21\"\" which is not really better.

leebyron: Agreed. Right now before this PR in the state where we accidentally implicitly support full Unicode and co-opt JSON.stringify to produce string literal grammars, we would render the string value [U+1F921] as the GraphQL string literal `"\uD83E\uDD21"`, which then serialized in a JSON body in an introspection query: "\"\\uD83E\\uDD21\"\"

After this diff with this manual version of string literal printing, we would render that same string value as the GraphQL string literal `"🤡"` which then serialized in a JSON body in an introspection query: "\"\uD83E\uDD21\"\".

IMHO this latter state is preferable. I just pushed a new version which explicitly includes this example in `introspection-test.js` to ensure it behaves as expected

andimarek: sounds good

src/language/lexer.js
```diff
@@ -104,18 +104,53 @@ export function isPunctuatorTokenKind(kind: TokenKindEnum): boolean {
...
 }
 
+/**
+ * The GraphQL specification defines source text as a sequence of unicode scalar
+ * values (which Unicode defines to exclude surrogate code points). However
+ * JavaScript defines strings as a sequence of UTF-16 code units which may
+ * include surrogates. A surrogate pair is a valid source character as it
+ * encodes a supplementary code point (above U+FFFF), but unpaired surrogate
+ * code points are not valid source characters.
+ */
```
andimarek, line 127: I think that is a great explanation.

src/language/__tests__/printString-test.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('does not escape space', () => {
+    expect(printString(' ')).to.equal('" "');
+  });
+
+  it('escapes all other control chars', () => {
+    for (let i = 1; i <= 0x9f; i++) {
+      const source = String.fromCharCode(i);
+      if (/[\b\f\n\r\t]/.test(source) || (i >= 0x0020 && i <= 0x007e)) {
```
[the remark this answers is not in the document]
IvanGoncharov: Anyway not a blocker

src/language/__tests__/printString-test.ts
```diff
@@ -0,0 +1,54 @@
...
+        continue;
+      }
+      expect(printString(source)).to.equal(
+        `"\\u00${i <= 0x000f ? '0' : ''}${i.toString(16).toUpperCase()}"`,
```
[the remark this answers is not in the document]
leebyron: I didn't explicitly use `padStart` in the source to avoid any browser compatibility issues, but using it in a test case seems like a good idea

src/language/__tests__/printString-test.ts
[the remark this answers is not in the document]
leebyron: I was also feeling uneasy about this. Let me try to do something more simple.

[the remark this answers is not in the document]
leebyron: Did something much simpler

IvanGoncharov (approved):
@leebyron Looks good!
Feel free to merge if it approved on today WG.

PR #3135, opened 2021-05-26
TS: Fix strict issues in src/validation

leebyron:
* Add `"strict": true` to tsconfig.json
* Fix all issues that arise within `src/validation`


saihaj (approved):
🎉

saihaj (approved):
👍🏼

PR #3136, opened 2021-05-26
TS: Fix strict issues in src/utilities

leebyron:
Depends on #3135

* Add "strict": true to tsconfig.json
* Fix all issues that arise within src/utilities

### DIFF
--- a/src/utilities/extendSchema.ts
+++ b/src/utilities/extendSchema.ts
@@ -379,7 +379,7 @@ export function extendSchemaImpl(
     return {
       ...field,
       type: replaceType(field.type),
-      args: mapValue(field.args, extendArg),
+      args: field.args && mapValue(field.args, extendArg),
     };
   }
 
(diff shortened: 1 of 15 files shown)


src/utilities/extendSchema.ts
```diff
@@ -393,24 +393,24 @@ export function extendSchemaImpl(
...
     const opTypes = {};
     for (const node of nodes) {
       // istanbul ignore next (See: 'https://github.com/graphql/graphql-js/issues/2203')
       const operationTypesNodes = node.operationTypes ?? [];
 
       for (const operationType of operationTypesNodes) {
+        // Note: While this could make early assertions to get the correctly
+        // typed values below, that would throw immediately while type system
+        // validation with validateSchema() will produce more actionable results.
+        // TODO ts-expect-error
```
saihaj, line 409: why are there `TODO` for `ts-expect-error`?

leebyron: Because we're not actually turning strict mode on yet so there isn't an error to expect, so TSC fails on seeing an "expect error" without an error. As soon as we turn on strict mode we expect an error here. 

This makes it easy to search and update all of these when we toggle strict on!

saihaj: makes sense thanks!

saihaj (approved):

PR #3142, opened 2021-05-27
TS: Fix strict issues in src/error

leebyron:
Depends on #3141

* Add "strict": true to tsconfig.json
* Fix all issues that arise within `src/error`

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -15,15 +15,6 @@ import { printLocation, printSourceLocation } from '../language/printLocation';
  */
 export class GraphQLError extends Error {
   /**
-   * A message describing the Error for debugging purposes.
-   *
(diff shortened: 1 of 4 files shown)


src/error/GraphQLError.ts
```diff
@@ -14,15 +14,6 @@ import { printLocation, printSourceLocation } from '../language/printLocation';
...
  */
 export class GraphQLError extends Error {
-  /**
-   * A message describing the Error for debugging purposes.
-   *
-   * Enumerable, and appears in the result of JSON.stringify().
-   *
-   * Note: should be treated as readonly, despite invariant usage.
-   */
-  message: string;
```
leebyron, line 24: Removed this since it's already included via `extends Error`

saihaj (approved):
🎉

PR #3145, opened 2021-05-28
Add coordinate field to schema element definitions

leebyron:
Depends on #3044

* Defines a `GraphQLSchemaElement` base class which defines a `.coordinate` property and `toString`/`toJSON` methods.
* Adds base class to types, fields, arguments, input fields, enum values, and directives.
* Uses this in validation error printing string templates.

Results in a net-reduction of code, minor simplification of definition.ts, and simplification/standardization of error messages


leebyron:
I made a loosely held decision to leave `.coordinate` off types, even though they are also schema elements since they have a trivial coordinate which is identical to `.name`. I'd love input on this.

src/type/definition.ts
```diff
@@ -1724,14 +1701,46 @@ export interface GraphQLInputFieldConfig {
 
 export type GraphQLInputFieldConfigMap = ObjMap<GraphQLInputFieldConfig>;
 
-export interface GraphQLInputField {
+export class GraphQLInputField extends GraphQLSchemaElement {
```
magicmark, line 1704: Does this become a breaking change? If folks were importing `GraphQLInputField` the interface - which may be different to `GraphQLInputField` the class?

magicmark: Actually I guess not, if even the function below still works:

```
export function isRequiredInputField(field: GraphQLInputField): boolean
```

I guess TS treats classes and interfaces the same here

magicmark (commented):
I've read over this and tried to understand what's going on. Adding the `GraphQLSchemaElement` base class, and having everything be a class that inherits it makes a lot of sense. I think all the elements are covered here.

Looks great to me! (But I'm still getting up to speed on this codebase)

src/type/introspection.ts
```diff
@@ -480,53 +480,24 @@ export const __TypeKind: GraphQLEnumType = new GraphQLEnumType({
...
-    {
-      name: 'name',
-      description: undefined,
-      type: new GraphQLNonNull(GraphQLString),
-      defaultValue: undefined,
-      deprecationReason: undefined,
-      extensions: undefined,
-      astNode: undefined,
-    },
-  ],
```
mjmahone, line 513: nice cleanup

src/type/validate.ts
```diff
@@ -184,17 +184,17 @@ function validateDirectives(context: SchemaValidationContext): void {
       // Ensure the type is an input type.
       if (!isInputType(arg.type)) {
         context.reportError(
-          `The type of @${directive.name}(${arg.name}:) must be Input Type ` +
+          `The type of ${arg} must be Input Type ` +
```
mjmahone, line 187: oooh this is a much nicer pattern too. Quite clever, that the arg will be schematized when it's stringified.

src/validation/rules/KnownArgumentNamesRule.ts
```diff
@@ -26,15 +26,14 @@ export function KnownArgumentNamesRule(context: ValidationContext): ASTVisitor {
     Argument(argNode) {
       const argDef = context.getArgument();
       const fieldDef = context.getFieldDef();
-      const parentType = context.getParentType();
 
-      if (!argDef && fieldDef && parentType) {
```
mjmahone, line 31: oh that is a much cleaner pattern: it's great that we can now remove the only-needed-for-the-message value.

src/type/definition.ts
mjmahone, line 1704: This change seems likely to break implementations that implement `GraphQLInputField` in a custom way. I don't know how common that is, but definitely not a purely safe adjustment.

src/validation/rules/custom/NoDeprecatedCustomRule.ts
```diff
@@ -24,11 +22,9 @@ export function NoDeprecatedCustomRule(context: ValidationContext): ASTVisitor {
       const fieldDef = context.getFieldDef();
       const deprecationReason = fieldDef?.deprecationReason;
       if (fieldDef && deprecationReason != null) {
-        const parentType = context.getParentType();
-        invariant(parentType != null);
         context.reportError(
           new GraphQLError(
-            `The field ${parentType.name}.${fieldDef.name} is deprecated. ${deprecationReason}`,
+            `The field ${fieldDef} is deprecated. ${deprecationReason}`,
```
mjmahone, line 27: Should this be `Field ${fieldDef} is deprecated.` similar to what you're doing in ValuesOfCorrectTypeRule?

mjmahone (approved):
I have no issue with switching the couple of schema interfaces to concrete classes, but if something would go wrong with this diff that's where I'd anticipate downstream issues.

PR #3146, opened 2021-05-29
build: add eslint-plugin-tsdoc

saihaj:
Add [`eslint-plugin-tsdoc`](https://tsdoc.org/pages/packages/eslint-plugin-tsdoc/) to start working on doc generation.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -449,9 +449,17 @@ overrides:
       project: ['tsconfig.json']
     plugins:
       - '@typescript-eslint'
+      - 'eslint-plugin-tsdoc'
     extends:
       - plugin:import/typescript
     rules:
+      ##########################################################################
+      # `eslint-plugin-tsdoc` rule list based on `v0.2.x`
+      # https://github.com/microsoft/tsdoc/tree/master/eslint-plugin
+      ##########################################################################
+
+      tsdoc/syntax: error
+
       ##########################################################################
       # `@typescript-eslint/eslint-plugin` rule list based on `v4.28.x`
       ##########################################################################
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,8 +18,10 @@
  * You may also import from each sub-directory directly. For example, the
  * following two import statements are equivalent:
  *
- *     import { parse } from 'graphql';
- *     import { parse } from 'graphql/language';
+ * ```ts
+ * import { parse } from 'graphql';
+ * import { parse } from 'graphql/language';
+ * ```
  */
 
 /** The GraphQL.js version info. */
--- a/src/jsutils/isIterableObject.ts
+++ b/src/jsutils/isIterableObject.ts
@@ -7,12 +7,13 @@
  * TypedArray, etc. but excludes string literals.
  *
  * @example
- *
+ * ```ts
  * isIterableObject([ 1, 2, 3 ]) // true
  * isIterableObject(new Map()) // true
  * isIterableObject('ABC') // false
  * isIterableObject({ key: 'value' }) // false
  * isIterableObject({ length: 1, 0: 'Alpha' }) // false
+ * ```
  */
 export function isIterableObject(
   maybeIterable: any,
--- a/src/language/lexer.ts
+++ b/src/language/lexer.ts
@@ -108,11 +108,13 @@ export function isPunctuatorTokenKind(kind: TokenKindEnum): boolean {
 }
 
 /**
+ * ```
  * SourceCharacter ::
  *   - U+0009 (Horizontal Tab)
  *   - U+000A (New Line)
  *   - U+000D (Carriage Return)
  *   - U+0020-U+FFFF
+ * ```
  */
 function isSourceCharacter(code: number): boolean {
   return (
@@ -296,9 +298,11 @@ function readNextToken(lexer: Lexer, start: number): Token {
 /**
  * Reads a comment token from the source file.
  *
+ * ```
  * Comment :: # CommentChar* [lookahead != CommentChar]
  *
  * CommentChar :: SourceCharacter but not LineTerminator
+ * ```
  */
 function readComment(lexer: Lexer, start: number): Token {
   const body = lexer.source.body;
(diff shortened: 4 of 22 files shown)


.eslintrc.yml
```diff
@@ -605,6 +606,9 @@ overrides:
       '@typescript-eslint/space-before-function-paren': off
       '@typescript-eslint/space-infix-ops': off
       '@typescript-eslint/type-annotation-spacing': off
+
+      # TS Doc
```
IvanGoncharov, line 610: Please match style of comments that precedes other rules

src/jsutils/isIterableObject.ts
```diff
@@ -7,12 +7,13 @@
  * TypedArray, etc. but excludes string literals.
  *
  * @example
- *
+ *```
```
IvanGoncharov, line 10: should be TS

src/language/lexer.ts
```diff
@@ -287,7 +287,7 @@ function unexpectedCharacterMessage(code: number): string {
 /**
  * Reads a comment token from the source file.
  *
- * #[\u0009\u0020-\uFFFF]*
+ * `#[\u0009\u0020-\uFFFF]*`
```
IvanGoncharov, line 290: It's a grammar so backtrick can be confusing here better to use code blocks for all grammar rules

src/language/parser.ts
```diff
@@ -88,12 +88,12 @@ export interface ParseOptions {
...
    *
-   * The syntax is identical to normal, query-defined variables. For example:
-   *
+   * @example The syntax is identical to normal, query-defined variables.
+   *```graphql
    *   fragment A($var: Boolean = false) on T  {
    *     ...
    *   }
-   *
+   *```
```
IvanGoncharov, line 96: please use consistent spacing

src/language/lexer.ts
```diff
@@ -692,7 +702,11 @@ function readName(
   );
 }
 
-// _ A-Z a-z
+/**
+ * ```
+ * _ A-Z a-z
+ * ```
+ */
```
IvanGoncharov, line 709: It's not exported function please leave it as is

src/language/parser.ts
```diff
@@ -346,7 +346,7 @@ export class Parser {
   }
 
   /**
-   * SelectionSet : { Selection+ }
+   * SelectionSet : `{ Selection+ }`
```
IvanGoncharov, line 349: It's a grammar rule please use same code block as you already used in lexer

src/index.ts
```diff
@@ -15,11 +15,12 @@
  * This also includes utility functions for operating on GraphQL types and
  * GraphQL documents to facilitate building tools.
  *
- * You may also import from each sub-directory directly. For example, the
- * following two import statements are equivalent:
- *
+ * You may also import from each sub-directory directly.
+ * @example the following two import statements are equivalent:
```
IvanGoncharov, line 19: Let's decide on `@example` latter, so for now just convert the code sample into markdown code blocks

src/index.ts
```diff
@@ -15,11 +15,12 @@
...
+ * ```ts
  *     import { parse } from 'graphql';
```
IvanGoncharov, line 21: Please convert it consistent indentation, because right now you have no spaces, 2 spaces, and 4 spaces variants.
Let's use no spaces variant across all code samples

src/language/lexer.ts
saihaj: the issue here is that eslint complains. So might need to look in how to escape it for non-exported members

.eslintrc.yml
```diff
@@ -593,8 +594,8 @@ overrides:
       '@typescript-eslint/no-use-before-define': off
       '@typescript-eslint/no-duplicate-imports': off # Superseded by `import/no-duplicates`
 
-      # Bellow rules are disabled because coflicts with Prettier, see:
-      # https://github.com/prettier/eslint-config-prettier/blob/master/%40typescript-eslint.js
+      # Below rules are disabled because they conflict with Prettier, see:
+      # https://github.com/prettier/eslint-config-prettier/blob/aeb4d52de54960ff48a8c627e6c25db9873d7719/index.js#L97-L110
```
IvanGoncharov, line 598: It's unrelated to this PR can you please extract it into a separate one?
Also please use non-versioned link https://github.com/prettier/eslint-config-prettier/blob/main/index.js

.eslintrc.yml
```diff
@@ -610,6 +611,11 @@ overrides:
       '@typescript-eslint/space-before-function-paren': off
       '@typescript-eslint/space-infix-ops': off
       '@typescript-eslint/type-annotation-spacing': off
+
+      # Validating TS Doc comments. see:
+      # https://tsdoc.org/pages/packages/eslint-plugin-tsdoc/
+      'tsdoc/syntax': error
```
IvanGoncharov, line 617: Can you please put it before `@typescript-eslint` rules and also make a header similar to this one:
https://github.com/graphql/graphql-js/blob/d82c8e2aa603536c6a305b1c66fb329c4d086302/.eslintrc.yml#L26-L29

src/language/parser.ts
IvanGoncharov: please put space before backtrick
Also please check all other places

src/language/parser.ts
IvanGoncharov: please use the same code block as you already used above

IvanGoncharov: same for the rest of the file

src/type/definition.ts
```diff
@@ -1093,15 +1093,15 @@ export interface GraphQLInterfaceTypeExtensions {
...
- * Example:
- *
- *     const EntityType = new GraphQLInterfaceType({
- *       name: 'Entity',
- *       fields: {
- *         name: { type: GraphQLString }
- *       }
- *     });
- *
+ * example
```
IvanGoncharov, line 1096: please keep it as it was before

PR #3151, opened 2021-06-01
CI: Add new job to check health of package-lock.json

IvanGoncharov:


saihaj:
@IvanGoncharov maybe we should explore Danger JS https://danger.systems/js/

IvanGoncharov:
@saihaj If you mean this example:
image

It's less safe since it doesn't check the content of `package-lock.json` just the fact it was changed.

saihaj:
Well we can add all the logic from workflow to the dangerfile and give feedback on PR by if anything fails or a warning.

PR #3155, opened 2021-06-02
Lexer: split part of refactoring from #3115

IvanGoncharov:
Extacted from #3115


leebyron:
Thanks for this alt. Since this is only half the changeset - let me fold some of these review ideas back into my total PR. I'll have that up in a couple hours

IvanGoncharov:
@leebyron Would be great thanks 👍
Can you please keep a separate commit from this PR?

PR #3159, opened 2021-06-04
execute: Rename resolveField function and update comments

yaacovCR:
The function implements the "Executing Field" section of the spec and should be called `executeField` rather than `resolveField` to reduce ambiguity.

Motivation: (1) correctness and (2) lays the groundwork for -- possibly -- providing a custom function for fields to optimize their execution, say, by not calling `getArgumentValues` if the arguments are passed unparsed to some external graphql service, or what have you.

Right now, if we would add a custom `execute` function, we would have to have a code block within a function called `resolveField` that calls the custom execute. Resolving is a part of executing, and so there is a semantic clash that could get pretty confusing (at least for me).

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -368,7 +368,7 @@ function executeOperation(
 
 /**
  * Implements the "Executing selection sets" section of the spec
- * for "write" mode.
+ * for fields that must be executed serially.
  */
 function executeFieldsSerially(
   exeContext: ExecutionContext,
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@yaacovCR Thanks a lot, especially for doing small commits with meaningful descriptions 👍

> Motivation: (1) correctness 

`#1` is totally enough for accepting any changes like this.
Moreover, as a reference implementation, we should always keep implementation in sync with a spec.

PR #3162, opened 2021-06-05
Create codeql-analysis.yml

IvanGoncharov:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
```
mjmahone, line 12: I'm not sure this belongs in `graphql-js`? There's no description on your diff, @IvanGoncharov, and I'm not always hip on the latest open source tooling, so I'm not totally sure how CodeQL's yml file getting added to the graphql-js repo benefits us.

https://securitylab.github.com/tools/codeql/ => it seems like anyone who *wants* to use CodeQL can, without defaulting everyone who downloads `graphql-js` to run it. I may be missing some of the benefits or purpose of CodeQL here, though, or what it's actually doing. Is this essentially just setting up a "code analysis" test, similar to our test coverage tests?

IvanGoncharov: @mjmahone Sorry, I opened this PR through GitHub UI (content is autogenerated), so it wasn't my intention to merge it just to give it a spin.

About why we need it, it is used by various tools to score the security of a particular OSS project, e.g. https://github.com/ossf/scorecard/blob/main/docs/checks.md#sast
Also, it can uncover some real security issues both in our sorce code and GitHub
workflows.


>  without defaulting everyone who downloads graphql-js to run it.

No, you don't need to install it or run it.
It just runs on CI, so it doesn't affect local workflow.

PR #3164, opened 2021-06-06
Refactor incremental

yaacovCR:
@robrichard @IvanGoncharov

Ideas


yaacovCR:
@IvanGoncharov  --- another code coverage upload issue....

IvanGoncharov:
@yaacovCR Looks like the coverage issue resolved itself.
Sorry, don't really in context for steam/defer to provide a meaningful review for this PR.
Fully trust @robrichard on this one.
Plan to work on merging this functionality in `main` after `16.0.0` release so happy with any refactoring/cleanup that will happen before that.

yaacovCR:
moved main changes to #3168, upon which this now depends

next goal is to refactor to use Repeaters @brainkim

https://repeater.js.org/

I think we need a Repeater on which to push DispatcherResults, and then we need to map that Repeater to add hasNext depending on whether there are more `subsequentPayloads` or `iterators`.

@brainkim, do you think the repeater executor could provide another argument function that provides the state of the Repeater, i.e. pending pushes? Or is that already available somehow?

brainkim:
@yaacovCR What specific state are you looking for? I think if you want whether or not the repeater has finished, you can just set a boolean variable in the scope. If you want things like whether next has been called, you can read the return value of the `push()` function. I haven’t looked through the code yet, but I’m happy to help here.

yaacovCR:
That is already helpful, will have to get some code up before asking more pointed questions :)

yaacovCR:
Hey @brainkim and any interested others. I tried reworking to use Repeaters, that way we could just "push" results to the repeater and automatically get an AsyncIterableIterator that is well tested, behaves similarly to generators, etc, etc, see https://repeater.js.org/ for more details.

But I hit some problems, namely that Repeaters give us a lazy-by-design AsyncIterableIterator and the existing graphql-js implementation is eager-not-lazy (except when streaming from an AsyncIterator), I found it not-so-straightforward to start pushing values to the Repeater, as the Repeater is not "primed," the executor is not started, and the assignment of the Repeater's push and stop arguments to the dispatcher's private variables does not fire early enough for the dispatcher to start pushing values. This is (maybe) surmountable.

But, more importantly, it does bring us around to a few implementation (not spec) questions:
1. Is current graphql-js behavior optimal, i.e. should the returned AsyncIterableIterator be lazy and start work on the next defer or stream payload only on request? Or should it be eager, as it is currently?
2. In particular, does it make sense that stream is eager when the underlying list field is represented by an Iterable, but not when it is represented by an AsyncIterable? I think it might, but also that it merits explicit discussion.

Namely, I can imagine a world in which defer fragments are assumed to be limited in number, and it makes sense for them to be eager, but stream payloads may be quite large, and so it makes sense for those to be eager -- in particular, it makes sense to be lazy when the underlying data is arriving as an AsyncIterable, in that it could be the very reason it is arriving that way is because the dataset is indeed quite large.

Anyways, I think the pros and cons should at least be briefly weighed or made explicit in terms of the path chosen, whether it's in the working group context, on Github, or even within code comments.

Any thoughts from others? @robrichard @brainkim

yaacovCR:
Although it’s difficult to switch implementation to be lazy and still retain hasNext false as merged into the last value containing payload instead of its own separate payload. The spec seems to prefer the former.

PR #3184, opened 2021-06-16
refactor: collectFields to separate utility

yaacovCR:

PR #3185, opened 2021-06-16
refactor: execution methods into Executor class

yaacovCR:
see: https://github.com/graphql/graphql-js/pull/3163#issuecomment-859546546

depends on #3184


yaacovCR:
Hi @IvanGoncharov

I see your point, that is functionally equivalent, and it does seem cleaner.

So....I think I have the wrapping execute and subscribe functions as slim as they can go, and I have combined the Executor and SubscriberExecutor classes into one.

This is ready again for review. Please feel free to edit/rename/reformat within this PR as necessary.

I am feeling very good about this! I think this will unlock a lot of potential for experimentation from which the ecosystem will benefit greatly. I'd love to see this within the next alpha.

yaacovCR:
In particular, I don't love how I left the organization of folders/directories. The breakdown in structure between `execution` and `subscription` does not seem to make sense anymore, as all the essential logic anyway is within the `Executor` class in the `executor.ts`. One thought would be to move all the `subscription` files into the `execution` folder also -- but it is somewhat helpful having more folders to help keep the tests organized. Back on the other hand, the subscription tests are really just testing the `Executor`, which doesn't live in that folder. I am going to push one more commit that just moves everything to `execution`, but feel free to change this PR at will.

yaacovCR:
merging this into branch on main repo, will split out into separate PRs as possible

PR #3186, opened 2021-06-17
Backport instanceOf Error Check Improvements

tubbo:
This backports #3172 to the `15.x.x` branch so that users of GraphQL
v15.x can take advantage of it. GraphQL v16.x has some breaking changes
and is not possible to run when you have other libraries depending on
the behavior of v15, such as GraphQL Code Generator.


tubbo:
@IvanGoncharov does this look ok to you?

IvanGoncharov:
@tubbo Thanks, super useful 👍 
Released as https://github.com/graphql/graphql-js/releases/tag/v15.5.1
cc: @benjamn @hwillson

benjamn:
Yay, thanks @tubbo and @IvanGoncharov!

Issue #205, opened 2015-10-13
Returning multiple errors


freiksenet:
Recent patch added ability to return `Error` from `resolve` to indicate failure. In a scenarios where validation happens in the `resolve`, it would be useful to be able to return multiple errors from one resolve.

I am not sure what would be the best way to do it, but otherwise I am willing to implement it. One way I see is to always check if result is a array and then check if elements of that list are all `Errors`.

leebyron:
Interesting. I'm curious what the use case is. I'm concerned about the burden this would put on the execution logic, but would entertain a PR here.

jonsharratt:
:+1:

We potentially have a `mutation` where we would return an array of promises in the `resolve` function in turn they call different services and endpoints of those services and we would want them all to at least try and resolve all of them and get their results (success or errornus).

At the moment we we map through our Promise array, catch and setup the error rather to be returned rather than thrown so we can eventually filter out and throw the array from the mutations `resolve`.

Is there way from a mutation `resolve` function to push the array of errors to the context in order for them to list?

From GraphQL Schema 7.2.2:
`If the data entry in the response is not null, the errors entry in the response may contain any errors that occurred during execution. If errors occurred during execution, it should contain those errors.`

freiksenet:
Sorry for the lack of activity here.

@leebyron one use case is custom validation, for example in mutations. In Reindex we have `migrate` mutation, that accepts a new schema that database should move to. We try to detect all the problems that schema has at once, so that user doesn't need to call the endpoint repeatedly to find all the errors. Currently we have to solve this by concatenating errors together and then throwing them as one error in resolve function. Another example we have in Reindex is permissions, there can be multiple permissions missing during one resolve, esp when mutations involve modifying connected types.

I think a simple solution could be to create a new error type - 'GraphQLMultipleError', which accepts list of errors, and then just manage that at top level of the graphql resolve function, converting it to multiple errors there. This way we won't make the logic overly convoluted, but still allow outputting multiple errors to the user.

leebyron:
Interesting - for custom validation, would it be possible to use the validator phase for this, which supports multiple errors per step? Do you need to execute the query to understand the issues, or can issues be determined statically?

amir-s:
@leebyron I ran into it and searched for it and I came here.

Basically I need to validate an input type based on some rules defined in the database, so the validation is at runtime. 

[...]

PR #3192, opened 2021-06-21
feat: introduce GraphQLAggregateError

yaacovCR:
Motivation:

1. This helper class allows reporting multiple distinct errors from within a resolver.
2. It may also be used to transform the ExecutionContext validation to a function that throws a collection of GraphQLErrors (which may or may not be more streamlined).

### DIFF
--- a/docs/APIReference-Errors.md
+++ b/docs/APIReference-Errors.md
@@ -3,7 +3,7 @@ title: graphql/error
 layout: ../_core/GraphQLJSLayout
 category: API Reference
 permalink: /graphql-js/error/
-sublinks: formatError,GraphQLError,locatedError,syntaxError
+sublinks: formatError,GraphQLError,locatedError,syntaxError,GraphQLAggregateError
 next: /graphql-js/execution/
 ---
 
@@ -107,3 +107,20 @@ type GraphQLErrorLocation = {
 
 Given a GraphQLError, format it according to the rules described by the
 Response Format, Errors section of the GraphQL Specification.
+
+### GraphQLAggregateError
+
+```js
+class GraphQLAggregateError extends Error {
+ constructor(
+   errors: Array<Error>,
+   message?: string
+ )
+}
+```
+
+A helper class for bundling multiple distinct errors. When a
+GraphQLAggregateError is thrown during execution of a GraphQL operation,
+a GraphQLError will be produced from each individual errors and will be
+reported separately, according to the rules described by the Response
+Format, Errors section of the GraphQL Specification.
--- a/src/error/GraphQLAggregateError.ts
+++ b/src/error/GraphQLAggregateError.ts
@@ -1,0 +1,36 @@
+/**
+ * A GraphQLAggregateError is a container for multiple errors.
+ *
+ * This helper can be used to report multiple distinct errors simultaneously.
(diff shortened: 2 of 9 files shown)


IvanGoncharov:
@yaacovCR Thanks a lot for PR 👍 
Sorry for the delay with the review, now I'm back and this PR stack is first on my list.

> It may also be used to transform the ExecutionContext validation to a function that throws a collection of GraphQLErrors (which may or may not be more streamlined).

Happy to merge this part of PR + transform of function ExecutionContext validation.
One thing, I like it to be clear that it's a temporary polyfill to `AggregateError` until we can drop Node v12/v14. So I'm against exporting it or including it as part of public API, also `GraphQL` prefix implies that it has some graphql-specific functionality. So I'm rather reintroduce `src/pollyfills` directory and have `import { AggregateErrorP } from 'src/pollyfills/AggregateError';` 

> This helper class allows reporting multiple distinct errors from within a resolver.

Don't agree with this one.
I think having multiple errors with the same `path` and same `locations` is confusing.
And create problematic DX, because we are losing the notion that these errors were once aggregated + we are losing the message of aggregated error.

It would be a way better solution if we add `why` property to graphql spec that will point to an array of simpler `message` + `extensions` objects. Similar to this stage 3 proposal: https://github.com/tc39/proposal-error-cause

We need this functionality anyway since I dislice how we join error messages with `;` ATM, for example:
https://github.com/graphql/graphql-js/blob/22b95040951348edff7934447a060b5a06295c9a/src/validation/rules/ValuesOfCorrectTypeRule.ts#L146-L147
Your idea of returning multiple errors from resolvers gives it a reason to make `why` an array.
I will work on a proposal for the WG (should be a pretty small change to the spec).
But if it is rejected I'm ok to merge the solution proposed in this PR (multiple errors for the same `path` & `locations`).

yaacovCR:
This is interesting feedback.

Some (numbered!) thoughts:
1. I debated whether to call this new helper class `AggregateError` or `ErrorSet`, eventually, for reasons I can't actually remember so much later, settling on the former. For our (later) purpose within this stack, we do not actually need the "wrapping" functionality of the "real" `AggregateError` or a polyfill, we just need to pass along a set of errors....more like an error set.
2. The addition of the word `GraphQL` was just meant to signify that objects of this special class trigger special processing when returned by resolvers -- if we are not exporting the helper/or not in favor of that ability, we can forgo that prefix.

After thinking about your feedback, I think using a non-exported `AggregateError` polyfill may not right even for the more limited purpose mentioned above, as the real `AggregateError` wraps errors with a message and that is not what we currently do with the errors possibly encountered when building an execution context. For example, we report variable coercion errors  as separate errors currently.

I am thinking maybe just to rethink the approach ==> more to what you mentioned about an operation context vs. an execution context, except that I think about it as more like a "public" execution context vs the "private" one used internally. The "public" one would include just operation, the fragment map, and the coerced variables, basically what we processed while validating.

I might think on that and reapproach this.

yaacovCR:
Closing this for now. I think it's fine to improve error handling in GraphQL -- whether with the above suggestions or any type of standardization. But I don't want to get too distracted from the purpose of this PR stack,

yaacovCR:
See #205

PR #3195, opened 2021-06-22
refactor: consolidate subscription into execution

yaacovCR:
`subscription` operations are described by the 'Execution' section of the spec.

Depends on #3192


yaacovCR:
@IvanGoncharov I have lifted the dependency for this PR on the #3192 

This is now just code reorganization/polish, and I believe can be released in v16.

src/execution/execute.ts
```diff
@@ -273,12 +282,13 @@ export function assertValidExecutionArguments(
 export function buildExecutionContext(
```
IvanGoncharov, line 282: Please remove exports on functions that what use only in subscriptions.

src/execution/README.md
```diff
@@ -7,3 +7,8 @@ fulfilling a GraphQL request.
 import { execute } from 'graphql/execution'; // ES6
 var GraphQLExecution = require('graphql/execution'); // CommonJS
 ```
+
+```js
+import { subscribe, createSourceEventStream } from 'graphql/execution'; // ES6
+var GraphQLSubscription = require('graphql/execution'); // CommonJS
+```
```
IvanGoncharov, line 14: We don't need two examples here.
Original one is more than enough to illustrate the point of using subfolders both as CommonJS and ESM modules.

IvanGoncharov (changes requested):
Can you please make a separate PR where you just merge the `subscription` folder to `execute`.
That way we don't lose history on `subscribe.ts` and also it would be way easier to see changes in that file.

IvanGoncharov:
> I believe can be released in v16.

One issue though is that people using `graphql/subscription` in imports and configs.
Can you please keep `src/subscription/index.ts` where you reexport necessary functions from `src/execute/index.ts` with a note that it's deprecated and will be removed in v17?

Issue #3169, opened 2021-06-11
Validation of variable values


dkbarn:
I'm curious why it's not possible to validate the values of variables using the ```validate``` framework.

For example, suppose I have a very simple validator that wants to assert the value for argument ```name``` contains only ascii characters, given this schema:
```
type Query {
    hello(name: String!): String!
}
```
I can implement this validator and it will work when the incoming request is this:
```
{
    'query': '{ hello(name: "foo") }'
}
```
but when the request is the following, I have no way of seeing the value for the VariableNode ```$myName``` from inside the validator:
```
{
    'query': 'query myQuery($myName: String!) { hello(name: $myName) }',
    'variables': {'myName': 'foo'}
}
```

It looks like it comes down to this bit of the code:
https://github.com/graphql/graphql-js/blob/main/src/graphql.ts#L118-L122

Is there any reason why ```variableValues``` cannot also be passed in and made available to the validate framework?

glasser:
My understanding is that the whole concept behind "validation" is that it is specifically a function of "operation + schema". So for example, you can cache the result of parsing and validation and not run it multiple times for the same operation and schema if only variables are changing. (Apollo Server does this!)

That said, I would love to be able to separate the `getVariableValues` stage out of `execute` so that we can report them with different kinds of error codes and plugin callbacks in our request processing pipeline.  @IvanGoncharov do you think that sort of change would be a reasonable PR? Would love to brainstorm exactly how to structure it.

ArchitectKatherineGraham:
I am interested in this as well.

From: David Glasser ***@***.***>
Sent: Friday, July 16, 2021 1:52 PM
To: graphql/graphql-js ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [Non-DoD Source] Re: [graphql/graphql-js] Validation of variable values (#3169)

All active links contained in this email were disabled. Please verify the identity of the sender, and confirm the authenticity of all links contained within the message prior to copying and pasting the address to a Web browser.
________________________________


My understanding is that the whole concept behind "validation" is that it is specifically a function of "operation + schema". So for example, you can cache the result of parsing and validation and not run it multiple times for the same operation and schema if only variables are changing. (Apollo Server does this!)

[...]

PR #3196, opened 2021-06-22
refactor: throw GraphQLAggregateErrors from buildExecutionContext

yaacovCR:
Rather than returning an ExecutionContext or an array of GraphQLErrors.

Motivation: this may facilitate refactoring of the executor into a class, in which the buildExecutionContext step is employed during a constructor, and requires definite assignment of the executionContext (or its member properties) into class members.


yaacovCR:
@IvanGoncharov , this is ready for review. based on the feedback to #3192, I have abandoned the throw on build error approach (despite the retained name of the branch) and am now basically going at least initially with @glasser suggestion https://github.com/graphql/graphql-js/issues/3169#issuecomment-881617623 to separate out the document normalization and variable coercion steps from the main execute flow.

PR #3207, opened 2021-06-30
ci: generate diff for NPM package

IvanGoncharov:


src/index.ts
```diff
@@ -3,7 +3,7 @@
  * but is also a useful utility for operating on GraphQL files and building
  * sophisticated tools.
  *
- * This primary module exports a general purpose function for fulfilling all
+ * Ths primary module exports a general purpose function for fulfilling all
```
nam-hle, line 6: Hi @IvanGoncharov you missed `i` 😅

IvanGoncharov: Yes, exectly.
this is a change to test how well this script works in CI pipeline

PR #3208, opened 2021-07-01
Add 'UniqueArgumentDefinitionNamesRule'

IvanGoncharov:
Background https://github.com/graphql/graphql-wg/issues/505


src/validation/rules/UniqueArgumentDefinitionNamesRule.ts
```diff
@@ -0,0 +1,91 @@
...
+
+function groupBy<K, T>(
+  list: ReadonlyArray<T>,
+  keyFn: (item: T) => K,
+): Map<K, Array<T>> {
+  const result = new Map<K, Array<T>>();
+  for (const item of list) {
+    const key = keyFn(item);
+    const group = result.get(key);
+    if (group === undefined) {
```
mjmahone, line 84: Given `![] === false`, I think this can just be:
```suggestion
    if (!group) {
```

src/validation/__tests__/UniqueArgumentDefinitionNamesRule-test.ts
```diff
@@ -0,0 +1,120 @@
...
+    `);
+  });
+
+  it('duplicating arguments', () => {
+    expectSDLErrors(`
+      type SomeObject {
+        someField(
+          foo: String
+          bar: String
+          foo: String
```
mjmahone, line 77: Do we require a test where the same argument is defined twice with different types?

src/validation/rules/UniqueArgumentDefinitionNamesRule.ts
```diff
@@ -0,0 +1,91 @@
...
+    // istanbul ignore next (See: 'https://github.com/graphql/graphql-js/issues/2203')
+    const fieldNodes = typeNode.fields ?? [];
+
+    for (const fieldDef of fieldNodes) {
+      const fieldName = fieldDef.name.value;
+
+      // istanbul ignore next (See: 'https://github.com/graphql/graphql-js/issues/2203')
+      const argumentNodes = fieldDef.arguments ?? [];
+
+      checkArgUniqueness(`${typeName}.${fieldName}`, argumentNodes);
```
mjmahone, line 49: I like the use of Schema Coordinates here!

src/validation/rules/UniqueArgumentDefinitionNamesRule.ts
```diff
@@ -0,0 +1,91 @@
...
+  function checkArgUniquenessPerField(typeNode: {
+    readonly name: NameNode;
+    readonly fields?: ReadonlyArray<FieldDefinitionNode>;
+  }) {
+    const typeName = typeNode.name.value;
+
+    // istanbul ignore next (See: 'https://github.com/graphql/graphql-js/issues/2203')
+    const fieldNodes = typeNode.fields ?? [];
+
+    for (const fieldDef of fieldNodes) {
```
mjmahone, line 43: Why split this out into its own, used-once const, rather than:
```suggestion
    for (const fieldDef of typeNode.fields ?? []) {
```

src/validation/rules/UniqueArgumentDefinitionNamesRule.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+export function UniqueArgumentDefinitionNamesRule(
+  context: SDLValidationContext,
+): ASTVisitor {
+  return {
+    DirectiveDefinition(directiveNode) {
+      // istanbul ignore next (See: 'https://github.com/graphql/graphql-js/issues/2203')
+      const argumentNodes = directiveNode.arguments ?? [];
+
+      return checkArgUniqueness(`@${directiveNode.name.value}`, argumentNodes);
```
mjmahone, line 26: Same idea as below, why this and not
```suggestion
      return checkArgUniqueness(`@${directiveNode.name.value}`, directiveNode.arguments ?? []);
```

mjmahone (approved):
This looks necessary and the implementation seems correct in my view.

PR #3209, opened 2021-07-08
Fix failing CI on main

IvanGoncharov:


mjmahone (approved):
I think I ran into the same issue testing today. No reason not to fix.

PR #3215, opened 2021-07-24
Export utilities for working with execution context

andyrichardson:
# About

While the `buildExecutionContext` function is exported, the utility functions which can consume the resulting response aren't.


yaacovCR:
I think we can close this PR now that we have #4210, #4211, and #4215 => please feel free to reopen if more exports are needed and we can consider them!

PR #3223, opened 2021-08-04
Disallow "true", "false", and "null" as enum values

### LINKED ISSUE #3221: Parsing SDL allows `true`, `false`, and `null` for enum value names
There is a schema validator that forbids `true`, `false`, and `null` as enum value names sinde https://github.com/graphql/graphql-js/pull/812/files. The `parse` function together with the `specifiedSDLRules` rules however allow these strings to be used as enum value names/values (they're the same there). This means you can parse and successfully validate an SDL which generates a validation error when creating the schema.

I'm not quite sure whether this is a bug in the parser or a missing SDL validator. `parseEnumValueDefinition` uses `parseName` and states that the `EnumValue` non-terminal is equivalent to the `name` non-terminal:

```javascript

  /**
   * EnumValueDefinition : Description? EnumValue Directives[Const]?
   *
   * EnumValue : Name
   */
  ;

  _proto.parseEnumValueDefinition = function parseEnumValueDefinition() {
    var start = this._lexer.token;
    var description = this.parseDescription();
    var name = this.parseName();
    var directives = this.parseDirectives(true);
    return {
      kind: Kind.ENUM_VALUE_DEFINITION,
      description: description,
      name: name,
      directives: directives,
      loc: this.loc(start)
    };
  }
```

There also is an `EnumValue` non-terminal which is defined as follows:

> EnumValue : Name but not `true`, `false` or `null`

So an enum declaration either should switch from `EnumName` to `EnumValue`, or there should be an SDL validator that rejects the three forbidden words when used as enum value names/values.

Yogu:
The specification disallows these in the SDL grammar definition, and using one of these enum values would result in a schema that would throw validation errors on creation.

Fixes #3221.

I also changed the type of `EnumValueDefinitionNode.name` from `NameNode` to `EnumValueNode`, so this is definitely breaking. We might even want to rename the `name` property to `value`, but that would be even more breaking.

### DIFF
--- a/src/language/__tests__/parser-test.ts
+++ b/src/language/__tests__/parser-test.ts
@@ -120,6 +120,26 @@ describe('Parser', () => {
     });
   });
 
+  it('does not allow "true", "false", or "null" as enum value', () => {
+    expectSyntaxError('enum Test { VALID, true }').to.deep.equal({
+      message:
+        'Syntax Error: Name "true" is reserved and cannot be used for an enum value.',
+      locations: [{ line: 1, column: 20 }],
+    });
+
+    expectSyntaxError('enum Test { VALID, false }').to.deep.equal({
+      message:
+        'Syntax Error: Name "false" is reserved and cannot be used for an enum value.',
+      locations: [{ line: 1, column: 20 }],
+    });
+
+    expectSyntaxError('enum Test { VALID, null }').to.deep.equal({
+      message:
+        'Syntax Error: Name "null" is reserved and cannot be used for an enum value.',
+      locations: [{ line: 1, column: 20 }],
+    });
+  });
+
   it('parses multi-byte characters', () => {
     // Note: \u0A0A could be naively interpreted as two line-feed chars.
     const ast = parse(`
--- a/src/language/parser.ts
+++ b/src/language/parser.ts
@@ -1036,13 +1036,11 @@ export class Parser {
 
   /**
    * EnumValueDefinition : Description? EnumValue Directives[Const]?
-   *
-   * EnumValue : Name
    */
   parseEnumValueDefinition(): EnumValueDefinitionNode {
     const start = this._lexer.token;
     const description = this.parseDescription();
-    const name = this.parseName();
+    const name = this.parseEnumValueName();
     const directives = this.parseConstDirectives();
     return this.node<EnumValueDefinitionNode>(start, {
       kind: Kind.ENUM_VALUE_DEFINITION,
@@ -1050,6 +1048,26 @@ export class Parser {
       name,
       directives,
     });
+  }
+
+  /**
+   * EnumValue : Name but not `true`, `false` or `null`
+   */
+  parseEnumValueName(): NameNode {
+    if (
+      this._lexer.token.value === 'true' ||
+      this._lexer.token.value === 'false' ||
+      this._lexer.token.value === 'null'
+    ) {
+      throw syntaxError(
+        this._lexer.source,
+        this._lexer.token.start,
+        `${getTokenDesc(
+          this._lexer.token,
+        )} is reserved and cannot be used for an enum value.`,
+      );
+    }
+    return this.parseName();
   }
 
   /**


IvanGoncharov (changes requested):
@Yogu Thanks for the change 👍 

Can you please make it more in line with what we already have for restricting fragment names: https://github.com/graphql/graphql-js/blob/6a5f51f3bd07f3914c7f1ec01908e9e325b77d58/src/language/parser.ts#L498-L506
No need for custom AST node type just apple restriction during parsing.

But please keep your custom error message, since it improves DX a lot.

src/language/__tests__/parser-test.ts
```diff
@@ -120,6 +120,13 @@ describe('Parser', () => {
     });
   });
 
+  it('does not allow "true" as enum value', () => {
+    expectSyntaxError('enum Test { VALID, true }').to.deep.equal({
+      message: 'Syntax Error: Name "true" is not a valid enum value.',
```
IvanGoncharov, line 125: Maybe we can be more specific and just say it's a restricted name that you can't use as a name of enum value.

src/language/__tests__/parser-test.ts
```diff
@@ -120,6 +120,13 @@ describe('Parser', () => {
     });
   });
 
+  it('does not allow "true" as enum value', () => {
```
IvanGoncharov, line 123: Also please add tests for `false` and `null`.
You can keep them under a single test case just rename it to something like: does not allow "true", "false", and "null" as enum values.

Yogu:
@IvanGoncharov 

> No need for custom AST node type just apple restriction during parsing.

It isn't a "custom" AST node type. `EnumValueNode` already existed and that's the name of the non-terminal used in the spec at the respective location. If we were to start fresh, it would probably be best to use `EnumValueNode` and rename `name` to `value`. Do you want to avoid the breaking change by continue using `NameNode` as the type for `EnumValueDefinitionNode`? I wouldn't have a problem with that, just checking that I understood you correctly.

src/language/__tests__/parser-test.ts
Yogu: > Name "true" is reserved and cannot be used as enum value.

I'd keep referring to it as "enum value" instead of "name of an enum value" because it's actually both the name and the value of the enum value (🤯) and calling it "value" is shorter.

Yogu: ah, I an stay ambiguous with that respect and (hopefully) clearer at the same time with

> Name "true" is reserved and cannot be used *for* an enum value.

src/language/parser.ts
```diff
@@ -1052,6 +1050,26 @@ export class Parser {
     });
   }
 
+  /**
+   * EnumValue : Name but not `true`, `false` or `null`
+   */
+  parseEnumValueName(): NameNode {
```
Yogu, line 1056: renamed this to `parseEnumValueName()` because it's no longer parsing into an `EnumValue`. but still having it in its own function keeps things clean.

src/language/__tests__/parser-test.ts
```diff
@@ -120,6 +120,24 @@ describe('Parser', () => {
     });
   });
 
+  it('does not allow "true", "false", or "null" as enum value', () => {
+    expectSyntaxError('enum Test { VALID, true }').to.deep.equal({
+      message:
+        'Syntax Error: Name "true" is reserved and cannot be used for an enum value.',
+      locations: [{ line: 1, column: 20 }],
+    });
```
IvanGoncharov, line 128: Can you please add empty lines in between to match other tests?

IvanGoncharov:
> It isn't a "custom" AST node type. EnumValueNode already existed and that's the name of the non-terminal used in the spec at the respective location. 

I checked the spec and you are completely right 👍 

> If we were to start fresh, it would probably be best to use EnumValueNode and rename name to value. 

I think it's would be surprising behaviors, e.g. you wrote your visitor for `EnumValue` so most devs including me will assume you would be triggered only on literal enum value (inside arguments or inside default values).
Vise versa if you register a visitor on `Name` you expect to be called on all names.
I think using `value` inside literals and `name` inside definition is pretty intuitive.

But I agree that this creates a mismatch with the spec.
However, I'm not sure how to address it in spec.

src/language/parser.ts
```diff
@@ -1052,6 +1050,26 @@ export class Parser {
...
+      this._lexer.token.value === 'true' ||
+      this._lexer.token.value === 'false' ||
+      this._lexer.token.value === 'null'
+    ) {
+      throw syntaxError(
+        this._lexer.source,
+        this._lexer.token.start,
+        `${getTokenDesc(
+          this._lexer.token,
+        )} is reserved and cannot be used for an enum value.`,
```
IvanGoncharov, line 1067: BTW. Adding a similar message to fragment name parsing is welcomed if you want to work on such PR.

PR #3231, opened 2021-08-11
execute: Forbid to return `null` from `serialize` function

IvanGoncharov:
Fixes #1579

### DIFF
--- a/src/execution/__tests__/executor-test.ts
+++ b/src/execution/__tests__/executor-test.ts
@@ -1123,7 +1123,7 @@ describe('Execute: Handles basic execution tasks', () => {
       errors: [
         {
           message:
-            'Expected a value of type "CustomScalar" but received: "CUSTOM_VALUE"',
+            'Expected `CustomScalar.serialize("CUSTOM_VALUE")` to return non-nullable value, returned: undefined',
           locations: [{ line: 1, column: 3 }],
           path: ['customScalar'],
         },
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -750,10 +750,10 @@ function completeLeafValue(
   result: unknown,
 ): unknown {
   const serializedResult = returnType.serialize(result);
-  if (serializedResult === undefined) {
+  if (serializedResult == null) {
     throw new Error(
-      `Expected a value of type "${inspect(returnType)}" but ` +
-        `received: ${inspect(result)}`,
+      `Expected \`${inspect(returnType)}.serialize(${inspect(result)})\` to ` +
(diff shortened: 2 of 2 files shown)


yaacovCR:
@IvanGoncharov I know a while back, but doesn’t this change (as opposed to #1579) raise field errors for null leaf values even when the types are nullable?

spec reference: first sentence here https://spec.graphql.org/draft/#sec-Non-Null.Result-Coercion

IvanGoncharov:
@yaacovCR It's forbidden by the spec here:
https://spec.graphql.org/draft/#sel-IANRLHHCDCECCGgB2xE

IvanGoncharov:
@yaacovCR Null is separate from scalars they are not converted to/from scalars.
Because it would cause them to not be reversable.

yaacovCR:
Thanks for quick response!

Ah, a reminder that nullable means that null is allowed/valid in response (does not cause error to bubble up), not that it’s actually not an error, which it sometimes isn’t (composite types) and sometimes is (for leaf values).

Makes sense in intent and practice now, appreciated!

yaacovCR:
I guess the error message should be “non-null” value not “non-nullable”.

IvanGoncharov:
> I guess the error message should be “non-null” value not “non-nullable”.

@yaacovCR It also includes `undefined` since it is nullable in JS (`undefined == null`).

yaacovCR:
I still think that’s a bit wrong grammar wise.

Undefined is not “nullable” . It’s not anything able, it’s just a set value, we could say the return value of serialize is not nullable if the parameter is not null, so it can’t return x value, but we can’t say the expected return value itself is a non nullable value.

anyway it’s splitting hairs, thanks for explaining!!!!

ephemer:
Hey @IvanGoncharov this PR changes behaviour that we've been relying on since 2017. We just ran into it by upgrading to Apollo Server 4, which uses graphql 16.x.

Basically we have a Scalar that – in serialize – looks to see if one of the object keys it was provided is `null`, and returns `null` in that case. We use this scalar in hundreds of places across our Schema, so pre-processing the value that is passed into serialize is not a good option.

```
    serialize({ relativeUrl, requestLocation } = {}) {
        if (!relativeUrl) {
            // allow for nullable ResourceURLs
            return null;
        }

        if (!requestLocation) {
            throw new Error("You must pass the user `requestLocation` when resolving ResourceURL");
        }

        return getResourceUrl(relativeUrl, requestLocation); // a string
    },
```

With the above we could just `return { relativeUrl: null, requestLocation }` for a nullable ResourceURL (the scalar type) and it will work fine. That is longer the case.

Can you think of a workaround?

PR #3238, opened 2021-08-17
Fix stream directive error on all directive declarations

andyrichardson:
# About

Minor bugfix - looks like the `@stream` directive validation wasn't being constrained exclusively to that directive.

I was seeing the following error 

```
// Query
{ articles { id ...ArticleFragment @anydirectivename(label: \"thing\") } } fragment ArticleFragment on Article { title }

// Error
"Stream directive cannot be used on non-list field \"articles\" on type \"Article\"."
```

### DIFF
--- a/src/validation/rules/StreamDirectiveOnListFieldRule.ts
+++ b/src/validation/rules/StreamDirectiveOnListFieldRule.ts
@@ -24,6 +24,7 @@ export function StreamDirectiveOnListFieldRule(
         fieldDef &&
         parentType &&
         isDirective(GraphQLStreamDirective) &&
+        node.name.value === "stream" &&
         !isListType(fieldDef.type)
       ) {
         context.reportError(


src/validation/rules/StreamDirectiveOnListFieldRule.ts
```diff
@@ -24,6 +24,7 @@ export function StreamDirectiveOnListFieldRule(
         fieldDef &&
         parentType &&
         isDirective(GraphQLStreamDirective) &&
```
andyrichardson, line 26: Not totally sure why this is here or if it is needed.

Assuming there is a reason for it :+1:

andyrichardson:
@robrichard do you mind giving this a quick pass? :pray:

src/validation/rules/StreamDirectiveOnListFieldRule.ts
robrichard: i think instead of this, i intended for a check to that the directive being validated is an instance of `GraphQLStreamDirective`

PR #3241, opened 2021-08-20
Test suggestionList more thoroughly

Cito:
Particularly, verify that it uses natural sort order instead of lexicographic sort order.


yaacovCR:
We need to make sure checks run before merging

PR #3246, opened 2021-08-26
re-export collectFields

PabloSzx:


ardatan:
Actually it would be nice to export `collectFields` directly from index file for the future.

saihaj:
> Actually it would be nice to export `collectFields` directly from index file for the future.

That would be better. One more breaking change but I think we should do this now

ardatan:
@saihaj If we both export `collectFields` from `execute` and index file, that won't be a breaking change, no?

saihaj:
Yep 👍🏼

PabloSzx:
> Yep 👍🏼

done :+1:

IvanGoncharov:
@PabloSzx Thanks for PR `collectFields` was internal function.
So v15 and v16 have different prototypes (last 2 arguments have different types) so they are incompatible anyway even if we export them.
I'm totally ok with making this function part of public API but before that, I need to do some changes to it.

PabloSzx:
@ardatan who is the lead maintainer of graphql-tools should be more interested in that comment

robrichard:
Calling out here that the experimental defer/stream branch also changes the api of `collectFields`. 

https://github.com/graphql/graphql-js/pull/2839/files#diff-b17f02a116457aa7749e04b3f306eba462f95979d4f76c313a6073fa201e4fe4R52

I think we need to carefully consider the api of this function before making it external, so that we can support defer/stream without introducing additional breaking changes.

Issue #3245, opened 2021-08-26
Feedback on v16.0.0-rc.1


IvanGoncharov:
Can you please test your lib/project against https://github.com/graphql/graphql-js/releases/tag/v16.0.0-rc.1 and leave feedback on this issue.

IvanGoncharov:
> I think we can export Visitor as an alias with a deprecation notice for now in order to keep backwards compatibility, no?

@ardatan `ASTVisitor` was added more than 2 years ago and is present in both v14 and v15, see:
https://github.com/graphql/graphql-js/commit/9366d6dfeda3458e3fa2135e39fe24d0a02e87ce#diff-559cd403e97ad511680f66b39577aff06d8b4e71349b7dc8a4732a31a43ddf75R9

IvanGoncharov:
Released a https://github.com/graphql/graphql-js/releases/tag/v16.0.0-rc.3 with all fixes included, big thanks for all PRs 👍 
Please feel free to report here or open a separate issue if you find some problem with RC3.
Also, I released v15.6.1 with some stuff backported from `main`, see: https://github.com/graphql/graphql-js/releases

Issue #3225, opened 2021-08-10
Refactor `visit` function


IvanGoncharov:
Extracted from #1145:

I very frequently use `visit` function in my work. However I never fully understood how it's working especially since entire `visitor.js` lacks typing.

I managed to figure out some things and add typings to entire file (with the exception of a few `any`) and all tests are passing. But I stuck since I can't understand original intentions for some of the features: 

@leebyron Could you please review this PR and also answer a few questions:
1) It [looks like](https://github.com/graphql/graphql-js/blob/master/src/language/visitor.js#L171)`visit` was designed to handle arrays in `root` as an argument, but this feature is not used anywhere. Is it required functionality?
2) `parent` of visitor callback is very strange. If the current node is not in an array it has a value of parent node as expected but if it's inside an array then `parent` is undefined. In this PR I made `parent` to always point to parent ASTNode as I think it's what's expected. Is it possible to merge such change?
3) `key` parameter is always equal to the last element of `path` array and can be either number or string depending on node location. I see that absolute majority of callback ignores this parameter, so maybe it's worth removing it (you always can get the same value from the path) to simplify callback signature?
4) `ancestors` is an array of both ASTNodes and arrays of ASTNodes which is very confusing + you can get the same array from root using `path` elements as indexes. In my PR I changed `ancestors` to be an array of parent ASTNodes. Is it possible to merge such change?

I saw you made a few small breaking changes in the recent PRs. I think it would be great to also include some interface cleanup of `visit` function. Since at the moment it scares away 3rd-party developers from writing tooling or contributing to existing projects. For example here is an issue reported on GraphQL Faker repository:
> We have done some investigation and it appears that the culprit is this method /src/proxy.ts@master#L122. Unfortunately, this is where we run out of depth, as we currently don't have the bandwidth to study the visit method / API properly.

IvanGoncharov:
Another idea it to replace `false` return value with something more readable, suggested by @mjmahone here:
https://github.com/graphql/graphql-js/pull/1466#pullrequestreview-149486078

yaacovCR:
Hi! Happy to continue discussion here about visitorKeys as mentioned in #3250. I think it’s useful to speed up the entire function to be able to specify what nodes should be visited. You can enter a node and return false to stop visiting, but then in certain scenarios I guess you would need to check the parent node type (like if you only want to enter certain name or value nodes). And it’s always a tiny bit faster to just not enter a node than you enter it and stop.

[...]

PR #3250, opened 2021-08-27
Bring `visitorKeys` back

ardatan:
See https://github.com/graphql/graphql-js/issues/3245#issuecomment-907177003

I checked `Allow edits by maintainers`. So you can make changes on this PR instead of creating a new one by checking out this branch using GitHub CLI. Also you can leave comments here then I can also do the changes you want :)

### DIFF
--- a/src/language/__tests__/visitor-test.ts
+++ b/src/language/__tests__/visitor-test.ts
@@ -896,7 +896,7 @@ describe('Visitor', () => {
 
   it('visits only the specified `Kind` in visitorKeyMap', () => {
     const visited: Array<[string, string, ReturnType<typeof getValue>]> = [];
-    
+
     const visitorKeyMap: VisitorKeyMap<ASTKindToNode> = {
       Document: ['definitions'],
       OperationDefinition: ['name'],
@@ -910,10 +910,10 @@ describe('Visitor', () => {
         visited.push(['leave', node.kind, getValue(node)]);
       },
     };
-    
+
     const exampleDocumentAST = parse(/* GraphQL */ `
-      query ExampleOperation { 
-        someField 
+      query ExampleOperation {
+        someField
       }
     `);
 
--- a/src/language/__tests__/visitor-test.ts
+++ b/src/language/__tests__/visitor-test.ts
@@ -3,10 +3,11 @@ import { describe, it } from 'mocha';
 
 import { kitchenSinkQuery } from '../../__testUtils__/kitchenSinkQuery';
 
-import type { ASTNode, SelectionSetNode } from '../ast';
+import type { ASTKindToNode, ASTNode, SelectionSetNode } from '../ast';
 import { isNode } from '../ast';
 import { Kind } from '../kinds';
 import { parse } from '../parser';
+import type { VisitorKeyMap, ASTVisitor } from '../visitor';
 import { visit, visitInParallel, BREAK } from '../visitor';
 
 function checkVisitorFnArgs(ast: any, args: any, isEdited: boolean = false) {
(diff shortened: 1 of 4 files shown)


yaacovCR:
Returning false means you no longer visit any of the children, what if you want to visit only some?

yaacovCR:
@IvanGoncharov  why exactly was it deprecated even in terms of its original rationale of custom ast?

in general, for use cases above, it lets you whitelist what you want to visit rather than entering each child node with a visitor function that returns false, the latter being slightly less performant, and sometimes quite more verbose.

yaacovCR:
Related: https://github.com/ardatan/graphql-tools/pull/3481

yaacovCR:
we can workaround the above bug secondary to lack of visitorKeys, but it doesn't change that we are doing too much visiting :(

we want to be speedy; if visitorKeys is dropped upstream, we will have to re-implement the v15 version -- please keep

src/language/__tests__/visitor-test.ts
```diff
@@ -893,6 +894,41 @@ describe('Visitor', () => {
...
+    const visitor: ASTVisitor = {
+      enter(node) {
+        visited.push(['enter', node.kind, getValue(node)]);
+      },
+      leave(node) {
+        visited.push(['leave', node.kind, getValue(node)]);
+      },
+    };
+
+    const exampleDocumentAST = parse(/* GraphQL */ `
```
IvanGoncharov, line 914: We don't use `/* GraphQL */` tag in our source code.
All tests in this file name their variable simply `ast`.
Always try to match the style of another test.

src/language/__tests__/visitor-test.ts
```diff
@@ -893,6 +894,41 @@ describe('Visitor', () => {
...
+      query ExampleOperation {
+        someField
+      }
+    `);
+
+    visit(exampleDocumentAST, visitor, visitorKeyMap);
+
+    expect(visited).to.deep.equal([
+      ['enter', Kind.DOCUMENT, undefined],
```
IvanGoncharov, line 923: All other tests use literal values instead of `Kind` properties.

src/language/__tests__/visitor-test.ts
```diff
@@ -893,6 +894,41 @@ describe('Visitor', () => {
     ]);
   });
 
+  it('visits only the specified `Kind` in visitorKeyMap', () => {
+    const visited: Array<[string, string, ReturnType<typeof getValue>]> = [];
```
IvanGoncharov, line 898: All other test use `Array<any>`.
If you want to improve typings it should be done in a separate PR and for the whole file.

src/language/visitor.ts
```diff
@@ -76,9 +76,14 @@ type ReducedField<T, R> = T extends null | undefined
   ? ReadonlyArray<R>
   : R;
 
-const QueryDocumentKeys = {
-  Name: [],
+/**
+ * A KeyMap describes each the traversable properties of each kind of node.
+ */
+export type VisitorKeyMap<KindToNode> = {
```
IvanGoncharov, line 82: Since we dropped support for custom ASTs we don't need `KindToNode` argument also I renamed `VisitorKeyMap` to `ASTVisitorKeyMap` and will provide an alias for `v15`.

IvanGoncharov:
> I checked Allow edits by maintainers. So you can make changes on this PR instead of creating a new one by checking out this branch using GitHub CLI. Also you can leave comments here then I can also do the changes you want :)

@ardatan That's great, however, I created new PR for a different reason.
In this and your previous PRs, you add code that doesn't match style of the whole file.
A test case that you added in this PR gives a good example of what I'm talking about.
All other test cases follow the same patterns and the same name convention so I expect one that you added to also do this.
I'm 100% up for providing necessary context when reviewing PR but I expect the PR authors to review files they are changing and code around the change.
Without that, I need to check every line of the proposed change and match it with the rest of the code (very hard in GitHub review UI) and spend time left review comments.
And that is a pretty time-consuming and exhausting process, so it's simpler for me to write code from scratch and create new PR.
For example, even though this PR is added single test case it took me way more time to review it and leave comments than to simply from it from scratch using other test cases as templates.

This situation repeats itself and was the major issue with all your previous PRs.
So if you want your future PRs to be merged please spend time and try to match the code style and conventions used in this project. If you to improve something (e.g. typings) do it in separate PR and for all places that have the pattern, you want to change.

P.S. I applied changes to this PR and merging it, but the above comment stays true for future PRs.

IvanGoncharov:
@yaacovCR I'm merging this one but I think it's not a good long-term solution to expose such a low-level mechanism and also think it's brittle and non-intuitive API. So it should be a part of #3225 and want to continue the discussion there.

IvanGoncharov:
@ardatan Added `ASTVisitorKeyMap` to v15 see #3267 will release new version of v15 shortly

leszekhanusz:
Another use for the visitorKeys is present in graphql-python/gql to [parse the results coming from the backend](https://github.com/graphql-python/gql/blob/47aaf92c995bcdfbdda38ffc9997bae6d64da479/gql/utilities/parse_result.py#L37)

PR #3251, opened 2021-08-27
backport parser typescript type-definitions for 15.x.x release

n1ru4l:
The 15.x.x release line did not have type definitions for the `Parser` class. We added type-definitions within our code-base, but with version 16.x.x being released that causes issues. Thus the easiest solution would be to have both type-definitions being provided from graphql-js.


n1ru4l:
@IvanGoncharov Any updates or ETA on the release of 15.5.2?

IvanGoncharov:
@n1ru4l Released

PR #3255, opened 2021-09-02
Add TResult to GraphQLFieldResolver signature

tgriesser:
Tried out the 16.0.0-rc and noticed the change from return type of `GraphQLFieldResolver` from [any in 15.x](https://github.com/graphql/graphql-js/blob/acc468bb550222d6ac0893d0655720d540e24566/src/type/definition.d.ts#L467-L476) to `unknown`. While this type change is correct, it'd be better if the generic could be specified by a consumer library, for instance if maintaining the previous behavior of `any` is desired in a library usage.

### DIFF
--- a/src/type/definition.ts
+++ b/src/type/definition.ts
@@ -965,12 +965,13 @@ export type GraphQLFieldResolver<
   TSource,
   TContext,
   TArgs = { [argument: string]: any },
+  TResult = unknown,
 > = (
   source: TSource,
   args: TArgs,
   context: TContext,
   info: GraphQLResolveInfo,
-) => unknown;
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@tgriesser Thanks a lot.

> While this type change is correct, it'd be better if the generic could be specified by a consumer library, for instance if maintaining the previous behavior of any is desired in a library usage.

Totally agree 👍 
Also, it should help server-side codegen.

I will release it in the next RC.
I also can add this argument to 15.x.x line, if you need to support both versions in your lib, just create PR against the `15.x.x` branch.

PR #3264, opened 2021-09-18
lexer: Remove superfluous statement in `readDigits`

Cito:
Setting `code = firstCode` does not make sense here, since it is immediately overwritten. Either just `let code;` or use an empty loop as I have done here.

### DIFF
--- a/src/language/lexer.ts
+++ b/src/language/lexer.ts
@@ -490,11 +490,10 @@ function readDigits(lexer: Lexer, start: number, firstCode: number): number {
 
   const body = lexer.source.body;
   let position = start;
-  let code = firstCode;
 
   do {
-    code = body.charCodeAt(++position);
-  } while (isDigit(code));
+    // empty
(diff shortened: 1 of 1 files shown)


saihaj (approved):

src/language/lexer.ts
```diff
@@ -490,11 +490,10 @@ function readDigits(lexer: Lexer, start: number, firstCode: number): number {
 
   const body = lexer.source.body;
   let position = start;
-  let code = firstCode;
 
   do {
-    code = body.charCodeAt(++position);
-  } while (isDigit(code));
+    // empty
```
IvanGoncharov: Empty loops can look suspicious in code.
I moved increate to be the loop body since it's the purpose of this code to calculate position.

IvanGoncharov:
@Cito
Code quality and readability are a top priority for reference implementation.
Thanks a lot for reviewing and contributing back  👍

PR #3278, opened 2021-09-28
lexer: fix expression to decode surrogate pairs

Cito:
When calculating the code point for a surrogate pair, the base code point of 0x10000 must be added, not bitwise ORed. Added some tests for catching this problem.

### DIFF
--- a/src/language/__tests__/lexer-test.ts
+++ b/src/language/__tests__/lexer-test.ts
@@ -1040,6 +1040,21 @@ describe('Lexer', () => {
       locations: [{ line: 1, column: 1 }],
     });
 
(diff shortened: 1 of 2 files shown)


Cito:
> Can you please give a link to some document referencing this is the correct way

Hi @IvanGoncharov - have a look at the C function `decode_utf16_pair` at the bottom of [this](https://unicodebook.readthedocs.io/unicode_encodings.html) page.

It doesn't matter whether you use '|' or '+' to combine the trailing and shifted leading values, because their bit patterns do not overlap. But you can't use '|' to add the `0x10000` offset, because its bit pattern overlaps with the shifted leading value.

PR #3282, opened 2021-09-30
Validate variables are known types

spawnia:
Adds a check to the validation rule VariablesAreInputTypes to ensure
that referenced types exist at all before checking they are input types.

This check is not explicitly described in the specification for validating
variables: http://spec.graphql.org/draft/#sec-Validation.Variables
I am assuming this is simply an oversight and the rule is implicit.
Still, it is valuable to check for it in order to provide a proper error to clients.

### DIFF
--- a/src/validation/__tests__/VariablesAreInputTypesRule-test.ts
+++ b/src/validation/__tests__/VariablesAreInputTypesRule-test.ts
@@ -21,6 +21,23 @@ describe('Validate: Variables are input types', () => {
     `);
   });
 
+  it('unknown types are invalid', () => {
+    expectErrors(`
+      query Foo($a: Unknown, $b: [[Unknown!]]!) {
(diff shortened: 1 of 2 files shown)


spawnia:
Is this rule the correct place to add this validation? Should we modify the spec to make this validation rule explicit?

IvanGoncharov:
@spawnia Thanks for opening PR 👍 
I simplified it a bit in d41e2b6 and almost merged, but notice it's duplicated in "KnownTypeNames" rules.
I used this opportunity to make it more explicit in #3284

Since code changes are not relevant anymore I'm closing this PR.

> Is this rule the correct place to add this validation? Should we modify the spec to make this validation rule explicit?

I did a quick cross-check and we have a bunch of discrepancies with the spec, so I opened a separate issue for that: #3286
If you are interested I can assign it to you?

PR #3288, opened 2021-10-05
Move validation of names into `GraphQL*` constructors

IvanGoncharov:


yaacovCR:
@IvanGoncharov do you remember the motivation for this change

tagging also those following #4262 @ab-pm @benjie

PR #3293, opened 2021-10-07
Align calls of buildExecutionContext

yaacovCR:
`buildExecutionContext` used to throw an error on failure, but was changed
to return an array of errors (enabling the function to return multiple
errors encountered during variable coercion).

As the function is no longer designed to throw, it no longer needs to be
wrapped in a try block when used within the `subscribe` function. This
change aligns the use of `buildExecutionContext between the `subscribe`
and `execute` functions.


saihaj:
@yaacovCR it is failing because of npm outage https://status.npmjs.org/

PR #3294, opened 2021-10-07
Properly name the root field execution functions

yaacovCR:
The codebase should refer to functions that execute the query, mutation, and/or subscription root fields as such, rather than as functions that execute the operations themselves. Executing a query or mutation returns a map of data and errors, while executing the root fields returns the completed value of the root fields.

Depends on #3306

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -166,7 +166,7 @@ export interface ExecutionArgs {
  * a GraphQLError will be thrown immediately explaining the invalid input.
  */
 export function execute(args: ExecutionArgs): PromiseOrValue<ExecutionResult> {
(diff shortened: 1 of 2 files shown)


yaacovCR:
That is, each could be merged into main consecutively. Maybe you meant to merge #3295 into main?

yaacovCR:
> I will ping you once finish, so you don't need to rebase multiple times.

I rebased everything. It's all ready to review and merge. This is the now the second PR in stack of current depth 9 numbered, beginning with #3306, and then #3294, followed by #3296, #3297, #3298, #3299, #3300, #3301, and #3302. They are ready for review!

PR #3300, opened 2021-10-07
Move executor implementation to separate file

yaacovCR:
Depends on #3299


yaacovCR:
As per our offline discussion, I am forking the executor only and will be pushing forward asap with class based Executor over there, in order to jumpstart experimentation. Hopefully whatever works at graphql-executor can be integrated upstream as desired.

yaacovCR#6

PR #3301, opened 2021-10-07
Refactor Executor implementation into class

yaacovCR:
Depends on #3300

PR #3302, opened 2021-10-07
Export internal Executor class for experimentation

yaacovCR:
This class, similar to the Parser class, is exported only to assist people in implementing their own executors without duplicating too much code and should be used only as last resort for cases such as experimental syntax or if certain features could not be contributed upstream.

It is still part of the internal API and is versioned, so any changes to it are never considered breaking changes. If you still need to support multiple versions of the library, please use the `versionInfo` variable for version detection.

Depends on #3301

### DIFF
--- a/src/execution/executor.ts
+++ b/src/execution/executor.ts
@@ -138,6 +138,16 @@ export interface ExecutionArgs {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


yaacovCR:
This is the last PR in stack of current depth 9 numbered consecutively from #3294 => #3302. They are ready for review!

PR #3305, opened 2021-10-11
Added 'GraphQLSchema.getRootType' and deprecate `getOperationRootType`

IvanGoncharov:
cc: @yaacovCR


yaacovCR:
Just a thought, maybe these new tests using `GraphQLSchema.getRootType` belong in the next PR with the fix.

PR #3312, opened 2021-10-12
language: change OperationTypeNode to enum

lekoaf:
Changes OperationTypeNode to enum instead of a union type.
This allows developers to use the type as `OperationTypeNode.QUERY`
instead of risking to msispell the string 'query'.

Example:

Instead of having to write

`operation === 'query'`

you can instead write

`operation === OperationTypeNode.QUERY`

(This is a continuation of #2887)


lekoaf:
Hmm, according to the linter, this may be a breaking change after all?

PR #3353, opened 2021-11-01
v16 regression: parsed GraphQL comments has invalid (negative) `token.column`

dotansimha:
While working on `graphql-eslint` (https://github.com/dotansimha/graphql-eslint/pull/724) and upgrading it to v16, we noticed that parsed comments has negative values in `startToken.column`. v15 was working fine.

Managed to add a failing test to ensure all columns are positive. I think it happens only when there are GraphQL comments at the beginning of the document. 

@IvanGoncharov @saihaj thoughts? 

> the test case I added isn't ideal, it just tries to confirm the existence of the bug, and not the correctness of the positions

### DIFF
--- a/src/language/__tests__/parser-test.ts
+++ b/src/language/__tests__/parser-test.ts
@@ -11,6 +11,7 @@ import { Kind } from '../kinds';
 import { Source } from '../source';
 import { TokenKind } from '../tokenKind';
 import { parse, parseValue, parseConstValue, parseType } from '../parser';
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@dotansimha Thanks for the test case. I will take a look.

IvanGoncharov:
@dotansimha Fixed in https://github.com/graphql/graphql-js/releases/tag/v16.0.1

PR #3355, opened 2021-11-02
Export GRAPHQL_MAX_INT and GRAPHQL_MIN_INT

### LINKED ISSUE #2524: Add max int to exports

tofran:
Closes #2524

(Replaces abandoned #2582)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -61,6 +61,9 @@ export {
   TypeKind,
   /** Constant Deprecation Reason */
   DEFAULT_DEPRECATION_REASON,
+  /** Int boundaries constants */
+  GRAPHQL_MAX_INT,
+  GRAPHQL_MIN_INT,
   /** GraphQL Types for introspection. */
   introspectionTypes,
   __Schema,
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,6 +51,9 @@ export {
   GraphQLString,
   GraphQLBoolean,
   GraphQLID,
+  /** Int boundaries constants */
+  GRAPHQL_MAX_INT,
+  GRAPHQL_MIN_INT,
   /** Built-in Directives defined by the Spec */
   specifiedDirectives,
   GraphQLIncludeDirective,
@@ -61,9 +64,6 @@ export {
   TypeKind,
   /** Constant Deprecation Reason */
   DEFAULT_DEPRECATION_REASON,
-  /** Int boundaries constants */
-  GRAPHQL_MAX_INT,
-  GRAPHQL_MIN_INT,
   /** GraphQL Types for introspection. */
   introspectionTypes,
   __Schema,
--- a/src/type/scalars.ts
+++ b/src/type/scalars.ts
@@ -9,13 +9,17 @@ import { GraphQLError } from '../error/GraphQLError';
 import type { GraphQLNamedType } from './definition';
 import { GraphQLScalarType } from './definition';
 
-// As per the GraphQL Spec, Integers are only treated as valid when a valid
-// 32-bit signed integer, providing the broadest support across platforms.
-//
-// n.b. JavaScript's integers are safe between -(2^53 - 1) and 2^53 - 1 because
-// they are internally represented as IEEE 754 doubles.
-export const MAX_INT = 2147483647;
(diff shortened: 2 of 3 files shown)


saihaj (approved):

src/type/scalars.ts
```diff
@@ -14,8 +14,8 @@ import { GraphQLScalarType } from './definition';
 //
 // n.b. JavaScript's integers are safe between -(2^53 - 1) and 2^53 - 1 because
 // they are internally represented as IEEE 754 doubles.
-const MAX_INT = 2147483647;
-const MIN_INT = -2147483648;
+export const MAX_INT = 2147483647;
```
IvanGoncharov, line 17: Please convert above comment to JSDoc
If you have time please write a separate docs for both min and max

IvanGoncharov: Also since we making it public please rename it to `GRAPHQL_INT_MAX` and `GRAPHQL_INT_MIN`.

IvanGoncharov (changes requested):
Please export it in `src/type/index.ts` and `src/index.ts`.
Also please check tests folder for the places where this number (or +-1) is used and replace them with this constant.

tofran:
Hello @IvanGoncharov , thank you for bringing these in and your detailed comment.
I have made the requested changes:

- [x] Checked the tests folder for the places where this number (or +-1) is used and replace them with this constant.*

```sh
$ grep -R "214748364" src  # Intentionally missing the last character
src/type/scalars.ts:export const GRAPHQL_MAX_INT = 2147483647;
src/type/scalars.ts:export const GRAPHQL_MIN_INT = -2147483648;
$ grep -R "2^31" src
src/type/scalars.ts:    'The `Int` scalar type represents non-fractional signed whole numeric values. Int can represent values between -(2^31) and 2^31 - 1.',
$ grep -R "2 \*\* 31" src

```

- [x] Converted above comment to JSDoc - separated and simplified
- [x] Exported it in src/type/index.ts and src/index.ts.
- [x] Renamed to `GRAPHQL_INT_MAX` and `GRAPHQL_INT_MIN`.

---

*Were you also suggesting that tests to these boundaries should be changed?
Ex:
```js
expect(() => parseLiteral('-9876504321')).to.throw(
  'Int cannot represent non 32-bit signed integer value: -9876504321',
);
```
Would now be
```js
expect(() => parseLiteral(GRAPHQL_MIN_INT-1)).to.throw(
  `Int cannot represent non 32-bit signed integer value: ${GRAPHQL_MIN_INT-1}`,
);
```

Should I make this change? Thank you.

saihaj (approved):

src/index.ts
```diff
@@ -61,6 +61,9 @@ export {
   TypeKind,
   /** Constant Deprecation Reason */
   DEFAULT_DEPRECATION_REASON,
+  /** Int boundaries constants */
+  GRAPHQL_MAX_INT,
+  GRAPHQL_MIN_INT,
```
IvanGoncharov, line 66: Please match the order in `types/index.ts`

IvanGoncharov: Everything else looks good.
Happy to merge once fixed.

tofran:
@IvanGoncharov @saihaj thank you.
Just pushed the requested changes, ready to merge.

saihaj (approved):

tofran:
@IvanGoncharov I don't think we will be able to merge until you re-review.
Thanks in advance.

IvanGoncharov:
@tofran Thanks 👍 
Merged 🎉

src/type/scalars.ts
jen801: 👍🏻

jen801 (commented):
#3355

Nexuscompute (approved):

saihaj:
Just to avoid noise from potential bot accounts.

Issue #3356, opened 2021-11-02
Upgrade issues with TypeScript enums and nominal typing


brainkim:
Hullo! I’m upgrading some Apollo projects to GraphQL 16 and running into slight difficulties with enums. The context is that we have utilities which create or modify AST nodes:

```diff
diff --git a/src/utilities/graphql/transform.ts b/src/utilities/graphql/transform.ts
index 2a5128d05..80db6d202 100644
--- a/src/utilities/graphql/transform.ts
+++ b/src/utilities/graphql/transform.ts
@@ -2,6 +2,7 @@ import { invariant } from '../globals';
 
 import {
   DocumentNode,
+  Kind,
   SelectionNode,
   SelectionSetNode,
   OperationDefinitionNode,
@@ -53,9 +54,9 @@ export type RemoveVariableDefinitionConfig = RemoveNodeConfig<
 >;
 
 const TYPENAME_FIELD: FieldNode = {
-  kind: 'Field',
+  kind: 'Field' as Kind.FIELD,
   name: {
-    kind: 'Name',
+    kind: 'Name' as Kind.NAME,
     value: '__typename',
   },
 };
```

The code is semantically sound, but TypeScript requires the assertion because you must reference enums directly wherever they are required (nominal typing).

```ts
enum Foo {
  a = "A",
  b = "B",
}

// This is a type error because we’re not referencing the enum directly.
const foo: Foo = "A";
// This is fine
const foo1: Foo = Foo.a;
```

The disadvantages of this change in GraphQL 16 are:
1. We have to add runtime dependencies on the files which contain the enums.
2. It’s slightly tricky to support ranges of graphql-js versions.

As a specific example of an enum-based difficulty, `OperationTypeNode` is an enum in GraphQL 16, but it’s not exported as a value in GraphQL 15, so there is no way to reference the enum value across 15 and 16, and you get type errors if you assign strings like `"query"` or `"mutation"` to the enum.

I think the above two concerns are reason enough to switch to using string unions, which are structurally typed, and don’t require direct reference. We can make this a non-breaking change by exporting a namespace instead of an enum, and shadowing the identifier on the type-level with a string union.

```ts
export type Kind = "Name" | "Document" /* ... */;
export namespace Kind {
  export const NAME = 'Name';
  export const DOCUMENT = 'Document';
  /* ... */
}
```

Thanks for reading. Let me know your thoughts!

saihaj:
Why not use the `enum` directly? 
```diff
[...]

PR #3360, opened 2021-11-03
avoid using enums in favor of frozen object literals

brainkim:
Implements the suggestion by @benjie (https://github.com/graphql/graphql-js/issues/3356#issuecomment-958969934)


IvanGoncharov:
@brainkim This is basically a revert of #3317 but with value and type sharing the same name.
After looking into it more I see that TS enums are broken.

That said, this is a breaking change so we need to release v17 for it.
So I suggest waiting a week and seeing if any other issue will come up that requires breaking release in order to be fixed.
Also please test this solution and be sure it fully meets your requirements since it would be very painful to change it after this PR is merged.

src/language/ast.ts
```diff
@@ -323,14 +323,18 @@ export interface OperationDefinitionNode {
...
-}
+export const OperationTypeNode = Object.freeze({
+  QUERY: 'query',
+  MUTATION: 'mutation',
+  SUBSCRIPTION: 'subscription',
+} as const);
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type OperationTypeNode =
+  typeof OperationTypeNode[keyof typeof OperationTypeNode];
```
IvanGoncharov, line 334: Please extract it into `jsutils` under descriptive name, e.g. `ValueTypes`
totally open to bike shedding

yaacovCR:
Why is it breaking if essentially relaxes type more than the auto type included within Enums?

brainkim:
So the only breaking part of this change is that there will no longer be type-level emits of the enum values (`type: OperationTypeNode.QUERY` -> `type: typeof OperationTypeNode.QUERY`). But this is easy to notice, easy to fix, and some of the enum types, specifically the legacy aliases, are slightly broken anyways, so I think there’s a way to sell this as a patch change.

benjie:
I'd sell it as a TypeScript fix for compatibility with v15 (assuming it improves compatibility with v15).

brainkim:
@IvanGoncharov Let me know what sort of tests you’re expecting if you get a chance!

IvanGoncharov:
> Let me know what sort of tests you’re expecting if you get a chance!

@brainkim No additional tests are needed, but please fix prettier and other linters.

IvanGoncharov:
> Why is it breaking if essentially relaxes type more than the auto type included within Enums?

@yaacovCR  It's breaking as Brian showed, also it breaking for a number of other reasons, e.g. enum merging, `keyof Kind` is not working, etc.

I think we should be truthful to the community and just release a new major version and not try to hide this change under the rug.

I also think this gives the opportunity to do a proper fix and not hack.
Having the same name for two exported things is bad, couple reasons from the top of my head:
* Tooltips in IDE would be confusing 
* We will have problems with generated docs (the plan was to require JSDoc comment on everything that is exported). 
* We need to have ESLint disable comment every time we define an enum
* It's super weird that `typeof Kind` and `Kind` are totally different types.

TS enums are bad but what leads us to a hacky and non-obvious solution that tries to imitate how should work in the first place.
Instead, I recognize that #3317 was a bad idea and we just need to partly revert it and have `Kind` (as const literal) and `KindEnum` (as type).
That makes the solution non-confusing and easier to understand and document.

yaacovCR:
Hi Ivan! So you are advocating for a partial reversion of #3317 in v16? And the full change here for v17?

brainkim:
@IvanGoncharov It’s technically a breaking change, but it’s unlikely to affect people according to analysis above. It’s your call, but if it’s going in a major release, I’d like to know the timeline for release so I can throw a bunch of `any`s into Apollo Client to avoid having to deal with this 😇

benjie:
I think it would definitely be worth fully evaluating the differences before committing. I have to agree that TypeScript enums are broken, and that `Kind` and `typeof Kind` differing is definitely undesirable (and liable to future breakage) - having thought about it more I'd probably go with `KindMap` (or similar) for the lookup object, and `Kind` for the actual kind (a string):

```ts
export const KindMap = {
  OBJECT: 'ObjectType' as const,
  OBJECT_FIELD: 'ObjectField' as const
};
export type Kind = typeof KindMap[keyof typeof KindMap];

const a: Kind = KindMap.OBJECT;
const b: Kind = 'ObjectType';

export interface ASTNode {
  kind: typeof KindMap.OBJECT
}
```

https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7wNIEtkBMCyAhmHALxwDeAUHHAPIBCAUgKIDCAKgFxwDkdAIwBWwBDA4BPMMF5xCqRCnQAaGvWbsOAfQBiASRYAZACI9+w0TB1ZgAG1yz5itDCoBfANxVQkWHBhSwHDYeGT+gRAAZsE4BMQA2gDWwBJR4dJpIXFgALpeVEgucjxZYVlEYAB0jKycXoXocAIlsWHmImKS0rz5PtDwODDAUJGECEEAggDKHAByELhB1LSJsTwBGdHlxNUanO5AA

brainkim:
Not sure why the linting is failing btw
Screen Shot 2021-11-08 at 8 31 22 AM

brainkim:
Another data point from @ferm10n https://github.com/graphql/vscode-graphql/pull/335#issuecomment-964669387

m1heng:
🤔 Just realized that if this merged, https://github.com/apollographql/federation/pull/1345 some of my current work maybe need revert too?

yaacovCR:
@brainkim main is now unstable, i.e. representing the next v17 alpha.

Do you have any further thoughts about the best path forward, if anything in your estimation has changed, etc, etc?

Would you be able to rebase this PR?

PR #3361, opened 2021-11-04
deploy full ESM with "type": "module" in "latest-esm" npm tag

PabloSzx:
I also just re-published graphql@16.0.1 as graphql-esm@16.0.1 https://npm.im/graphql-esm, tested it in an example project using libraries with full ESM support https://github.com/PabloSzx/test-graphql-esm, everything works perfectly 👌 

It can be tested right now doing this:
```json
{
  "dependencies": {
    "graphql": "npm:graphql-esm@^16.0.1"
  }
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,6 @@
 /node_modules
 /reports
 /npmDist
+/npmEsmDist
 /denoDist
 /websiteDist
--- a/integrationTests/ts/package.json
+++ b/integrationTests/ts/package.json
@@ -7,6 +7,7 @@
   },
   "dependencies": {
     "graphql": "file:../graphql.tgz",
+    "graphql-esm": "file:../graphql-esm.tgz",
     "typescript-4.4": "npm:typescript@4.4.x",
     "typescript-4.5": "npm:typescript@4.5.x",
     "typescript-4.6": "npm:typescript@4.6.x",
--- a/integrationTests/webpack/webpack.config.json
+++ b/integrationTests/webpack/webpack.config.json
@@ -1,8 +1,11 @@
 {
   "mode": "production",
-  "entry": "./entry.js",
+  "entry": {
+    "cjs": "./entry.js",
+    "mjs": "./entry-esm.mjs"
+  },
   "output": {
-    "filename": "main.cjs",
+    "filename": "main-[name].cjs",
     "library": {
       "type": "commonjs2"
     }
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,6 +5,7 @@
 /node_modules
 /reports
 /npmDist
+/npmEsmDist
 /denoDist
 /websiteDist
 
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -473,6 +473,10 @@ rules:
   yield-star-spacing: off
 
 overrides:
+  - files:
+      - 'integrationTests/node-esm/**/*.js'
+    parserOptions:
+      sourceType: module
   - files: '**/*.ts'
     parser: '@typescript-eslint/parser'
     parserOptions:
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,5 +6,6 @@
 /node_modules
 /reports
 /npmDist
+/npmEsmDist
 /denoDist
 /websiteDist
--- a/integrationTests/node/index.mjs
+++ b/integrationTests/node/index.mjs
@@ -2,13 +2,13 @@
 import assert from 'assert';
 import { readFileSync } from 'fs';
 
-import { graphqlSync } from 'graphql';
-import { buildSchema } from 'graphql/utilities';
-import { version } from 'graphql/version';
+import { graphqlSync } from 'graphql-esm';
+import { buildSchema } from 'graphql-esm/utilities';
+import { version } from 'graphql-esm/version';
 
 assert.deepStrictEqual(
-  version,
-  JSON.parse(readFileSync('./node_modules/graphql/package.json')).version,
+  version + '+esm',
+  JSON.parse(readFileSync('./node_modules/graphql-esm/package.json')).version,
 );
 
 const schema = buildSchema('type Query { hello: String }');
--- a/integrationTests/node/package.json
+++ b/integrationTests/node/package.json
@@ -6,6 +6,7 @@
     "test": "node test.js"
   },
   "dependencies": {
-    "graphql": "file:../graphql.tgz"
+    "graphql": "file:../graphql.tgz",
+    "graphql-esm": "file:../graphql-esm.tgz"
   }
 }
--- a/integrationTests/node/test.js
+++ b/integrationTests/node/test.js
@@ -14,7 +14,12 @@ for (const version of nodeVersions) {
   console.log(`Testing on node@${version} ...`);
 
   childProcess.execSync(
-    `docker run --rm --volume "$PWD":/usr/src/app -w /usr/src/app node:${version}-slim node ./index.js`,
+    `docker run --rm --volume "$PWD":/usr/src/app -w /usr/src/app node:${version}-slim node ./index.cjs`,
+    { stdio: 'inherit' },
+  );
+
+  childProcess.execSync(
+    `docker run --rm --volume "$PWD":/usr/src/app -w /usr/src/app node:${version}-slim node ./index.mjs`,
     { stdio: 'inherit' },
   );
 }
--- a/integrationTests/ts/esm.ts
+++ b/integrationTests/ts/esm.ts
@@ -1,0 +1,38 @@
+import type { ExecutionResult } from 'graphql-esm/execution';
+
+import { graphqlSync } from 'graphql-esm';
+import {
+  GraphQLString,
+  GraphQLSchema,
+  GraphQLObjectType,
+} from 'graphql-esm/type';
+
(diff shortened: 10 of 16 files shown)


resources/build-npm-esm.js
```diff
@@ -0,0 +1,166 @@
+'use strict';
```
IvanGoncharov, line 1: since most of the code is shared with `build-npm.js` it makes sense to just add a cli option to it.
Also please check all places where `build-npm` is used directly or through NPM cli and adjust them accordingly.

IvanGoncharov (commented):
Please add `integrationTests/esm` for this package.
Try to test as many esm-specific things as possible.

IvanGoncharov:
Also please ensure that this package works with upcoming TS 4.5 that should support esm natively and add integration tests for that.

integrationTests/ts-esm/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "private": true,
+  "description": "graphql-js ESM should compile with all supported TS versions",
+  "scripts": {
+    "test": "node test.js"
+  },
+  "dependencies": {
+    "graphql": "file:../graphql-esm.tgz",
+    "typescript-4.2": "npm:typescript@4.2.x",
```
PabloSzx, line 9: For some reason Typescript@4.1.x didn't work, I'm not completely sure the reason :man_shrugging:

resources/build-npm.js
```diff
@@ -8,30 +8,50 @@ const ts = require('typescript');
...
+    const destPath = path.join(distDirectory, filepath);
 
     fs.mkdirSync(path.dirname(destPath), { recursive: true });
     if (filepath.endsWith('.ts')) {
-      const cjs = babelBuild(srcPath, { envName: 'cjs' });
-      writeGeneratedFile(destPath.replace(/\.ts$/, '.js'), cjs);
-
-      const mjs = babelBuild(srcPath, { envName: 'mjs' });
-      writeGeneratedFile(destPath.replace(/\.ts$/, '.mjs'), mjs);
+      if (isFullESM) {
```
IvanGoncharov, line 40: Please rename it to `isESMOnly`

resources/build-npm-esm.js
```diff
@@ -0,0 +1,9 @@
...
+
+const { buildNPM } = require('./build-npm.js');
+
+if (require.main === module) {
+  buildNPM({
+    isFullESM: true,
+  });
+}
```
IvanGoncharov, line 9: please use CLI flag (e.g. `--esm-only`) to `build-npm`

integrationTests/webpack/webpack.config.json
```diff
@@ -2,6 +2,8 @@
   "mode": "production",
   "entry": "./entry.js",
   "output": {
-    "libraryTarget": "commonjs2"
+    "library": {
+      "type": "commonjs2"
+    }
```
IvanGoncharov, line 7: Is it a fix?
If so please extract it in a separate PR and provide an explanation for it.

integrationTests/ts/package.json
```diff
@@ -9,6 +9,8 @@
     "typescript-4.1": "npm:typescript@4.1.x",
     "typescript-4.2": "npm:typescript@4.2.x",
     "typescript-4.3": "npm:typescript@4.3.x",
-    "typescript-4.4": "npm:typescript@4.4.x"
+    "typescript-4.4": "npm:typescript@4.4.x",
+    "typescript-4.5": "npm:typescript@4.5.1-rc",
+    "typescript-next": "npm:typescript@next"
```
IvanGoncharov, line 14: please don't do it next, if TS releases a new version it will break CI.

integrationTests/ts-esm/package.json
```diff
@@ -0,0 +1,15 @@
+{
```
IvanGoncharov, line 1: AFAIK, ESM changes only imports, it doesn't change any other aspect.
Please don't duplicate the entire test folder for it.
Just create a separate ESM-specific test file and alias ESM packages as `graphql-esm`.

IvanGoncharov: Please do the same for webpack

integrationTests/node-esm/index.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable node/no-missing-import, import/no-unresolved */
```
IvanGoncharov, line 1: Please don't just duplicate `node` test here.
Please try to test ESM-specific things here (e.g. importing `package.json`, dynamic import, etc.).
Also please create a test setup where you have an intermediate package with `peerDepenedicies` for `graphql` and test how it works with ESM package.

package.json
```diff
@@ -56,6 +56,7 @@
     "@babel/plugin-transform-typescript": "7.15.8",
     "@babel/preset-env": "7.15.8",
     "@babel/register": "7.15.3",
+    "@types/babel__core": "^7.1.16",
```
IvanGoncharov, line 59: Please explain this change?
Please do that for other non-straight forward changes, beforehand.

integrationTests/webpack/webpack.config.json
PabloSzx: image


I will remove it, but I won't make a separate PR just for

integrationTests/ts/package.json
PabloSzx: > please don't add next, if TS releases a new version it will break CI.

No problem, but I was just doing what you told me to, keep in mind that the new typescript ESM is no longer in 4.5.1, it's in @ next, see https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-rc/#beta-delta

package.json
PabloSzx: image

vs

image

IvanGoncharov: Yes, it's a JS file, not TS.
Converting those files to TS is not part of this PR.

integrationTests/webpack/webpack.config.json
IvanGoncharov: It's totally worth it, same as any other unrelated change.
Every change that makes sense on its own should be a separate PR.

package.json
PabloSzx: it doesn't have to be TS for it to have autocompletion, and this change was not making this file typescript, but whatever you want 👍

saihaj: > Converting those files to TS is not part of this PR.

don’t think anything is changed. adding @types shouldn’t be an issue it just helps improve DX in vscode at least.

.gitignore
```diff
@@ -14,3 +14,4 @@
 /denoDist
 /npm
 /deno
+/npmEsmDist
```
IvanGoncharov, line 17: Can you please move it to the same place as in `.eslintignore`?
All ignore files should have the same order.

resources/gen-version.js
```diff
@@ -4,21 +4,31 @@ const fs = require('fs');
 
 const { version } = require('../package.json');
 
-const versionMatch = /^(\d+)\.(\d+)\.(\d+)-?(.*)?$/.exec(version);
-if (!versionMatch) {
-  throw new Error('Version does not match semver spec: ' + version);
+if (require.main === module) {
```
IvanGoncharov, line 7: @PabloSzx Please revert, we shouldn't run `npm version` twice and create two sets of tags and duplicate other steps.
It just that `npm run build:npm` should now produce two folder standard and ESM.
And `16.0.1` is `16.0.1` even if it's published as ESM.
If devs for some reason want to know if it's ESM or not they should import `package.json`

IvanGoncharov (changes requested):

package.json
IvanGoncharov: @saihaj Yes, it makes sense, but as separate PR.
It has nothing to do with ESM specifically.

Every commit/PR should be topic-based and revertable.
Also, it doesn't help if `git blame` shows that this package was added for the ESM package.

DX improvements can be done but they should be done in separate PRs with explanations not as random changes on unrelated PRs

yaacovCR (approved):
Looks good to me!

looks like esm tag makes sense out of band from canaries and all that

yaacovCR:
> Looks good to me!
> 
> looks like esm tag makes sense out of band from canaries and all that

@IvanGoncharov  is this ready for re-review?

saihaj (approved):

IvanGoncharov:

IvanGoncharov:
I rebased this PR to the latest main as the solution for #3603

yaacovCR:
@IvanGoncharov @PabloSzx  et al

question about the current state => the build scripts add a build qualifier of "+esm" and a tag suffix of "-esm", but do not rename the package to `graphql-esm` => the integration tests that use `graphql-esm` just point to a local tarball with the distribution, presumably where the name of the package is still `graphql`.

Over at the `graphql-esm` on npm -- it doesn't seem like we have used the exact machinery within this repo to publish as on npm `graphql-esm` is up to 16.5 and the tag is just 'latest' not 'latest-esm'.

So, where are we with this? Some options:

(1) we can get rid of the build qualifiers and tag suffix, and have a package name suffix of "-esm" and every time we publish to `graphql`, we should also publish to `graphql-esm`....
(2) we can get rid of the cjs version and also get rid of the build qualifiers and tag suffix.

So either way, the end goal will be to get rid of the build qualifier and tag suffix, right?

But have we decided on what will happen?

PR #3362, opened 2021-11-05
Add option printDirective to printSchema

spawnia:
Resolves https://github.com/graphql/graphql-js/issues/2020

I initially thought about implementing this as https://github.com/graphql/graphql-js/issues/2020#issuecomment-866798901,
but really could not think of a good use case where
a directive should be printed in a modified way.

Thus, I went for a simpler signature:

```ts
shouldPrintDirective?: (directiveNode: DirectiveNode) => boolean
```

This is only a partial implementation for now, there are more possible
locations in a schema that can have a directive. Before I finish the
implementation, I would like to validate the direction is right.


IvanGoncharov:
@spawnia Thanks for opening draft PR.
Any solution that we merge should work with a code-first approach and `extensions`.
I want to push your previous proposal https://github.com/graphql/graphql-js/issues/2020#issuecomment-866798901 even further and add callback as part of `GraphQLDirective` that accepts `GraphQLField`, `GraphQLType`, etc., and returns an array of directive args.

Example:
```ts
export const GraphQLDeprecatedDirective: GraphQLDirective =
  new GraphQLDirective({
    name: 'deprecated',
    locations: [
      DirectiveLocation.FIELD_DEFINITION,
      // ...
    ],
    args: {
      reason: {
        type: GraphQLString,
        defaultValue: DEFAULT_DEPRECATION_REASON,
      },
    },
    notSureWhatNameToChoose(obj) {
      if (obj.deprecationReason != null) {
        return [
          obj.deprecationReason === DEFAULT_DEPRECATION_REASON
          ? {}
          : { reason: obj.deprecationReason },
        ];
      }
    }
  });
```

That way directives can leave in separate packages and be fully self-contained.

spawnia:
Hey @IvanGoncharov, thanks for the feedback on the PR.

It does surface a very interesting point, that is how we can bridge the gap for features such as `@deprecated`, `@specifiedBy` and possibly future features such as `@oneOf`.

### Representation

- SDL: ✔️ covered by the spec
- Code: ✔️ mechanism is up to the implementation
- Introspection : ✔️ covered by the spec

### Conversion

- SDL to Code: ✔️ supported in `buildSchema`
- Introspection to Code: ✔️ supported in `buildClientSchema`
- Code to Introspection: ✔️supported by introspection mechanism
- SDL to Introspection: ✔️ works through SDL -> Code -> Introspection
- Code to SDL: ⛔ not supported
- Introspection to SDL: ⛔ would work through Introspection -> Code -> SDL

With those built-in features that are also available in introspection, I do not see a use case why they should ever not be present in a printed schema. They are already exposed through introspection anyways, and thus not expected to contain internal-only information. Thus, I think we could simply always print them. If there are concerns regarding backwards compatiblity, we could add a config setting such as `printBuiltInDirectives: bool`.

> I want to push your previous proposal #2020 (comment) even further and add callback as part of GraphQLDirective that accepts GraphQLField, GraphQLType, etc., and returns an array of directive args.

Your example makes it quite clear why it would be useful for directives to influence how they are printed. Howerver, in a purely SDL driven implementation, not every used directive necessarily even has a code based definition associated with it. The main project I am building (https://github.com/nuwave/lighthouse) only defines directives through SDL, which has been sufficient for validation and all other purposes. We would require a default implementation of this method anyways - probably just returning all the arguments? I would argue that this functionality is related but not essential to the problem I am trying to solve here.

> Any solution that we merge should work with a code-first approach and extensions.

If it must exclusively work with `extensions`, that would force directive users into having to specify the following for each directive:
1. Method to convert directive into extensions
2. Method to convert extensions back into directive

We should offer a generalized implementation of this that smooths the path from SDL -> Code -> SDL for implementations that are primarily SDL-driven.

IvanGoncharov:
> With those built-in features that are also available in introspection, I do not see a use case why they should ever not be present in a printed schema. They are already exposed through introspection anyways, and thus not expected to contain internal-only information. Thus, I think we could simply always print them. If there are concerns regarding backwards compatiblity, we could add a config setting such as printBuiltInDirectives: bool.

@spawnia Not sure I understand your arguments, yes built-in stuff should be printable.
If you look into source code they are already printed with hand-written code for each directive.
I just suggest a mechanism that will work for both built-in and custom directives.

> The main project I am building (nuwave/lighthouse) only defines directives through SDL, which has been sufficient for validation and all other purposes. We would require a default implementation of this method anyways - probably just returning all the arguments? I would argue that this functionality is related but not essential to the problem I am trying to solve here.

`astNode` were added just to enhance error messages we can't use them for anything else.
I totally agree with Lee on this one: https://github.com/graphql/graphql-js/issues/869#issuecomment-309899380
The only way forward that I see is `@directive => change object (field, type, etc.) including extensions => @directive`.

spawnia:
> If you look into source code they are already printed with hand-written code for each directive.

You are right, I completely missed that.

Since `extensions` is completely custom, I don't think we can build abstractions on it to make directive printing any easier. I came back around to https://github.com/graphql/graphql-js/issues/2020 and now implemented a POC for an API that does not have any assumptions about SDL usage, extensions, presence or implementation of directive classes, etc.

src/utilities/printSchema.ts
```diff
@@ -7,47 +7,97 @@ import { print } from '../language/printer';
...
+export interface PrintSchemaOptions {
+  printDirectives?: (
+    definition:
+      | GraphQLSchema
+      | GraphQLType
+      | GraphQLField<unknown, unknown>
+      | GraphQLEnumValue
+      | GraphQLInputField
+      | GraphQLArgument,
+  ) => ReadonlyArray<ConstDirectiveNode>;
```
spawnia, line 55: Is there perhaps a better data structure to return? I don't want to go with `string`, as that puts the burden of proper formatting on the end users.

src/utilities/printSchema.ts
```diff
@@ -7,47 +7,97 @@ import { print } from '../language/printer';
...
+    isIntrospectionType,
+    options,
+  );
 }
 
+/**
+ * Useful implementation of PrintSchemaOptions.printDirectives for users who
+ * define their schema through SDL and simply want to print all included directives.
+ */
+export const directivesFromAstNodes: PrintSchemaOptions['printDirectives'] = (
```
spawnia, line 86: I am fine with omitting the addition of this function if you don't think it is useful. It would certainly solve the use case I have and allow me to do a simple one-liner: `printSchema(source, { printDirectives: directivesFromAstNodes })`.

src/utilities/printSchema.ts
```diff
@@ -7,47 +7,97 @@ import { print } from '../language/printer';
...
+import { isSpecifiedScalarType } from '../type/scalars';
+
+import type { ConstDirectiveNode } from '../language/ast';
+import { isConstDirectiveNode } from '../language/ast';
 
 import { astFromValue } from './astFromValue';
 
-export function printSchema(schema: GraphQLSchema): string {
+export interface PrintSchemaOptions {
+  printDirectives?: (
```
spawnia, line 47: Perhaps splitting this into separate monomorphic functions would make it easier for the end user and avoid them having to do type checks:

```ts
printDirectivesForSchema: (schema: GraphQLSchema): ReadonlyArray<ConstDirectiveNode>;
printDirectivesForType: (type: GraphQLType): ReadonlyArray<ConstDirectiveNode>;
...
```

src/utilities/printSchema.ts
IvanGoncharov: @spawnia It is 100% useful, but it doesn't fit the intended usage of SDL directives.
So it's ok if users implement it themself but as reference implementation, we can't encourage it.

spawnia:
> It should be tied to directive objects, otherwise it's not modular.

That implies we tie directive objects to definitions in a unified manner. That seems to contradict the philosophy of not letting directives bleed into the programmatic representation of a schema.

The implementation I propose is agnostic as to if or how directives are tied into the schema. It can leverage AST nodes if the schema came from an SDL source, it can use something the developer placed in `extensions`, or it can use any other mechanism.

> Otherwise schema created by one tool/library is not fully printable by another library.

I am not convinced that schema-agnostic and universally uniform printing is needed. It may even be necessary to do custom printing, for instance in Apollo federation: https://www.apollographql.com/docs/federation/federation-spec/#fetch-service-capabilities

IvanGoncharov:
> I am not convinced that schema-agnostic and universally uniform printing is needed. It may even be necessary to do custom printing, for instance in Apollo federation: apollographql.com/docs/federation/federation-spec/#fetch-service-capabilities

@spawnia It's exactly a problem we have right now.
Apollo has all these directives in their schema but you can't print them because:
1. You can't print directives at all and your solution solves that.
2. Code for printing is Apollo-specific, your solution doesn't fix that.

If we truly want to solve this problem, the schema should be self-printable.

> That implies we tie directive objects to definitions in a unified manner. That seems to contradict the philosophy of not letting directives bleed into the programmatic representation of a schema.

Directive definitions are already part of the schema even SDL directives (you can check introspection).
If you read spec:
> Directives can also be used to annotate the type system definition language as well, which can be a useful tool for supplying additional metadata in order to generate GraphQL execution services, produce client generated runtime code, or many other useful extensions of the GraphQL semantics.

So "supplying additional metadata" is directly mapped to `extensions` as it the only "metadata-like" thing you can change in `GraphQL*` classes. Also, it's perfectly fine to have directive objects in the schema (and we have them today). What's not ok (and Lee was referring to it) is to have "directive values" (not definitions) being associated with schema.
The directive should be used as "transformations" (the same way as decorators used in any language) and should not persist in the schema (but transformation can do anything including writing to "extensions").

Back to your proposal, let's use your framework as an example.
Disclaimer: Don't know much about `graphql-php` but I assume it work the same way as `graphql-js`.
So you defined your directives in SDL and process users SDL, now you get schema which you can pass to any other tool. Those 3rd-party tools see SDL directives in your schema (`schema.getDirectives()`) but see the same schema as your framework use but at the same time can't print schema in the same way as you do (because code for printing directives is not part of the schema).

spawnia:
I removed everything that was tied directly to the SDL source. The implementation is now fully agnostic to what will result in the printed directives in the final schema.

I think this implementation is useful on its own. `graphql-js` being the fundamental base library for all GraphQL implementations does not need to be any more opinionated than this. Imposing any further structure also imposes additional constraints on potential use cases, so it should be left up to users how they want to decide the directives that are printed.

yaacovCR:
Closing this for now. Note that graphql-tools has a https://github.com/ardatan/graphql-tools/blob/master/packages/utils/src/print-schema-with-directives.ts tool.

The reference implementation has decided not to support it, as technically it's unsafe, we can't safely recreate the original directives from a given schema, because the directives could be theoretically used in any which way by a schema builder, i.e. could modify the schema. My understanding is that the function in `graphql-tools` simply does best-effort, but that would lead to confusion in this repository in terms of spec-mandated behavior.

See related discussion in terms of introspection and a still potentially in-progress effort there:

https://github.com/graphql/graphql-js/pull/3047

Issue #3229, opened 2021-08-11
Disable suggestions


sandorvasas:
Recently we had a security test of our application and although I had disabled introspection on our GraphQL API, the `didYouMean` suggestions came up as a vulnerability. I kind of agree it's not a good idea to show around blackhats what methods our API exposes, and propose a `disableSuggestions: boolean` option.

sahanatroam:
related: https://github.com/apollographql/apollo-server/issues/3919

IvanGoncharov:
Closing in favor of #2247

PR #3363, opened 2021-11-06
fix: remove security vulnerability of exposing graph in production environment

sahanatroam:
Resolves and related to: 
https://github.com/graphql/graphql-js/issues/3229
https://github.com/graphql/graphql-js/issues/2247

Following Node.js best practices of setting NODE_ENV=production we could use the same pattern to disable this security vulnerability in graphql-js at the core.

**Open for feedback to resolve this issue ASAP**

### DIFF
--- a/src/jsutils/didYouMean.ts
+++ b/src/jsutils/didYouMean.ts
@@ -12,7 +12,7 @@ export function didYouMean(
   firstArg: string | ReadonlyArray<string>,
   secondArg?: ReadonlyArray<string>,
 ) {
-  if (process.ENV.NODE_ENV === 'production') {
+  if (process.env.NODE_ENV === 'production') {
     return ''
   };
 
--- a/src/jsutils/didYouMean.ts
+++ b/src/jsutils/didYouMean.ts
@@ -12,6 +12,10 @@ export function didYouMean(
   firstArg: string | ReadonlyArray<string>,
   secondArg?: ReadonlyArray<string>,
 ) {
+  if (process.ENV.NODE_ENV === 'production') {
+    return ''
+  };
+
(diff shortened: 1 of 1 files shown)


saihaj (changes requested):
This is a node specific fix. As suggested in that issue i think it would be better to have a Boolean configuration.

src/jsutils/didYouMean.ts
```diff
@@ -12,6 +12,10 @@ export function didYouMean(
   firstArg: string | ReadonlyArray<string>,
   secondArg?: ReadonlyArray<string>,
 ) {
+  if (process.env.NODE_ENV === 'production') {
```
sahanatroam, line 15: @saihaj Thanks for the review, I wonder if by default the library should enforce good practices so that we do not leave vulnerable code in production systems?

saihaj: Yes but this isn’t a platform agnostic fix. This library can be used in non-node environments. So it is something in the environment you are implementing that this can be handled.

sahanatroam: Understood. Keen to see how this progress and I could help with the fix: https://github.com/graphql/graphql-js/issues/2247#issuecomment-962397993

IvanGoncharov:
It shouldn't be a dev vs production configuration.
Production servers can support both introspection and `didYouMean` (e.g. public APIs, or anyone who does not follow the "security through obscurity" approach).

Having global flags (though `env` or something else) is also bad since it leads to very non-intuitive behavior (e.g. if calls some function in web worker it will behave differently).

I want to keep #2247 as the single discussion for discussing solutions so closing this one.
I will explore some ideas I have next work week and write my findings in #2247

PR #3367, opened 2021-11-09
Allow to print custom amount of lines before and after the provided location

kamilkisiela:


IvanGoncharov:
@kamilkisiela Can you please describe the purpose?
Do you need it to comment descriptions, directives, or something else?

kamilkisiela:
We use it internally to print a location of an error but the default `1` line before and after is not that helpful.

```
schema.graphql:2783:9
    2782 |           
    2783 |         }
         |         ^
    2784 |
```

```
schema.graphql:2783:9
    2780 |       
    2781 |         input BatchInput {
    2782 |           
    2783 |         }
         |         ^
    2784 |         
    2785 |       
    2786 |         type SingleInput {
```

I don't mind applying this change only to the lines "before" the actual line of an error.

```
schema.graphql:2783:9
    2781 |         input BatchInput {
    2782 |           
    2783 |         }
         |         ^
    2784 |         
```

I think even changing the default to `2` lines instead of `1` is also acceptable on my side (I can prepare another PR).

IvanGoncharov:
@kamilkisiela Thanks for the explanation 👍 
Makes sense.
Quick question why do you point to the end of the node and not to the beginning of the node?

IvanGoncharov:
@kamilkisiela I missed that it's an empty type so you probably get this error from the parser.
Do you use `printSourceLocation` directly or you are using `GraphQLError.toString`?

PR #3370, opened 2021-11-18
feat: allow passing a contextValue to validation rules

n1ru4l:
It should be possible to pass a `contextValue` into validation rules that can be used within for complex logic such as query complexity calculation or other more complex validation rules that require sharing/modifying a `contextValue` provided from the outside.

### DIFF
--- a/src/validation/ValidationContext.ts
+++ b/src/validation/ValidationContext.ts
@@ -267,4 +267,7 @@ export class ValidationContext extends ASTValidationContext {
   }
 }
 
-export type ValidationRule = (context: ValidationContext) => ASTVisitor;
+export type ValidationRule = (
+  context: ValidationContext,
+  contextValue?: unknown,
+) => ASTVisitor;
--- a/src/validation/validate.ts
+++ b/src/validation/validate.ts
@@ -43,6 +43,7 @@ export function validate(
 
   /** @deprecated will be removed in 17.0.0 */
   typeInfo: TypeInfo = new TypeInfo(schema),
+  contextValue?: unknown,
 ): ReadonlyArray<GraphQLError> {
   const maxErrors = options?.maxErrors ?? 100;
 
@@ -72,7 +73,9 @@ export function validate(
 
   // This uses a specialized visitor which runs multiple visitors in parallel,
   // while maintaining the visitor skip and break API.
-  const visitor = visitInParallel(rules.map((rule) => rule(context)));
+  const visitor = visitInParallel(
+    rules.map((rule) => rule(context, contextValue)),
+  );
 
   // Visit the whole document with each instance of all provided rules.
   try {


dotansimha (approved):

dotansimha:
Agree with that. Seems like the spec doesn't refer to what should be passed to validation as input so that's probably an implementation detail?
We saw a need to access the `contextValue` during validation rules also in `envelop`, where some plugins need to implement custom validations that need to access `context` data (for example: for accessing the incoming request, the authenticated user, or any other dynamic data that might come from the context).

Also, the spec doesn't refer to when the context needs to be built and how, so I think keeping that as optional is a good way to go.

saihaj (approved):
Agreed this can be useful and I don’t see why not allow this. We can even backport this to `v15.x`. 🤔 not sure if there is a test we can add for this

benjie:
One particularly relevant section of the GraphQL spec states:

> As GraphQL type system schema evolves over time by adding new types and new fields, it is possible that a request which was previously valid could later become invalid. Any change that can cause a previously valid request to become invalid is considered a breaking change. GraphQL services and schema maintainers are encouraged to avoid breaking changes, however in order to be more resilient to these breaking changes, sophisticated GraphQL systems may still allow for the execution of requests which at some point were known to be free of any validation errors, and have not changed since.
> -- https://spec.graphql.org/draft/#sec-Validation.Type-system-evolution

It's clear to me that the GraphQL spec intends to you to be able to validate a GraphQL _document_ once and execute it via a _request_ multiple times. It's important to note that nothing in section 5 of the GraphQL spec refers to request-specific items such as `operationName`, `variableValues` or `initialValue` used in [`ExecuteRequest`](https://spec.graphql.org/draft/#ExecuteRequest()) - section 5 purely relates to the `document` and the `schema`.

A big advantage of this approach is that the result of parsing and validation can be cached - parse/validate it once and then execute requests using it many times without having to do that work again. In fact, if you're using persisted operations only then you could perform the parse and validate when the operation is persisted (before the server even starts up, maybe even months ago), and at runtime you can jump straight to `execute` without any need to re-parse or re-validate (depending on the storage format you use for persisted operations).

---

All of the above said, performing request-specific validation is useful - the values stored in `variableValues` and `contextValue` and (to a lesser extend) `initialValue` may need to be factored in to some validation you do. Examples include query complexity and pagination limit validations that need to know what the user has asked for _precisely_ before they can do their work.

Therefore, I think a better way to serve this purpose would be to leave `validate` as I believe section 5 intends, and potentially introduce a new "second phase" of validation that can occur alongside the request, such as `validateRequest`. This new method could accept `contextValue`, `variableValues` and (maybe) `initialValue` and run _separate_ validation rules which have more context.

So in this case you'd perform `parse`, then `validate`, then cache the result. When you receive a request, you could then take this pre-parsed, pre-validated document and then execute `validateRequest` against it, followed by `execute` or `subscribe` as appropriate. I'd imagine that `validateRequest` is a no-op of the vast majority of GraphQL schemas out there currently.

I'm definitely opposed to the signature for `ValidationRule` changing like this because it would prevent the validation caching that's so critical to some speedups available in the GraphQL.js ecosystem - if you introduced a `validateRequest` then I think you'd also add a new `RequestValidationRule` type that allows for these extra parameters.

IvanGoncharov:
@n1ru4l First I agree with Benjie that all standard rules should be fully static (has access only to schema and query).
Second I don't see why you can't use existing `validate` and pass context yourself:
```js
function ruleThatNeedContext(contextValue) {
  return (context) => {
    return {
      Field(node) {
        // you have access both to contextValue and context
      }
    };
  }
}
```
It simple high-order function, you want to generalize that you can always add the wrapper curry function. 
If you need help please ping me and I will help you to write it.

In general, JS/TS has good support for high-order functions so we should use it.

IvanGoncharov:
@n1ru4l Do you agree with solution I propose ⬆️ ?
If so can I close this PR?

n1ru4l:
I would prefer to keep this open and use it as potential discussion material for introducing an optional extended validation phase within the GraphQL specification.

IvanGoncharov:
> I would prefer to keep this open and use it as potential discussion material for introducing an optional extended validation phase within the GraphQL specification.

@n1ru4l If it's related to GraphQL specificaton it makes more sense to open issue against spec repo and link this one for context.
Closing this PR since it contains code that contradicts spec and can't be merged.

PR #3384, opened 2021-11-26
Preserve non-error values thrown from resolvers

IvanGoncharov:
Fixes #3379


IvanGoncharov:
Need to add some basic tests.

yaacovCR:
Similarly, as far as I can tell, in react, componentDidCatch gets a parameter of type mixed…. So….

https://github.com/facebook/react/blob/0ddd69d1226c660ee4930b0dce7860b706e548bc/packages/react-reconciler/src/ReactFiberThrow.new.js#L143

PR #3385, opened 2021-11-27
fix type for first argument to isTypeOf

ryym:
Hi,

Currently the type of `source` argument in `isTypeOf` is `TSource` (the source type of `GraphQLObjectType`), but I think it is wrong since `isTypeOf` may be called with a source value of any type that implements the same interface or that belongs to the same union. Therefore I changed the type of `source` to `unknown`.

For example, the following code passes type check but results in an error at runtime.

```typescript
import {
  graphql,
  GraphQLSchema,
  GraphQLObjectType,
  GraphQLString,
  GraphQLInterfaceType,
} from "graphql";

class Human {
  constructor(readonly name: string) {}
}

class Droid {
  constructor(readonly name: string, readonly primaryFunction: string) {}
  serialNumber = () => "-";
}

const CharacterType = new GraphQLInterfaceType({
  name: "Character",
  fields: { name: { type: GraphQLString } },
});

const HumanType = new GraphQLObjectType<Human>({
  name: "Human",
  fields: {
    name: { type: GraphQLString },
  },
  interfaces: [CharacterType],
  isTypeOf: (source) => {
    return source instanceof Human;
  },
});

const DroidType = new GraphQLObjectType<Droid>({
  name: "Droid",
  interfaces: [CharacterType],
  fields: {
    name: { type: GraphQLString },
    primaryFunction: { type: GraphQLString },
  },
  isTypeOf: (source) => {
    // XXX: Here the type of source is inferred as Droid, but it can be Human.
    // In that case this results in error at runtime.
    console.log("debug", source.serialNumber());
    return source instanceof Droid;
  },
});

const QueryType = new GraphQLObjectType({
  name: "Query",
  fields: {
    anyCharacter: {
      type: CharacterType,
      resolve: () => new Human("Han Solo"),
    },
  },
});

const schema = new GraphQLSchema({ query: QueryType, types: [DroidType, HumanType] });
const query = "query Test { anyCharacter { name } }";
graphql({ schema, source: query, rootValue: {} }).then((res) => {
  console.log(res);
  //=> res.errors: [TypeError: source.serialNumber is not a function]
});
```

### DIFF
--- a/src/execution/__tests__/abstract-test.ts
+++ b/src/execution/__tests__/abstract-test.ts
@@ -158,7 +158,7 @@ describe('Execute: Handles execution of abstract types', () => {
     const DogType = new GraphQLObjectType<Dog, Context>({
       name: 'Dog',
       interfaces: [PetType],
-      isTypeOf(_source, context) {
+      isTypeOf(_value, context) {
         const error = new Error('We are testing this error');
         if (context.async) {
           return Promise.reject(error);
@@ -240,7 +240,7 @@ describe('Execute: Handles execution of abstract types', () => {
     const DogType = new GraphQLObjectType<Dog, Context>({
       name: 'Dog',
       interfaces: [PetType],
-      isTypeOf(_source, context) {
+      isTypeOf(_value, context) {
         return context.async ? Promise.resolve(false) : false;
       },
       fields: {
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
Changing type to unknown is correct and improves DX however it's a breaking change so I can't merge it in v16.
The good news, we will switch `main` to "v17 alpha" this Wednesday so I will merge it after the switch.

yaacovCR:
Flagging this for v17.

yaacovCR:
I added another generic parameter TAbstract defaulted to unknown on top of this.

PR #3393, opened 2021-11-29
Update doc examples to reflect the current API

paulserraino:
hello, I was trying out graphql-js today and noticed that the Readme documentation doesn't match the current object-based interface


saihaj (approved):
Thanks!

PR #3397, opened 2021-12-01
TS: Use both input and output types in scalars parse/serialize

honzajerabek:
Hi, currently the `parseValue` and `serialize` functions in `GraphQLScalarTypeConfig` are receiving `unknown` type in the value arg, but the type is known - it's the inverted type of the function to each other.

```
...
serialize: (val: TInternal) => TExternal
parseValue: (val: TExternal) => TInternal
...
```

With current `unknown` typing there are TS errors when trying to have it strongly typed.

This minimal custom scalar

```
export const Date = new GraphQLScalarType({
  name: 'Date',
  serialize: (value: Date): number => value.valueOf(),
  parseValue: (value: number): Date => new Date(value),
});
```
causes these errors:

**serialize:**
```
TS2322: Type '(value: Date) => number' is not assignable to type 'GraphQLScalarSerializer<number>'.   Types of parameters 'value' and 'outputValue' are incompatible.     Type 'unknown' is not assignable to type 'Date'
```

**parseValue**
```
TS2322: Type '(value: number) => Date' is not assignable to type 'GraphQLScalarValueParser<Date>'.   Types of parameters 'value' and 'inputValue' are incompatible.     Type 'unknown' is not assignable to type 'number'. 
```

### DIFF
--- a/src/type/definition.ts
+++ b/src/type/definition.ts
@@ -584,8 +584,8 @@ export class GraphQLScalarType<TInternal = unknown, TExternal = TInternal> {
   name: string;
   description: Maybe<string>;
   specifiedByURL: Maybe<string>;
-  serialize: GraphQLScalarSerializer<TExternal>;
-  parseValue: GraphQLScalarValueParser<TInternal>;
+  serialize: GraphQLScalarSerializer<TExternal, TInternal>;
+  parseValue: GraphQLScalarValueParser<TInternal, TExternal>;
   parseLiteral: GraphQLScalarLiteralParser<TInternal>;
   extensions: Readonly<GraphQLScalarTypeExtensions>;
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@honzajerabek Thanks for reporting.
It's `unknown` on purpose since you need to do validation inside both functions.
`parseValue` should validate query variable and `serialize` should validate return of resolvers.
`graphql-js` do all necessary validation for all types except, standard scalars so for standard scalars you need to do all validation yourself that's why we need to have `unknown` there.

honzajerabek:
Okay thanks for explanation

yaacovCR:
See https://github.com/graphql/graphql-js/issues/3451#issuecomment-1008114612 for some follow-on thoughts, perhaps the serializer could be made type-safe....

PR #3406, opened 2021-12-04
Simplify code by replacing Object.entries with Object.keys

IvanGoncharov:


src/jsutils/mapValue.ts
```diff
@@ -10,8 +10,8 @@ export function mapValue<T, V>(
 ): ObjMap<V> {
   const result = Object.create(null);
 
-  for (const [key, value] of Object.entries(map)) {
-    result[key] = fn(value, key);
```
saihaj, line 14: I liked this more :)

PR #3281, opened 2021-09-29
RFC: Client Controlled Nullability Operators Implementation

twof:
Implementation of https://github.com/graphql/graphql-spec/issues/867 along with the suggested `?` syntax. 

[Here's a doc with a list of some of the important test cases and the justifications for their behavior.](https://docs.google.com/document/d/1relgXUdhJrlaSR70lrITQKmcIHXjIVyWxDCKRuzMEvs/edit#heading=h.gm7m1frqgtc4)

This PR includes work done by @xuewei8910, @aprilrd, @magicmark, and @lizjakubowski.

PR #3418, opened 2021-12-08
[RFC] Client Controlled Nullability experiment implementation w/o execution

twof:
This branch is a subset of https://github.com/graphql/graphql-js/pull/3281 with a new flag introduced to mark the new operators experimental so they can be merged into master. 

To set up the experiment gates, inspiration was drawn from https://github.com/graphql/graphql-js/pull/1141

For additional context see https://github.com/graphql/graphql-js/pull/3281#issuecomment-987855321

State of the experiment:

We're merging in two parts of the original RFC and holding off on some of the more controversial bits that are still heavily in flux.

This PR only introduces parsing, lexing, printing, and visiting for the `?` and `!` operators and list syntax. It does not include validation or execution.


src/language/__tests__/parser-test.ts
```diff
@@ -224,6 +228,196 @@ describe('Parser', () => {
     ).to.not.throw();
   });
 
+  it('parses required field', () => {
+    const result = parse(
```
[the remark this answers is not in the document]
IvanGoncharov: Clarification: please move the text into `document` variable and you can use dedent if you think it will import readability.

src/language/__tests__/parser-test.ts
```diff
@@ -224,6 +228,196 @@ describe('Parser', () => {
...
+      dedent`
+      query {
```
IvanGoncharov, line 234: Since you are testing fields specifically, please use the shorter form without operation prefix:
```
{ 
   requiredField!
}
```

src/language/ast.ts
```diff
@@ -360,8 +375,26 @@ export interface FieldNode {
...
+export interface OptionalModifierNode {
+  readonly kind: Kind.OPTIONAL_DESIGNATOR;
+  readonly loc?: Location;
 }
 
+// modifiers can be ! or ?
+export type NullabilityModifierNode =
+  | RequiredModifierNode
+  | OptionalModifierNode;
+
```
IvanGoncharov, line 397: @twof Did you drop support for changing nullability of array items?

IvanGoncharov (changes requested):

src/language/ast.ts
twof: Not officially. I'd interpreted [your comment](https://github.com/graphql/graphql-js/pull/3281#issuecomment-987855321) as you asking for only `!` and `?` to be in this PR. Should I pull the array syntax over too?

twof: I'd also be open to merging in this first, so array syntax can be reviewed in a separate PR. The diff created by introducing the array syntax will be quite a bit larger than this one, and I don't want to make you review too much code at once if that would be helpful.

twof: @IvanGoncharov I've prepared a list syntax PR, so I can merge that into this branch if you tell me that's what you want: https://github.com/twof/graphql-js/pull/2

It actually wasn't as big as I thought it would be.

IvanGoncharov: @twof Technically it may be easier to review them separately, but practically it makes hard to understand some of the reasoning in the code. 
For example, without list syntax we don't really need 2 new node kinds in AST we can just have enum with 3 values as in your original proposal.

> Not officially. I'd interpreted your comment as you asking for only ! and ? to be in this PR. Should I pull the array syntax over too?

Sorry, about that. I would try to communicate my suggestion more cleanly.

twof: @IvanGoncharov All good! I've merged in the list syntax.

twof:
Updates: I've merged in the list syntax, rebased on `main`, and I've got all tests passing. @IvanGoncharov let me know if there's anything I can do to help with the experiment flag implementation.

twof:
New behavior was selected for null propagation at the last WG meeting, there are going to be some pretty major changes incoming. We're sticking with the current list syntax.

src/language/ast.ts
```diff
@@ -358,10 +375,36 @@ export interface FieldNode {
...
   readonly name: NameNode;
   readonly arguments?: ReadonlyArray<ArgumentNode>;
+  // Note: Client Controlled Nullability is experimental
+  // and may be changed or removed in the future.
+  readonly required?: ListNullabilityNode | NullabilityDesignatorNode;
   readonly directives?: ReadonlyArray<DirectiveNode>;
   readonly selectionSet?: SelectionSetNode;
 }
 
+export interface RequiredDesignatorNode {
```
twof, line 385: We might want to change the name of these nodes to match the new null propagation behavior. I'll have to come back to this.

twof:
@saihaj are you or the other folks on the review team going to have time to take a look at it this week? If so, I can rebase on main and make sure it's up to date with the other branch with execution: https://github.com/graphql/graphql-js/pull/3510

twof:
@IvanGoncharov Alright I've moved over the code that Apollo-iOS needs, as we discussed. Thanks again!

src/language/__tests__/parser-test.ts
```diff
@@ -224,6 +228,463 @@ describe('Parser', () => {
...
+                  kind: Kind.NAME,
+                  loc: { start: 2, end: 7 },
+                  value: 'field',
+                },
+                arguments: [],
+                directives: [],
+                selectionSet: undefined,
+                required: {
+                  kind: Kind.REQUIRED_DESIGNATOR,
+                  loc: { start: 9, end: 10 },
```
IvanGoncharov, line 506: Invalid, should be
```suggestion
                  loc: { start: 7, end: 10 },
```
An outer node should have a wider range and include all its inner nodes.

IvanGoncharov: I working on a review commit that will fix that.

PR #3423, opened 2021-12-11
assertName-test: test new instead of deprecated function

Cito:


saihaj (approved):

src/type/__tests__/assertName-test.ts
```diff
@@ -2,10 +2,9 @@ import { expect } from 'chai';
 import { describe, it } from 'mocha';
 
 import { assertName, assertEnumValueName } from '../assertName';
-import { assertValidName } from '../../utilities/assertValidName';
 
-describe('assertValidName()', () => {
-  assertValidName('test');
+describe('assertName()', () => {
+  assertName('test');
```
IvanGoncharov, line 7:
```suggestion
```

IvanGoncharov: This one was a mistake, so just removing

PR #3435, opened 2021-12-20
gh/actions: run benchmark &amp; diff-npm-package only on PRs

IvanGoncharov:


saihaj:
I still don’t understand the point with this CI job and imho not much time should be spent on this.

PR #3439, opened 2021-12-21
Make GraphQLFormattedError.extensions have type GraphQLErrorExtensions

simon-abbott:
The `GraphQLFormattedError.extensions` field comes from the `GraphQLError.extensions`, which is of type `GraphQLErrorExtensions`. Having the mismatch means that trying to access the extensions produces weird type errors, even if 3rd party libraries correctly extend `GraphQLErrorExtensions`.

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -231,7 +231,7 @@ export interface GraphQLFormattedError {
    * Reserved for implementors to extend the protocol however they see fit,
    * and hence there are no additional restrictions on its contents.
    */
-  readonly extensions?: { [key: string]: unknown };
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@simon-abbott Thanks for taking the time to report this 👍 

Can you please explain your use case in more detail?

Also, let me explain the original intention for `GraphQLFormattedError`.
It's intended to represent errors transmitted to the client in serialized GraphQL response.
Also, it's intended to be used in clients that should work with any spec compliment server and can't make assumptions on the content of `extensions`.

Some of the "extensions" are designed to be used only on the server-side (e.g. stuff specific to logging or tracing) and never send to the client.

Issue #2185, opened 2019-09-19
Proposal: Faster Algorithm For OverlappingFieldsCanBeMerged


SimonAdameit:
At [XING](https://www.xing.com) we're using the [Sangria](https://sangria-graphql.org) implementation of GraphQL. The implementation the OverlappingFieldsCanBeMerged validation of Sangria follows the one in this reference implementation. We experienced performance problems with it related to the number of fragments. After investing some time to solve it, we came up with a better algorithm.

- [Here](https://tech.xing.com/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01) is the description of the algorithm.
- And [here](https://github.com/sangria-graphql/sangria/pull/458) is our PR for Sangria.

The proposal would be to use this algorithm in the reference implementation. Below is an example benchmark graph showing the performance improvement for multiple fragments in the same selection set.

![2019-09-16 NoOverlapFrag](https://user-images.githubusercontent.com/93972/65224756-19095d00-dac4-11e9-99ad-03f64c6d4459.png)

IvanGoncharov:
@SimonAdameit Thanks a lot for sharing 👍 Its super useful ATM we preparing for TS conversion so we in a feature freeze state but it would be at the top queue.
Also, I think we need to use this opportunity and address edge cases raised in #1497 at the same time.

yaacovCR:
> but it would be at the top queue after conversion is finished.
> Also, I think we need to use this opportunity and address edge cases raised in #1497 at the same time.

with TS conversion finished, this should be in the news!

yaacovCR:
updated link to article: https://tech.new-work.se/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01

PR #3442, opened 2021-12-24
 Prevent Infinite Loop in OverlappingFieldsCanBeMergedRule

nicolaslt:
We have spotted some `RangeError: Maximum call stack size exceeded` exceptions while testing our own setup, and traced it to this validation rule.

Note: I don't feel 100% confident that I understand the logic flow and intent of this code in detail. Perhaps the bugfix I'm providing is isolated, or perhaps it's a whole class of use case that has been omitted in tests in which cases this might break some valid use cases but not be detected?

### DIFF
--- a/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
+++ b/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
@@ -1023,7 +1023,7 @@ describe('Validate: Overlapping fields can be merged', () => {
 
   it('does not infinite loop on immediately recursive fragment mentioned in queries', () => {
     expectValid(`
-      query myQuery {
+      {
         ...fragA
       }
 
@@ -1031,8 +1031,54 @@ describe('Validate: Overlapping fields can be merged', () => {
     `);
   });
 
+  it('does not infinite loop on recursive fragment with a field named after fragment', () => {
+    expectValid(`
+      {
+        ...fragA
+        fragA
+      }
+
+      fragment fragA on Query { ...fragA }
+    `);
+  });
+
+  it('finds invalid cases even with field named after fragment', () => {
+    expectErrors(`
+      {
+        fragA
+        ...fragA
+      }
+
+      fragment fragA on Type { 
+        fragA: b
+      }
+    `).toDeepEqual([
+      {
+        message:
+          'Fields "fragA" conflict because "fragA" and "b" are different fields. Use different aliases on the fields to fetch both if this was intentional.',
+        locations: [
+          { line: 3, column: 9 },
+          { line: 8, column: 9 },
+        ],
+      },
+    ]);
+  });
+
   it('does not infinite loop on transitively recursive fragment', () => {
     expectValid(`
+      fragment fragA on Human { name, ...fragB }
+      fragment fragB on Human { name, ...fragC }
+      fragment fragC on Human { name, ...fragA }
+    `);
+  });
+
+  it('does not infinite loop on transitively recursive fragment mentioned in queries', () => {
+    expectValid(`
+      {
+        ...fragA
+        fragB
+      }
+
       fragment fragA on Human { name, ...fragB }
       fragment fragB on Human { name, ...fragC }
       fragment fragC on Human { name, ...fragA }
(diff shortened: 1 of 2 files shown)


Cito:
Great find - important to fix this!

nicolaslt:
@Cito I've just realised the error still happens when the cyclical reference is transitive. I'm going to try to fix this as well within the next 3h, but I will have limited availability after that until the 4th of January.

Cito:
@nicolaslt Yes, the algorithm is a bit complex. Will try to examine it more thoroughly and compare it with the spec where it looks simpler.

nicolaslt:
I've worked out that we might be able to simple re-use `comparedFragmentPairs`, if we extend its scope to keep track of `fields` as well as `fragment` (in which case I'd rename it, eh).

The only reason I can think of not to do that is if it's valid to have a field named after a fragment. I'm on my way to test this.

nicolaslt:
Yep, it's valid GraphQL it seems! (though tested with version 15.2.0, not with latest)

SDL:
```gql
type Query {                      
  getObject: Object               
}                                 

type Object {        
  field: Boolean     
  otherField: Boolean
}
```

Query:
```gql
query myQuery {
  getObject {
    ...field
    field
  }
}
                          
fragment field on Object {
   otherField             
}
```

Results (with a dummy resolver):
```json
{
  "data": {
    "getObject": {
      "otherField": null,
      "field": null
    }
  }
}
```


So the proper fix I think would be to use `comparedFragmentPairs` and make `getReferencedFieldsAndFragmentNames` return type something that lets like 270 realise when the fieldOrFragmentName is actually a fragment or a field. If it's a fragment, read&write to `comparedFragmentPairs`, if it's a field carry on as usual.

nicolaslt:
I'm starting to understand the codebase a bit better. I think just using `comparedFragmentPairs` is actually ok.

I'm about to push a new version with extra tests added for good measure.

nicolaslt:
@Cito I think this is ready again

Cito:
@nicolaslt Thanks - will look into this after the festivities.

Cito:
@nicolaslt Had another look and I think breaking the recursion in `collect_conflicts_between_fields_and_fragment` like you do here is the proper solution. Nice!

Still, I wonder whether the validation algorithm could be simplified and made less complex, in terms of code and runtime complexity, which is quadratic in both number of fields in the selection sets, and in the levels of nested fragments. This might be exploited for denial of service attacks, by sending requests with many overlapping fields and/or deeply nested fragments - if you do both, you can construct queries that are not too large, but still would take an extremely long time to validate. However, this is essentially the algorithm from the spec, just a bit more complicated because it also provides a list of error messages for the individual conflicts. I think it should be possible to store a codified "response shape" for every field in a selection set with the same response name and check if that is unique in linear time using a Map, instead of checking whether the response shapes are the same for all pairs of such fields.

Cito:
Oops - just noticed that regarding the runtime complexity of the algorithm, we already have #2185.

nicolaslt:
Yeah, I didn't spot that other PR but it did feel like the implementation could be improved.

What's next for both PRs? Should we port all missing tests (if any) to the other PR, or merge mine in the interim?

Cito:
I think we should merge this PR to fix the immediate problem and to have the additional tests until we change the implementation as suggested in #2185.

nicolaslt:
That sounds good to me. I'm guessing there's isn't anything for me to do,
but ping if needed!

n1ru4l (approved):

dotansimha (approved):

nicolaslt:
Hi all, is there any ETA for when this could be merged/released?

I'm quite eager to patch my servers.

dotansimha:
@IvanGoncharov any thoughts before merging this one?

ardatan (approved):

saihaj (approved):

IvanGoncharov:
@nicolaslt Thanks for contributing this fix 👍
I did some small tweaks in the tests before merging it.

@Cito Big thanks for the excellent review!

nicolaslt:
Nice, thanks @IvanGoncharov.

Would you have an estimate on when this would be released?

PR #3449, opened 2022-01-07
make schema creation lazier

yaacovCR:
if getPossibleTypes and/or getImplementations are never called on a schema, the schema does not have calculate the implementation map at startup

### DIFF
--- a/src/type/schema.ts
+++ b/src/type/schema.ts
@@ -142,10 +142,12 @@ export class GraphQLSchema {
   private _directives: ReadonlyArray<GraphQLDirective>;
   private _typeMap: TypeMap;
   private _subTypeMap: ObjMap<ObjMap<boolean>>;
-  private _implementationsMap: ObjMap<{
-    objects: Array<GraphQLObjectType>;
-    interfaces: Array<GraphQLInterfaceType>;
-  }>;
+  private _implementationsMap: Maybe<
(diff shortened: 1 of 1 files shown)


yaacovCR:
[graphql-executor](https://github.com/yaacovCR/graphql-executor) never calls these subfunctions, as there are equivalent (cross realm compatible) versions

IvanGoncharov:
@yaacovCR What's your motivation for this change?
Lazy evaluations make code more complicated and in general, we try to keep stuff immutable.
Performance-wise I don't see any difference even with the huge GitHub's schema we use in benchmark:https://github.com/graphql/graphql-js/runs/4736844456?check_suite_focus=true

yaacovCR:
The motivation is as above. I want eventually a way to quickly serialize schemas and their metadata. I basically have one within executor schema construct in graphql executor, but it would help if schema metadata was only generated if needed.

yaacovCR:
In terms of code complexity, I believe the PR demonstrates that this is not a complex change. Is there a particular part that bothers you?

yaacovCR:
Hmmmm will have to look closer as to whether benchmark GitHub schema uses a lot of interfaces, maybe not a big deal

yaacovCR:
Closing for now, just a reminder to me to benchmark before proposing any optimization

PR #3450, opened 2022-01-07
Add enableDeferStream to GraphQLSchema::toConfig()

benjie:
schema.toConfig() doesn't carry through the `enableDeferStream` setting currently; this fixes that omission.

Note: this is a PR to the defer-stream branch.


robrichard:
@benjie there's currently some debate on the best way to enable/disable defer & stream. If you have any thoughts there is an ongoing discussion here: https://github.com/robrichard/defer-stream-wg/discussions/12

PR #3453, opened 2022-01-10
fix(validate): no reusing root types

thomasheyenbrock:
Hello `graphql-js` folks 👋 While playing with schema and SDL validation, I noticed that the reference implementation does not follow the following rule from the [latest spec](https://spec.graphql.org/draft/#sec-Root-Operation-Types):

> The query, mutation, and subscription root types must all be different types if provided.

Not sure if that's an error in the spec or in the reference implementation 🤔 I went ahead and proposed a fix in here, but happy to open a PR on the spec instead if that's the preferred. This is after all a breaking change, even though I doubt that it would have a big impact. (At least I've never seen a schema that does this.)

### DIFF
--- a/src/type/__tests__/validation-test.ts
+++ b/src/type/__tests__/validation-test.ts
@@ -435,7 +435,19 @@ describe('Type System: A Schema must have Object root types', () => {
 
 describe('Type System: Root types must all be different if provided', () => {
   it('rejects a Schema where the same type is used for the "query" and "mutation" root types', () => {
-    const schema = buildSchema(`
+    const schema = new GraphQLSchema({
+      query: SomeObjectType,
+      mutation: SomeObjectType,
+    });
+    expectJSON(validateSchema(schema)).toDeepEqual([
+      {
+        message:
+          'All root types must be different, SomeObject is already used for "query" and cannot also be used for "mutation".',
+        locations: [{ line: 6, column: 3 }],
+      },
+    ]);
+
+    const schemaFromSDL = buildSchema(`
       type SomeObject {
         f: SomeObject
       }
@@ -445,7 +457,7 @@ describe('Type System: Root types must all be different if provided', () => {
         mutation: SomeObject
       }
     `);
-    expectJSON(validateSchema(schema)).toDeepEqual([
+    expectJSON(validateSchema(schemaFromSDL)).toDeepEqual([
       {
         message:
           'All root types must be different, SomeObject is already used for "query" and cannot also be used for "mutation".',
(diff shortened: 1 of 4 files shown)


src/type/__tests__/validation-test.ts
```diff
@@ -341,6 +341,12 @@ describe('Type System: A Schema must have Object root types', () => {
       input SomeInputObject {
         test: String
       }
+      input SomeInputObject2 {
+        test: String
+      }
+      input SomeInputObject3 {
+        test: String
+      }
```
thomasheyenbrock, line 349: For context: I changed this test to avoid getting an error here for reusing root types as well, to keep the scope of the test as tight as possible.

src/type/validate.ts
```diff
@@ -144,6 +144,44 @@ function validateRootTypes(context: SchemaValidationContext): void {
...
+        ' is already used for "query" and cannot also be used for "mutation".',
+      getOperationTypeNode(schema, OperationTypeNode.MUTATION) ??
+        mutationType.astNode,
+    );
+  }
+
+  if (
+    queryType &&
+    subscriptionType &&
+    queryType.name === subscriptionType.name
```
n1ru4l, line 161: While I also prefer the name equal check, it seems like a `queryType  === subscriptionType` might also be sufficient?

thomasheyenbrock: I wasn't sure about if `graphql-js` always keeps the same reference to the type objects, and I thought checking the name of the type is safer as it also prevents issues when doing stuff like this for whatever reason:
```js
const query = new GraphQLObjectType(...);
const anotherQuery = new GraphQLObjectType(query.toConfig());
expect(query === anotherQuery).toBeFalse();
```

saihaj (approved):

src/type/validate.ts
yaacovCR: There is a separate check for different objects of the same type name, so we can/should have this check be of object identity imo

src/type/__tests__/validation-test.ts
```diff
@@ -365,7 +371,7 @@ describe('Type System: A Schema must have Object root types', () => {
       schema,
       parse(`
         extend schema {
-          subscription: SomeInputObject
+          subscription: SomeInputObject3
```
yaacovCR, line 374: Maybe we could change the other types to be something other than Input types while we were at it, not critical

src/type/validate.ts
thomasheyenbrock: Good point @yaacovCR, you can't construct a `GraphQLSchema` object when two types have the same name. I changed it to check for referential equality here.

src/type/__tests__/validation-test.ts
thomasheyenbrock: Done ✅

yaacovCR (approved):

IvanGoncharov:
@thomasheyenbrock Thanks for PR 👍 
I applied review changes in 095517abee2751c04ba04d2baee9e2cc5504821a

PR #3454, opened 2022-01-11
feat: allow providing an object to the GraphQLError constructor

n1ru4l:
When using the `GraphQL` error within my schema code I usually only use the `message` and `extensions` parameters. However, this requires passing 4 undefined values to the constructor.

```ts
throw new GraphQLError("Oops.", undefined, undefined, undefined, undefined, { code: "SERVER_GOES_BRRT" })
```

This API allows writing this code, which feels less verbose to me.

```ts
throw new GraphQLError({
  message: "Oops.",
  extensions: { code: "SERVER_GOES_BRRT" },
})
```

Given that the `execute` and `subscribe` function signatures also evolved into a single object, this change feels natural to me. Please let me know what you think.

An alternative API might be the following, where the message is ALWAYS the first argument.

```ts
throw new GraphQLError("Oops.", {
  extensions: { code: "SERVER_GOES_BRRT" },
})
```

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -20,7 +20,7 @@ export interface GraphQLErrorExtensions {
   [attributeName: string]: unknown;
 }
 
-interface GraphQLErrorArgs {
+export interface GraphQLErrorArgs {
   nodes?: ReadonlyArray<ASTNode> | ASTNode | null;
   source?: Maybe<Source>;
   positions?: Maybe<ReadonlyArray<number>>;
@@ -111,14 +111,17 @@ export class GraphQLError extends Error {
    */
   readonly extensions: GraphQLErrorExtensions;
 
+  /**
+   * @deprecated Please use the `GraphQLErrorArgs` constructor overload instead.
+   */
   constructor(
     message: string,
-    nodes?: GraphQLErrorArgs['nodes'],
-    source?: GraphQLErrorArgs['source'],
-    positions?: GraphQLErrorArgs['positions'],
-    path?: GraphQLErrorArgs['path'],
-    originalError?: GraphQLErrorArgs['originalError'],
-    extensions?: GraphQLErrorArgs['extensions'],
+    nodes?: ReadonlyArray<ASTNode> | ASTNode | null,
+    source?: Maybe<Source>,
+    positions?: Maybe<ReadonlyArray<number>>,
+    path?: Maybe<ReadonlyArray<string | number>>,
+    originalError?: Maybe<Error & { readonly extensions?: unknown }>,
+    extensions?: Maybe<GraphQLErrorExtensions>,
   );
   constructor(message: string, args?: GraphQLErrorArgs);
   constructor(message: string, ...rawArgs: BackwardsCompatibleArgs) {
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -20,6 +20,43 @@ export interface GraphQLErrorExtensions {
   [attributeName: string]: unknown;
 }
 
+interface GraphQLErrorArgs {
+  message: string;
+  nodes?: ReadonlyArray<ASTNode> | ASTNode | null;
+  source?: Maybe<Source>;
+  positions?: Maybe<ReadonlyArray<number>>;
+  path?: Maybe<ReadonlyArray<string | number>>;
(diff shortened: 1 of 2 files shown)


dotansimha (approved):

saihaj (approved):

timsuchanek (approved):
Nice! That's a great api improvement!

src/error/GraphQLError.ts
```diff
@@ -76,15 +113,16 @@ export class GraphQLError extends Error {
...
+  constructor(...rawArgs: BackwardsCompatibleArgs) {
+    const {
+      message,
+      nodes,
+      source,
+      positions,
+      path,
+      originalError,
+      extensions,
+    } = toNormalizedArgs(rawArgs);
```
IvanGoncharov, line 125: Current form is bad for doc generation and DX.
Please use TS overloading with two separate signatures.

IvanGoncharov (changes requested):
> An alternative API might be the following, where the message is ALWAYS the first argument.

Please use this option since it follow JS convention for custom errors.

src/error/GraphQLError.ts
```diff
@@ -20,6 +20,41 @@ export interface GraphQLErrorExtensions {
...
+      source?: GraphQLErrorArgs['source'],
+      positions?: GraphQLErrorArgs['positions'],
+      path?: GraphQLErrorArgs['path'],
+      originalError?: GraphQLErrorArgs['originalError'],
+      extensions?: GraphQLErrorArgs['extensions'],
+    ];
+
+function toNormalizedArgs(args: BackwardsCompatibleArgs): GraphQLErrorArgs {
+  const firstArg = args[0];
+  if (firstArg == null || 'kind' in firstArg || 'length' in firstArg) {
```
n1ru4l, line 45: I used `'length' in firstArg` instead of `Array.isArray` here as the latter did not properly discriminate the type (https://stackoverflow.com/a/57343333)

src/error/GraphQLError.ts
n1ru4l: @IvanGoncharov Done.

src/error/GraphQLError.ts
```diff
@@ -78,13 +113,17 @@ export class GraphQLError extends Error {
...
-    originalError?: Maybe<Error & { readonly extensions?: unknown }>,
-    extensions?: Maybe<GraphQLErrorExtensions>,
-  ) {
+    nodes?: GraphQLErrorArgs['nodes'],
+    source?: GraphQLErrorArgs['source'],
+    positions?: GraphQLErrorArgs['positions'],
+    path?: GraphQLErrorArgs['path'],
+    originalError?: GraphQLErrorArgs['originalError'],
+    extensions?: GraphQLErrorArgs['extensions'],
+  );
```
IvanGoncharov, line 122: Please keep the old type "as is".
It's temporary anyway until we remove the position-based one in the next release.
BTW, you can add deprecation notice for this signature.

src/error/GraphQLError.ts
IvanGoncharov: Not ideal, but we can leave with it since it temporary code anyway.

src/error/__tests__/GraphQLError-test.ts
```diff
@@ -353,4 +353,29 @@ describe('toJSON', () => {
...
+      nodes: [operationNode],
+      source,
+      positions: [6],
+      path: ['path', 2, 'a'],
+      originalError: new Error('I like turtles'),
+      extensions: { hee: 'I like turtles' },
+    });
+
+    expect(error.toJSON()).to.deep.equal({
+      extensions: {
```
IvanGoncharov, line 368: please order fields based on the order of `GraphQLError` properties in definitions.
Please compact short properties into a single line as with other tests in that file.

src/error/GraphQLError.ts
n1ru4l: done

src/error/__tests__/GraphQLError-test.ts
n1ru4l: done!

IvanGoncharov:
@n1ru4l Merged!
If you have time can you please, switch all the existing code in this library to use the new constructor signature (including the tests, just leave one test checking that GraphQLError can accept positional arguments).

n1ru4l:
@IvanGoncharov Yes I will send one sometime this week!

PR #3466, opened 2022-01-27
feat: setup canary releases using changesets

### LINKED ISSUE #3383: Canary releases
As discussed on graphql-js-wg we decided to switch to unstable `main` and provide canary release every week.

Currently, there are a couple of blockers we need to address:
- [ ] Allow publishing NPM packages without updating the `main` (otherwise we will add a bunch of useless commits in git history).
   - [ ] Don't commit `graphql/version.ts` and only generate it during the build process.
   - [ ] Figure out a way to safely release versions without updating `package.json`
- [x] Setup docs site for v16 we should have docs for "stable" version.
- [x] Release v16.1.0
- [x] Switch main to v17.0.0-alpha.0

cc: @saihaj

saihaj:
Need to setup GitHub secret for NPM token `NODE_AUTH_TOKEN` and Bot token to comment `GH_API_TOKEN`

closes #3383


.github/workflows/canary.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Publish a message
+        if: steps.canary.outputs.released
+        uses: 'kamilkisiela/pr-comment@master'
+        with:
+          message: |
+            The latest changes of this PR are available as alpha in npm (based on the declared `changesets`):
+            ```
+            ${{ steps.canary.outputs.changesetsPublishedPackages}}
+            ```
+          bot-token: ${{ secrets.GH_API_TOKEN }}
```
dotansimha, line 52: you can use `${{ secrets.GITHUB_TOKEN }}` here, check the setup in `graphiql` repo

.github/workflows/canary.yml
```diff
@@ -0,0 +1,64 @@
...
+          bot: 'theguild-bot'
```
dotansimha, line 53: i think you can use `github-actions`

.github/workflows/canary.yml
```diff
@@ -0,0 +1,64 @@
+name: Canary Release
```
dotansimha, line 1: I guess we need the 2nd workflow for release? this way we can release using `changeset` as well. What do you think @IvanGoncharov ? 
(Changesets collects all changeset files and keep an updated Release PR, so when it's merged, you get the actual version bump and NPM release)

resources/canary-release.js
```diff
@@ -0,0 +1,99 @@
+/* eslint-disable */
```
dotansimha, line 1: This is optional, we are using that to do custom pre-release and not use the default way of doing that with `changesets`...

.github/workflows/canary.yml
```diff
@@ -0,0 +1,64 @@
...
+        id: canary
+        uses: 'kamilkisiela/release-canary@master'
+        with:
+          npm-token: ${{ secrets.NODE_AUTH_TOKEN }}
+          npm-script: 'npm run release:canary'
+          changesets: true
+
+      - name: Publish a message
+        if: steps.canary.outputs.released
+        uses: 'kamilkisiela/pr-comment@master'
```
IvanGoncharov, line 45: This custom guild stuff, right?
I think idea was that we use some popular tool for doing canary releases?
Those look like there are guild internal scripts?
I'm ok with automation but it should be something popular and widespread.

IvanGoncharov (changes requested):

.github/workflows/canary.yml
kamilkisiela: I can make it `facebook/pr-comment`, does it change anything?

kamilkisiela: how it's different from https://github.com/graphql/graphql-js/blob/main/resources/build-deno.js ?

.github/workflows/canary.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/setup-node@v2
+        with:
+          cache: npm
+          node-version: ${{ env.NODE_VERSION_USED_FOR_DEVELOPMENT }}
+
+      - name: Install Dependencies
+        run: npm ci --ignore-scripts
+
+      - name: Setup NPM credentials
+        run: echo "//registry.npmjs.org/:_authToken=$NODE_AUTH_TOKEN" >> ~/.npmrc
```
IvanGoncharov, line 31: I'm not ok with putting NPM token on FS, it means all subsequent scripts have access to it.
Also, I'm not ok passing NPM token to anything that is not official NPM or GitHub tool/lib/etc.

.github/workflows/canary.yml
IvanGoncharov: We discuss canary release so let's address that and see how it goes.

PR #3467, opened 2022-01-28
feat: disable 'did you mean x' during validation

n1ru4l:
This PR introduces a new property `didYouMean` on the `ValidationContext`, that can be used by validation rules for determining whether suggestions should be added to the error message.

**Example Usage:**
```ts
import { validate } from "graphql";

const errors = validate(schema, documentAST, undefined, {
  didYouMean: false
});
```

I did not add the test yet as I first want to start the discussion on whether this is how this should be implemented.

---

Related issues:

- https://github.com/graphql/graphql-js/issues/2247
- https://github.com/redwoodjs/redwood/issues/4275

### DIFF
--- a/src/validation/ValidationContext.ts
+++ b/src/validation/ValidationContext.ts
@@ -162,6 +162,8 @@ export class SDLValidationContext extends ASTValidationContext {
 export type SDLValidationRule = (context: SDLValidationContext) => ASTVisitor;
 
 export class ValidationContext extends ASTValidationContext {
+  public readonly didYouMean: boolean;
+
   private _schema: GraphQLSchema;
   private _typeInfo: TypeInfo;
   private _variableUsages: Map<
@@ -179,12 +181,15 @@ export class ValidationContext extends ASTValidationContext {
     ast: DocumentNode,
     typeInfo: TypeInfo,
     onError: (error: GraphQLError) => void,
+    /** Whether the "did you mean x" suggestions should be enabled. */
+    didYouMean?: boolean,
   ) {
     super(ast, onError);
     this._schema = schema;
     this._typeInfo = typeInfo;
     this._variableUsages = new Map();
     this._recursiveVariableUsages = new Map();
+    this.didYouMean = didYouMean == null ? true : didYouMean;
   }
 
   get [Symbol.toStringTag]() {
--- a/src/validation/rules/FieldsOnCorrectTypeRule.ts
+++ b/src/validation/rules/FieldsOnCorrectTypeRule.ts
@@ -42,15 +42,19 @@ export function FieldsOnCorrectTypeRule(
           const schema = context.getSchema();
           const fieldName = node.name.value;
 
-          // First determine if there are any suggested types to condition on.
-          let suggestion = didYouMean(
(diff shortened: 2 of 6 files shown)


yaacovCR:
I think this is great!

would suggest option name to be enableSuggestions with default of true, perhaps might eventually default to false

see https://github.com/graphql/graphql-js/issues/2247

in that issue, consensus seems to be around disabling introspection to be one and the same with disabling suggestions, but I think this PR might be about the lower level building blocks that allow that direction

yaacovCR:
There are also some didYouMean calls in enum parseLiteral, I think we can add an execution option in parallel to the one till validate that then has to be passed to parseLiterals?

n1ru4l:
> There are also some didYouMean calls in enum parseLiteral, I think we can add an execution option in parallel to the one till validate that then has to be passed to parseLiterals?

I only focused on the validation rules for now. I am happy to also add this to further parts that require it, once there is a consensus on how we should tackle this!

saihaj (commented):
Looks like right way forward to me. 
Side note: we should make validate also take one argument which is an object.

n1ru4l:
> Side note: we should make validate also take one argument which is an object.

I had a discussion with @benjie about this before in the context of adding `contextValue` to validation rules. One takeaway was "The currently validate signature is useful for the cache use case, [...]". Please correct me if I interpreted this wrong.

````ts
validate(a, b, c)
````

is easier cachable than

```ts
validate({
  a,
  b,
  c,
})
```

For internal reference: https://tortilla-hq.slack.com/archives/CAY2119MX/p1637243621148400

Though I am not against changing the function signature, however, this might be out of the scope of this PR.

benjie:
I meant that if your validation result is valid independent of context and variables then you can cache the validation result and the next time you see the query you can know it's valid independent of the _new_ context and variables that it comes along with. The actual signature of the function is unimportant so long as context and variables are not included (caching it based on the list of args vs the extracted list of properties isn't much different).

benjie:
(Also the GraphQL spec itself makes it very clear that validation does not factor in context or variables - anything that does require that (e.g. coercing the variable values) is done during ExecuteRequest.)

dotansimha (approved):

saihaj (approved):

IvanGoncharov (changes requested):
I fully agree wit Lee's coment here: https://github.com/graphql/graphql-js/issues/2247#issuecomment-815430294

It should be configured by schema and should disable the execution of introspection and even access to introspection types from `GraphQLSchema`.
Also, there are other places using `didYouMean` outside of validation, e.g. https://github.com/graphql/graphql-js/blob/e3aaab0e1172136321cce492ccf9b0bbdf10aebe/src/utilities/coerceInputValue.ts#L140

So it should be a solution that can disable introspection leaking all over the library, not only for validation.

PabloSzx (approved):

andrewmcgivery:
Are there any plans to complete this PR or is this one stale at this point and would need to start over?

JoviDeCroock:
Implemented in https://github.com/graphql/graphql-js/pull/4214

PR #3487, opened 2022-02-08
feat: expose getArgumentValues

### LINKED ISSUE #2719: Expose getArgumentValues from root
`getArgumentValues` is marked as internal and only exposed when you import from `graphql/execution/values`.

If you could expose it from the root `index.js` it would make importing it far easier in contexts without imports to relative files (a cdn for my case.)

The function is already used in production ready code in joinmonster and is necessary to get arguments from resolve info in field handlers. I don't see a good reason not to expose it.

saihaj:
Closes #2719
﻿

### DIFF
--- a/src/execution/index.ts
+++ b/src/execution/index.ts
@@ -16,7 +16,7 @@ export type {
 export { subscribe, createSourceEventStream } from './subscribe';
 
 export {
-  getVariableValues,
   getArgumentValues,
+  getVariableValues,
   getDirectiveValues,
 } from './values';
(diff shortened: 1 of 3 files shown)


n1ru4l (approved):

src/execution/index.ts
```diff
@@ -15,4 +15,8 @@ export type {
 
 export { subscribe, createSourceEventStream } from './subscribe';
 
-export { getVariableValues, getDirectiveValues } from './values';
+export {
+  getVariableValues,
+  getArgumentValues,
```
IvanGoncharov, line 20:
```suggestion
  getArgumentValues,
  getVariableValues,
```

@saihaj Please use consistent order in different `index.ts` files.
Changed to follow the same order as `src/index.ts`

Issue #2974, opened 2021-03-18
Resolvers keep running even after one of them throws/rejects - consider Promise.allSettled?


olsavmic:
Hi,
I recently came across this issue while trying to create a DB transaction per request.

The reason for using a DB transaction per request (which means 1 connection per client) is to keep the returned data consistent (`ISOLATION READ COMMITED`) as with increased traffic some inconsistencies started to occur on certain queries.

The second reason is having more control over the db connections to support load balancing with dynamic number of read replicas (as the only available solution unfortunately does not support connection pool). However I suppose this can be solved with some effort.

The problem is that **when one of the resolvers fails with error** (which is not a casual case but happens sometimes), **other resolvers keep running**. However I need to **release the db connection** before I send the response which sometimes happens **before the rest of resolvers finishes** (most of them are data loaders which makes the problem more obvious as the default behaviour is to wait for the next tick until the batch operation runs).

The problem could be resolved by using `Promise.allSettled` instead of `Promise.all`. I can't come up with any problem this change could cause except longer time till response in case of error.

Can you please correct me if I'm mistaken or consider this change? Or maybe transactional processing of resolvers is a bad idea in general (although I don't see any reason except the performance gain from using multiple connections per request)

Thanks a lot!

IvanGoncharov:
@olsavmic I think it's a more general issue than DB transactions since it will affect any resources with reference counting. 
Also, it can potentially DDoS problem, if you can find a request that fails fast but provokes a lot of memory allocation referenced by long-running resolvers you can run out of memory on the server.

But we need to be extra careful with `execute` code to not affect performance or correctness.
So let's postpone this fix until we finish TS conversion and release it in the upcoming `16.0.0`.

olsavmic:
Good point, thank you!

olsavmic:
Hi @IvanGoncharov, I see this issue in the 16.0.0-alpha.1 milestone which has already been released. Is it still planned for the v16 or did you come over to some issues with this change? 

Thanks a lot! :)

hamzahamidi:
This feature is useful when the tasks are not dependent on each other.

olsavmic:
@hamzahamidi Can you elaborate on this? I can't think of such situation as I'm not saying to run these resolvers in sequence but rather just way for all resolvers to finish before sending a response which allows for proper resource cleanup and prevents possible DDoS as @IvanGoncharov stated.

yaacovCR:
[...]

PR #3494, opened 2022-02-11
fix: improve handling of promises in defaultTypeResolver

hayes:
Currently it is possible to have unhandled promise rejections that arise from a mix of sync and async isTypeOf checks.

This should fix that issue.

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -1823,6 +1823,12 @@ export const defaultTypeResolver: GraphQLTypeResolver<unknown, unknown> =
         if (isPromise(isTypeOfResult)) {
           promisedIsTypeOfResults[i] = isTypeOfResult;
         } else if (isTypeOfResult) {
+          if (promisedIsTypeOfResults.length > 0) {
+            Promise.all(promisedIsTypeOfResults).then(undefined, () => {
+              /* ignore errors */
+            });
+          }
+
           return type.name;
         }
       }
--- a/src/execution/__tests__/union-interface-test.ts
+++ b/src/execution/__tests__/union-interface-test.ts
@@ -1,6 +1,8 @@
 import { expect } from 'chai';
 import { describe, it } from 'mocha';
 
+import { expectJSON } from '../../__testUtils__/expectJSON.js';
+
 import { parse } from '../../language/parser.js';
 
 import {
@@ -12,7 +14,7 @@ import {
 import { GraphQLBoolean, GraphQLString } from '../../type/scalars.js';
 import { GraphQLSchema } from '../../type/schema.js';
 
-import { executeSync } from '../execute.js';
+import { execute, executeSync } from '../execute.js';
 
 class Dog {
   name: string;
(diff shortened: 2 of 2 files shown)


saihaj (approved):

yaacovCR:
This means that a mixture of sync/async calls becomes sync unnecessarily, as we return sync value…. Maybe catch handlers should be added to the promises so they reject silently? Is that possible?

hayes:
I had considered that, but silently swallowing errors seems like a bad idea.  I think it makes sense for execution to become async if any async functions are called.  This should be a pretty rare case, but if these errors were swallowed silently it could become very tricky to debug.

yaacovCR:
A side channel might be useful as well in context of https://github.com/graphql/graphql-js/issues/2974 to report the first error to client and potentially log all errors

yaacovCR:
But as there currently is no side channel, I think I would approve these changes, and then refactor later

hayes:
kinda forgot about this PR, but I think this change would still be good to get in

src/execution/execute.ts
```diff
@@ -1618,6 +1618,10 @@ export const defaultTypeResolver: GraphQLTypeResolver<unknown, unknown> =
         if (isPromise(isTypeOfResult)) {
           promisedIsTypeOfResults[i] = isTypeOfResult;
         } else if (isTypeOfResult) {
+          if (promisedIsTypeOfResults.length > 0) {
```
yaacovCR, line 1621:
```suggest
if (promisedIsTypeOfResults.length > 0) {
  Promise.all(promisedIsTypeOfResults).then(undefined, () => {
    /* ignore errors */
  });
}
```

yaacovCR, line 1621: I think this should probably silently ignore the errors....

Something like:

```ts
if (promisedIsTypeOfResults.length > 0) {
  Promise.all(promisedIsTypeOfResults).then(undefined, () => {
    /* ignore errors */
  });
}
```

yaacovCR: Seeing as one functioning isTypeOf resolver is all one should have and all one needs.

Compare also to a sync case with multiple sync isTypeOf resolvers, one that returns true, and multiple that throw instead of returning false. If the order is serendipitous, the non-erroring one will return true and mask all the other potential errors.

In #3706, we convert a sync error to an async error, but that's the not happy path of an error. Here, this change will convert every synchronous isTypeOf to a non-synchronous....

yaacovCR (commented):
There are 2 different questions:

1. Should there ever be unhandled promise rejections?
2. Must all errors be reported?

There probably should never be unhandled promise rejections, but we do not necessarily need to report all errors.

src/execution/execute.ts
yaacovCR: Compare also to multiple async errors, when we return immediately with the first async error, without awaiting the others. I would argue that #3706 is a bit funny, in that if we have a sync error, there is no need for:

```ts
return promiseForObject(...).finally(() => throw originalError);
```

we should just have:
```ts
promiseForObject(....).then(undefined, () =>{ /*ignore*/ });
throw originalError;
```

hayes: makes sense, updated

hayes:
Forgot about this PR, but wanted to check in and see if this can get merged

JoviDeCroock (approved):
This looks like the right change to me in terms of error-resilience - there is an unwanted commit in the history here though

hayes:
Good catch! I removed the extra commit and rebased the pr

JoviDeCroock:
@hayes is the intention of landing this in 16? Else we might need to target the 16.x.x branch.

CC @yaacovCR are you okay with the changes as they are currently lined out, wanted to follow up on your concerns.

hayes:
@JoviDeCroock I don't feel strongly about which release this makes it into, for me its not urgent, it just seems like it would be good to land eventually

src/execution/execute.ts
yaacovCR: Looks good!

PR #3499, opened 2022-02-13
extend breaking change detection by returning affected astnodes

mmuth:
When comparing two GraphQL schemas and finding breaking changes between them, I'd propose to not only describe the found changes in a textual way (`myType.field changed type from String to Int`), but also add the technical background in forms of affected AST nodes.

I am just implementing a validation solution that tries to detect in detail, if a GQL client is affected by a breaking change. With this simple (and backwards compatible) addition, it is possible. Thought it could be useful in general :).

### DIFF
--- a/src/utilities/__tests__/findBreakingChanges-test.ts
+++ b/src/utilities/__tests__/findBreakingChanges-test.ts
@@ -1,6 +1,15 @@
 import { expect } from 'chai';
 import { describe, it } from 'mocha';
 
+import type {
+  DirectiveDefinitionNode,
+  EnumTypeDefinitionNode,
+  InputObjectTypeDefinitionNode,
+  InterfaceTypeDefinitionNode,
+  ObjectTypeDefinitionNode,
+} from '../../language/ast';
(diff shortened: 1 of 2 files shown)


IvanGoncharov (changes requested):

IvanGoncharov (changes requested):
@mmuth My main concern with this PR is that it's too tied to SDL.
Since you can construct from many different sources and these schemas would miss AST Nodes.

Looking at [how it's done in GraphQL Inspector](https://github.com/kamilkisiela/graphql-inspector/blob/6b02e14a4b75524cb5a92684d1944b7deaa08a4f/packages/core/src/diff/changes/argument.ts#L24) I think it makes sense to just capture "schema coordinates" of a change.
See #3044 and #3145.
Also, #3044 makes it super easy to resolve coordinates back to astNode.
Also it would be super easy to test (no need to dig up AST nodes in tests).

Let me think about this feature for some time and please ignore my review comments until we finalize API design.

PR #3504, opened 2022-02-17
Declare fieldNodes always has at least 1 FieldNode

mattkrick:
fieldNodes will never be an empty array.

```ts
// Before
const node = info.fieldNodes[0] // undefined | FieldNode

// After
const node = info.fieldNodes[0] // FieldNode
```

This is helpful when noUncheckedIndexAccess is turned on.


Ramifaria1984 (changes requested):
- [ ] @ #

PR #3513, opened 2022-03-18
Implement OneOf Input Objects via `@oneOf` directive

erikkessler1:
This PR implements graphql/graphql-spec#825 to add a `@oneOf` directive that indicates ~~an Object is a OneOf Object or~~ an Input Object is a OneOf Input Object. We then validate ~~OneOf Objects and~~ OneOf Input Objects conform to the spec during the validation and execution phases.

Closes graphql/graphql-wg#648


src/execution/__tests__/oneof-test.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      expectJSON(result).toDeepEqual({
+        data: {
+          test: null,
+        },
+        errors: [
+          {
+            locations: [{ column: 23, line: 3 }],
+            message:
+              'Argument "input" of non-null type "TestInputObject!" must not be null.',
```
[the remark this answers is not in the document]
benjie: Ah; we're not actually validating the document so this is irrelevant. In which case, yes, what you have here seems correct 👍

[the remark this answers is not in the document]
benjie: Ah; we're not actually validating the document so this is irrelevant. In which case, yes, what you have here seems correct 👍

erikkessler1:
> As for oneOf input objects, this is all the spec PR addresses currently; would it make sense to cut down this PR to only deal with oneOf input objects to get this closer to review?

Thank you for taking a look! I agree that focusing on oneOf input objects makes sense, so I've stripped out everything around oneOf objects.

benjie:
Odd, all my comments seem to have been duplicated; I've deleted the duplicates.

benjie (changes requested):
This is looking good :+1:

My main concern right now is very JS-specific, and it's that we've not tested `undefined` and I think that might cause a whole host of issues with the `Object.keys` strategy. Though GraphQL typically uses JSON (which drops `undefined`) this is not always the case; I'm concerned that `variables: { a: "abc", b: undefined }` which _should_ be valid might not be seen as such. Please add some tests for this so we can check :pray:

src/utilities/coerceInputValue.ts
```diff
@@ -142,6 +142,28 @@ function coerceInputValueImpl(
...
+        onError(
+          pathToArray(path),
+          inputValue,
+          new GraphQLError('Exactly one key must be specified.'),
+        );
+      }
+
+      const key = keys[0];
+      const value = coercedValue[key];
+      if (value === null) {
```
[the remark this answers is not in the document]
erikkessler1: I believe checking `value === null` is the correct behavior. If the value is `undefined` that means the value couldn't be coerced (e.g., wrong type or not a field that exists).

The following tests demonstrate this and will fail if we use `value == null` 
https://github.com/graphql/graphql-js/blob/36f44d36724612fba8f5d15b8ce8b6c524d3b78d/src/utilities/__tests__/coerceInputValue-test.ts#L313-L375

erikkessler1:
I've added some test coverage for those cases including the `variables: { a: "abc", b: undefined }` case.

benjie (approved):
Looks good to me - excellent work @erikkessler1! Will need a proper review from someone more familiar with GraphQL.js than I, but it feels like this implements the behaviour in the oneof spec accurately.

GraphQL.js reviewer: please pay special attention to the error messages, I'm not sure of the conventions here. Feel free to ping me (here or via DM) to discuss if needed.

benjie:
Pushed a commit that hopefully deals with the code coverage issue :+1:

benjie:
Merged main; CI passing now :raised_hands:

src/type/introspection.ts
```diff
@@ -323,6 +323,16 @@ export const __Type: GraphQLObjectType = new GraphQLObjectType({
...
         resolve: (type) => ('ofType' in type ? type.ofType : undefined),
       },
+      oneOf: {
+        type: GraphQLBoolean,
+        resolve: (type) => {
+          if (isInputObjectType(type)) {
+            return type.isOneOf;
+          }
+
+          return null;
```
IvanGoncharov, line 333: You don't need to return explicit null.
See the above fields, e.g. `enumValues`.

src/type/introspection.ts
```diff
@@ -323,6 +323,16 @@ export const __Type: GraphQLObjectType = new GraphQLObjectType({
         type: __Type,
         resolve: (type) => ('ofType' in type ? type.ofType : undefined),
       },
+      oneOf: {
+        type: GraphQLBoolean,
+        resolve: (type) => {
+          if (isInputObjectType(type)) {
+            return type.isOneOf;
```
IvanGoncharov, line 330: We should have matching names in introspection and in JS types.
So we either should have both be `oneOf` or both being `isOneOf`.
Personally, I like `isOneOf` + it matching the `@deprecated`/`isDeprecated` pair that we already have.

IvanGoncharov (changes requested):

src/type/introspection.ts
benjie: Updated in spec and implementation to use `isOneOf` for consistency with `isDeprecated` :+1:

src/type/introspection.ts
benjie: Fixed :+1:

benjie:
Great feedback, thanks @IvanGoncharov!

benjie:
@graphql/graphql-js-reviewers I believe this is ready for review and possibly merge into v17 👍

seamusml:
Hi 👋, is there any outlook on when this might get merged and published? :)

benjie:
@graphql/graphql-js-reviewers I've merged in the latest `main`, so it should be good for review 👍 

@IvanGoncharov Can you dismiss your stale review, please?

saltman424:
@IvanGoncharov any chance you can dismiss the review @benjie mentioned above? I know this has been a long awaited feature for many

src/execution/__tests__/oneof-test.ts
IvanGoncharov, line 1: Can you please rename it to `oneOf-test.ts`
We use camelcase in names

src/execution/__tests__/oneof-test.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  type TestObject {
+    a: String
+    b: Int
+  }
+
+  schema {
+    query: Query
+  }
```
IvanGoncharov, line 29: Not needed since `Query` is already a shorthand.

src/execution/__tests__/oneof-test.ts
```diff
@@ -0,0 +1,187 @@
...
+  variableValues?: { [variable: string]: unknown },
+): ExecutionResult | Promise<ExecutionResult> {
+  return execute({ schema, document: parse(query), rootValue, variableValues });
+}
+
+describe('Execute: Handles OneOf Input Objects', () => {
+  describe('OneOf Input Objects', () => {
+    const rootValue = {
+      test({ input }: { input: { a?: string; b?: number } }) {
+        return input;
```
IvanGoncharov, line 44: We have a special function for this case
https://github.com/graphql/graphql-js/blob/main/src/jsutils/identityFunc.ts

src/execution/__tests__/oneof-test.ts
```diff
@@ -0,0 +1,187 @@
...
+          }
+        }
+      `;
+      const result = executeQuery(query, rootValue);
+
+      expectJSON(result).toDeepEqual({
+        data: {
+          test: {
+            a: 'abc',
+            b: null,
```
IvanGoncharov, line 63: It's a confusing test that `b` is not null in resolver's `args`.
It's better for the resolver to return the `String` that contains `inspect` of `args`.

src/execution/__tests__/oneof-test.ts
```diff
@@ -0,0 +1,187 @@
...
+      const result = executeQuery(query, rootValue, {
+        input: { a: 'abc', b: 123 },
+      });
+
+      expectJSON(result).toDeepEqual({
+        errors: [
+          {
+            locations: [{ column: 16, line: 2 }],
+            message:
+              'Variable "$input" got invalid value { a: "abc", b: 123 }; Exactly one key must be specified.',
```
IvanGoncharov, line 157: It doesn't mention the name of the type and name of conflicting properties.
Other errors have this info, e.g. https://github.com/graphql/graphql-js/blob/e17a0897f67305626c6090ce0174f101b7a96fc4/src/execution/__tests__/variables-test.ts#L506

src/execution/__tests__/oneof-test.ts
IvanGoncharov, line 1: We generally don't create per-language feature test files.
Instead, tests are grouped per library functionality.
All similar tests leave in this file:
https://github.com/graphql/graphql-js/blob/e17a0897f67305626c6090ce0174f101b7a96fc4/src/execution/__tests__/variables-test.ts
Do you think you need a separate file?

src/execution/__tests__/oneof-test.ts
```diff
@@ -0,0 +1,187 @@
...
+    query: Query
+  }
+`);
+
+function executeQuery(
+  query: string,
+  rootValue: unknown,
+  variableValues?: { [variable: string]: unknown },
+): ExecutionResult | Promise<ExecutionResult> {
+  return execute({ schema, document: parse(query), rootValue, variableValues });
```
IvanGoncharov, line 37: You use exactly the same value in all your test cases, why do you need to pass it every time?

src/execution/__tests__/oneof-test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      expectJSON(result).toDeepEqual({
+        data: {
+          test: null,
+        },
+        errors: [
+          {
+            locations: [{ column: 23, line: 3 }],
+            message:
+              'Argument "input" of non-null type "TestInputObject!" must not be null.',
```
IvanGoncharov, line 88: It's a missleading error message.
Is it specific just for `@oneOf`, or do we have a general issue here?

src/type/__tests__/introspection-test.ts
```diff
@@ -1519,6 +1536,105 @@ describe('Introspection', () => {
...
+        anotherField(anotherArg: AnotherInputObject): String
+      }
+    `);
+
+    const source = `
+      {
+        a: __type(name: "SomeInputObject") {
+          isOneOf
+        }
+        b: __type(name: "AnotherInputObject") {
```
IvanGoncharov, line 1561: nit: please use meaningful names for types and aliases as in other tests.
`oneOfObject: __type(name: "OneOfObject")`

src/type/__tests__/introspection-test.ts
```diff
@@ -1519,6 +1536,105 @@ describe('Introspection', () => {
...
+        b: {
+          isOneOf: false,
+        },
+      },
+    });
+  });
+
+  it('returns null for oneOf for other types', () => {
+    const schema = buildSchema(`
+      type A implements I {
```
IvanGoncharov, line 1581: Please use the standard name for dummy types, not letters. See other tests in this file.

src/type/__tests__/introspection-test.ts
```diff
@@ -1519,6 +1536,105 @@ describe('Introspection', () => {
...
+        }
+        u: __type(name: "U") {
+          isOneOf
+        }
+      }
+    `;
+
+    expect(graphqlSync({ schema, source })).to.deep.equal({
+      data: {
+        a: {
```
IvanGoncharov, line 1619: nit: In this and previous test you can shorten it by doing
```suggestion
        a: { isOneOf: null },
```
as a single line

IvanGoncharov (changes requested):
It's a partial review. I will find time for other files soon.

Also, please update these files to have the new type with `@oneOf`: https://github.com/graphql/graphql-js/blob/e17a0897f67305626c6090ce0174f101b7a96fc4/src/__testUtils__/kitchenSinkSDL.ts

IvanGoncharov:
Sorry missed that spec proposal in stage 2, so this PR can actually be merged.
Also missed all the comments (my inbox is in bad shape atm).
@benjie Thank you for the ping 👍 

@erikkessler1 If you don't mind, I will address stylistic changes by just committing on top of your PR.
I think it would be faster that way, it takes a lot of time to comment on a lot of tiny changes.

erikkessler1:
> If you don't mind, I will address stylistic changes by just committing on top of your PR.
I think it would be faster that way, it takes a lot of time to comment on a lot of tiny changes.

@IvanGoncharov that sounds good to me!

src/execution/__tests__/oneof-test.ts
erikkessler1: This isn't quite identity as we need to destructure the argument to get `input`.

src/execution/__tests__/oneof-test.ts
erikkessler1: This sort of error would be caught at validation time and give a better message. The default value is invalid so it gets coerced to null.

src/type/__tests__/introspection-test.ts
erikkessler1: Updated.

src/execution/__tests__/oneof-test.ts
erikkessler1: I was keeping it as a generic helper in case we want to pass different values down the line. Similar to: https://github.com/graphql/graphql-js/blob/e17a0897f67305626c6090ce0174f101b7a96fc4/src/execution/__tests__/nonnull-test.ts#L111-L116

src/execution/__tests__/oneof-test.ts
erikkessler1: I'm not sure I understand what you mean by "b is not null in resolver's args". Passing `{ a: "abc", b: null }` would be invalid.

IvanGoncharov:
@erikkessler1 Thanks for addressing my review comments, and sorry for the delay 🙇 
I'm still planning to review/merge this PR, and I will try to find time this weekend.

IvanGoncharov:
@erikkessler1 Sorry, I didn't find time for a quality review.
That said, the last time I reviewed it, all issues were non-critical and could be fixed later.
Since there were no other objections I will merge it as is.

PR #3516, opened 2022-03-20
fix: 🏷️ Updated types reported by deno

hinogi:


yaacovCR:
Closing this PR for now. With current deno supporting node package, our use in v17 of conditional exports, and our deno github branch, I am hoping any additional deno customization is no longer necessary. We can reopen this if there is a need.

PR #3527, opened 2022-04-05
allow unions to declare implementation of interfaces

yaacovCR:
See: https://github.com/graphql/graphql-spec/pull/939


yaacovCR:
https://github.com/graphql/graphql-spec/pull/939#issuecomment-1092282972

yaacovCR:
Should queries be allowed to select fields directly on the union now that the union definitely implements the field?

Some exploration within the reference implementation yielded that queries of this type would execute anyway without a problem, but fail validation because of the FieldsOnCorrectTypeRule.

So, this rule would possibly need to be updated for a union to find the field on one of the interfaces it implements!

yaacovCR:
*Decision Record* We decided [at the last WG](https://github.com/graphql/graphql-wg/blob/main/agendas/2022/2022-05-05.md) that if unions implement interfaces, then they have directly querable fields. This PR and [the spec PR](https://github.com/graphql/graphql-spec/pull/939) have been updated with the changes for introspection, validation.

Now all composite types have a `getFields` method!

yaacovCR:
Actually, this naive field copying from the interfaces to the Union does not account for different interfaces with valid overlapping fields like new arguments with defaults. Work to do!

yaacovCR:
closing this old, stale spec PR. still available for reference

PR #3537, opened 2022-04-19
ESLint: disallow using node globals in src/tests

IvanGoncharov:


robrichard:
@IvanGoncharov can we no longer use `setTimeout`/`setImmediate` in tests? After this change they are lint errors from `eslint/no-undef`?

PR #3543, opened 2022-04-24
pass valid value to codecov config

is2ei:
This PR updates the Codecov config to pass the valid option.

https://docs.codecov.com/docs/pull-request-comments#disable-comment

> To disable the comment, simply set the top-level comment key to false in your codecov.yml, as shown below.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,10 +1,10 @@
 codecov:
   notify:
-    require_ci_to_pass: yes
+    require_ci_to_pass: true
 
 parsers:
   javascript:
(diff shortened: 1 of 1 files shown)


saihaj (approved):

IvanGoncharov:
YAML 1.1 supports `yes` and `no` as aliases for `true` & `false`.

IvanGoncharov:
And codecov docs used both true/false and yes/no:
https://docs.codecov.com/docs/codecov-yaml

But I agree that true/false is less confusing.

PR #3511, opened 2022-03-15
feat: allow symbols as GraphQLSchema extension fields

n1ru4l:

PR #3546, opened 2022-04-26
Drop support for node 12 and 17, update rest to latest minor version

IvanGoncharov:


n1ru4l:
@IvanGoncharov Could we also drop support for old TypeScript versions as part of this pull request (typescript-4.3)?

n1ru4l:
@IvanGoncharov See https://github.com/graphql/graphql-js/pull/3511#issuecomment-1110015285

PR #3548, opened 2022-04-27
ci: disable Codecov and measure test coverage on every run

IvanGoncharov:
Motivation: codecov fails from time to time creating false alarms for
our CI. For example, https://github.com/graphql/graphql-js/runs/6158401816?check_suite_focus=true


yaacovCR:
@IvanGoncharov i don't know if c8 has a watch option, I couldn't find one -- preserving a testonly script without c8 is helpful to just throw on --watch on the command line.

Or, maybe we could separately add a testonly:watch script?

PR #3550, opened 2022-04-29
introduce new Intersection type

yaacovCR:
Intersections of unions and interfaces can be considered to "implement" their unions and interface members.

A type with a field of type Intersection will satisfy an interface where the field defined in the interface is one of the member types of the intersection.

Alternative to #3527


yaacovCR:
https://github.com/graphql/graphql-spec/pull/939#issuecomment-1113270431

yaacovCR:
This needs a separate spec PR, will post link when ready.

yaacovCR:
spec PR: https://github.com/graphql/graphql-spec/pull/941

yaacovCR:
discussion @ https://github.com/graphql/graphql-wg/discussions/944

yaacovCR:
Summary of discussion from WG https://github.com/graphql/graphql-wg/blob/main/agendas/2022/2022-05-05.md :

`union HousePet implements Node` was overwhelmingly felt to be best because it is (1) expressive (2) concise (3) less complex (4) will lead to early validation error, which overall would be most valuable. Intersections were felt to be powerful tools that would then propagate a small number of breaking changes to a large number of dangerous and --more importantly-- possibly silent errors!

The action item will then be to move forward as possible with https://github.com/graphql/graphql-spec/pull/939 . Immediate next steps is to refine the implementation at https://github.com/graphql/graphql-js/pull/3527 with further checking for all the necessary validation and test changes.

PR #3552, opened 2022-05-01
Switch NPM package to support only ESM

IvanGoncharov:
As disscussed during graphql-js-wg:
https://github.com/graphql/graphql-js-wg/blob/main/agendas/2022-03-30.md

### DIFF
--- a/.babelrc-npm.json
+++ b/.babelrc-npm.json
@@ -1,27 +1,10 @@
 {
-  "plugins": [
-    "@babel/plugin-transform-typescript",
-    "./resources/inline-invariant"
+  "presets": [
+    ["@babel/preset-env", { "modules": false, "targets": { "node": "14" } }]
   ],
-  "env": {
-    "cjs": {
-      "presets": [
-        [
-          "@babel/preset-env",
-          { "modules": "commonjs", "targets": { "node": "14" } }
-        ]
-      ],
-      "plugins": [
-        ["./resources/add-extension-to-import-paths", { "extension": "js" }]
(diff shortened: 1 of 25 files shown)


saihaj:
@IvanGoncharov what about #3361 ?

IvanGoncharov:
> @IvanGoncharov what about #3361 ?

@saihaj As we agreed, let's try with full ESM first, and if the community strongly asks for CJS we can go the #3361 approach with dual packages.

saihaj:
> > @IvanGoncharov what about #3361 ?
> 
> @saihaj As we agreed, let's try with full ESM first, and if the community strongly asks for CJS we can go the #3361 approach with dual packages.

I mean the work was already done there and we just need to that `esm` only package under main `graphql` npm

IvanGoncharov:
> I mean the work was already done there and we just need to that esm only package under main graphql npm

@saihaj Can't do that, benchmark and integration tests need to use ESM, it's the absolute majority of changes in this PR.
https://github.com/graphql/graphql-js/pull/3361 don't do that.

saihaj:
> > I mean the work was already done there and we just need to that esm only package under main graphql npm
> 
> @saihaj Can't do that, benchmark and integration tests need to use ESM, it's the absolute majority of changes in this PR. #3361 don't do that.

@IvanGoncharov we have tests in that PR 
image

saihaj:
ok but we can just push a commit on initial work done there and wait for @graphql/graphql-js-reviewers team for review

PR #3529, opened 2022-04-05
Fix crash in node when mixing sync/async resolvers

asztal:
Fixes #3528

Ensures that if `executeFields` encounters a mix of promises and thrown errors, the promises will be awaited before throwing the error.

This does technically change `executeFields` to return a rejected promise in this scenario as opposed to throwing the not-null error synchronously, but I figured if any of the resolvers are asynchronous then returning a promise will be expected anyway, and this was better than crashing the process.

Note: I had to add an `unhandledRejection` event listener myself, as it seems mocha doesn't do it.

PR #3573, opened 2022-05-11
Fix crash in node when mixing sync/async resolvers (backport of #3529)

### LINKED ISSUE #3528: Unhandled rejection if an object has both a synchronous and asynchronous "Cannot return null for non-nullable field" error
Sorry about the title but it's a pretty specific scenario.

It looks like this error occurs because of how errors are handled for non-nullable fields. (I saw this with the `Cannot return null for non-nullable field` error, but it seems if the resolver throws an error it would be handled the same way.)

The problem appears to be here: https://github.com/graphql/graphql-js/blob/main/src/execution/execute.ts#L450
Let's say in the example below, the City object type has a resolver for `country`, which is asynchronous, and a resolver for `population`, which is synchronous, and they _both_ return null (which they shouldn't, of course). The resolver for the `country` field is run, and returns a promise. Then the resolver for the `population` field is run, and because it's synchronous, the error is thrown straight away, and the promise returned by the `country` resolver is discarded. When that promise eventually rejects, the process crashes with an `unhandledRejection` error. 

```js
// File: testcase.mjs
import { graphql, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLSchema, GraphQLString } from "graphql";

process.on("unhandledRejection", reason => {
    console.error("Unhandled rejection!", reason);
});

const City = new GraphQLObjectType({
    name: "City",
    fields: {
        name: { type: new GraphQLNonNull(GraphQLString) },
        country: { type: new GraphQLNonNull(GraphQLString), resolve: () => Promise.resolve(null) },
        population: { type: new GraphQLNonNull(GraphQLInt), resolve: () => null },
    }
});

    name: "Query",
    fields: {
        cities: {
            type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(City))),
            resolve() {
                return [{ name: "London", country: "GB", population: 8_982_000 }];
            }
        }
    }
});

    types: [City],
    query: Query,
});

try {
    graphql(schema, `
        query {
            cities { name country population }
        }
    `).then(result => {
        if (result.errors)
            console.error("Error returned in result.errors[]:", result.errors);
        else
            console.log("Result:", result.data);
    }, error => {
        console.error("graphql() rejected:", error);
    });
} catch(err) {
    console.error("Error caught by try/catch:", err);
}
```

The output of this is:

```
Error returned in result.errors[]: [
  Error: Cannot return null for non-nullable field City.population.
      at completeValue (/home/lee/restapi/node_modules/graphql/execution/execute.js:566:13)
      at resolveField (/home/lee/restapi/node_modules/graphql/execution/execute.js:478:19)
      at executeFields (/home/lee/restapi/node_modules/graphql/execution/execute.js:292:18)
      at collectAndExecuteSubfields (/home/lee/restapi/node_modules/graphql/execution/execute.js:755:10)
      at completeObjectValue (/home/lee/restapi/node_modules/graphql/execution/execute.js:745:10)
      at completeValue (/home/lee/restapi/node_modules/graphql/execution/execute.js:597:12)
      at completeValue (/home/lee/restapi/node_modules/graphql/execution/execute.js:563:21)
      at /home/lee/restapi/node_modules/graphql/execution/execute.js:627:25
      at Array.map (<anonymous>)
      at safeArrayFrom (/home/lee/restapi/node_modules/graphql/jsutils/safeArrayFrom.js:36:23) {
    locations: [ [Object] ],
    path: [ 'cities', 0, 'population' ]
  }
]
Unhandled rejection! Error: Cannot return null for non-nullable field City.country.
    at completeValue (/home/lee/restapi/node_modules/graphql/execution/execute.js:566:13)
    at /home/lee/restapi/node_modules/graphql/execution/execute.js:475:16 {
  locations: [ { line: 3, column: 27 } ],
  path: [ 'cities', 0, 'country' ]
}
```

chrskrchr:
This is a backport of ~#3529~ #3651 to the `15.x` branch that fixes #3528.  

Full credit to @asztal for identifying the fix and submitting the original PR.


src/__testUtils__/expectJSON.js
```diff
@@ -0,0 +1,51 @@
+import { expect } from 'chai';
```
chrskrchr, line 1: Backported and stripped type information from this module on `main`:

https://github.com/graphql/graphql-js/blob/main/src/__testUtils__/expectJSON.ts

chrskrchr:
Moving this PR back to draft while I sort out the typing issues.

IvanGoncharov (changes requested):
Also please backport it to v16.
It would be strange to have this fix in v15 and v17 but not v16 :)

chrskrchr:
> Also please backport it to v16. It would be strange to have this fix in v15 and v17 but not v16 :)

Thanks for the heads up!  I wasn't aware that `main` had moved on to v17.  Submitted a similar PR for `16.x.x` here:

https://github.com/graphql/graphql-js/pull/3576

chrskrchr:
@IvanGoncharov - I finally worked through the EasyCLA issues, would you mind taking another look at this PR?

chrskrchr:
Closing this and re-submitting as a new PR per the guidance in #3651.

PR #3586, opened 2022-05-12
Fix performance issue with ESLint

IvanGoncharov:


IvanGoncharov:
Before ~5m
After ~20s
🎉 🎉 🎉

PR #3593, opened 2022-05-16
Add `GraphQLValidationError` as a subclass of `GraphQLError`

IvanGoncharov:
Also deprecated `ValidationContext.reportError` in favor of new `report`
function with named argument.

Motivation: subclassing allows you to distinguish between errors produced by `graphql-js`.
Idea is to subclass all errors into different types e.g. GraphQLValidationError, GraphQLSyntaxError, etc.
That way clients can be sure what type of error it is.

```js
const { data, errors, extensions } = graphql({ /* ... */ });

const response = {
  data,
  errors.map((error) => {
    if (error instanceof GraphQLValidationError) {
      // do a special handling of validation errors
    } else if (error instanceof GraphQLSyntaxError) {
      // do a special handling of syntax errors
    }
  }),
  extensions,
};
```

Without this change, you can't distinguish one type of error from other types of errors.
Subclassing errors for that purpose is a standard approach in JS (e.g. `TypeError` is a subclass of `Error`) and other OOP languages.

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -237,3 +237,21 @@ export interface GraphQLFormattedError {
    */
   readonly extensions?: { [key: string]: unknown };
 }
+
+export interface GraphQLValidationErrorOptions {
+  nodes: ReadonlyArray<ASTNode> | ASTNode;
+  originalError?: Error | undefined;
+  extensions?: GraphQLErrorExtensions | undefined;
+}
+
+export class GraphQLValidationError extends GraphQLError {
+  constructor(message: string, options: GraphQLValidationErrorOptions) {
+    super(message, options);
+
+    this.name = 'GraphQLValidationError';
+  }
+
+  get [Symbol.toStringTag](): string {
+    return 'GraphQLValidationError';
+  }
+}
--- a/src/error/index.ts
+++ b/src/error/index.ts
@@ -1,6 +1,7 @@
-export { GraphQLError } from './GraphQLError';
+export { GraphQLError, GraphQLValidationError } from './GraphQLError';
 export type {
   GraphQLErrorOptions,
+  GraphQLValidationErrorOptions,
   GraphQLFormattedError,
   GraphQLErrorExtensions,
 } from './GraphQLError';
--- a/src/index.ts
+++ b/src/index.ts
@@ -377,10 +377,16 @@ export {
 export type { ValidationRule } from './validation/index';
 
 // Create, format, and print GraphQL errors.
-export { GraphQLError, syntaxError, locatedError } from './error/index';
+export {
+  GraphQLError,
+  GraphQLValidationError,
+  syntaxError,
+  locatedError,
+} from './error/index';
 
 export type {
   GraphQLErrorOptions,
+  GraphQLValidationErrorOptions,
   GraphQLFormattedError,
   GraphQLErrorExtensions,
 } from './error/index';
--- a/src/validation/ValidationContext.ts
+++ b/src/validation/ValidationContext.ts
@@ -1,9 +1,14 @@
 import type { Maybe } from '../jsutils/Maybe';
 import type { ObjMap } from '../jsutils/ObjMap';
 
-import type { GraphQLError } from '../error/GraphQLError';
+import type {
+  GraphQLError,
+  GraphQLErrorExtensions,
+} from '../error/GraphQLError';
+import { GraphQLValidationError } from '../error/GraphQLError';
 
 import type {
+  ASTNode,
   DocumentNode,
   FragmentDefinitionNode,
   FragmentSpreadNode,
@@ -35,6 +40,13 @@ interface VariableUsage {
   readonly defaultValue: Maybe<unknown>;
 }
 
+interface ValidationReportOptions {
+  message: string;
+  nodes: ReadonlyArray<ASTNode> | ASTNode;
+  originalError?: Error | undefined;
+  extensions?: GraphQLErrorExtensions | undefined;
+}
+
 /**
  * An instance of this class is passed as the "this" context to all validators,
  * allowing access to commonly useful contextual information from within a
@@ -62,7 +74,19 @@ export class ASTValidationContext {
     return 'ASTValidationContext';
   }
 
+  // TODO: when remove change `onError` to use GraphQLValidationError type instead
+  /* c8 ignore next 4 */
+  /** @deprecated Use `report` instead, will be removed in v18 */
   reportError(error: GraphQLError): void {
+    this._onError(error);
+  }
+
+  report(options: ValidationReportOptions): void {
+    const error = new GraphQLValidationError(options.message, {
+      nodes: options.nodes,
+      originalError: options.originalError,
+      extensions: options.extensions,
+    });
     this._onError(error);
   }
 
--- a/src/validation/__tests__/validation-test.ts
+++ b/src/validation/__tests__/validation-test.ts
@@ -3,8 +3,6 @@ import { describe, it } from 'mocha';
 
 import { expectJSON } from '../../__testUtils__/expectJSON';
 
-import { GraphQLError } from '../../error/GraphQLError';
-
 import type { DirectiveNode } from '../../language/ast';
 import { parse } from '../../language/parser';
 
@@ -76,11 +74,10 @@ describe('Validate: Supports full validation', () => {
       return {
         Directive(node: DirectiveNode) {
           const directiveDef = context.getDirective();
-          const error = new GraphQLError(
-            'Reporting directive: ' + String(directiveDef),
-            { nodes: node },
-          );
-          context.reportError(error);
+          context.report({
+            message: 'Reporting directive: ' + String(directiveDef),
+            nodes: node,
+          });
         },
       };
     }
--- a/src/validation/rules/ExecutableDefinitionsRule.ts
+++ b/src/validation/rules/ExecutableDefinitionsRule.ts
@@ -1,5 +1,3 @@
-import { GraphQLError } from '../../error/GraphQLError';
-
 import { Kind } from '../../language/kinds';
 import { isExecutableDefinitionNode } from '../../language/predicates';
 import type { ASTVisitor } from '../../language/visitor';
@@ -26,11 +24,10 @@ export function ExecutableDefinitionsRule(
             definition.kind === Kind.SCHEMA_EXTENSION
               ? 'schema'
               : '"' + definition.name.value + '"';
-          context.reportError(
-            new GraphQLError(`The ${defName} definition is not executable.`, {
-              nodes: definition,
-            }),
-          );
+          context.report({
+            message: `The ${defName} definition is not executable.`,
+            nodes: definition,
+          });
         }
       }
       return false;
--- a/src/validation/rules/FieldsOnCorrectTypeRule.ts
+++ b/src/validation/rules/FieldsOnCorrectTypeRule.ts
@@ -2,8 +2,6 @@ import { didYouMean } from '../../jsutils/didYouMean';
 import { naturalCompare } from '../../jsutils/naturalCompare';
 import { suggestionList } from '../../jsutils/suggestionList';
 
-import { GraphQLError } from '../../error/GraphQLError';
-
 import type { FieldNode } from '../../language/ast';
 import type { ASTVisitor } from '../../language/visitor';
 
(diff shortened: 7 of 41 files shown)


n1ru4l (commented):
Why is this subclass necessary? What value does it provide over the normal `GraphQLError` class? Can you please update the PR description so people can understand why this change is necessary and what kind of issue it solves? I see a lot of PRs by you without a proper description merged quickly (no reviews from additional people) and I really struggle being able to follow the progress and motivation behind a lot of these changes. I don't think that is healthy for an open-source project that serves as a reference implementation..

IvanGoncharov:
>  I see a lot of PRs by you without a proper description merged quickly (no reviews from additional people) and I really struggle being able to follow the progress and motivation behind a lot of these changes. 

Most of them related to internal scripts, removing deprecated stuff, dropping old dependencies, or "polishing changes" without API change.
For example, I don't know what additional description should PR have if it drops an already deprecated API, switches some internal scripts to TS, or add support for "mutation testing".

But I agree that providing more context for changes related to public API is necessary.
Thanks for reminding me that I updated the PR description.

yaacovCR:
@IvanGoncharov thanks for additional context. I think more information is necessary to determine whether a type hierarchy and its maintenance burden for caught errors is necessary.

I understand from abbreviated review that all errors returned by validate are essentially validation errors. Inasmuch as we should be encouraging sophisticated clients to use the individual life cycle methods including parse, validate, execute, isn't the error type self evident from the stage?

What specific use case is desired here? A local subgraph for Apollo Federation is one possible use case where the subgraph does not make clear the stage at which the pipeline failed. But even that seems far fetched as a gateway could ensure proper syntax and the presence of data signifies that execution has begun.

I would reject this change without a compelling use case. I bet there could be one I'm missing, and I'd love to hear more.

IvanGoncharov:
All implementation of GraphQL server I know provide `formatError`, and client code doesn't have any means to distinguish errors of different types.
For example, envelope see here: https://github.com/dotansimha/envelop/blob/9449ef36f61d4d9561f88a87204b3a16f3c6f98e/packages/core/src/plugins/use-masked-errors.ts#L13

Apollo Server has custom code to handle that by injecting `extensions.code` but this requires recreating errors with a bunch of hacks: https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-errors/src/index.ts
The entire file is dedicated to that purpose and is full of hacks.

I don't think it's worth having this hack in every package implementing pipeline, e.g. `expess-graphql` that also has `formatError` callback.

Also as a bonus now you will have a more meaningful error name (`GraphQLError => GraphQLValidationError`) in console/log without changing anything in your code.

P.S. Currently, I'm working on a spec proposal to standardize "error category" field on errors but even if it is accepted we still need to have a way to set it up without passing additional arg to every constructor call.

IvanGoncharov:
@yaacovCR Also why do we need to force API users to wrap everything step in the pipeline into try/catch if we can have a single try/catch for the entire pipeline.
Making code that interacts with `graphql-js` easier is the good goal by itself.

yaacovCR:
> All implementation of GraphQL server I know provide `formatError`, and client code doesn't have any means to distinguish errors of different types.
> 
> For example, envelope see here: https://github.com/dotansimha/envelop/blob/9449ef36f61d4d9561f88a87204b3a16f3c6f98e/packages/core/src/plugins/use-masked-errors.ts#L13

I am not familiar with envelop internals but looking further down in that file, it looks like the plug-in determines the error type as expected from the lifecycle stage. It looks like formatError can add some additional pan-error formatting but that the plug-in itself is handling the errors as appropriate based on the stage, which is what I would expect.

Maybe @n1ru4l can add some context there if interested.

> Apollo Server has custom code to handle that by injecting `extensions.code` but this requires recreating errors with a bunch of hacks: https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-errors/src/index.ts
> 
> The entire file is dedicated to that purpose and is full of hacks.

I am not familiar with their error hierarchy, but just from looking at some of the custom error types, it seems more involved than just the lifecycle stages, for better or for worse.

You do not link to Apollo code where parse, validate, execute is called, so I cannot say much more about where Apollo Server loses (or do not lose!) the error context.

> I don't think it's worth having this hack in every package implementing pipeline, e.g. `expess-graphql` that also has `formatError` callback.

I think envelop and express-graphql use the formatError callback very differently. I don't think either is a "hack" although I am not sure what you mean by that.

> Also as a bonus now you will have a more meaningful error name (`GraphQLError => GraphQLValidationError`) in console/log without changing anything in your code.

I don't understand why that is necessary. Are errors typically logged without the message? These are not thrown errors...

> P.S. Currently, I'm working on a spec proposal to standardize "error category" field on errors but even if it is accepted we still need to have a way to set it up without passing additional arg to every constructor call.

Certainly if this becomes part of the spec, it makes sense to implement it within the reference implementation as part of the spec rfc process. In general, errors are probably underspecified and deserves attention. But that makes it all the more puzzling why you would choose to implement it first within the reference implementation and only later bring a spec proposal, especially if the two may clash. Or perhaps this simply is the spec rfc PR and I misunderstood the timeline of when you plan to merge?

n1ru4l:
Regarding `envelop`s `useMaskedErrors` plugin, @yaacovCR is correct here. We only specifically wrap the `execute` and `subscribe` functions (and the `buildContext` function, which is not a core part of graphql-js), and only mask those types of errors. `execute` would never raise a `GraphQLValidationError`, so I think this subclassing is not needed at all. You can just rely on the fact that any `GraphQLError` returned from `validate` can be treated as if it was a `GraphQLValidationError`. In general, you would not want to mask validation or parse errors as the GraphQL API consumers would have a real hard time figuring out what they are doing wrong.

I don't see any strong arguments or use-cases that justify introducing this sub-classing.

A bit off-topic, but I would also highly advice against using the `graphql` function over `parse`, `validate`, and `execute`/`subscribe` as it does not allow doing any optimizations such as parse and validation caching, which is a no-go for any production GraphQL server. 🤔 

Another off-topic, since the `validate` and `execute`/`subscribe` function return errors as `{ errors: [GraphQLError] }`, it would make sense to alter the `parse` function to return the error in the same shape. 🤔 Happy to discuss the pros and cons fo this in another issue.

IvanGoncharov:
I propose a simpler solution to this discussion. I plan to present "error categories" (in response) on the next WG. If accepted subclassing errors is a way forward since it's the most ergonomic way to set the error categories.

So after finishing spec PR I will work on a matching graphql-js PR and hopefully will finish it in time before the next WG.

If rejected, I agree that having "graphql-js" specific "error categories" via subclassing doesn't make sense.

> A bit off-topic, but I would also highly advise against using the graphql function over parse, validate, and execute/subscribe as it does not allow doing any optimizations such as parse and validation caching, which is a no-go for any production GraphQL server.

I want to change that, but it's a long-term plan and nothing concrete yet.
Having a pluggable pipeline as part of the reference implementation can be very valuable the entire ecosystem.

yaacovCR:
> I propose a simpler solution to this discussion. I plan to present "error categories" (in response) on the next WG. If accepted subclassing errors is a way forward since it's the most ergonomic way to set the error categories.

Any additional specification on errors is long overdue and kudos on tackling it. Looking forward.

yaacovCR:
> Most of them related to internal scripts, removing deprecated stuff, dropping old dependencies, or "polishing changes" without API change. For example, I don't know what additional description should PR have if it drops an already deprecated API, switches some internal scripts to TS, or add support for "mutation testing".
> 
> But I agree that providing more context for changes related to public API is necessary. Thanks for reminding me that I updated the PR description.

@IvanGoncharov even non-public API changes should have descriptions that are as detailed as practical. `graphql-js` is a community-wide open-source project relied on by a large community with many interested contributors, many of whom would happily take larger roles as maintainers.

Changes that seem obvious to you because of your many years of familiarity with the code base are not necessarily straightforward to all of the interested parties following along and hoping to contribute.

Some more numbered thoughts:
1. Even when simply completing pre-planned deprecation, a PR would be more informative to the community with a sentence or two explaining why the bits were deprecated to begin with.
2. Internal tooling in some ways requires even more description than the public API; the public API usually has a counterpart within the spec, and so is sometimes much better documented. The internal tooling is completely sans documentation. Adding descriptions to PRs helps understand the specific code changes, but more often helps interested parties understand the medium if not long-term overall direction of the code base.
3. Just as an example, take the PR adding support for "mutation testing" with Stryker. This change to me was very puzzling. Although experimenting with mutation testing is very interesting, the PR lacks context for understanding how "mutations testing" will be used. Is there a particular coverage target we are hoping for? Will we be rejecting PRs that change this target? Do everyday contributors need to be aware of this new type of testing? What alternatives to Stryker exist? Is the maintenance burden of the additional dev dependency worth it? A few sentences could have been very explanatory, something like "Adding support for mutations testing; Stryker is a popular choice that is easily to configure, open to alternatives, no specific target in mind at this time, but important to understand how sound our code base is, will work when possible on decreasing the number of surviving mutants, no plans at this time to integrate into PR checks."
4. I would suggest that every PR should be required to undergo review. This will help you in the short run, as it will lead to fewer instances when PRs must be reverted, as occurred recently. It will also provide a forcing function to make sure that the quick couple of explanatory sentences for each PR always make it in. More importantly, it will help you in the long-run by fostering a healthier multi-maintainer setup for graphql-js that decreases your workload, so you can work on the open-source parts that you most enjoy.

yaacovCR:
There is no review on #3596 because reviews of your PRs are not required. In fact, you explicitly reject that above. I as well am busy and I will not signal explicit approval if you are not blocked by it. That’s why forcing functions are important.

I don’t think you should block your team, but I don’t think that means you should onboard another maintainer from Apollo. You have many interested individuals from all sorts of companies and Apollo is already represented.

I read from the above that you want to move fast. In fact, I sympathize! But others want you to move slower so they have time to process and understand your changes. Others want to move fast when it comes to their changes, after all. You want to carefully review them. I suggest a balance where the same process is applied for all maintainers.

yaacovCR:
Explicitly: I do not think we should adopt an experimental period with your suggestions above. I think we should equalize how all PRs are treated, from all members of the community and they should all be allowed to merge to unstable main if they have an approving review and no registered dissent.

Other forcing functions that I would suggest: an issue should be started for every PR: could be even multiple PRs for one “polish” issue so that the intent of PRs over the medium term can be tracked.

a roadmap  file should be introduced and issues/PRs should be required to correspond to items in the roadmap so long term planning could be clear.

i think you may find it frustrating considering your expertise to introduce these checks because it will slow your ability to make changes. But consider the broader needs of the graphql community, which would benefit from a broader cross section of institutional representation. Also consider the fact that actual public api and hot fixes account for a minority of your changes, such that the community will in my opinion scarcely be slowed, and the fact that adopting balanced consistent standards for all contributors will make this GitHub repository a more welcoming place for open source development which will ensure the health of the community over the long term and its resilience to scary shocks (!!!!!!!!!) to individual personnel.

graphql is a very special open source technology and much of that is because of your own diligent and excellent work! I thank you personally for it. I make the above suggestions because despite the limitation on your personal velocity, I think it will be in the community’s and your own long term interests. It is worth sacrificing some of your own productivity to make the overall project environment healthier.

IvanGoncharov:
> Explicitly: I do not think we should adopt an experimental period with your suggestions above. I think we should equalize how all PRs are treated, from all members of the community and they should all be allowed to merge to unstable main if they have an approving review and no registered dissent.

Ok, it was measured I was willing to do as a compromise.
I will implement your and @n1ru4l into suggestions in my workflow.
A mandatory and detailed description of PRs and giving the community time to feedback on them.

> graphql is a very special open source technology and much of that is because of your own diligent and excellent work! I thank you personally for it. I make the above suggestions because despite the limitation on your personal velocity, I think it will be in the community’s and your own long term interests. It is worth sacrificing some of your own productivity to make the overall project environment healthier.

I don't understand why "improving community/onboarding" should first be done on the most technically challenging project under the GraphQL Foundation.
The guild has full control over https://github.com/graphql/graphql.github.io (nothing get done except for merging update to yoga docs owned by the guild) and https://github.com/graphql/dataloader (no review procedure on every commit being merged and no updates in a month).
There is a bunch of work the long overdue on both projects.
And those projects can be used as a good testing ground for every proposal you made.

In a previous discussion we had, I was named the "bottleneck" but the guild and was pressured to onboard guild maintainers to review the group (with the full merge rights) which I did: https://youtu.be/S-VoLuwW1sM?t=3003

Now it looks like it situation is repeating and I don't want to go through another hour of pressure.

Suggesting stuff is easy but implementing them is hard.
So if you truly want to help the entire GraphQL community I suggest you implement those suggestions that you made on https://github.com/graphql/graphql.github.io to solve the "bottleneck" there since Guild has full control and all technical knowledge there (the whole site was rewritten by the guild).

I think that leading by example, is an instrument here.
So before spending my time on a proposal that looked good on the surface I want to see it working on a similar project but on a smaller scale. And I see https://github.com/graphql/graphql.github.io as an excellent opportunity to do that with similar onboarding and a way lower price of doing mistakes.

That said I listen to your @n1ru4l feedback about the process and suggest real steps based on them. I just don't want "change everything" in one go and evolve our processes over time.

yaacovCR:
@IvanGoncharov should we make this one a draft as well?

yaacovCR:
Closing this old PR. Feel free to reopen!

PR #3596, opened 2022-05-18
benchmark: test performance cost of re-creating schema

IvanGoncharov:
Nessary as part of a current and possible future disscussion on perfomance penalties of schema
transformation.

### DIFF
--- a/benchmark/GraphQLSchema-benchmark.js
+++ b/benchmark/GraphQLSchema-benchmark.js
@@ -1,0 +1,15 @@
+import { GraphQLSchema } from 'graphql/type/schema.js';
+import { buildClientSchema } from 'graphql/utilities/buildClientSchema.js';
+
+import { bigSchemaIntrospectionResult } from './fixtures.js';
+
+const bigSchema = buildClientSchema(bigSchemaIntrospectionResult.data);
+
+export const benchmark = {
+  name: 'Recreate a GraphQLSchema',
+  count: 40,
+  measure() {
+    // eslint-disable-next-line no-new
+    new GraphQLSchema(bigSchema.toConfig());
+  },
+};


yaacovCR:
@github-actions run-benchmark

yaacovCR:
Hmmm. Am I write in reading this as about 450 ops/s = about 2.2 ms to recreate a schema? One note is that this is additive if a pipeline does this repeatedly.

Note that at the last WG, we discussed not just simple schema recreation, but also potentially "rewiring" it, i.e. recreating all types that reference changed types. In `graphql-tools`, `mapSchema`  changes only the requested types, and then calls `rewireTypes` which unfortunately recreates ALL schema types. This ensures that all fields on all types are pointing to the new types, but it may often be overkill.

Within `graphql-js`, the analogue would be `lexicographicallySortSchema` which does the same thing, I believe.

IvanGoncharov:
Point of this benchmark is to prove that you can attach configuration to GraphQLSchema and force users to re-create schema if they want to change config.
So it's related to `getFieldDef' discussion, discussion about new flag to disable introspection.
And few other related proposal I have, it's not directly related to assigning resolvers.

> Within graphql-js, the analogue would be lexicographicallySortSchema which does the same thing, I believe.

`lexicographicallySortSchema` doing sorting which computation intensive on itself and scales non-linear with number of items so comporation with simply assigning properties is not fair.
Also it sort bunch of things like "input fields", args, etc. that don't have resolvers.
Even under this condition `lexicographicallySortSchema` takes 10ms on my machine (probably 20ms on GitHub's worker).
So I think properly implemented resolver attaching function will take 5ms tops on a huge schema.

> Hmmm. Am I write in reading this as about 450 ops/s = about 2.2 ms to recreate a schema? One note is that this is additive if a pipeline does this repeatedly.

Yes, it's additive but you can use caching (ES6 Map of transformation) if you want to do it per request.
Also note it's worse case performance since it run on default configuration of GitHub CI inside docker (probably inside VM). On my machine it's stable `970 ops/s` which is ±1 ms.

Also please note it's done for absolutely massive GitHub schema which is way way bigger that most users will ever have.

IvanGoncharov:
> We are optimizing for cold starts for lambda-like environments where caching is not possible.

So if it's a per-request transformation (e.g. different schema based on headers, URL, etc.) when it can be cached. 

If transformation is done before starting the server, I think 2.2ms on a weak CI runner for GitHub size schema is not a big price to pay.

yaacovCR:
I guess it's important to keep in mind the baseline cost of a cold start for the execution environment to spin up.

https://mikhail.io/serverless/coldstarts/aws/

If it's usually 400 ms, then an extra 2 ms could represent some huge increase with respect to the user code, but still less than a 0.5% increase compared to the overall startup delay, not so important of an optimization after all!

yaacovCR:
Whoops. Other solutions measure cold starts in ms time. https://blog.cloudflare.com/eliminating-cold-starts-with-cloudflare-workers/ Sorry for the multiple comments.

PR #3597, opened 2022-05-19
Fix typing for `isNonNullType`

yaacovCR:
= If a type is a non-null type, then the wrapped type must be nullable.

= Introduces `GraphQLNullableInputType` and `GraphQLNullableOutputType` to properly represent the result when `isNonNullType` is called on parameters of type `GraphQLInputType` or `GraphQLOutputType`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -147,6 +147,8 @@ export type {
   GraphQLAbstractType,
   GraphQLWrappingType,
   GraphQLNullableType,
+  GraphQLNullableInputType,
+  GraphQLNullableOutputType,
   GraphQLNamedType,
   GraphQLNamedInputType,
   GraphQLNamedOutputType,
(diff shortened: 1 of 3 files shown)


yaacovCR:
My guess is that this qualifies as a "bug fix" and can go into a patch release.

IvanGoncharov:
Solid improvement.
Some quirks and limitations like that are mostly legacy of Flow deficiencies.

> My guess is that this qualifies as a "bug fix" and can go into a patch release.

Yes, but the patch releases don't make sense in alpha :)

IvanGoncharov (approved):
My guess is that this qualifies as a "bug fix" and can go into a patch release.

src/index.ts
```diff
@@ -147,6 +147,8 @@ export type {
   GraphQLAbstractType,
   GraphQLWrappingType,
   GraphQLNullableType,
+  GraphQLNullableInputType,
+  GraphQLNullableOutputType,
```
IvanGoncharov, line 151: Adds a new type so it's a "feature".
Please change the label.

Anything that adds new public import is a "feature".
and require new minor version, see: https://semver.org/#spec-item-7

Please change the label.

PR #3598, opened 2022-05-19
Refactor `GraphQL*Type` TS types to be more DRY

yaacovCR:
Motivation:
1. Makes it easier to understand the type hierarchy.
2. Adding a new type, were it ever to happen, would require fewer code changes.

Depends on #3597

### DIFF
--- a/src/type/definition.ts
+++ b/src/type/definition.ts
@@ -51,22 +51,9 @@ import type { GraphQLSchema } from './schema';
  * These are all of the possible kinds of types.
  */
 export type GraphQLType =
-  | GraphQLScalarType
-  | GraphQLObjectType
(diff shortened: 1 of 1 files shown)


IvanGoncharov (approved):
> Adding a new type, were it ever to happen, would require fewer code changes.

I would say chances are close to zero. But who knows.

Overall I think it's an improvement.

P.S. "stacked" PRs are not mandatory if the overall size of PR is small like here. Just create separate commits and merge PR with "merge and rebase". One caveat, if you do that please don't do fix as separate commits (e.g. `add tests`, `review fixes`, `improve coverage`) please update commits directly using `git rebase -i`

Issue #3519, opened 2022-03-23
consider not deprecating typeInfo and getFieldDef options to validate function and TypeInfo constructor


yaacovCR:
https://github.com/graphql/graphql-js/blob/e1726dfea66979bfe7ad1c0b0834613e4b6ce4b4/src/validation/validate.ts#L44-L45
https://github.com/graphql/graphql-js/blob/e1726dfea66979bfe7ad1c0b0834613e4b6ce4b4/src/utilities/TypeInfo.ts#L66-L67

From what I understand, by preserving these options, we add the ability for a server to add custom meta fields such as `__fulfilled` or `__directive` and still validate properly.

yaacovCR:
https://github.com/graphql/graphql-js/blob/e1726dfea66979bfe7ad1c0b0834613e4b6ce4b4/src/utilities/TypeInfo.ts#L311-L322

The default function here is what explicitly allows the meta fields included within the spec. Adding a custom version of this function would allow for custom behavior.

yaacovCR:
Alternatively, we could allow passing a list of pairs/tuples of check functions and meta fields instead of an entire custom function.

yaacovCR:
Related: https://github.com/yaacovCR/graphql-executor/commit/00f54c0df07ff6317c640077669a339da250fe35

yaacovCR:
@IvanGoncharov i believe #3575 completes some of the above planned deprecation

While expected, I think that’s unfortunate, for the reasons outlined above.

Could you comment here when you get a chance? Maybe there are alternatives I have not considered?

IvanGoncharov:
@yaacovCR I'm working on moving this function into `GraphQLSchema.
It would allow you to subclass `GraphQLSchema` and override this function.
Would this work for you?

yaacovCR:
Maybe, but why is that better? It would mean schemas have to be customized to add custom meta fields which by definition exist outside the schema and could be executor specific.

Could this also be restored until there is a replacement?

yaacovCR:
Maybe if there is not consensus we can discuss further at next js wg?

saihaj:
The PRs that are made there is little to no context on why things are deprecated. PRs have no description for why we are making changes. We should be documenting all the changes we make so it is important that we first create an issue for broader discussion and then use that as an item in the JS WG.

yaacovCR:
Some thoughts on this issue.

[...]

PR #3599, opened 2022-05-19
Allow interface resolveType functions to resolve to child interfaces

yaacovCR:
= The child interfaces must eventually resolve to a runtime object type.

implements: #3253

### DIFF
--- a/src/execution/__tests__/abstract-test.ts
+++ b/src/execution/__tests__/abstract-test.ts
@@ -669,6 +669,10 @@ describe('Execute: Handles execution of abstract types', () => {
         name: String
         isFriendly: Boolean
       }
+
+      type Person implements Named {
+        name: String
+      }
     `);
 
     const document = parse(`
@@ -706,6 +710,13 @@ describe('Execute: Handles execution of abstract types', () => {
     );
 
     const petType = assertInterfaceType(schema.getType('Pet'));
+    // FIXME: workaround since we can't inject resolveType into SDL
+    namedType.resolveType = () => 'Pet';
+    petType.resolveType = () => 'Person';
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Runtime Object type "Person" is not a possible type for encountered abstract type "Pet".',
+    );
+
     // FIXME: workaround since we can't inject resolveType into SDL
     namedType.resolveType = () => 'Pet';
     petType.resolveType = () => undefined;
--- a/src/execution/__tests__/abstract-test.ts
+++ b/src/execution/__tests__/abstract-test.ts
@@ -271,7 +271,7 @@ describe('Execute: Handles execution of abstract types', () => {
       errors: [
         {
           message:
-            'Abstract type "Pet" must resolve to an Object type at runtime for field "Query.pet". Either the "Pet" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.',
+            'Abstract type resolution for "Pet" for field "Query.pet" failed. Encountered abstract type "Pet" must resolve to an Object or Interface type at runtime. Either the "Pet" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.',
           locations: [{ line: 3, column: 9 }],
           path: ['pet'],
         },
@@ -610,26 +610,26 @@ describe('Execute: Handles execution of abstract types', () => {
     }
 
     expectError({ forTypeName: undefined }).toEqual(
-      'Abstract type "Pet" must resolve to an Object type at runtime for field "Query.pet". Either the "Pet" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.',
+      'Abstract type resolution for "Pet" for field "Query.pet" failed. Encountered abstract type "Pet" must resolve to an Object or Interface type at runtime. Either the "Pet" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.',
     );
 
     expectError({ forTypeName: 'Human' }).toEqual(
-      'Abstract type "Pet" was resolved to a type "Human" that does not exist inside the schema.',
+      'Abstract type resolution for "Pet" for field "Query.pet" failed. Encountered abstract type "Pet" was resolved to a type "Human" that does not exist inside the schema.',
     );
 
     expectError({ forTypeName: 'String' }).toEqual(
-      'Abstract type "Pet" was resolved to a non-object type "String".',
+      'Abstract type resolution for "Pet" for field "Query.pet" failed. Encountered abstract type "Pet" was resolved to a non-object type "String".',
     );
 
     expectError({ forTypeName: '__Schema' }).toEqual(
-      'Runtime Object type "__Schema" is not a possible type for "Pet".',
+      'Abstract type resolution for "Pet" for field "Query.pet" failed. Runtime Object type "__Schema" is not a possible type for encountered abstract type "Pet".',
     );
 
     // FIXME: workaround since we can't inject resolveType into SDL
     // @ts-expect-error
     assertInterfaceType(schema.getType('Pet')).resolveType = () => [];
     expectError({ forTypeName: undefined }).toEqual(
-      'Abstract type "Pet" must resolve to an Object type at runtime for field "Query.pet" with value { __typename: undefined }, received "[]".',
+      'Abstract type resolution for "Pet" for field "Query.pet" with value { __typename: undefined } failed. Encountered abstract type "Pet" must resolve to an Object or Interface type at runtime, received "[]".',
     );
 
     // FIXME: workaround since we can't inject resolveType into SDL
@@ -638,6 +638,108 @@ describe('Execute: Handles execution of abstract types', () => {
       schema.getType('Cat');
     expectError({ forTypeName: undefined }).toEqual(
       'Support for returning GraphQLObjectType from resolveType was removed in graphql-js@16.0.0 please return type name instead.',
+    );
+  });
+
+  it('hierarchical resolveType with Interfaces yields useful error', () => {
+    const schema = buildSchema(`
+      type Query {
+        named: Named
+      }
+
+      interface Named {
+        name: String
+      }
+
+      interface Animal {
+        isFriendly: Boolean
+      }
+
+      interface Pet implements Animal & Named {
+        name: String
+        isFriendly: Boolean
+      }
+
+      type Cat implements Pet & Named & Animal {
+        name: String
+        isFriendly: Boolean
+      }
+
+      type Dog implements Pet & Animal & Named {
+        name: String
+        isFriendly: Boolean
+      }
+    `);
+
+    const document = parse(`
+      {
+        named {
+          name
+        }
+      }
+    `);
+
+    function expectError() {
+      const rootValue = { named: {} };
+      const result = executeSync({ schema, document, rootValue });
+      return {
+        toEqual(message: string) {
+          expectJSON(result).toDeepEqual({
+            data: { named: null },
+            errors: [
+              {
+                message,
+                locations: [{ line: 3, column: 9 }],
+                path: ['named'],
+              },
+            ],
+          });
+        },
+      };
+    }
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Named')).resolveType = () => 'Animal';
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Interface type "Animal" is not a subtype of encountered interface type "Named".',
+    );
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Named')).resolveType = () => 'Pet';
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => undefined;
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Encountered abstract type "Pet" must resolve to an Object or Interface type at runtime. Either the "Pet" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.',
+    );
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => 'Human';
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Encountered abstract type "Pet" was resolved to a type "Human" that does not exist inside the schema.',
+    );
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => 'String';
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Encountered abstract type "Pet" was resolved to a non-object type "String".',
+    );
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => '__Schema';
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Runtime Object type "__Schema" is not a possible type for encountered abstract type "Pet".',
+    );
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    // @ts-expect-error
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => [];
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" with value {} failed. Encountered abstract type "Pet" must resolve to an Object or Interface type at runtime, received "[]".',
+    );
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => 'Pet';
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Encountered abstract type "Pet" resolved to "Pet", causing a cycle.',
     );
   });
 });
--- a/src/execution/__tests__/abstract-test.ts
+++ b/src/execution/__tests__/abstract-test.ts
@@ -698,46 +698,48 @@ describe('Execute: Handles execution of abstract types', () => {
       };
     }
 
+    const namedType = assertInterfaceType(schema.getType('Named'));
     // FIXME: workaround since we can't inject resolveType into SDL
-    assertInterfaceType(schema.getType('Named')).resolveType = () => 'Animal';
+    namedType.resolveType = () => 'Animal';
     expectError().toEqual(
       'Abstract type resolution for "Named" for field "Query.named" failed. Interface type "Animal" is not a subtype of encountered interface type "Named".',
     );
 
+    const petType = assertInterfaceType(schema.getType('Pet'));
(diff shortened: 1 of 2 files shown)


src/execution/execute.ts
```diff
@@ -826,32 +825,73 @@ function completeAbstractValue(
...
-    ),
+    runtimeType,
     fieldNodes,
     info,
     path,
     result,
   );
 }
 
+function resolveType(
```
saihaj, line 836: I would suggest making it one argument that is an object

yaacovCR: Hmmm, I think most of the code in this file does not have named arguments, see for example nearby `ensureValidRuntimeType` and many others.

The public API has been moving in the direction of named parameters, probably for good reason. The functions in this file by contrast are not exported and only used a few times, so the benefits of named parameters that leap to mind (any order, can save a set of arguments and pass them around) do not seem as straightforward.

But I defer to the group, thanks for the eyeballs.

yaacovCR: Actually, I think you're right, in this case we have a recursive function that is passing the exact same set of arguments save one, they should be combined.

yaacovCR: Hmmm, I'm looking at it, and it seems like similar refactoring could occur in so many places. I'm going to leave this as is for now.

src/execution/execute.ts
```diff
@@ -826,32 +825,73 @@ function completeAbstractValue(
...
+    returnType,
+    abstractType,
+    fieldNodes,
+    info,
+    result,
+    encounteredTypeNames,
+  );
+}
+
 function ensureValidRuntimeType(
```
IvanGoncharov, line 880: this function is called "ensure" (and previously it just validated type) but now actually recursively resolves type. Please change its name?

src/execution/execute.ts
```diff
@@ -826,32 +825,73 @@ function completeAbstractValue(
...
+  const contextValue = exeContext.contextValue;
+  const possibleRuntimeType = resolveTypeFn(
+    result,
+    contextValue,
+    info,
+    abstractType,
+  );
+
+  if (isPromise(possibleRuntimeType)) {
+    return possibleRuntimeType.then((resolvedPossibleRuntimeType) =>
```
IvanGoncharov, line 855: Do you really need to add `resolveType` function to implement this feature or it's just refactoring mixed in?

The problem here is that adding one more `then` looks unnecessary.
Every `then` adds a trip to the microtasks queue even if the promise is already resolved and that significantly affects performance.
Because of that code, some time can look repetitive but it is done on purpose to minimize the number of `then` and improve performance.

IvanGoncharov (changes requested):

src/execution/execute.ts
```diff
@@ -866,29 +906,62 @@ function ensureValidRuntimeType(
...
+      `Abstract type resolution for "${returnType.name}" for field "${info.parentType.name}.${info.fieldName}" failed. ` +
+        `Encountered abstract type "${abstractType.name}" was resolved to a non-object type "${runtimeTypeName}".`,
       { nodes: fieldNodes },
     );
   }
 
   if (!exeContext.schema.isSubType(returnType, runtimeType)) {
     throw new GraphQLError(
-      `Runtime Object type "${runtimeType.name}" is not a possible type for "${returnType.name}".`,
+      `Abstract type resolution for "${returnType.name}" for field "${info.parentType.name}.${info.fieldName}" failed. ` +
```
IvanGoncharov, line 963: You still need to check if the intermediate interface is a sub-type of `returnType` and currently you bypassing this check.
Also please add a test for it.

IvanGoncharov (changes requested):

src/execution/execute.ts
yaacovCR: Sure => what do you suggest? I would like something that implies ensure + recursion as necessary. I guess I thought ensure was good enough, but if you have some candidates that you would prefer, that would be fine.

src/execution/execute.ts
yaacovCR: @IvanGoncharov, I totally agree with you, but that leaves me a bit puzzled.

This code actually is repetitive; I switch on whether the type resolver for the given abstract type (or the default type resolver) returned a promise, and only call `.then` when it is a promise. For exactly the reason you mention above, to not use then unnecessarily.

For that reason, I did need the `resolveType` function -- or, at least I thought I did.

Maybe you could be more specific about where you think the extra trip to the microtask queue is happening?

src/execution/execute.ts
yaacovCR: I gave it a go => I was more explicit that we start off with the type name and changed the function to `deriveRuntimeType`.

6c63b4f2

src/execution/__tests__/abstract-test.ts
```diff
@@ -271,7 +271,7 @@ describe('Execute: Handles execution of abstract types', () => {
       errors: [
         {
           message:
-            'Abstract type "Pet" must resolve to an Object type at runtime for field "Query.pet". Either the "Pet" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.',
+            'Abstract type resolution for "Pet" for field "Query.pet" failed. Encountered abstract type "Pet" must resolve to an Object or Interface type at runtime. Either the "Pet" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.',
```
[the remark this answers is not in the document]
yaacovCR: The error messages have to be changed (in my opinion) because there is now a multi-step process for deriving the runtime type, and we want to be as helpful as possible by pointing out what type we started with and where we failed. It's not a separate change => it's just part and parcel of what we're doing here, it wouldn't make sense on its own, it would just be repetitive unless we have the multi-step process.

src/execution/execute.ts
yaacovCR: There is a test a few lines above for the intermediate interface.

How is the test being bypassed?

src/execution/__tests__/abstract-test.ts
```diff
@@ -640,4 +640,106 @@ describe('Execute: Handles execution of abstract types', () => {
...
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => '__Schema';
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Runtime Object type "__Schema" is not a possible type for encountered abstract type "Pet".',
+    );
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    // @ts-expect-error
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => [];
```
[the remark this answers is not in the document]
yaacovCR: This is a child interface rather than the initial abstract type, even though I know the same function is being used, I am not making assumptions, and adding a test for that as well.

src/execution/__tests__/abstract-test.ts
```diff
@@ -640,4 +640,106 @@ describe('Execute: Handles execution of abstract types', () => {
...
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Named')).resolveType = () => 'Animal';
+    expectError().toEqual(
+      'Abstract type resolution for "Named" for field "Query.named" failed. Interface type "Animal" is not a subtype of encountered interface type "Named".',
+    );
+
+    // FIXME: workaround since we can't inject resolveType into SDL
+    assertInterfaceType(schema.getType('Named')).resolveType = () => 'Pet';
+    assertInterfaceType(schema.getType('Pet')).resolveType = () => undefined;
```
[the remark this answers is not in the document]
yaacovCR: sure thing  46ae4b19

src/execution/execute.ts
yaacovCR: ah, you meant that i am checking against `returnType` but not the current abstract type. fixed, new test added.

We should get an error if we go from Named => Pet => Person, even though Person is Named, it's not a Pet.

df8f15c7

yaacovCR: 7ceffbe8

yaacovCR: 7ceffbe8

yaacovCR: 1e2988e

src/execution/execute.ts
yaacovCR: feel free to directly modify this PR if you have something else in mind.

src/execution/__tests__/abstract-test.ts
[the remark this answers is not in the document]
IvanGoncharov: In that case, please add the same test for intermediate interfaces.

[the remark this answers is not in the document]
IvanGoncharov: got confused, now I see `Pet` is intermediate interface
👍

src/execution/execute.ts
```diff
@@ -866,29 +906,62 @@ function ensureValidRuntimeType(
...
+      `Abstract type resolution for "${returnType.name}" for field "${info.parentType.name}.${info.fieldName}" ` +
+        `with value ${inspect(result)} failed. ` +
+        `Encountered abstract type "${currentAbstractType.name}" must resolve to an Object or Interface type at runtime, ` +
+        `received "${inspect(runtimeTypeName)}".`,
+    );
+  }
+
+  if (encounteredTypeNames.has(runtimeTypeName)) {
+    throw new GraphQLError(
+      `Abstract type resolution for "${returnType.name}" for field "${info.parentType.name}.${info.fieldName}" failed. ` +
```
IvanGoncharov, line 918: We can't have cycles, they were possible only because previously you didn't do a subtype check (previous review comment).
Schema specifically validated to not have interface cycles, see schema validation code.
I already fixed this as part of previous comment will push on top of your PR.

IvanGoncharov (dismissed):

src/execution/execute.ts
IvanGoncharov: Pushed, please review.

yaacovCR: Some schemas are not validated i.e. "assumeValid" is true, and so I imagined we might still need the cycle check. I could be off-base.

yaacovCR:
hi @IvanGoncharov just checking on the status here, no rush, just curious!

src/execution/execute.ts
yaacovCR: summary of offline conversation:

> I am checking if we have a promise appropriately at each resolveType call, but because I refactored a recursive function that returned a PromiseOrValue, when there is no extra recursive steps, I ended up calling `then` twice: once for the only call to the possibly async type resolver, but also once on the entire recursive function.
>
> The recursion can be refactored (actually much closer to the original) so that if there is no need for recursion, the recursive part is not called.

src/execution/execute.ts
yaacovCR: i added a test for an unvalidated schema to show why this is still necessary.

yaacovCR:
This is ready for re-review, thanks.

src/execution/execute.ts
IvanGoncharov: As the comment says it's up to a person who sets the `assumeValid` flag is a guarantee that schema is valid.
https://github.com/graphql/graphql-js/blob/7532a5c8eb4001b726a2b45c8760bb71a6e10f2e/src/type/schema.ts#L416-L418

We can't assume something is broken when a person explicitly states it's valid.
Especially in performance-sensitive code.

src/execution/__tests__/abstract-test.ts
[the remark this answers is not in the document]
IvanGoncharov: > and we want to be as helpful as possible by pointing out what type we started with

Please note that this is not a **client** error, the client can't trigger that error unless there is a bug in type resolvers. So for a client having this info is useless and it should matter only for people working on a schema.

Moreover, for clients, this detail is already accessible through the `path` and `locations`.
GraphiQL will highlight the field you get an error on and you will see its type in a tooltip.

I'm not sure having it explicitly in the message just for schema developers justifies an additional complexity of passing additional arguments just for error reporting.

In general, type resolvers should be independent of each other same as field resolvers. So it shouldn't matter what was previous type resolver since it's doesn't affect the resolver's argument.
If the schema author really wants to debug it when he can easily use the available type name and field name to quickly find initial type in a chain.

src/execution/execute.ts
yaacovCR: I am basing off how unvalidated fragment cycles are not allowed to infinite loop.

src/execution/__tests__/abstract-test.ts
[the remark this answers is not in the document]
yaacovCR: From my survey of the repo and at GraphQL-Tools, there are many schema developers out there who get confused by cryptic error messages from this (and many other!) packages. I am in favor of keeping more details in the error message, perhaps I can refactor per @saihaj  suggestion to just pass around one object that bundles some of the constant parameters?

yaacovCR:
Hey @IvanGoncharov  looks like we both solved the extra then but I overwrote your changes with mine. I am away from computer until about 15 hours from now, sorry about that and feel free to restore yours. Thanks for your help advancing this

yaacovCR:
> Hey @IvanGoncharov looks like we both solved the extra then but I overwrote your changes with mine. I am away from computer until about 15 hours from now, sorry about that and feel free to restore yours. Thanks for your help advancing this

I was able to fix before work yay

src/execution/execute.ts
```diff
@@ -795,63 +797,63 @@ function completeLeafValue(
...
     fieldNodes,
     info,
     path,
     result,
   );
 }
 
-function ensureValidRuntimeType(
-  runtimeTypeName: unknown,
+function resolveRuntimeTypeName(
```
yaacovCR, line 827: This function now essentially completes, suggest renaming to completeAbstractValueImpl or something else.

yaacovCR:
good news! i rebased on main and after #3624 , the report uploads correctly!

src/execution/__tests__/abstract-test.ts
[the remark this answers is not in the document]
yaacovCR: Actually, I think it just makes sense to go with the shorter messages. We can always change it later.

src/execution/execute.ts
yaacovCR: i pushed a change renaming it

IvanGoncharov (changes requested):
This PR still has some unfinished discussions.
But I discovered one major roadblock, it requires spec PR, because otherwise, it contradicts spec:
https://spec.graphql.org/draft/#sec-Value-Completion.Resolving-Abstract-Types

I suggest opening spec PR that changes the algorithm to be recursive.
\+ I will vote for fast-track it to stage2 if you present it on the next WG.

yaacovCR:
Two things:

### First, what other discussions are pending?

The two pending issues I had seen were the content of the error messages and whether cycles should be allowed, and I have changed this PR to correspond to your suggestions. Are you not sure of the optimal behavior with respect to one or both of these issues? Do you feel that you need a certain amount of time to think about these issues, or do these issues need feedback from the graphql-js-wg or the main graphql-wg? I will address your comment in a second, but we can work on these issues in parallel, as although async discussions have their drawback, they do allow us to open up multiple threads and make progress in parallel.

### Second, I think this PR does not contradict spec, as the spec says we can resolve by whatever means we want.

In particular, consider the evolution of this PR. A previous version had a separate resolve method corresponding to the ResolveAbstractType step you have linked. But:
1. that method returned a value or promise, and
2. refactoring any functionality returning a promise or value into a separate function forces the calling function to perform an additional isPromise check and `.then` call
3. `graphql-js` has the aspirational policy of not only not entering the task queue unnecessarily, but of never adding an additional job to job queue (see https://blog.greenroots.info/task-queue-and-job-queue-deep-dive-into-javascript-event-loop-model and https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide/In_depth#run_javascript_run)

And so we had to mix the separate `resolveAbstractType` method with abstract value completion using recursive `completeAbstractValue` and `completeAbstractValueImpl` inner functions and an incomplete `resolveAbstractType` portion linking between the two.

So, this change does not contradict the spec, but the performance (?) constraints of `graphql-js` have causes us to make a change that makes the one-to-one match between the implementation and the spec more difficult to grasp.

[Perhaps there are 3 tenets of `graphql-js`, [not 2!](https://github.com/graphql/graphql-js/issues/3519#issuecomment-1129850480): (1) a reference implementation, (2) a production-ready JavaScript graphql pipeline, (3) a fully-fleshed out exercise in avoiding any extra jobs in the job queue. And perhaps there is some sort of inchoate rule where for any given issue, only two of the three `graphql-js` tenets may be upheld!]

Bottom line, I don't think we should necessarily force all complying implementations to implement this feature in JavaScript. I think it's a useful feature, but I don't know that in every language it should be mandated.

However, I am fine:

1. working on a spec PR
2. presenting it at the next WG
3. adding that my own opinion is that we should implement this in `graphql-js` without adopting the spec PR, and then
4. accepting whatever happens

Although please see above -- let's work out the other issues in the meantime. I am leaving this as open, please follow-up when possible.

yaacovCR:
See:
https://github.com/graphql/graphql-spec/pull/960
https://github.com/graphql/graphql-wg/pull/1016

src/execution/execute.ts
```diff
@@ -790,106 +791,128 @@ function completeLeafValue(
...
+  if (possibleRuntimeType == null) {
     throw new GraphQLError(
-      `Abstract type "${returnType.name}" was resolved to a type "${runtimeTypeName}" that does not exist inside the schema.`,
+      `Abstract type "${abstractType.name}" was resolved to a type "${possibleRuntimeTypeName}" that does not exist inside the schema.`,
       { nodes: fieldNodes },
     );
   }
 
-  if (!isObjectType(runtimeType)) {
+  if (isInterfaceType(possibleRuntimeType)) {
```
leebyron, line 877: what about union types?

src/execution/execute.ts
```diff
@@ -790,106 +791,128 @@ function completeLeafValue(
...
+        path,
+        result,
+      ),
+    );
+  }
+
+  if (possibleRuntimeTypeName == null) {
     throw new GraphQLError(
-      `Abstract type "${returnType.name}" must resolve to an Object type at runtime for field "${info.parentType.name}.${info.fieldName}". Either the "${returnType.name}" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.`,
+      `Abstract type "${abstractType.name}" must resolve to an Object type or an intermediate Interface type at runtime for field "${info.parentType.name}.${info.fieldName}". Either the "${abstractType.name}" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.`,
```
leebyron, line 845: What do you think about only altering this error message in the case that an intermediate Interface type actually exists as a potential case?

Looking at the unit test impact, my sense is the primary downside of this change is the additional complication to this error message that makes it slightly harder to understand.

Ideally for the large majority common case where there is no intermediate interface, we don't burden the error message

src/execution/execute.ts
yaacovCR: Yes, if unions ever implement interfaces, they could potentially be a link in the chain!

src/execution/execute.ts
yaacovCR: 👍  Tailoring the error message to the interface at hand seems pretty reasonable.

yaacovCR:
Closing this, can be used as base if anyone wants to champion the associated RFC

Issue #3539, opened 2022-04-21
scalars with access to `context` re #2663


iambumblehead:
I'd like my graphql service to return internationalized error messages for invalid scalars

Following the example from this related ticket https://github.com/graphql/graphql-js/issues/2663, I'd like the service to do something like this,
```javascript
  parseValue: (value, context) => {
    const lang = context.get( 'accept-language' ); // somehow access express/koa language header
    const rounded = round(value);
    if (value !== rounded) {
      // throw new GraphQLError("Please provide a rounded number");
      throw new GraphQLError( i18nMessage( lang, 'please_provide_rounded_number' ) );
    }
    return value;
  },
```
Is there a known way to achieve the result described above? Because the scalar type does not have access to the context, I can't imagine how to achieve this.

iambumblehead:
related https://github.com/apollographql/apollo-server/issues/4010

apollo documentation vaguely recommendation a mixture of plugins and hooks that are ineffective and require a lot of boilerplate code

saihaj:
Yeah we don't really have a way to pass context to type system. You could try something with [`envelop`](http://envelop.dev) and hijack the results on after hooks in lifecycle methods

iambumblehead:
context is widely needed and perhaps it should be passed through all parts of the response lifecycle, to be shared with any possible userland hooks that would need to modify or customize anything.

IvanGoncharov:
@iambumblehead Thanks for providing the code sample, it makes super clear what you want to achieve.
Sadly scalars are used in many situations where context is not available (e.g. default values in SDL, introspection, etc.).

You can solve your use case by using `formatError` (or similarly named) callback provided by your framework and `extensions` field. 

```js
  parseValue: (value, context) => {
    const rounded = round(value);
    if (value !== rounded) {
      throw new GraphQLError({
        message: "Please provide a rounded number",
        extensions: { i18nMessage: 'please_provide_rounded_number' },
      });
    }
    return value;
  },
```

For the server part, I will use https://github.com/graphql/express-graphql for this example, but almost every other popular GraphQL server allows the same.
```js
app.use(
  '/graphql',
  graphqlHTTP(async (request, response, graphQLParams) => ({
    schema: MyGraphQLSchema,
    // ...
    customFormatErrorFn(error) {
      const formattedError = error.toJSON();
      
      const { i18nMessage } = error.extensions;
      if (i18nMessage != null) {
        const lang = request.headers['accept-language'];
[...]

PR #3600, opened 2022-05-19
Allow scalar `parse*` and `serialized` methods to access context

yaacovCR:
See #3539 for more details.

Even when provided, `context` must be optional because:
1. Default values for scalar are parsed when building a schema, outside of the execution context.
2. Default values for scalar are serialized during findBreakingChanges and other schema operations, also be outside of the execution context.

BREAKING CHANGES are included for argument lists of several functions to include the optional context parameter:
1. `getVariableValues`
2. `getArgumentValues`
3. `coerceInputValue`
4. `valueFromAST`
5. `astFromValue`

### DIFF
--- a/src/execution/__tests__/executor-test.ts
+++ b/src/execution/__tests__/executor-test.ts
@@ -3,6 +3,7 @@ import { describe, it } from 'mocha';
 
 import { expectJSON } from '../../__testUtils__/expectJSON';
 
+import { identityFunc } from '../../jsutils/identityFunc';
 import { inspect } from '../../jsutils/inspect';
 
 import { Kind } from '../../language/kinds';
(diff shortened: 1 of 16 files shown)


saihaj (approved):

yaacovCR:
Updated the PR with necessary changes to `astFromValue`.

IvanGoncharov (changes requested):
Strongly oppose this.
Scalars are used in many situations that don't have access to context and you made it optional for that reason. Scalars should behave in exact same way no matter where they are used.

Moreover, we deliberately do not pass context in functions like `validate` that was already discussed and rejected in another issue.
In use case outlined in #3539 that means if you specify scalar as literal you will get not localize error message (through validate) but if you pass it as the variable it became localized.

Same with other situations where scalars used outside of variables, e.g. default values in SDL you will get non-localized values.
I think localization usecase should be solved through error's `extensions` and `formatError` function from GraphQL server lib as outlined here: https://github.com/graphql/graphql-js/issues/3539

yaacovCR:
I don't think I am as dogmatic about scalars always behaving exactly--exactly--exactly the same in every context.

But, I do think that the solution that @IvanGoncharov has suggested works better in the proposed use case, and so this PR is currently a solution in search of a problem.... or.... more specifically, a use case that justifies the complexity and caveats.

Thanks for the review!

PR #3604, opened 2022-05-24
benchmark: various improvments to benchmark

IvanGoncharov:
This is stacked PR, for descriptions of various changes please look into the description of underlying commits.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -708,7 +708,6 @@ overrides:
       import/no-extraneous-dependencies: [error, { devDependencies: true }]
       import/no-nodejs-modules: off
       no-console: off
-      no-await-in-loop: off
   - files: 'resources/eslint-internal-rules/**'
     env:
       node: true
--- a/.github/workflows/cmd-run-benchmark.yml
+++ b/.github/workflows/cmd-run-benchmark.yml
@@ -32,14 +32,30 @@ jobs:
 
       - name: Run Benchmark
         run: |
-          cat <<EOF >> replyMessage.txt
-          <details>
-          <summary> Benchmark output </summary>
-          EOF
-          npm run benchmark -- --revs HEAD BASE | tee -a replyMessage.txt
-          cat <<EOF >> replyMessage.txt
-          </details>
-          EOF
+          npm run benchmark -- --revs HEAD BASE
+
+      - name: Create replyMessage
+        uses: actions/github-script@v5
+        with:
+          script: |
+            const fs = require('fs');
+
+            // GH doesn't expose job's id so we need to get it through API, see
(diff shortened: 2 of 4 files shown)


yaacovCR:
> Sadly, I need to figure out a way to run benchmark reliably on CI after #3604

So it seems like this PR broke benchmarking? Perhaps by making it overly restrictive in terms of discarded samples? Or, at least that's my understanding.

1. Should we revert at least until fixed?
2. Might we consider a 3rd party benchmarking solution as an additional dev-dependency so that we can outsource some of the benchmarking work?
3. How do other libraries control for context-switching? Do they just increase the number of runs, perhaps randomizing the order of the runs when doing A/B tests?

IvanGoncharov:
> Might we consider a 3rd party benchmarking solution as an additional dev-dependency so that we can outsource some of the benchmarking work?

If you find one. 
Everything I tried is really bad, no lib can reliably measure a 1-2% difference.
The numbers differ a lot between runs.
It's easy to measure the performance of some microbenchmarks it's hard to have repeated results on functions like `execute`, `buildSchema`, etc.

AFAIK, all big project have their own custom benchmark.

> How do other libraries control for context-switching? Do they just increase the number of runs, perhaps randomizing the order of the runs when doing A/B tests?

There is some tech for compiled languages that allow you to get super reliable measures.
But I don't know about anything that can do this for JS.

Moreover, involuntary context switching is an external source of noise for your measurement.
Plus it's totally unpredictable how long it will take + it's unpredictable how much CPU cache will be lost during the context switch.
So I don't think it's possible to account for this with purely statistical methods.
For example, if you run a benchmark before this PR with Youtube playing in the background you will see how severely it affects measurements.

> Should we revert at least until fixed?

If it will block someone.
You can still run the benchmark locally if your computer is not doing anything else and you get a reliable benchmark.
I did this change specifically to measure the effect of AsyncIteratable and other changes to the executor.
Without this change, I can't reliably measure the effects of those changes even locally.

So I would say now it gives reliable numbers if it works, previously it just gave unreliable numbers.

IvanGoncharov:
@yaacovCR Above comment was intended to explain why benchmarking is hard and not discourage you from looking at existing 3rd-party solutions.
It would be great if you look into what is available in the JS ecosystem right now.
Especially since I did my own research many years ago.

yaacovCR:
I think we are not too concerned about "clean room" benchmarking, but rather about real world benchmarking and if context switching introduces a level of noise beyond which the benchmark can't reliably distinguish, I'm not sure the measurable clean room difference would be meaningful. It seems to me that the "solution" as it were would be just to increase the number of runs and randomize the order.

Perhaps anyway this change could be reverted for CI and then enabled locally so CI would at least work even if the variability is higher than local.

PR #3605, opened 2022-05-25
Add new 'GraphQLSchema.getField' method

IvanGoncharov:
Motivation: generalize it and remove code dublication in two places and
also allow to use as public API outside of graphql-js

### DIFF
--- a/src/type/schema.ts
+++ b/src/type/schema.ts
@@ -16,6 +16,8 @@ import { OperationTypeNode } from '../language/ast';
 
 import type {
   GraphQLAbstractType,
+  GraphQLCompositeType,
+  GraphQLField,
   GraphQLInterfaceType,
   GraphQLNamedType,
   GraphQLObjectType,
@@ -30,7 +32,12 @@ import {
 } from './definition';
 import type { GraphQLDirective } from './directives';
 import { isDirective, specifiedDirectives } from './directives';
-import { __Schema } from './introspection';
+import {
+  __Schema,
+  SchemaMetaFieldDef,
+  TypeMetaFieldDef,
+  TypeNameMetaFieldDef,
+} from './introspection';
 
 /**
  * Test if the given value is a GraphQL schema.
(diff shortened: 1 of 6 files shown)


src/type/schema.ts
```diff
@@ -350,6 +357,42 @@ export class GraphQLSchema {
...
+      case TypeMetaFieldDef.name:
+        return this.getQueryType() === parentType
+          ? TypeMetaFieldDef
+          : undefined;
+      case TypeNameMetaFieldDef.name:
+        return TypeNameMetaFieldDef;
+    }
+
+    // this function is part "hot" path inside executor and check presence
+    // of 'getFields' is faster than to use `!isUnionType`
```
yaacovCR, line 389: how much faster? just in general? is there a reference as to why? just curious....

IvanGoncharov: @yaacovCR 
image

IvanGoncharov:
> @IvanGoncharov Please, see benchmark results here: [graphql/graphql-js/runs/6598090837?check_suite_focus=true#step:6:1](https://github.com/graphql/graphql-js/runs/6598090837?check_suite_focus=true#step:6:1)

Sadly, I need to figure out a way to run benchmark reliably on CI after #3604 
But here is data from my machine (HEAD is this PR.
image

So performance slightly increase by ~1% due to using switch instead of multiple `if`.

src/type/schema.ts
IvanGoncharov: very interesting, I tried it with naked `instanceof` and performance regression is still there just significantly smaller:
image

yaacovCR: What about memoizied instanceOf or instanceof? Just curious about all this for discussion about replacing instanceof

IvanGoncharov: image

yaacovCR:
Because this PR contains two changes ie change to switch and addition of unnecessary check for unions during execution, the benchmark results are not so accurate.

i have suggested adding two helpers. See  https://github.com/graphql/graphql-js/issues/3519#issuecomment-1129829416

getObjectField and getCompositeField maybe?

yaacovCR:
> Sadly, I need to figure out a way to run benchmark reliably on CI after #3604

So it seems like #3604 broke benchmarking? Can we revert at least until fixed? Can we consider a 3rd party benchmarking solution as an additional dev-dependency so that we can outsource some of the benchmarking work?

PR #3614, opened 2022-05-28
Lint GitHub Actions

IvanGoncharov:
Right now we pass NPM token in github action and plan to have automatic
releases that is why it is critical to lint action files.
I didn't find any JS-based solution and the only validator I could find
is https://github.com/rhysd/actionlint

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,11 @@ jobs:
       - name: Spellcheck
         run: npm run check:spelling
 
+      - name: Lint GitHub Actions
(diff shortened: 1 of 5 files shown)


yaacovCR:
This seems to be causing tests to be skipped silently. See #3616 where tests are not run.

Available message in log:
> docker://rhysd/actionlint:latest is not allowed to be used in graphql/graphql-js. Actions in this workflow must be: within a repository owned by graphql, created by GitHub, or matching the following: codecov/codecov-action@v1.

Issue #2801, opened 2020-09-17
Getting multiple instances of graphql error despite only having a single version


link2cory:
I wasn't sure whether or not to make a new issue or comment on https://github.com/graphql/graphql-js/issues/594 but that one is so old I figured it would be preferable to start anew

In my project I get the "ensure that there are not multiple versions of GraphQL installed in your node_modules directory" error despite having done so.

yarn list output: 
```
yarn list --pattern graphql                   
yarn list v1.22.5
├─ @apollographql/apollo-tools@0.4.8
├─ @apollographql/graphql-playground-html@1.6.26
├─ @types/graphql-upload@8.0.4
├─ apollo-graphql@0.5.0
├─ graphql-extensions@0.12.4
├─ graphql-scalars@1.2.7
├─ graphql-tag@2.11.0
├─ graphql-tools@4.0.8
├─ graphql-upload@8.1.0
└─ graphql@15.3.0
Done in 0.87s.
```
and for posterity:

```
find node_modules -name graphql
node_modules/graphql
```
My tech stack includes apollo-server-lambda and nexus-schema.  Not sure if/how one of them could be behind this, but the issue only cropped up after switching from apollo-server-express to apollo-server-lambda.  Nevertheless, it would seem that graphql-js is reporting this error erroneously as I have proven there is only one instance of graphql in my node_modules directory.

exact error message:

```
Error: Cannot use GraphQLObjectType "Bio" from another module or realm.                                  
                                                                                                         
Ensure that there is only one instance of "graphql" in the node_modules                                  
directory. If different versions of "graphql" are the dependencies of other                              
relied on modules, use "resolutions" to ensure only one version is installed.                            
                                                                                                         
https://yarnpkg.com/en/docs/selective-version-resolutions                                                
                                                                                                         
Duplicate "graphql" modules cannot be used at the same time since different                              
versions may have different capabilities and behavior. The data from one                                 
version used in the function from another could produce confusing and                                    
spurious results.
```

Link to my project repo:  https://github.com/link2cory/portfolio-backend/tree/serverless
a quick note: although it is not in the current version of my repo, I have tried making use of the resolutions yarn with exactly the same results.  Plus the evidence I have provided above suggests to me that it should not be necessary based on my project dependencies.  

Please correct me if I am wrong in any of these assumptions!

onhate:
hi @link2cory I'll check what is going on

link2cory:
Wow thank you so much for your help!

[...]

PR #3616, opened 2022-05-29
Separate interfaces from implementations for exported classes

yaacovCR:
= allows for custom implementations of the GraphQL entity implementations, e.g. the `execute` function from GraphQL can be used for any object that implements the `GraphQLSchema` interface, even if does not conform to the default `GraphQLSchemaImpl` class.
= currently provides symbols for predicates for the following interfaces: `GraphQLScalarType`, `GraphQLObjectType`,
`GraphQLInterfaceType`, `GraphQLUnionType`, `GraphQLEnumType`, `GraphQLInputObjectType`, `GraphQLList`, `GraphQLNonNull`, `GraphQLDirective`, `GraphQLSchema`, and `Source`, i.e. all classes that made internal use of the `instanceOf` method.
= the `instanceOf` method now uses symbols instead of `instanceof`.
= the exported BREAK object is now passed into the visitor function to allow for cross-compatibility between `graphql` library instances.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,8 @@ import {
   GraphQLString,
 } from 'graphql';
 
-var schema = new GraphQLSchema({
-  query: new GraphQLObjectType({
+var schema = new GraphQLSchemaImpl({
+  query: new GraphQLObjectTypeImpl({
     name: 'RootQueryType',
     fields: {
       hello: {
--- a/docs-old/APIReference-GraphQL.md
+++ b/docs-old/APIReference-GraphQL.md
@@ -32,8 +32,8 @@ _Schema_
 
 <ul class="apiIndex">
   <li>
-    <a href="../type/#graphqlschema">
-      <pre>class GraphQLSchema</pre>
+    <a href="../type/#graphqlschemaimpl">
+      <pre>class GraphQLSchemaImpl</pre>
       A representation of the capabilities of a GraphQL Server.
     </a>
   </li>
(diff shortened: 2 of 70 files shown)


yaacovCR:
This PR is meant to provide a direction and proof of for removing the reliance on `instanceof` within this library. The changed within likely probably could be broken up into several different smaller PRs.

For example, moving BREAK into the visitor method argument certainly could be its own change.

yaacovCR:
tests not running, see https://github.com/graphql/graphql-js/pull/3614#issuecomment-1140431232

yaacovCR:
Note that we still warn (in development) if the objects do not conform to the GraphQL version within the peer dependency, by branding the version into the above object instances. The check is performed only at development, but the branding, for simplicity, happens in production as well.

It would be good to run benchmarks. It might pay to research whether we can skip the branding in production, which might require a build step to remove the branding.

yaacovCR:
Note that the current PR changes the class names to `GraphQL*TypeImpl` and has the class-free TS interfaces use just `GraphQL*Type`.

To lighten the breaking change set, we could keep the class names as `GraphQL*Type` and have the class-free TS interfaces use `IGraphQL*Type`.

yaacovCR:
Note that I removed the comment about how this can enable cross-realm graphql.

That is because I was referring to constructing a schema in one thread, executing in another. It turns out that (with more research I could have discovered that) it is impossible to pass functions to a worker, so although a worker thread can receive a request and return a result, there is no way to pass GraphQL*Type objects across threads. See https://github.com/graphql/graphql-js/issues/2801#issuecomment-1154000057 -- thanks @IvanGoncharov for pointing this out.

This PR could be used to enable esm/cjs compatibility (avoiding/embracing the dual package hazard), but that can be saved in a more lightweight way by #3617 or some other solution.

So the only remaining real reason to enable this PR is to allow drop in replacements for schema creation, validation, execution, etc.

So, I am closing this PR for now, because the value of this has significantly decreased in my mind, and if I am able to get in some of the remaining refactors from `graphql-executor`, I have less of a need for it myself. We always have it as a reference if need be.

IvanGoncharov:
Just a note, in TS you can "implement class":
```
class A {
  a: string | undefined;
}
class B implements A {
  a: string | undefined;
}

const a: A = new B();
```

The downside, you need to implement all private properties.
But schema objects don't have that many private properties.
That said maybe one day we will switch to ECMAScript private properties when TS will learn to ignore them in the "implements" check.

Issue #3603, opened 2022-05-24
Provide a clear path on how to support graphql-js 16 and 17 at the same time as a library author


n1ru4l:
As it seems right now graphql-js 17 will ship without a commonjs solution.

Since ESM modules can only be loaded with a dynamic import within a commonjs context that will make a lot of libraries that support multiple graphql-js versions (commonjs and ESM in GraphQL.js 16 and ESM only in GraphQL.js 17) a pain to use.

Before deciding the only ship ESM, there should be a clear path for those libraries. Otherwise, the ecosystem around graphql-js will result in kind of a broken state.

yaacovCR:
see #3616 which could also be a workaround for the dual-package hazard with cjs/esm compatibility

jaydenseric:
> Provide a clear path on how to support graphql-js 16 and 17 at the same time as a esm and cjs library author

We shouldn't. Pure ESM is a good and necessary major change; it's ok for the ecosystem to upgrade to v17 when ready and leave v16 behind.

See these comments for detailed reasons why pure ESM is good and attempting both ESM and CJS is not:

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#gistcomment-3850849

I'm really looking forward to the pure ESM `graphql` v17 release; as soon as it happens I’ll swap [`graphql-upload`](https://www.npmjs.com/package/graphql-upload) over to pure ESM also:

https://github.com/jaydenseric/graphql-upload/pull/290#pullrequestreview-981670520

If that takes too long I'll move it to pure ESM regardless.

There is no need to overthink and over-engineer all this. CJS has overstayed its welcome and we’re collectively torturing ourselves with it for no particular reason; the time is now to switch to pure ESM. There is never going to be a better time. Many of Sindre Sorhus's packages are pure ESM, as are most of mine, packages like `node-fetch`, the remark/rehype family of packages, and more. `graphql` is not the first to move to pure ESM and it sure won't be the last. Very soon people won't be able to support or use CJS even if they wanted to.

jaydenseric:
> If that takes too long I'll move it to pure ESM regardless.

It took too long 😂 [`graphql-upload`](https://www.npmjs.com/package/graphql-upload) is pure ESM from [v16.0.0](https://github.com/jaydenseric/graphql-upload/releases/tag/v16.0.0), so don't worry about downstream effects of moving `graphql` to pure ESM for that package as we only need/want ESM moving forwards.

[...]

PR #3617, opened 2022-05-29
feat: replace instanceOf with unique Symbol checks

n1ru4l:
This is the first step for a potential reintroduction of CommonJS alongside ESM that avoids the dual package hazard `instanceOf` issues.

~~In addition, I think it would make sense to introduce and assert a flag on `globalThis` in order to avoid multiple GraphQL.js versions being used together (unless there is a valid use case) that should be importet at the top of each entry-point file.~~

**entry-shim.ts**
```ts
const versionSymbol = Symbol.for("loaded-graphql-version")

if (globalThis[versionSymbol] && globalThis[versionSymbol] !== version) {
  throw new Error("Tried loading two different GraphQL versions. Please resolve your resolutions.")
}

globalThis[Symbol.for("graphql-version")] = version
```

**Update:** I don't think we should manually assert the GraphQL version with an `entry-shim` as described above. Resolving a single version of GraphQL is the responsibility of the package manager.

The motivation behind this PR is the reality that CommonJS and ESM will co-exist for some more time until the whole ecosystem catches up. See https://github.com/graphql/graphql-js/issues/3603

____

Related:
- https://github.com/graphql/graphql-js/issues/3603
- https://github.com/graphql/graphql-js/pull/3616


src/type/definition.ts
```diff
@@ -319,6 +360,7 @@ export function assertAbstractType(type: unknown): GraphQLAbstractType {
  * ```
  */
 export class GraphQLList<T extends GraphQLType> {
+  readonly [isGraphQLListTypeSymbol]?: true = true;
```
n1ru4l, line 363: TypeScript went wild without the optional, anyone having an idea why?

yaacovCR: Hmmm. What is the error message?

yaacovCR:
Makes sense to me in terms of separating version checks from instanceOf checks, the two are related, but ultimately separate concerns. I guess the version check changes should be part of this PR if old instanceOf behavior is being deprecated.

[Fundamentally, I don't think we should be doing version checks. If the TS type signatures align, imo that is the package manager's responsibility and graphql-js should get out of the way.

But considering we currently do, I guess it is reasonable to continue.]

src/type/definition.ts
n1ru4l: image

```
Type 'GraphQLType' does not satisfy the constraint 'GraphQLNullableType'.ts(2344)
```

at https://github.com/n1ru4l/graphql-js/blob/dc18ec7c0c5ba0c89a5e074102b784ba57c1dc53/src/type/definition.ts#L441

n1ru4l: image

```
Type 'GraphQLType' does not satisfy the constraint 'GraphQLNullableType'.
  Type 'GraphQLNonNull<GraphQLNullableType>' is not assignable to type 'GraphQLNullableType'.
    Property '[isGraphQLListTypeSymbol]' is missing in type 'GraphQLNonNull<GraphQLNullableType>' but required in type 'GraphQLList<GraphQLType>'.ts(2344)
```

at https://github.com/n1ru4l/graphql-js/blob/dc18ec7c0c5ba0c89a5e074102b784ba57c1dc53/src/type/definition.ts#L227

yaacovCR: I think that's a separate error in definition of `GraphQLWrappingType` somehow uncovered only by this PR. I have submitted a separate fix #3622.

yaacovCR:
Review requested? Approach seems great -- although we also need to decide about whether/where/how to enforce common version.

Generally, I think we need feedback from @IvanGoncharov or perhaps the larger graphql-js-wg or even the main graphql-wg to figure out which way forward on esm only.

src/error/GraphQLError.ts
```diff
@@ -29,13 +29,22 @@ export interface GraphQLErrorOptions {
...
+}
+
 /**
  * A GraphQLError describes an Error found during the parse, validate, or
  * execute phases of performing a GraphQL operation. In addition to a message
  * and stack trace, it also includes information about the locations in a
  * GraphQL document and/or execution result that correspond to the Error.
  */
 export class GraphQLError extends Error {
+  [isGraphQLErrorSymbol]: true = true;
```
IvanGoncharov, line 47: I suggest making it a static property otherwise it consumes memory on every instance.
Also, I suggest having a single static property with symbol `graphql@{git_hash}` as key and enum similar to what @yaacovCR has in #3616

IvanGoncharov:
We have two use cases that this change potentially can solve:
1. CJS/ESM
2. passing Schema to workers, see #2801

However `#2` is need to be confirmed, see: https://github.com/graphql/graphql-js/issues/2801#issuecomment-1154000057

Do you know any other use case this change solves beyond `ESM/CJS`?

IvanGoncharov:
I'm asking because if we just addressing CJS/MJS here, we need to compare it with the solution from #3361 or other similar solutions.

n1ru4l:
@IvanGoncharov I am not a big fan of https://github.com/graphql/graphql-js/pull/3361 as it would require using resolution overwrites. On yarn using resolutions is straightforward and handy. On npm, you will have to use third-party packages. 🙁 (https://www.npmjs.com/package/npm-force-resolutions)

src/error/GraphQLError.ts
n1ru4l: @yaacovCR I am not sure whether making it a static property would make sense as another issue this might solve in the future is to have GraphQLError like-objects that are not instances of the GraphQLErro class.

The following seems way too specific for classes:

```ts
maybeError.constructor?.[isGraphQLErrorSymbol] === true
```

vs

```ts
maybeError[isGraphQLErrorSymbol] === true
```


I am not sure whether a boolean per instance is that big of an issue and a memory monster. 🤔 

____

Regarding: "I suggest having a single static property with symbol graphql@{git_hash} as key and enum similar to what @yaacovCR has in https://github.com/graphql/graphql-js/pull/3616"

I am not sure whether I get this right, but do you assume that each symbol should be scoped to a git hash commit in order to avoid cross-version usage?

 In that case, I would rather propose the solution mentioned in https://github.com/graphql/graphql-js/issues/3603#issue-1246255464 and raise an error as soon as a second GraphQL.js version is loaded which IMHO should be a separate pull request.

> Instead could be another solution to add something like this to the "entry" modules to ensure only one version of GraphQL.js is loaded? Are there use-cases in which you would want to run two graphql-js versions alongside each other?
>
>```ts
>if (globalThis[`_______graphqlVersion`] && globalThis[`_______graphqlVersion`] !== version) {
>  throw new Error("Please fix your resolutions, multiple graphql-js versions are loaded at the same time. <explanation and link for setting yarn and npm resolution>")
>}
>
>globalThis[`_______graphqlVersion`] = version
>```

n1ru4l: I am not sure whether we need this check at all. Our runtime code should not address wrongly configured/installed dpendencies. Only installing and loading a single version of graphql-js should be a package manager concern, not ours.

yaacovCR: But I generally agree with @n1ru4l that we should seriously consider removing this check. It made sense for the time when the library moved from being a regular dependency to a peer dependency, but that time was long ago.

saihaj (approved):
I do like the idea of `Symbol` also being used for version checks

dotansimha (approved):
I like the Symbol solution. Seems like a pretty straightforward solution. 
Since `Symbol.for` will work regardless of where it was executed from, I think it will also solve the multiple instances issue.

yaacovCR:
the cjs/mjs issues should be resolved in v17 with our conditional exports map. v17 does use symbol checks within instanceOf for performance reasons (instead of removing it):
-  https://github.com/graphql/graphql-js/pull/4468

PR #3618, opened 2022-05-30
extendSchema: Do not modify standard directives

IvanGoncharov:
Without this change:
```
const extendedSchema = extendSchema(schema, parse(extensionSDL));
extendSchema.getDirective('skip') === GraphQLSkipDirective
// false
```

Maybe in future we need more comprehensive solution, but for now it's enouch to just match behaviour that we already have to standard types.

### DIFF
--- a/src/utilities/__tests__/extendSchema-test.ts
+++ b/src/utilities/__tests__/extendSchema-test.ts
@@ -17,7 +17,7 @@ import {
   assertScalarType,
   assertUnionType,
 } from '../../type/definition';
-import { assertDirective } from '../../type/directives';
+import { assertDirective, specifiedDirectives } from '../../type/directives';
 import {
   GraphQLBoolean,
   GraphQLFloat,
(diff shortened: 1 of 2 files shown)


yaacovCR:
CI broken :(

Change looks ok but somewhat unnecessary as recreated specified directive objects are identical to old directives, only necessary if for example schema modifiers are checking for specified directives via object identity rather than name, which should probably be avoided, as this "mistake" might be made by others.

For example. mapSchema from graphql tools recreates specified directives...

IvanGoncharov:
> CI broken :(

Fixed, you can now rerun CI on your PRs.

IvanGoncharov:
> Change looks ok but somewhat unnecessary as recreated specified directive objects are identical to old directives, only necessary if for example schema modifiers are checking for specified directives via object identity rather than name, which should probably be avoided, as this "mistake" might be made by others.

Yes, but it would be a global, even in our own code:
https://github.com/graphql/graphql-js/blob/main/src/utilities/astFromValue.ts#L128

IvanGoncharov:
> > CI broken :(
> 
> Fixed, you can now rerun CI on your PRs.

Sadly, GitHub UI doesn't allow to rerun it.
So I just force pushed to trigger it.

yaacovCR:
Uncovered problems with #3611 ?

yaacovCR (approved):

yaacovCR:
CI working again -- if you want me to rebase on main and merge, just let me know.

PR #3619, opened 2022-05-31
use named parameters for node visit context info

yaacovCR:
Motivation:

1. Some visit context information is rarely used, this could be pathway toward refactoring, future-proofing any breaking changes to come. (see: https://github.com/graphql/graphql-js/issues/3225)
2. Eventually, BREAK and/or other visit controllers could be added as third argument directly to the visitor method rarely than exported from the library. (see https://github.com/graphql/graphql-js/pull/3616)

### DIFF
--- a/docs-old/APIReference-Language.md
+++ b/docs-old/APIReference-Language.md
@@ -215,7 +215,7 @@ visit function.
 
 ```js
 var editedAST = visit(ast, {
-  enter(node, key, parent, path, ancestors) {
+  enter(node, { key, parent, path, ancestors }) {
     // @return
     //   undefined: no action
     //   false: skip visiting this node
@@ -223,7 +223,7 @@ var editedAST = visit(ast, {
     //   null: delete this node
     //   any value: replace this node with the returned value
   },
-  leave(node, key, parent, path, ancestors) {
+  leave(node, { key, parent, path, ancestors }) {
     // @return
     //   undefined: no action
     //   false: no action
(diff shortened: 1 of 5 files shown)


yaacovCR:
Motivation:

1. Some visit context information is rarely used, this could be pathway toward refactoring, future-proofing any breaking changes to come. (see: https://github.com/graphql/graphql-js/issues/3225)
2. Eventually, BREAK and/or other visit controllers could be added as third argument directly to the visitor method rarely than exported from the library. (see https://github.com/graphql/graphql-js/pull/3616)

saihaj (approved):

IvanGoncharov (changes requested):
@yaacovCR I'm totally for this change 👍 
But it makes a breaking change without migration path and means you can't support v16 and v17 in the same code base.
Are we sure we want to do this?

yaacovCR:
Backwards compatible version is easier said than done. We can't just change the second argument to keyOrContext, because the directionality is actually opposite, our users are supplying the function, rather than us writing it. How would we know which type of function we were supplied and whether to pass the key or the context?

What seems possible, but not necessarily the best choice: we could make our users specify in some way which type of visitor function they are passing, and then require them to pass it that way. Easiest but ugliest, we could create an entire new function called `visitWithNamedVisitorFnArguments` that supercedes `visit` and then deprecate `visit`.

Adding to next js-wg. If you are reading this, please feel free to chime in.

yaacovCR:
closing this old, stale PR, still available for reference if seems of greater utility.

PR #3620, opened 2022-05-31
align return types from execution and subscription

yaacovCR:
with respect to possible promises

motivations:
1. no need to enter event loop unnecessarily
2. further step toward integration of `subscribe` workflow into `execute`

### DIFF
--- a/src/execution/__tests__/subscribe-test.ts
+++ b/src/execution/__tests__/subscribe-test.ts
@@ -125,23 +125,45 @@ function createSubscription(pubsub: SimplePubSub<Email>) {
   return subscribe({ schema: emailSchema, document, rootValue: data });
 }
 
-async function expectPromise(promise: Promise<unknown>) {
-  let caughtError: Error;
-
-  try {
-    /* c8 ignore next 2 */
-    await promise;
-    expect.fail('promise should have thrown but did not');
-  } catch (error) {
-    caughtError = error;
-  }
+function expectPromise(maybePromise: unknown) {
+  assert(isPromise(maybePromise));
 
   return {
-    toRejectWith(message: string) {
+    toResolve() {
+      return maybePromise;
+    },
+    async toRejectWith(message: string) {
+      let caughtError: Error;
+
+      try {
+        /* c8 ignore next 2 */
+        await maybePromise;
+        expect.fail('promise should have thrown but did not');
+      } catch (error) {
+        caughtError = error;
+      }
+
       expect(caughtError).to.be.an.instanceOf(Error);
       expect(caughtError).to.have.property('message', message);
     },
   };
+}
+
+function expectEqualPromisesOrValues<T>(
+  value1: PromiseOrValue<T>,
+  value2: PromiseOrValue<T>,
+): PromiseOrValue<T> {
+  if (isPromise(value1)) {
+    assert(isPromise(value2));
+    return Promise.all([value1, value2]).then((resolved) => {
+      expectJSON(resolved[1]).toDeepEqual(resolved[0]);
+      return resolved[0];
+    });
+  }
+
+  assert(!isPromise(value2));
+  expectJSON(value2).toDeepEqual(value1);
+  return value1;
 }
 
 const DummyQueryType = new GraphQLObjectType({
@@ -153,7 +175,7 @@ const DummyQueryType = new GraphQLObjectType({
 
 function subscribeWithBadFn(
   subscribeFn: () => unknown,
-): PromiseOrValue<ExecutionResult> {
+): PromiseOrValue<ExecutionResult | AsyncIterable<unknown>> {
   const schema = new GraphQLSchema({
     query: DummyQueryType,
     subscription: new GraphQLObjectType({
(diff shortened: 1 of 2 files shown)


yaacovCR:
because this depends on #3630 the diff to review (until #3630 is merged) would be here:

https://github.com/yaacovCR/graphql-executor/compare/extract...align

src/execution/subscribe.ts
```diff
@@ -220,22 +262,44 @@ async function executeSubscription(
...
+  if (eventStream instanceof Error) {
+    throw locatedError(eventStream, fieldNodes, pathToArray(path));
+  }
+
+  // Assert field returned an event stream, otherwise yield an error.
+  if (!isAsyncIterable(eventStream)) {
+    throw locatedError(
+      new GraphQLError(
+        'Subscription field must return Async Iterable. ' +
+          `Received: ${inspect(eventStream)}.`,
```
IvanGoncharov, line 297: nit: It looks like it can fit the single line now.

IvanGoncharov: scratch it, I checked it doesn't :(

src/execution/subscribe.ts
```diff
@@ -220,22 +262,44 @@ async function executeSubscription(
...
 }
+
+function ensureAsyncIterable(
+  eventStream: unknown,
+  fieldNodes: ReadonlyArray<FieldNode>,
+  path: Path,
+): AsyncIterable<unknown> {
+  if (eventStream instanceof Error) {
+    throw locatedError(eventStream, fieldNodes, pathToArray(path));
+  }
```
IvanGoncharov, line 290: this part doesn't fit under name of the function.
Maybe `assertEventStream`?

src/execution/subscribe.ts
```diff
@@ -220,22 +261,32 @@ async function executeSubscription(
     // Call the `subscribe()` resolver or the default resolver to produce an
     // AsyncIterable yielding raw payloads.
     const resolveFn = fieldDef.subscribe ?? exeContext.subscribeFieldResolver;
-    const eventStream = await resolveFn(rootValue, args, contextValue, info);
+    const eventStream = resolveFn(rootValue, args, contextValue, info);
 
-    if (eventStream instanceof Error) {
-      throw eventStream;
+    if (isPromise(eventStream)) {
+      return eventStream.then(assertEventStream).then(undefined, (error) => {
```
IvanGoncharov, line 267: It's a direct match on how this is done in `execute` just in one line:
https://github.com/graphql/graphql-js/blob/ea1894ae22bd732f1dc7fa4f7bf21f1525688e9a/src/execution/execute.ts#L516-L518

https://github.com/graphql/graphql-js/blob/ea1894ae22bd732f1dc7fa4f7bf21f1525688e9a/src/execution/execute.ts#L533-L536

IvanGoncharov: There additional cost of micro queue trip only if the stuff in `assertEventStream` throws.
Otherwise, nothing changed.

src/execution/__tests__/subscribe-test.ts
```diff
@@ -135,9 +137,6 @@ async function expectPromise(promise: Promise<unknown>) {
   }
 
   return {
-    toReject() {
-      expect(caughtError).to.be.an.instanceOf(Error);
-    },
```
IvanGoncharov, line 140: Yes, it's a very bad practice to match random errors, because it results in a lot of false positive tests.

src/execution/__tests__/subscribe-test.ts
```diff
@@ -152,9 +151,9 @@ const DummyQueryType = new GraphQLObjectType({
   },
 });
 
-async function subscribeWithBadFn(
+function subscribeWithBadFn(
   subscribeFn: () => unknown,
-): Promise<ExecutionResult> {
+): PromiseOrValue<ExecutionResult> {
```
IvanGoncharov, line 156: We still have `await subscribeWithBadFn` in a few places.
It was a non-issue when we always returned promises, but now it means we don't check if it's a promise before resolving it.

IvanGoncharov: to clarify I'm speaking about "await" working with both promises and non-promises.

IvanGoncharov: As a temporary, solution you can do the same approach of saving to result and checking it with `isPromise`.

src/execution/__tests__/subscribe-test.ts
```diff
@@ -165,13 +164,28 @@ async function subscribeWithBadFn(
...
+  return expectEquivalentStreamErrors(subscribeResult, streamResult);
+}
+
+function expectEquivalentStreamErrors(
+  subscribeResult: ExecutionResult | AsyncGenerator<ExecutionResult>,
+  createSourceEventStreamResult: ExecutionResult | AsyncIterable<unknown>,
+): ExecutionResult {
+  assert(!isAsyncIterable(subscribeResult));
+  expectJSON(createSourceEventStreamResult).toDeepEqual(subscribeResult);
+  return subscribeResult;
```
IvanGoncharov, line 188: I understand why this code is needed to preserve old tests intact.
In a long run, it's too much complexity for tests.
I'm ok merging it for now but please try to simplify it in a separate PR.

Or if you think this code will change with upcoming refactoring please feel free to leave a 'TODO' comment and finish it after you finish with refactoring.

yaacovCR: Simplified tests in [f88b523](https://github.com/graphql/graphql-js/pull/3620/commits/f88b523c175b1c1280dc4659be12426f8f9ea07f)

src/execution/__tests__/subscribe-test.ts
yaacovCR: fixed awaits, and used `expectPromise.toResolve()` to ensure is promise where appropriate. [f88b523](https://github.com/graphql/graphql-js/pull/3620/commits/f88b523c175b1c1280dc4659be12426f8f9ea07f)

src/execution/__tests__/subscribe-test.ts
yaacovCR: And added TODOs

yaacovCR:
See further discussion [with respect to advantages of using repeaters](https://repeater.js.org/) about general anti pattern of `Promise<AsyncIterator>`

yaacovCR:
@brainkim

PR #3621, opened 2022-05-31
change incorrect subscribe return type to a GraphQLError

yaacovCR:
…rather than a systems error

to match spec (in parallel to field-level resolver errors in execute)

note the addition of the path, as execution has begun

### DIFF
--- a/src/execution/__tests__/subscribe-test.ts
+++ b/src/execution/__tests__/subscribe-test.ts
@@ -454,9 +454,16 @@ describe('Subscription Initialization Phase', () => {
   });
 
   it('throws an error if subscribe does not return an iterator', async () => {
-    (await expectPromise(subscribeWithBadFn(() => 'test'))).toRejectWith(
-      'Subscription field must return Async Iterable. Received: "test".',
-    );
+    expectJSON(await subscribeWithBadFn(() => 'test')).toDeepEqual({
+      errors: [
+        {
+          message:
+            'Subscription field must return Async Iterable. Received: "test".',
+          locations: [{ line: 1, column: 16 }],
+          path: ['foo'],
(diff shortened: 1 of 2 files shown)


saihaj (approved):

IvanGoncharov (approved):
I'm for it 👍 
Currently, the subscription algorithm doesn't mention this check at all:
https://spec.graphql.org/draft/#sec-Source-Stream

I don't see how this change is contradict spec, so I'm for merging it.

But if you think previous behavior was contradicting the spec it's valuable to document please update the description with links to the spec and more details on what contradiction is.

@yaacovCR Otherwise feel free to merge.

yaacovCR:
I think fixing the entire spec in terms of how it clarifies errors is beyond the scope of this PR. See also https://github.com/graphql/graphql-spec/pull/894

And the entire `Response` section of the spec doesn't take into account subscriptions, so it's tough to say exactly what's going on, but more on that below.

Fundamentally, however, I imagine that the spec purports there to be unified behavior as much as possible between all operations, as within the spec they are all considered different forms of execution, with subscriptions simply a different operation type. That is why this PR, which is one of several on the road toward integrating subscriptions into `execute`, should require no spec change and actually is more "correct" by the spec.

More specifically, there is a rough equivalence given between `ResolveFieldEventStream` and `ResolveFieldValue` and insofar as errors within root field resolvers are field errors for `ResolveFieldValue` with a path ([see your comment, actually](https://github.com/graphql/graphql-spec/pull/894#issuecomment-953037385), they are presumably field errors also for subscriptions. Just as request errors get a map of just errors for subscriptions, so, too, should root field errors.

In my opinion. :)

PR #3622, opened 2022-05-31
fix: update remaining `GraphQLNonNull<GraphQLType>` codepoints to `GraphQLNonNull<GraphQLNullableType>` 

yaacovCR:
This is related to #3597, in the sense that #3597 made the `GraphQLNonNull<GraphQL*Type>` => `GraphQLNonNull<GraphQLNullable*Type>` change in other code points related to the `isNonNullType` function, but not within `GraphQLWrappingType` or assertNonNullType.

This PR was prompted by the uncovering of a bug by a different PR, see: https://github.com/graphql/graphql-js/pull/3617#discussion_r885544662

### DIFF
--- a/src/type/definition.ts
+++ b/src/type/definition.ts
@@ -183,7 +183,9 @@ export function isNonNullType(
   return instanceOf(type, GraphQLNonNull);
 }
 
-export function assertNonNullType(type: unknown): GraphQLNonNull<GraphQLType> {
+export function assertNonNullType(
+  type: unknown,
+): GraphQLNonNull<GraphQLNullableType> {
   if (!isNonNullType(type)) {
     throw new Error(`Expected ${inspect(type)} to be a GraphQL Non-Null type.`);
   }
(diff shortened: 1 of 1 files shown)


n1ru4l (approved):

saihaj (approved):

IvanGoncharov (changes requested):
Looks good!

But as previously discussed you need to provide a description of your change, currently, it only contains links.

Also, bug fixes should include tests, otherwise, we have no way to detect regressions.

yaacovCR:
We all can and do make mistakes, especially when busy. An extra note explicitly stating that this was a long standing bug could not have hurt, sure.

yaacovCR:
I updated the description above. For good measure, I did a quick search for additional code points to correct and udpated the title of this PR.

dotansimha (approved):

dotansimha:
@IvanGoncharov Any chance you can review this PR again? 
At the moment, we don't have TypeScript types testing setup at the moment, and from my experience with codegen and other libraries, it is not always easy to add TypeScript types testing. 
Do you think we can merge this one and maybe open an issue to improve TS types testing later? (also because this PR is fairly small and simple, and it's really addressing a real TS types bug)

IvanGoncharov:
> At the moment, we don't have TypeScript types testing setup, and from my experience with codegen and other libraries, it is not always easy to add TypeScript types testing.
Do you think we can merge this one and maybe open an issue to improve TS types testing later? (also because this PR is fairly small and simple, and it's really addressing a real TS types bug)

Agree, answered here https://github.com/graphql/graphql-js/pull/3622#issuecomment-1143484274

Simpdaddyjosh (approved):

PR #3623, opened 2022-05-31
diff-npm-package: create reports folder

yaacovCR:

PR #3624, opened 2022-06-01
diff-npm-package: fix reports dir creation

yaacovCR:
#3623 creates the reports folder via the report path, rather than the reports folder path. This did not trigger an error, because no changes were found.

### DIFF
--- a/resources/diff-npm-package.ts
+++ b/resources/diff-npm-package.ts
@@ -32,10 +32,11 @@ const diff = execOutput(`npm diff --diff=${fromPackage} --diff=${toPackage}`);
 if (diff === '') {
   console.log('No changes found!');
 } else {
-  const reportPath = localRepoPath('reports', 'npm-dist-diff.html');
-  if (!fs.existsSync(reportPath)) {
-    fs.mkdirSync(reportPath);
+  const reportsDir = localRepoPath('reports');
+  if (!fs.existsSync(reportsDir)) {
+    fs.mkdirSync(reportsDir);
   }
+  const reportPath = path.join(reportsDir, 'npm-dist-diff.html');
   fs.writeFileSync(reportPath, generateReport(diff));
   console.log('Report saved to: ', reportPath);
 }


yaacovCR:
History:
1. I noticed that there was a problem with pull requests in terms of creating the report file: See, for example: https://github.com/graphql/graphql-js/runs/6670599072?check_suite_focus=true#step:6:12.
2. I surmised this was because of the recent PR  https://github.com/graphql/graphql-js/pull/3611 changing the reports directory.
3. I assumed that the problem was that the directory was not being created, so I submitted https://github.com/graphql/graphql-js/pull/3623 to try to fix that.
4. @IvanGoncharov noticed that there was something wrong with the check, and @saihaj submitted a fix for it, and #3623 got merged.
5. But, the truth is, that the report was not being created at all because no diff was found, so the check I added should never have been hit.
6. I am unclear why the former PRs found a diff, but with the creation of #3623, no diff was found. I find this very strange.
7. Anyway, I am submitting what I hope is a better fix to #3623, as the logic there unfortunately was still off, as the "reports" directory was being created with the same path later used as the report filename -- there needs to be two different paths, one for the creation of the folder, one for the creation of the file.
8. I am not sure how to test this, as the report is not being created, and I am not sure how to actually force the report to be created.

yaacovCR:
I am not sure why, but luckily #3622 still triggers a diff, and is still failing, so can be used to test this PR.

After rebasing, as expected, now mkdir is failing because we are calling it with the wrong path. Merging this PR should fix that, and so at least after we merge this, we can use that PR to see if the change works.

At that point, CI from #3622 should give us the diff, and we might be able to see why there is one!

IvanGoncharov:
Thanks for the detailed, investigation 👍 
Your description gives a lot of contexts.

> I am not sure how to actually force the report to be created.

Any change that affects the content of the NPM file (add stuff to package.json, LICENSE, README, or any files under 'src/').

IvanGoncharov:
Context: I need this script to see what changed inside the NPM package.
It's relevant on PRs that update `tsconfig.json`, `resources/build-npm.js`, etc.
It was super useful during TS migration.
Before that, I was doing the same by cloning PRs and generating before/after builds and running diff on it.
For example, it would be relevant in upcoming CJS/MJS discussions.

PR #3625, opened 2022-06-02
benchmark: replace failing on context switching with warning

IvanGoncharov:
Consequently failed measurements due to context switches are bad.
It mean something is bad with our setup:
* Our test are taking too long to run (e.g. due to `count` being too
  big)
* Machine that we measure on is doing some heave background task.
* Some other factors, that are hard to predict.

Currently, it's easy to solve this issue locally just by disabling
background task on your machine, but it is non-trivial task to solve on
CI.

### DIFF
--- a/resources/benchmark.ts
+++ b/resources/benchmark.ts
@@ -122,9 +122,12 @@ function collectSamples(modulePath: string) {
 
     if (sample.involuntaryContextSwitches > 0) {
       numOfConsequentlyRejectedSamples++;
-      if (numOfConsequentlyRejectedSamples > 5) {
-        throw Error(
-          'Can not sample benchmark due to 5 consequent runs beings rejected because of context switching',
+      if (numOfConsequentlyRejectedSamples === 5) {
+        console.error(
+          yellow(
+            '  Five or more consequent runs beings rejected because of context switching.\n' +
+              '  Measurement can take a significantly longer time and its correctness can also be impacted.',
+          ),
         );
       }
       continue;


IvanGoncharov:
cc: @yaacovCR

resources/benchmark.ts
```diff
@@ -122,9 +122,12 @@ function collectSamples(modulePath: string) {
...
     if (sample.involuntaryContextSwitches > 0) {
       numOfConsequentlyRejectedSamples++;
-      if (numOfConsequentlyRejectedSamples > 5) {
-        throw Error(
-          'Can not sample benchmark due to 5 consequent runs beings rejected because of context switching',
+      if (numOfConsequentlyRejectedSamples === 5) {
+        console.error(
+          yellow(
+            '  Five or more consequent runs beings rejected because of context switching.\n' +
+            '  Measurement can take a significantly longer time and its correctness can also be impacted.',
```
yaacovCR, line 129: Question: if we still throw out bad samples, why is measurement correctness impacted when it finally completes?

yaacovCR: I guess you're just saying you don't know for sure, and that's fine.

yaacovCR (approved):

resources/benchmark.ts
IvanGoncharov: Yes, exactly. 
If you running a heavy background process it's causing all other kinds of issues, e.g. your CPU can thermal throttle (relevant only for local runs), shared caches, and memory access is affected.
This notice here is basically to say if we running on a PC with a heavy load we shouldn't try to figure out how it affects measurements. 
You simply should solve the root cause, which is not yet possible on our CI.

PR #3627, opened 2022-06-06
Reuse GraphQLWrappingType in 'GraphQLType' definition

IvanGoncharov:
Now GraphQLType is union of named types and wrapping types.
Spotted during review of #3622


yaacovCR (approved):
Nice catch!

PR #3628, opened 2022-06-06
benchmark: add Windows compatibility

yaacovCR:
1. use local git clone instead of git archive | tar
2. normalize module path within sampleCode

Motivation: Windows users should not be required to use WSL to run benchmark (or any other scripts). The library should be cross-platform as possible both at runtime and in development.

Furthermore: it may be valuable to benchmark in multiple environemnts, including unix-type environments, WSL, and native Windows runtimes.

### DIFF
--- a/resources/benchmark.ts
+++ b/resources/benchmark.ts
@@ -95,7 +95,8 @@ function prepareBenchmarkProjects(
     const repoDir = path.join(tmpDir, hash);
     fs.rmSync(repoDir, { recursive: true, force: true });
     fs.mkdirSync(repoDir);
-    exec(`git archive "${hash}" | tar -xC "${repoDir}"`);
+    exec(`git clone --quiet "${localRepoPath()}" "${repoDir}"`);
+    exec(`git checkout --quiet --detach "${hash}"`, { cwd: repoDir });
     exec('npm --quiet ci --ignore-scripts', { cwd: repoDir });
     fs.renameSync(buildNPMArchive(repoDir), archivePath);
     fs.rmSync(repoDir, { recursive: true });
@@ -111,14 +112,14 @@ function prepareBenchmarkProjects(
   }
 }
 
-function collectSamples(modulePath: string) {
+function collectSamples(projectPath: string, benchmark: string) {
   let numOfConsequentlyRejectedSamples = 0;
   const samples = [];
 
   // If time permits, increase sample size to reduce the margin of error.
   const start = Date.now();
   while (samples.length < minSamples || (Date.now() - start) / 1e3 < maxTime) {
-    const sample = sampleModule(modulePath);
+    const sample = sampleModule(projectPath, benchmark);
 
     if (sample.involuntaryContextSwitches > 0) {
       numOfConsequentlyRejectedSamples++;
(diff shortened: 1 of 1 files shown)


yaacovCR:
@github-actions run-benchmark

resources/benchmark.ts
```diff
@@ -373,11 +373,18 @@ interface BenchmarkSample {
   involuntaryContextSwitches: number;
 }
 
-function sampleModule(modulePath: string): BenchmarkSample {
+function normalizePath(rawPath: string): string {
+  if (process.platform === 'win32') {
+    return path.posix.join(...rawPath.split(path.sep));
+  }
+  return rawPath;
+}
```
IvanGoncharov, line 381: > normalize benchmark path within sampleCode

Can you please explain the problem that you trying to solve with this code?
Can it be done with https://nodejs.org/api/path.html#pathnormalizepath ?

IvanGoncharov (dismissed):

resources/benchmark.ts
yaacovCR: path.normalize would be perfect, thank you!

IvanGoncharov: Is this happening because of `localRepoPath()`, if so please add `normalize` call in that specific function?
If not can please explain where non-normalized path is being created?

resources/benchmark.ts
```diff
@@ -95,7 +95,8 @@ function prepareBenchmarkProjects(
     const repoDir = path.join(tmpDir, hash);
     fs.rmSync(repoDir, { recursive: true, force: true });
     fs.mkdirSync(repoDir);
-    exec(`git archive "${hash}" | tar -xC "${repoDir}"`);
```
IvanGoncharov, line 98: I thought that Win10 supports `tar`, is it a problem with using pipe?
https://pureinfotech.com/tar-curl-windows-10/

yaacovCR:
> > 3\. spawn from projectPath to avoid need to handle drive letter
> 
> @yaacovCR I don't understand this one, probably because I'm not using Windows for development. Can you please explain?

Windows paths have colons in them, which node interprets as a protocol like “node:” or “file:” i don’t see the value in introducing the correct presumably file: prefix to the import that would handle this when we have the option to run from working directory and use a relative path. If there is an advantage to using an absolute path, I could reconsider.

> Don't have actual stats but I expect 99% of users use some variant of Linux in production. 

Unless they are running in browser.

> I assumed WSL is a good solution for Windows developers but if you feel strongly that we should support native Windows workflow I'm ok to merge it.

I feel strongly.

resources/benchmark.ts
yaacovCR: I guess so! I can update the description above.

Overall it's not clear to me why creating and then extracting a tar file is better than just creating a local clone, but if you feel we should create two different piping commands lines depending on the platform, that's fine as well.

As long as it works, I will be happy.

resources/benchmark.ts
yaacovCR: Actually, sorry normalize isn’t exactly what we want. We want to convert a windows style path to a posix style path because even windows wants posix style paths in source code. So we should be able to use path.posix.normalize in the source code

yaacovCR: fixed with [`url.pathToFileURL`](https://nodejs.org/api/url.html#urlpathtofileurlpath)

yaacovCR:
> Windows paths have colons in them, which node interprets as a protocol like “node:” or “file:” i don’t see the value in introducing the correct presumably file: prefix to the import that would handle this when we have the option to run from working directory and use a relative path. If there is an advantage to using an absolute path, I could reconsider.

Fixed with [`url.pathToFileURL`](https://nodejs.org/api/url.html#urlpathtofileurlpath) 

edited list above

resources/benchmark.ts
```diff
@@ -377,7 +379,7 @@ function sampleModule(modulePath: string): BenchmarkSample {
   const sampleCode = `
     import fs from 'node:fs';
 
-    import { benchmark } from '${modulePath}';
+    import { benchmark } from '${url.pathToFileURL(modulePath)}';
```
IvanGoncharov, line 382: Looks great, now 👍 
Just add a comment explaining why we need conversion, otherwise, we risk someone removing it in the future.

PR #3633, opened 2022-06-07
benchmark: increase process priority

yaacovCR:
...to try to avoid involuntary context switches.

This seems to me to be pretty TETAW ("too easy to actually work").

### DIFF
--- a/resources/benchmark.ts
+++ b/resources/benchmark.ts
@@ -381,8 +381,11 @@ function sampleModule(modulePath: string): BenchmarkSample {
 
   const sampleCode = `
     import fs from 'node:fs';
+    import os from 'node:os';
 
     import { benchmark } from '${moduleURL}';
+
+    os.setPriority(os.constants.priority.PRIORITY_HIGHEST);
 
     // warm up, it looks like 7 is a magic number to reliably trigger JIT
     benchmark.measure();
--- a/resources/benchmark.ts
+++ b/resources/benchmark.ts
@@ -435,6 +435,7 @@ function sampleModule(modulePath: string): BenchmarkSample {
     {
       stdio: ['inherit', 'inherit', 'inherit', 'pipe'],
       env: { NODE_ENV: 'production' },
+      uid: process.getuid(),
     },
   );
 
--- a/resources/benchmark.ts
+++ b/resources/benchmark.ts
@@ -417,8 +417,10 @@ function sampleModule(modulePath: string): BenchmarkSample {
   `;
 
   const result = cp.spawnSync(
-    process.execPath,
+    'sudo',
     [
+      '--close-from=4',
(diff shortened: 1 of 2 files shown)


yaacovCR:
@github-actions run-benchmark

yaacovCR:
Not sure if sudo can help us here. if i were github actions, it wouldn't!

yaacovCR:
@github-actions run-benchmark

yaacovCR:
@github-actions run-benchmark

yaacovCR:
@github-actions run-benchmark

yaacovCR:
@github-actions run-benchmark

yaacovCR:
sudo and setPriority seem to be working! although another erupts when writing for some reason...

I can read more about affinity

yaacovCR:
I am guessing when using a shell, piping via 3 does not work

https://github.com/graphql/graphql-js/blob/b91f51b8f9c6bc7a2d0b5d1ef7b6121bf094cac4/resources/benchmark.ts#L408

IvanGoncharov:
@yaacovCR You need to pass `sudo -C4` 
https://superuser.com/questions/394522/sudo-preserve-non-std-file-descriptors

resources/benchmark.ts
```diff
@@ -376,9 +376,12 @@ interface BenchmarkSample {
 function sampleModule(modulePath: string): BenchmarkSample {
   const sampleCode = `
     import fs from 'node:fs';
+    import os from 'node:os';
 
     import { benchmark } from '${modulePath}';
 
+    os.setPriority(os.constants.priority.PRIORITY_HIGHEST);
+
```
IvanGoncharov, line 384: @yaacovCR Also, you can try setting priority for benchmark itself to the lowest.
That will force context switches to occur before/after measurments.

yaacovCR: Or maybe benchmark will never finish? I think we can see what happens with this as a first step and take it from there.

yaacovCR:
> @yaacovCR Simpler solution would be to run the entire benchmark under sudo.

That works for me.

As a side note, I want to make sure this works cross platform. I would assume that ci would add sudo to the whole command and local users would be required to do so on their own or use the windows equivalent.

yaacovCR:
@github-actions run-benchmark

yaacovCR:
@github-actions run-benchmark

yaacovCR:
@IvanGoncharov 

> sudo: you are not permitted to use the -C option

Needs further research

IvanGoncharov:
> Needs further research

@yaacovCR Can I quickly try to run some experiments on top of your PR?

yaacovCR:
Closing for now. Would be on wishlist to revisit in terms of our overall benchmarking picture.

PR #3634, opened 2022-06-10
createSourceEventStream: introduce named arguments and deprecate positional arguments

yaacovCR:
Deprecates the positional arguments to `createSourceEventStream`, to be removed in the next major version, in favor of named arguments.

Motivation:
1. aligns `createSourceEventStream` with the other exported entrypoints `graphql`, `execute`, and `subscribe`
2. allows simplification of `mapSourceToResponse`

suggested by @IvanGoncharov

### DIFF
--- a/src/execution/__tests__/subscribe-test.ts
+++ b/src/execution/__tests__/subscribe-test.ts
@@ -458,6 +458,7 @@ describe('Subscription Initialization Phase', () => {
       'Must provide document.',
     );
 
+    // @ts-expect-error
     expect(() => createSourceEventStream(schema)).to.throw(
       'Must provide document.',
     );
--- a/src/execution/__tests__/subscribe-test.ts
+++ b/src/execution/__tests__/subscribe-test.ts
@@ -191,7 +191,7 @@ function subscribeWithBadFn(
 
   return expectEqualPromisesOrValues(
     subscribe({ schema, document }),
-    createSourceEventStream(schema, document),
+    createSourceEventStream({ schema, document }),
   );
 }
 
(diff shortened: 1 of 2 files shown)


yaacovCR:
To be backported to v16 with later PR to remove positional arguments in v17

n1ru4l (approved):

src/execution/__tests__/subscribe-test.ts
```diff
@@ -421,6 +421,48 @@ describe('Subscription Initialization Phase', () => {
     expect(() => subscribe({ schema })).to.throw('Must provide document.');
   });
 
+  it('allows positional arguments to createSourceEventStream', () => {
```
IvanGoncharov, line 424: Please switch all tests to use named argument, and make this a test to support positional arguments.

src/execution/subscribe.ts
```diff
@@ -163,14 +120,50 @@ function mapSourceToResponse(
  * "Supporting Subscriptions at Scale" information in the GraphQL specification.
  */
 export function createSourceEventStream(
-  schema: GraphQLSchema,
-  document: DocumentNode,
+  argsOrSchema: ExecutionArgs | GraphQLSchema,
+  /** Note: document argument is required if positional arguments are used */
+  /** @deprecated will be removed in next major version in favor of named arguments */
```
IvanGoncharov, line 125: Please make separate TS signatures as we did for GraphQLErrors.
That way single `deprecate` would be enough, also it will be less confusing.

src/execution/__tests__/subscribe-test.ts
yaacovCR: Hmmm, I may need more coffee, but I think that's done already. There are very few direct tests of `createSourceEventStream`, as the code is mostly tested through calls to `subscribe`.

src/execution/subscribe.ts
yaacovCR: Great idea. [e8a2a74](https://github.com/graphql/graphql-js/pull/3634/commits/e8a2a7405c921b549124db8f7a53c241f8b375b2)

src/execution/__tests__/subscribe-test.ts
IvanGoncharov: wow, we use it only in one place in whole tests.
When totally make sense.
Just got confused because I expected more changes.

When everything is good, just small improvement to the description
```suggestion
  it('Deprecated: allows positional arguments to createSourceEventStream', () => {
```

Similar to this one
https://github.com/graphql/graphql-js/blob/75635f03ba950a5f0aa212bad2052297e9a59725/src/validation/__tests__/validation-test.ts#L61

src/execution/subscribe.ts
```diff
@@ -170,7 +131,45 @@ export function createSourceEventStream(
...
+  argsOrSchema: ExecutionArgs | GraphQLSchema,
+  document?: DocumentNode,
+  rootValue?: unknown,
+  contextValue?: unknown,
+  variableValues?: Maybe<{ readonly [variable: string]: unknown }>,
+  operationName?: Maybe<string>,
+  subscribeFieldResolver?: Maybe<GraphQLFieldResolver<any, any>>,
 ): PromiseOrValue<AsyncIterable<unknown> | ExecutionResult> {
+  if (isSchema(argsOrSchema)) {
+    invariant(document != null, 'Must provide document.');
```
IvanGoncharov, line 145: Do you still need this `invariant` if you make separate TS signatures?

IvanGoncharov: In general, I think his approach is cleaner than doing "Impl" function.

yaacovCR: I tried that approach but couldn’t get rid of invariant! @n1ru4l  any ideas?

src/execution/__tests__/subscribe-test.ts
yaacovCR: Done!

src/execution/subscribe.ts
yaacovCR: I think unfortunately a TS limitation see [c39e0f3](https://github.com/graphql/graphql-js/pull/3634/commits/c39e0f350feb7f6be8b67f4da98885ced7991ca5)

IvanGoncharov (approved):
Looks good!

dotansimha (approved):

PR #3635, opened 2022-06-10
createSourceEventStream: remove deprecated positional argument overload

yaacovCR:
See #3634, which introduced a named argument form for the exported function and deprecated the original positional argument overload. This is to be backported to v16.

This PR removes the deprecated format from v17. It is a BREAKING CHANGE and cannot be backported.

This PR depends on #3634.

### DIFF
--- a/src/execution/__tests__/subscribe-test.ts
+++ b/src/execution/__tests__/subscribe-test.ts
@@ -421,49 +421,6 @@ describe('Subscription Initialization Phase', () => {
     expect(() => subscribe({ schema })).to.throw('Must provide document.');
   });
 
-  it('Deprecated: allows positional arguments to createSourceEventStream', () => {
-    async function* fooGenerator() {
-      /* c8 ignore next 2 */
-      yield { foo: 'FooValue' };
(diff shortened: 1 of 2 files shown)


src/execution/__tests__/subscribe-test.ts
```diff
@@ -421,6 +421,28 @@ describe('Subscription Initialization Phase', () => {
     expect(() => subscribe({ schema })).to.throw('Must provide document.');
   });
 
+  it('throws an error if document is missing when using positional arguments', async () => {
```
IvanGoncharov, line 424: Did you forget to remove this one?
Or am I missing something?
Probably worth to also mark this test as "Deprecated:"

IvanGoncharov (dismissed):
@yaacovCR With exception of an issue with "positional test" it's ready to be merged.

src/execution/__tests__/subscribe-test.ts
yaacovCR: Forgot to remove. And now done in #3634

PR #3636, opened 2022-06-10
move `subscribe` code to `execute` file

yaacovCR:
= makes no other changes
= paves the way for further changes


src/execution/execute.ts
```diff
@@ -234,10 +236,8 @@ function buildResponse(
 /**
  * Essential assertions before executing to provide developer feedback for
  * improper use of the GraphQL library.
- *
- * @internal
  */
-export function assertValidExecutionArguments(
```
IvanGoncharov, line 240: Can you please quickly look if someone is using those functions with removed "export"?
I usually use https://cs.github.com/ for that:
https://cs.github.com/?q=buildExecutionContext%20graphql%20language%3ATypeScript&scopeName=All%20repos&scope=#

It's not a reason to not remove "export" it's just if you see some popular library using it we need to contact the author at some point.

IvanGoncharov:
@yaacovCR Thanks a lot for doing it as a separate step 👍 
Save a lot of effort during the review!
Changes look good once you remove the "positional test".
If you find that some popular libraries were using our "internal" functions please open a separate issue to discuss what to do and communicate with their authors.
Otherwise this PR is ready to be merge 👍

src/execution/execute.ts
yaacovCR: I do not have access to that technology preview. I will revert the removed exports, and we can consider removing them in a separate PR.

PR #3639, opened 2022-06-12
refactor: subscribe: introduce `buildPerEventExecutionContext`

yaacovCR:
= introduces `buildPerEventExecutionContext` that creates an `ExecutionContext` for each subscribe event from the original `ExecutionContext` used to create the event stream
= `subscribe` now directly builds the `ExecutionContext` instead of relying on `createSourceEventStream`
= introduces `createSourceEventStreamImpl` and `executeImpl` functions that operate on the built `ExecutionContext` rather the `ExecutionArgs`
= `subscribe` calls the `createSourceEventStreamImpl` function on the original context and eventuallys calls `executeImpl` on the per event context created by `buildEventExecutionContext`.

Motivation:
1. remove unnecessary `buildExecutionContext` call, reducing duplicate work
2. paves the way for easily enhancing the `buildPerEventExecutionContext` to add a new `perEventContextFactory` argument to augment the context argument passed to resolvers as need per event.

depends on #3654

### DIFF
--- a/src/execution/__tests__/subscribe-test.ts
+++ b/src/execution/__tests__/subscribe-test.ts
@@ -14,7 +14,7 @@ import { GraphQLList, GraphQLObjectType } from '../../type/definition';
 import { GraphQLBoolean, GraphQLInt, GraphQLString } from '../../type/scalars';
 import { GraphQLSchema } from '../../type/schema';
 
-import type { ExecutionResult } from '../execute';
+import type { ExecutionArgs, ExecutionResult } from '../execute';
 import { createSourceEventStream, subscribe } from '../execute';
 
 import { SimplePubSub } from './simplePubSub';
@@ -192,6 +192,15 @@ function subscribeWithBadFn(
   return expectEqualPromisesOrValues(
     subscribe({ schema, document }),
     createSourceEventStream({ schema, document }),
+  );
+}
+
+function subscribeWithBadArgs(
+  args: ExecutionArgs,
+): PromiseOrValue<ExecutionResult | AsyncIterable<unknown>> {
+  return expectEqualPromisesOrValues(
+    subscribe(args),
+    createSourceEventStream(args),
   );
 }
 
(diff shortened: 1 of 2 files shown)


saihaj (approved):

yaacovCR:
Not the best diagrams (fault my own) screenshotted from [yEd](https://www.yworks.com/products/yed-live):

## Original:

original

## New:
new

Note that per event, we are now no longer re-performing the portions of `BuildExecutionContext` that do not need to be reperformed.

yaacovCR:
Note that these changes do imply that the `*Impl` functions could be methods on an `Execution` option or the need for some other abstraction.

But that is already implied by the passing of `ExecutionContext` to the majority of functions in the file.

These functions may be renamed/reorganized in later refactoring!

yaacovCR:
Offline, @IvanGoncharov added a few notes:

1. The spec also indicates that execution context should be built only once for subscriptions in that coercion of variables, for instance, is described only once, not per event.
2. At the same time, when CreateSorceEventStream and ExecuteSubscriptionEvent are managed by two separate services, as suggested by the spec, coercion will end up happening per event. This PR cannot, should not, and does not change that, because the provided implementation of the full subscription operation utilizes only a single service, exposing CreateSourceEventStream and Execute to do so.

yaacovCR:
I rebased this on top of the simpler changes introduced in #3654

src/execution/__tests__/subscribe-test.ts
```diff
@@ -195,6 +195,15 @@ function subscribeWithBadFn(
...
 }
 
+function subscribeWithBadArgs(
+  args: ExecutionArgs,
+): PromiseOrValue<ExecutionResult | AsyncIterable<unknown>> {
+  return expectEqualPromisesOrValues(
+    subscribe(args),
+    createSourceEventStream(args),
+  );
+}
```
IvanGoncharov, line 205: Can `subscribeWithBadFn` now use this function?

yaacovCR: Yes good idea!

IvanGoncharov (approved):
This change solves the real issue.

I don't like some of the solutions here like "executeImpl" or "per even context", but I can't propose any other solution that doesn't involve bigger refactoring.

P.S. I'm working on such bigger refactoring in parallel

src/execution/__tests__/subscribe-test.ts
yaacovCR: https://github.com/graphql/graphql-js/pull/3639/commits/03f1a1ed3e546c5d8edbed84e2c41ecfa6a98f64

PR #3640, opened 2022-06-12
subscribe: once execution starts....

yaacovCR:
data should return null?

### DIFF
--- a/src/execution/__tests__/subscribe-test.ts
+++ b/src/execution/__tests__/subscribe-test.ts
@@ -427,6 +427,7 @@ describe('Subscription Initialization Phase', () => {
 
     const result = subscribe({ schema, document });
     expectJSON(result).toDeepEqual({
+      data: null,
       errors: [
         {
           message:
@@ -451,6 +452,7 @@ describe('Subscription Initialization Phase', () => {
 
     const result = subscribe({ schema, document });
     expectJSON(result).toDeepEqual({
+      data: null,
       errors: [
         {
           message: 'The subscription field "unknownField" is not defined.',
@@ -477,6 +479,7 @@ describe('Subscription Initialization Phase', () => {
 
   it('throws an error if subscribe does not return an iterator', async () => {
     const expectedResult = {
+      data: null,
       errors: [
         {
           message:
@@ -498,6 +501,7 @@ describe('Subscription Initialization Phase', () => {
 
   it('resolves to an error for subscription resolver errors', async () => {
     const expectedResult = {
+      data: null,
       errors: [
         {
           message: 'test error',
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -1141,12 +1141,15 @@ export function createSourceEventStream(
   try {
     const eventStream = executeSubscription(exeContext);
     if (isPromise(eventStream)) {
-      return eventStream.then(undefined, (error) => ({ errors: [error] }));
+      return eventStream.then(undefined, (error) => ({
+        data: null,
+        errors: [error],
+      }));
     }
 
     return eventStream;
   } catch (error) {
-    return { errors: [error] };
+    return { data: null, errors: [error] };
   }
 }
 


yaacovCR:
See some discussion in https://github.com/graphql/graphql-spec/pull/894

See additional rough thoughts on how subscriptions fit in at: https://github.com/graphql/graphql-js/pull/3621#issuecomment-1150280432

benjie:
The transition between "request error" (payloads like `{errors: GraphQLError[]}`) and "field error" (payloads like `{data: object | null, errors?: GraphQLError[]}`) in the spec doesn't seem to happen at explicitly designated boundaries, but here's my take:

If a "request error" is raised, then that's seen as being "before execution" (even if it's part of ExecuteRequest(), for example the "request error" raised by [CoerceVariableValues()](https://spec.graphql.org/draft/#CoerceVariableValues()).

Asserts don't define an error type because they should not happen (due to validation, etc); but I see the assert in ExecuteQuery() and ExecuteMutation() as raising a "request error" (since there is no field at that point). Once these algorithms have moved onto the `ExecuteSelectionSet` phase we're out of "request error" phase and into "field error" phase.

For subscriptions the definition is somewhat harder to discern. Do we introduce the "boundary" before or after step 10 of [CreateSourceEventStream()](https://spec.graphql.org/draft/#CreateSourceEventStream())? What happens if the call to `ResolveFieldEventStream` results in an error?

One option is that _because_ this is an error in a field method, it _is_ a field error, and thus the division between "request error phase" and "field error phase" is between steps 9 and 10 of CreateSourceEventStream:

> 9. Let argumentValues be the result of [CoerceArgumentValues](https://spec.graphql.org/draft/#CoerceArgumentValues())(subscriptionType, field, variableValues)
> ***\~\~BOUNDARY\~\~***
> 10. Let fieldStream be the result of running [ResolveFieldEventStream](https://spec.graphql.org/draft/#ResolveFieldEventStream())(subscriptionType, initialValue, fieldName, argumentValues).

So an error raised whilst creating the stream _for a field_ should be treated as a field error, and thus should result in the field itself being null. However, this may be surprising to the user who is expecting a stream not a singular payload.

It seems to me that if you fail to create the event stream then that should be a request error (you've failed to set up a subscription, so the operation has failed). I.e. rather than putting the boundary at the aforementioned position, I think it should probably be once you enter [MapSourceToResponseEvent()](https://spec.graphql.org/draft/#MapSourceToResponseEvent()). But honestly, I could go either way on this. There does seem to be ambiguity here, but I'd have to read the spec more fully again to figure out if this was defined or not.

yaacovCR:
> It seems to me that if you fail to create the event stream then that should be a request error (you've failed to set up a subscription, so the operation has failed). I.e. rather than putting the boundary at the aforementioned position, I think it should probably be once you enter [MapSourceToResponseEvent()](https://spec.graphql.org/draft/#MapSourceToResponseEvent()). But honestly, I could go either way on this.

I guess any type of error for CreateSourceEventStream leads to failed input for MapSourceToResponseEvent which is a kind of request error…

I think this needs more hardening in the spec and likely a better way to distinguish between error types than the existence of data:null which doesn’t make much sense for subscriptions.

i am going to close this for now, but it’s here if we want to revisit

benjie:
In summary, my current opinion is that "execution begins"

- for queries: when you enter [ExecuteQuery()](https://spec.graphql.org/draft/#ExecuteQuery())
- for mutations: when you enter [ExecuteMutation()](https://spec.graphql.org/draft/#ExecuteMutation())
- for subscriptions: when you enter [MapSourceToResponseEvent()](https://spec.graphql.org/draft/#MapSourceToResponseEvent())

yaacovCR:
That makes data:null match current behavior, but I think a spec improvement would make all this explicit and also explicitly acknowledge that with subscriptions, there are two rounds of execution, one that sets up the source stream and one for each event....

benjie:
Yeah, I agree that clarity here should come from the spec.

PR #3642, opened 2022-06-13
Remove 'devAssert' checks that duplicate TS types

IvanGoncharov:
`graphql` provides TS types since `14.5.0` (released 3 years ago)
and we fully switched to TS in `15.0.0` so I think it's time to drop
runtime typechecks.

Motivations: This type checks were added long time ago since we shifted
towards TS we just maintained them without adding new ones.
In general, this check increase bundle size add runtime cost and we
can't realistically check all arguments to all functions.
Instead we should focus on adding more asserts on stuff that can't be
checked by TS.

### DIFF
--- a/src/execution/__tests__/executor-test.ts
+++ b/src/execution/__tests__/executor-test.ts
@@ -22,54 +22,6 @@ import { GraphQLSchema } from '../../type/schema';
 import { execute, executeSync } from '../execute';
 
 describe('Execute: Handles basic execution tasks', () => {
-  it('throws if no document is provided', () => {
-    const schema = new GraphQLSchema({
-      query: new GraphQLObjectType({
-        name: 'Type',
-        fields: {
-          a: { type: GraphQLString },
-        },
-      }),
-    });
-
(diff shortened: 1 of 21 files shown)


yaacovCR:
Not that I am against — but I believe the reason devAssertions have been retained is to help out those coding in JS without access to IDE help from TS. I don’t believe we ”forgot“ to remove them. We have chosen until now to keep supporting JS at this level.

I do agree, however, that in terms of our cost/benefit analysis, we should remove them — in favor of more generic help to JavaScript developers via the exported types. I hope our JavaScript users agree!

IvanGoncharov:
> I don’t believe we ”forgot“ to remove them. We have chosen until now to keep supporting JS at this level.

Yes, you're right it wasn't "forgotten" moreover I maintained them and migrated from generic `invariant` to more specific `devAssert` some time ago. But I think as time passes they become less useful for the ecosystem, so it's time to drop them.
That said we still have some of the runtime checks to enforce stuff not supported by the TS type system, e.g. exact types.

PR #3644, opened 2022-06-14
execute: integrate subscriptions and refactor pipeline

yaacovCR:
`execute` no longer runs the query algorithm for subscription operations. Rather, subscription operations are performed, as per the spec. `subscribe` and `createSourceEventStream` are removed.

See additional comments below.

### DIFF
--- a/integrationTests/ts/basic-test.ts
+++ b/integrationTests/ts/basic-test.ts
@@ -23,12 +23,13 @@ const queryType: GraphQLObjectType = new GraphQLObjectType({
 
 const schema: GraphQLSchema = new GraphQLSchema({ query: queryType });
 
-const result: ExecutionResult = graphqlSync({
-  schema,
-  source: `
+const result: ExecutionResult | AsyncGenerator<ExecutionResult, void, void> =
+  graphqlSync({
+    schema,
+    source: `
     query helloWho($who: String){
       test(who: $who)
     }
   `,
-  variableValues: { who: 'Dolly' },
-});
+    variableValues: { who: 'Dolly' },
+  });
--- a/src/__tests__/starWarsIntrospection-test.ts
+++ b/src/__tests__/starWarsIntrospection-test.ts
@@ -1,5 +1,7 @@
-import { expect } from 'chai';
+import { assert, expect } from 'chai';
 import { describe, it } from 'mocha';
+
+import { isAsyncIterable } from '../jsutils/isAsyncIterable';
 
 import { graphqlSync } from '../graphql';
 
@@ -8,6 +10,7 @@ import { StarWarsSchema } from './starWarsSchema';
 function queryStarWars(source: string) {
   const result = graphqlSync({ schema: StarWarsSchema, source });
   expect(Object.keys(result)).to.deep.equal(['data']);
+  assert(!isAsyncIterable(result));
   return result.data;
 }
 
--- a/src/execution/__tests__/executor-test.ts
+++ b/src/execution/__tests__/executor-test.ts
@@ -20,7 +20,7 @@ import {
 import { GraphQLBoolean, GraphQLInt, GraphQLString } from '../../type/scalars';
 import { GraphQLSchema } from '../../type/schema';
 
-import { execute, executeSync } from '../execute';
+import { execute, executeSubscriptionEvent, executeSync } from '../execute';
 
 describe('Execute: Handles basic execution tasks', () => {
   it('executes arbitrary code', async () => {
@@ -907,6 +907,19 @@ describe('Execute: Handles basic execution tasks', () => {
     expectJSON(
       executeSync({ schema, document, operationName: 'S' }),
     ).toDeepEqual({
+      errors: [
+        {
+          message:
+            'Schema is not configured to execute subscription operation.',
+          locations: [{ line: 4, column: 7 }],
+        },
+      ],
+    });
+
+    expectJSON(
+      executeSubscriptionEvent({ schema, document, operationName: 'S' }),
+    ).toDeepEqual({
+      data: null,
       errors: [
         {
           message:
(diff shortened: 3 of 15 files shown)


n1ru4l (approved):

yaacovCR:
Note that this PR:

= removes the `executeImpl` function introduced by #3639 in favor of `executeQueryOrMutation` and `executeSubscription` functions.
= renames the previous `executeSubscription` field to `executeSubscriptionRootField` both because it is more expressive and to prevent a clash.
= introduces new `rootFieldsExecutor` argument set to `executeQueryOrMutation` and `executeOperation`. Considering that the operation is an existing argument to these functions, it can be presumed that the caller knows best which algorithm to use, see also below.
= when implementing MapSourceToResponseStream, the algorithm calls `executeQueryOrMutation` with the new `rootFieldsExecutor` argument set to `executeFields`.
= the remaining `*Impl` function from #3639 is `createSourceEventStreamImpl`. As noted within #3639, this function may also be removed in later refactoring.

yaacovCR:
@graphql/graphql-js-reviewers @IvanGoncharov 

Note: 2, 3, and 4 above can be extracted as separate commits or as separate PRs, if desired. A portion of 1 can also be extracted, i.e. the renaming of `executeImpl`.

yaacovCR:
We may want to introduce the ability to force the algorithm irrespective of the operation type defined by the document.

The spec presents the subscription algorithm of CreateSourceEventStream and MapSourceEventToResponse as being handled by a single service, but then also notes that at scale it may be helpful to have them managed and scaled separately, as one is stateful (the former), while ExecuteSubscriptionEvent, the main component of the latter, is not.

It is much more common in practice within the JavaScript ecosystem for the two parts of the algorithm to be managed as a single service, but we do want to ensure that they can still be managed separately.

The “execute” function should be able to be used to implement the ExecuteSubscriptionEvent algorithm for the two service approach even after we integrate “subscribe” into “execute”. This new PR means that any operation with a subscription will be interpreted as the full subscription algorithm, so we need another option to “execute” like “forceQueryAlgorithm” so that two service subscription providers can still use the “execute” function.

yaacovCR:
This was already done in graphql-executor; I simply forgot to do so here; I can hopefully add it back tomorrow.

yaacovCR:
Another option is to include a new top level executeSubscriptionEvent export that can be used with two service subscriptions, without introducing a new argument to execute. This would mean that the service for execution of subscription events could not handle mutations, and the service for execution of mutations would never handle subscription events…

Yet another export, executeNonSubscriptions? executeQueryOrMutation could be used for queries and mutations not related to subscriptions. It would have similar behavior as to current export, but do nothing / fail when fed a subscription.

The different possible approaches mean that for now I have to hold off on this PR until I get more feedback either asynchronously or at the next js-wg.

yaacovCR:
I got some feedback offline from @IvanGoncharov positive in terms of the `executeSubscriptionEvent` approach. Introducing that new method will hopefully follow in a different PR.

yaacovCR:
Now that #3658 suggests that we may be awaiting a serious refactor, this PR has been reworked to represent a least change approach to the existing API. Rather than list the individual steps to the granularity of every renamed function, I will just list broadly what the PR accomplishes, after which I will list what the PR does not do.

1. updates `execute` to run the full ExecuteSubscription algorithm rather than just ExecuteSubscriptionEvent, updating the return types of `execute` and `graphql`
2. removes subscribe/createSourceEventStream (these would have to be deprecated in a 16.x release.
3. introduces `DocumentInfo` and `ExecutionInfo` interfaces that advanced clients can use to analyze a document and set up execution parameters. Internal `errors` variable -- the only necessarily private information required during execution -- is left within `ExecutionContext`. This is generally similar to the setup within defer/stream with payload context objects, so the separate variable is not much of an issue (in my opinion).
4. export helpers `buildDocumentInfo`, `coerceVariableValues`, as well as `buildExecutionInfo` so that advanced clients can have some help creating parts or all of the `DocumentInfo`/`ExecutionInfo`.
5. expose helpers `executeQuery`, `executeMutation`, `executeSubscription`, and `createSourceEventStream` which take `ExecutionInfo` rather than `ExecutionContext` objects. These functions return appropriately typed versions, i.e. `executeQuery`/`executeMutation` do not ever return streams (until defer/stream) merged.
6. no need anymore for the newly introduced `*Impl` functions within `execute`

What this PR does not do:
1. compile schemas/documents/field collections/variables/arguments
2. provide runtime or type checks for the arguments to `executeQuery`, etc. I.e. `executeQuery` can only take an `ExecutionInfo` object with an operation that is of type QUERY (for ExecuteQuery) or SUBSCRIPTION (for ExecuteSubscriptionEvent). `executeMutation` should only work for info objects containing a MUTATION, and `executeSubscription` and `createSourceEventStream` should only work for info objects containing a SUBSCRIPTION. That is because that TS does not (yet?) support [discriminated unions via nested properties](https://github.com/microsoft/TypeScript/issues/18758) and so it's not as easy to do without introducing repetitive type guards that drill down into the `ExecutionInfo` object or an additional somewhat unnecessary property on `ExecutionInfo` itself. Considering the type safety comes at a runtime expense, I am forgoing this for now.

yaacovCR:
The new `exported` helpers have not yet been exported from main, and are only available via deep imports. That can/should change once the API is finalized.

yaacovCR:
Ah => I forgot to mention that I have also snuck in here a new argument to allow total customization of the executor used by EXECUTE_SUBSCRIPTION_EVENT.

As well as a [default executor](https://github.com/yaacovCR/graphql-executor/blob/3f69fb3581931a9e6ac459430f0ba5d01332121f/src/execution/execute.ts#L1264-L1271): 

```ts
export const defaultSubscriptionEventExecutor = (
  exeInfo: ExecutionInfo,
  payload: unknown,
): PromiseOrValue<ExecutionResult> =>
  executeQuery({
    ...exeInfo,
    rootValue: payload,
  });
```

This workflow allows the `buildPerEventExecutionContext` to be removed.

yaacovCR:
Closing this for now in favor of plan in #4210

PR #3649, opened 2022-06-17
Use a type-safe way is differentiating ExecutionResult from Array<Error>

matthargett:
### Motivation

Checking a string key on an array is not valid, correctly yielding a warning on a TS branch I'm playing in.
Narrow the type with Array.isArray instead.

### Performance
Running `yarn benchmark` (which is amazing, btw) on my 2019 MacBook Pro running macOS 12.4 and node 18.2.0, I see a few slight increases and slight decreases of my local vs HEAD. I exited all other user processes on the system, but there was still some slight variability from run to run regardless.

Here's a screenshot of the run, let me know if I need to dive into CPU profiler data to find another type-safe alternative that's cheaper at runtime.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,9 +85,9 @@
     "prism-react-renderer": "1.3.3",
     "react": "17.0.2",
     "react-dom": "17.0.2",
-    "ts-node": "10.8.0",
-    "typedoc": "0.22.15",
-    "typescript": "4.6.4",
+    "ts-node": "10.8.1",
+    "typedoc": "0.22.17",
+    "typescript": "4.7.3",
     "url-loader": "4.1.1"
   },
   "publishConfig": {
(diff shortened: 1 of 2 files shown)


yaacovCR:
Checks failing because lockfile does not match, our ci requires you to update this yourself and then we check to make sure it matches.

Putting together your typescript version changes and your proposed code change, it sounds like you are saying that on newest version of typescript, we get a warning on our explicit property check?

It could be the simplest thing is to just disable the warning for that line.

yaacovCR:
Could be also that this is not secondary to the version change and just a different tsconfig in your project? I skimmed the TS 4.7 release notes and didn't note anything at first glance.

saihaj:
@github-actions run-benchmark

matthargett:
Superseded by #3670

PR #3655, opened 2022-06-21
polish: add tests for `expectPromise` and `expectEqualPromisesOrValues`

yaacovCR:
adds tests for `expectPromise` and `expectEqualPromisesOrValues`

#3620 introduced `expectEqualPromisesOrValues` with a TODO of moving to testUtils and testing.

This PR accomplishes that, and also moves the related `expectPromise` function, adding tests for that utility function as well.

### DIFF
--- a/src/__testUtils__/expectEqualPromisesOrValues.ts
+++ b/src/__testUtils__/expectEqualPromisesOrValues.ts
@@ -3,7 +3,7 @@ import { assert } from 'chai';
 import { isPromise } from '../jsutils/isPromise';
 import type { PromiseOrValue } from '../jsutils/PromiseOrValue';
 
-import { expectJSON } from './expectJSON';
+import { expectMatchingValues } from './expectMatchingValues';
 
 export function expectEqualPromisesOrValues<T>(
   items: ReadonlyArray<PromiseOrValue<T>>,
@@ -20,12 +20,4 @@ export function expectEqualPromisesOrValues<T>(
   }
 
   assert(false, 'Received an invalid mixture of promises and values.');
-}
-
-function expectMatchingValues<T>(values: ReadonlyArray<T>): T {
-  const remainingValues = values.slice(1);
-  for (const value of remainingValues) {
-    expectJSON(value).toDeepEqual(values[0]);
-  }
-  return values[0];
 }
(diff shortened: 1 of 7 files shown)


saihaj (approved):

src/__testUtils__/expectEqualPromisesOrValues.ts
```diff
@@ -0,0 +1,23 @@
...
+import { isPromise } from '../jsutils/isPromise';
+import type { PromiseOrValue } from '../jsutils/PromiseOrValue';
+
+import { expectMatchingValues } from './expectMatchingValues';
+
+export function expectEqualPromisesOrValues<T>(
+  items: ReadonlyArray<PromiseOrValue<T>>,
+): PromiseOrValue<T> {
+  const remainingItems = items.slice();
+  const firstItem = remainingItems.shift();
```
IvanGoncharov, line 12:
```suggestion
  const [firstItem, ...remainingItems] = items;
```

src/__testUtils__/expectMatchingValues.ts
```diff
@@ -0,0 +1,9 @@
+import { expectJSON } from './expectJSON';
+
+export function expectMatchingValues<T>(values: ReadonlyArray<T>): T {
+  const remainingValues = values.slice(1);
```
IvanGoncharov, line 4: same here

IvanGoncharov (approved):
I left a minor comment about using ES6 syntax otherwise ready to be merged 👍 

Note: In the long run, I don't think need to test both `subscribe` and `createSourceEventStream`. Currently, both functions are messy, but after the upcoming refactoring hopefully, we should have more modular code and don't need to check both. Similar to how we don't check the `graphql` function with all possible scenarios of the underlying function.

src/__testUtils__/expectEqualPromisesOrValues.ts
yaacovCR: cfefdc35

src/__testUtils__/expectMatchingValues.ts
yaacovCR: cfefdc35

PR #3657, opened 2022-06-22
introduce executeSubscriptionEvent

yaacovCR:
to replace the old exported functionality removed from execute by #3644

depends on #3644

### DIFF
--- a/integrationTests/ts/basic-test.ts
+++ b/integrationTests/ts/basic-test.ts
@@ -23,12 +23,13 @@ const queryType: GraphQLObjectType = new GraphQLObjectType({
 
 const schema: GraphQLSchema = new GraphQLSchema({ query: queryType });
 
-const result: ExecutionResult = graphqlSync({
-  schema,
-  source: `
+const result: ExecutionResult | AsyncGenerator<ExecutionResult, void, void> =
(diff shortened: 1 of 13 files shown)


IvanGoncharov (changes requested):
So we are back to square one where we duplicate work done by `createSourceEventStream`.
Adding a new API means we can't easily change or remove it so we need to solve the problem of duplicating work.
In #3658 I proposed exact same API but without this problem.
Merging this PR restricts design space for figuring out how to remove this duplication.
So I want to resolve the problem of duplication before merging it, so we sure API is compatible with this solution.

I think it's fair to block this change since a full alternative is provided in #3658

yaacovCR:
I don't understand, this is for use in a separate service that is set up to only run this function to manage subscriptions, so there is duplicative work of coercing because it is done by a separate service. We have discussed already how CreateSourceEventStream is not commonly used but this is available now for those who need it.

I do not see how your PR provides an alternative without this with

yaacovCR:
Edited above

yaacovCR:
I am folding this into #3644 because it provides functionality that was removed from by #3644. It makes sense to make it a separate commit within that PR.

Issue #3314, opened 2021-10-13
Query planning


IvanGoncharov:
# Problem

`execute` was designed as a simple implementation that follows GraphQL specification.
I think we should definitely for implementing anything that helps make it more usable in a typical graphql server.

However, there is growing demand to use `execute` in more complex tools/libraries to do stuff it wasn't designed to do (e.g. stitching, jit, etc.). Developers working on such libraries face a problem where they need to replace some parts of `execute` but don't want to support their own copy of `execute`.

One of the solutions is to convert the implementation of `execute` (as a class or separate function) into the framework with "hooks". It not only complicates code but also doesn't solve problems of libraries that want to do alternative execution concepts, for example, `graphql-jit`.

I want to propose an alternative solution that keeps `execute` as a simple library function but allows libraries to reuse the most complex parts and offer any "customization" they want to end-users.

It's sims that all other major implementations of GraphQL switched to query plans and are happy with it.
Having query plans have its drawbacks (it is slow and uses memory) but it's cachable and that mitigates them. 
Yes, it's a big shot to implement this feature but it solves the core issue that we discussed recently.

Idea is to split `execute` into two parts:
1. Generate a cachable query plan (with the query as a cache key) that has all possible things done ahead of execution (coercing args, collecting fields, preparing "static" part of the `info` object, etc.)
2. Simple executor that executes query plan, since all work is done ahead of time it should be faster especially for cases where you have nested lists types in your query.

And `#1` should solve the issue you have at the moment since even if you don't need all that data it's cachable so you can use only stuff that you need.
I expect all non-trivial libraries/tools (stitching, federation, graphql-jit) to replace `#2` with custom code so idea is to make it as small as possible move as much stuff as possible into `#1`.

Query plans can also be used in other use cases, for example, you can implement client-side code generation as "reducing visitor" on the query plan.

IvanGoncharov:
We can run this plan as a separate `experimentalExecutor` folder meaning it doesn't touch production code (a way simpler review process, mostly about code style and general feedback) and it doesn't interfere with spec proposals like `steam/defer`.

Please note: This is not a trivial task so it should be worked by a person deeply familiar both with graphql and `graphql-js` internals. Also, the goal is to develop this feature in collaboration with one of the tools/libraries that experience problems that I mentioned in the above comment.

IvanGoncharov:
@yaacovCR Recently we had a discussion about that recently it also related to the issues and PRs you opened recently.

[...]

PR #3658, opened 2022-06-22
Separate validation of operation and coercion of variables into separate steps

IvanGoncharov:
This draft PR shows what I'm currently working on.
It passes tests but fails due to lower coverage.
I will try to split it into smaller commits with proper descriptions and separate discussions.
The purpose of this Draft PR is to show the end goal of these smaller comments.

Any feedback/review is welcome but please keep in mind this PR is in the exploration phase (and code is still raw) so please focus only on high-level stuff.

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -1127,27 +1127,29 @@ function executeSubscription(
     rootType,
     operation.selectionSet,
   );
+
   const [responseName, fieldNodes] = [...rootFields.entries()][0];
-  const fieldName = fieldNodes[0].name.value;
-  const fieldDef = schema.getField(rootType, fieldName);
-
-  if (!fieldDef) {
-    throw new GraphQLError(
-      `The subscription field "${fieldName}" is not defined.`,
-      { nodes: fieldNodes },
-    );
-  }
-
   const path = addPath(undefined, responseName, rootType.name);
-  const info = buildResolveInfo(
-    exeContext,
-    fieldDef,
-    fieldNodes,
-    rootType,
-    path,
-  );
 
   try {
+    const fieldName = fieldNodes[0].name.value;
+    const fieldDef = schema.getField(rootType, fieldName);
+
+    if (!fieldDef) {
+      throw new GraphQLError(
+        `The subscription field "${fieldName}" is not defined.`,
+        { nodes: fieldNodes },
+      );
+    }
+
+    const info = buildResolveInfo(
+      exeContext,
+      fieldDef,
+      fieldNodes,
+      rootType,
+      path,
+    );
+
     // Implements the "ResolveFieldEventStream" algorithm from GraphQL specification.
     // It differs from "ResolveFieldValue" due to providing a different `resolveFn`.
 
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -174,34 +174,7 @@ export function execute(args: ExecutionArgs): PromiseOrValue<ExecutionResult> {
     return { errors: exeContext };
   }
 
-  // Return a Promise that will eventually resolve to the data described by
-  // The "Response" section of the GraphQL specification.
-  //
-  // If errors are encountered while executing a GraphQL field, only that
-  // field and its descendants will be omitted, and sibling fields will still
-  // be executed. An execution which encounters errors will still result in a
-  // resolved Promise.
-  //
-  // Errors from sub-fields of a NonNull type may propagate to the top level,
-  // at which point we still log the error and null the parent field, which
-  // in this case is the entire response.
-  try {
-    const { operation } = exeContext;
-    const result = executeOperation(exeContext, operation);
-    if (isPromise(result)) {
-      return result.then(
-        (data) => buildResponse(data, exeContext.errors),
-        (error) => {
-          exeContext.errors.push(error);
-          return buildResponse(null, exeContext.errors);
-        },
-      );
-    }
-    return buildResponse(result, exeContext.errors);
-  } catch (error) {
-    exeContext.errors.push(error);
-    return buildResponse(null, exeContext.errors);
-  }
+  return executeOperation(exeContext, exeContext.operation);
 }
 
 /**
@@ -330,12 +303,14 @@ export function buildExecutionContext(
 function executeOperation(
   exeContext: ExecutionContext,
   operation: OperationDefinitionNode,
-): PromiseOrValue<ObjMap<unknown>> {
+): PromiseOrValue<ExecutionResult> {
   const rootType = exeContext.schema.getRootType(operation.operation);
   if (rootType == null) {
-    throw new GraphQLError(
-      `Schema is not configured to execute ${operation.operation} operation.`,
-      { nodes: operation },
+    return buildErrorResponse(
+      new GraphQLError(
+        `Schema is not configured to execute ${operation.operation} operation.`,
+        { nodes: operation },
+      ),
     );
   }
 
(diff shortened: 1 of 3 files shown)


IvanGoncharov:

src/execution/execute.ts
```diff
@@ -220,41 +215,32 @@ export function executeSync(args: ExecutionArgs): ExecutionResult {
...
+  schema: GraphQLSchema;
+  operation: OperationDefinitionNode;
+  fragments: ObjMap<FragmentDefinitionNode>;
+  rootType: GraphQLObjectType;
+  fieldResolver: GraphQLFieldResolver<any, any>;
+  typeResolver: GraphQLTypeResolver<any, any>;
+  subscribeFieldResolver: GraphQLFieldResolver<any, any>;
+}
 
+type ResultOrGraphQLErrors<T> =
```
yaacovCR, line 234: Is this helpful in terms of https://github.com/graphql/graphql-js/pull/3649 ?? -- I'm afraid I did not try to reproduce that error, and have not yet hit it with my TS version.

src/execution/execute.ts
```diff
@@ -220,41 +215,32 @@ export function executeSync(args: ExecutionArgs): ExecutionResult {
...
+  | { errors: ReadonlyArray<GraphQLError>; result?: never }
+  | { result: T; errors?: never };
+
+export function makeExecutionPlan(
+  schema: GraphQLSchema,
+  document: DocumentNode,
+  operationName: Maybe<string>,
+  options: GraphQLExecutionPlanOptions = {},
```
yaacovCR, line 242: Why are some of these named options, but not others. Is the function going to be memoized?

src/execution/execute.ts
```diff
@@ -286,87 +272,321 @@ export function buildExecutionContext(
...
+    this.coercedValues = objMap;
+  }
+}
+
+export type ExecutableRequest =
+  | ExecutableQueryRequest
+  | ExecutableMutationRequest
+  | ExecutableSubscriptionRequest;
+
+export interface ExecutableQueryRequest {
```
yaacovCR, line 349: @IvanGoncharov  This is the key comment I believe in my review of this approach. It's more of a question than a comment.

I know I have asked elsewhere to export interfaces so that third parties can customize, add hooks, etc. I am generally happy to see the suggestion adopted, but I have to say that I don't know if the added complexity in this particular instance is so helpful. We currently have a pipeline of schema generation => request parse => request validate => request execute. Hidden in there also as a step is that without "assumeValid", schema validation is automatically performed.

My thinking in overall direction was to break out additional recommended steps, i.e., "validateSchema" would be its own step, request document parsing would be its own step, variable coercion would be its own step, and execute would then operation on the results of the previous stages, such that ExecutionContext would be broken down into its constituent parts, some cacheable (document), some not cacheable/not as cacheable (variables). Perhaps this interface/implementation approach has some additional benefits I am not appreciating?

src/execution/execute.ts
```diff
@@ -220,41 +215,32 @@ export function executeSync(args: ExecutionArgs): ExecutionResult {
...
-    operationName,
-    fieldResolver,
-    typeResolver,
-    subscribeFieldResolver,
-  } = args;
+interface ExecutionPlan {
+  schema: GraphQLSchema;
+  operation: OperationDefinitionNode;
+  fragments: ObjMap<FragmentDefinitionNode>;
+  rootType: GraphQLObjectType;
```
yaacovCR, line 228: It's nice that this is cacheable at this point, so that we don't have to repeatedly check that it exists for the same schema/operation.

src/execution/execute.ts
```diff
@@ -286,87 +272,321 @@ export function buildExecutionContext(
...
+  operationType: OperationTypeNode.QUERY;
+
+  coerceVariableValues: (
+    rawVariableValues: Maybe<{ readonly [variable: string]: unknown }>,
+  ) => ResultOrGraphQLErrors<CoercedVariableValues>;
+
+  /**
+   * Implements the "ExecuteQuery" algorithm described in the GraphQL specification.
+   */
+  executeOperation: (
```
yaacovCR, line 359: see above, I like how executeOperation now actually executes every operation (whereas previously the term was used for retrieving the data-only for queries/mutations). I just think we can deprecate the execute stage of the pipeline and utilize executeOperation, which would take a set of cacheable parameters (schema/document) as well as the 3 arguments here).

src/execution/execute.ts
IvanGoncharov: Yes, exactly.
I missed #3649 but I always dislike this particular check.
I think we need to have a generic convention and this one seems like one that suits our use case the best since the error response is a valid GraphQL response.

src/execution/execute.ts
IvanGoncharov: > I just think we can deprecate the execute stage of the pipeline and utilize executeOperation, which would take a set of cacheable parameters (schema/document) as well as the 3 arguments here).

We can do that as an intermediate step.
Just to clarify you are proposing using `WeakMap` internally to cache stuff based on schema/document/operationName as you do in `graphql-executor`?
Or are you proposing to use externally catchable `ExecutionPlan`as a replacement for schema/document/operationName?

src/execution/execute.ts
IvanGoncharov: > Hidden in there also as a step is that without "assumeValid", schema validation is automatically performed.

It's not hidden, it's "optional" you can perform it explicitly if you want.
All GraphQL servers I saw just call `validateSchema` and even our own `graphql` function does that, for example, https://github.com/graphql/express-graphql/blob/f4414b44996f25a5328e523d9a4b213fd1d70b16/src/index.ts#L268
But I agree it's a bad API design so I plan to work on the `GraphQLExecutableSchema` class to make it more explicit from an API standpoint.

> I know I have asked elsewhere to export interfaces so that third parties can customize, add hooks, etc. I am generally happy to see the suggestion adopted, but I have to say that I don't know if the added complexity in this particular instance is so helpful.

I still have the same position about hooks, this PR adds **interfaces** that can be used by external libraries to customize executor.
Different here is that we don't have any single line of code that can run 3rd-party `ExecutableRequest`, so it's not a hook.

I'm for adding configuration/interfaces if the below rules are satisfied:
1. Use case for which we add it is fully spec-compliant.
2. We can't incorporate this use case in graphql-js. For example, there is a "one size fits all" solution.

Custom GraphQL executors satisfy both rules, JIT is a valid use case but JIT is not a universal one (not everyone trusts the code that `eval`) it also applies some limitations: https://github.com/zalando-incubator/graphql-jit#differences-to-graphql-js
Even JIT aside, custom execute engines purposely built for complex scenarios like federation, and the stitching will always outperform generic ones even with all necessary hooks provided (hooks add their own cost and prevent certain JIT/compilation optimizations).
So it's a valid use case for adding interfaces.

This PR basically implements the https://github.com/graphql/graphql-js/issues/3314 proposal and defines clean interfaces that can be used by GraphQL servers to support specialized GraphQL executors.

> I have to say that I don't know if the added complexity in this particular instance is so helpful.

Just to be clear, are you referring to interfaces specifically or your comment is about the entire approach of using class instances vs having functions accept query plan as an argument?

src/execution/execute.ts
yaacovCR: The latter.

yaacovCR: But as I wrote above, although the beginnings of #3314 r here, there is very little that is being actually done and stored within the plan (afaict) besides storing the a validated root type. In terms of the exported interface, it also would not be enough (afaict) to build a customized stitching executor which needs access to later steps in the execution algorithm. Although I think I see the direction, in terms of functionality, I am not sure this as yet brings enough minimum value to be worth the complexity.

src/execution/execute.ts
yaacovCR: I don't think we need an ExecutableSchema class. I think if validateSchema is not called, the schema should not be validated, plain and simple.

In terms of your last question -- I am not sure I understand the distinction. I commented below in terms of relationship to #3314.

Basically, if the goal is to separate out coercion, and storing result of document analysis into separate step, I think that can be done as separate exported functions, like analyzeDocument and coerceVariables, and then the user can cache them. We actually don't have to deprecate execute, we can still include it for those who want it but just be explicit about which exported pipeline steps it performs: analyzeDocument, coerceVariables, getRootType, executeOperation...

src/execution/execute.ts
yaacovCR: I was constrained within graphql-executor because I wanted it to be a drop in replacement for graphql-js. So there, I used a simple memoization strategy that assumed if you have cached a document, everything about it should be cached.

If we are talking about changing the graphql-js api - I think exporting a few more low level building blocks is the low-hanging fruit.

yaacovCR:
Closing this old draft PR for now, we can always reopen to iterate again from here.

PR #3662, opened 2022-06-28
refactor: introduce GraphQLExecutableSchema

yaacovCR:
Storing the resolve function leads to a 30x performance boost for introspection execution (on my local machine).


yaacovCR:
yeah, when you're hitting errors!

PR #3665, opened 2022-07-04
[RFC] CCN: Validation

twof:
Validation implementation for the [Client Controlled Nullability RFC](https://github.com/graphql/graphql-spec/pull/895)

Follow up to https://github.com/graphql/graphql-js/pull/3418


twof:
Are we holding off on reviewing this until more discussions are resolved, or are we good to go on this PR?

PR #3666, opened 2022-07-05
Add parent name to KnownTypeNamesRule error message

blake-transcend:
## Description
Currently while validating graphql, the parent type is not specified in the error message when validation fails due to a missing type. Without the parent name this can be very difficult to track down in large projects, as the missing type could be used in many places.

## Contribution checklist
- [x] prettier
- [x] npm test

## Examples:
Before:
```
Unknown type "CommunicationMessage". Did you mean "CommunicationMetadata", "Communication", or "CommunicationInput"?
```
After:
```
Unknown type "CommunicationMessage" in type "Communication". Did you mean "CommunicationMetadata", "Communication", or "CommunicationInput"?
```

### DIFF
--- a/src/validation/__tests__/KnownTypeNamesRule-test.ts
+++ b/src/validation/__tests__/KnownTypeNamesRule-test.ts
@@ -64,11 +64,11 @@ describe('Validate: Known type names', () => {
       }
     `).toDeepEqual([
       {
-        message: 'Unknown type "JumbledUpLetters".',
+        message: 'Unknown type "JumbledUpLetters" in type "Foo".',
         locations: [{ line: 2, column: 24 }],
       },
       {
-        message: 'Unknown type "Badger".',
+        message: 'Unknown type "Badger" in type "Foo".',
         locations: [{ line: 5, column: 25 }],
       },
       {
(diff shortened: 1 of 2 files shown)


blake-transcend:
@saihaj Any idea on when this will be merged?

IvanGoncharov (changes requested):
> Currently while validating graphql, the parent type is not specified in the error message when validation fails due to a missing type. Without the parent name this can be very difficult to track down in large projects, as the missing type could be used in many places.

But you have the exact location embedded in error itself, also it nicely printed as code trace:
https://github.com/graphql/graphql-js/blob/67aefd9c1daefceacf937363c2da9e1117e550d9/src/error/__tests__/GraphQLError-test.ts#L271-L285

@blake-transcend Can you please explain why this is not enough in your use case?

> Any idea on when this will be merged?

This PR in its current form is problematic since if unknown type names are used in directives, fragments, and operation definitions, it is still referred to them as `in type`.

It can be fixed, but I don't think it worth added complexity.

yaacovCR:
@IvanGoncharov  Do u get the parent type in a code first schema?

blake-transcend:
Hey @IvanGoncharov , thanks for reviewing the PR and maintaining this library! There's a couple common use cases where this is beneficial:

1. Large types - When there are a large number of fields on the type, then the printed excerpt & location numbers will only contain a subset of the type, which can make it hard to identify which type they belong to. For example, if many of my types have the same field names (createdAt, updatedAt, id, etc), it would be hard to track down which type the surrounding snippet belongs to if the fields are all generically named.
2. Transpiled/generated schemas - It is relatively common for schemas to be generated/transpiled at runtime, which is problematic for this error because the location numbers don't correspond to any actual files and potentially aren't even static (so are basically worthless) and the errors that are printed may be swallowed by 3rd party tooling. Additionally, the ordering of fields may change between runs which makes the snippets and line numbers even more useless. This means that in order to debug which parent type actually needs to be fixed, you would need to either add console logs or debug inside of this library to see what was actually going on, which makes it really difficult to debug.

Updating the phrasing of the error message to be more generic or replace `type` with the actual type would both still retain the improvement in debuggability that this PR adds, however I'm not sure how to implement the latter (not very familiar with this code). Having this change implemented really helped me to debug this error when it happened while I was developing a feature, so I am pretty confident that it will also improve the experience of others as well.

PR #3671, opened 2022-07-10
Allow type overriding

stevefan1999-personal:
This means whichever last type that has the same name will triumph, if it follows a certain order.

Fixes https://github.com/axe-me/vite-plugin-node/issues/52

### DIFF
--- a/src/type/schema.ts
+++ b/src/type/schema.ts
@@ -212,7 +212,7 @@ export class GraphQLSchema {
       }
 
       const typeName = namedType.name;
-      if (this._typeMap[typeName] !== undefined) {
+      if (this._typeMap[typeName] !== undefined && !config.allowOverride) {
         throw new Error(
           `Schema must contain uniquely named types but contains multiple types named "${typeName}".`,
         );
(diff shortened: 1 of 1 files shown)


yaacovCR:
Thanks for starting this PR. Type overriding/merging is an interesting topic, so interesting that a [new working group subcommittee](https://github.com/graphql/composite-schemas-wg/) has been added to try to bring it into the spec.

QUESTION:

From what I can tell, the linked project has multiple __identical types__ of the same name that are __different__ instances, and you propose allowing "overriding" as a way around the error, which happens to be safe, because you happen to know the types are identical.

This PR, however, seems too broad a stroke to fix that source project problem, in that it allows type overriding even when it would be unsafe.

What do I mean by unsafe? If two types allow overriding, but actually differ, then the schema returned by introspection might not always be identical to the actual accessible fields at runtime, which would depend on which instance of the type is specified by the field return type, not by the instance of the type in the type map.

Reasoning a bit, it seems like it would be better to somehow ensure that you don't have different instances of identical types in your source project, i.e. that the "overrides" happen outside of `graphql-js` before the types are passed to `graphql-js` rather than within our code.

Please add more context, let us know your thoughts!

PR #3673, opened 2022-07-12
Add Anonymizer class

IvanGoncharov:
A super early draft of a class to Anonymise graphql documents and matching values in 3rd-party files.

Context: I have a report of performance degradation but can't access schema and queries to reproduce it. This PR attempts to make a utility class that obfuscates all names by using their salted hashes so I can get access to the "anonymized" version of files I need to reproduce this issue.

Note: **Not ready for review**, I'm publishing this PR just to generate a "canary NPM package" and pass it to the person that reported "performance degradation". After this design is proven I will work on cleaning up this PR.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -394,6 +394,7 @@ export type {
 
 // Utilities for operating on GraphQL type schema and parsed sources.
 export {
+  Anonymizer,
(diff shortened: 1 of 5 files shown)


IvanGoncharov:

PR #3674, opened 2022-07-13
fix vitest incompatibility

capaj:
ATM vitest/vite cannot load graphql correctly. It always ends up loading both CJS and ESM source in my projects.
I have reported this to vite here: https://github.com/vitejs/vite/issues/7879
but I think a simple change here could resolve the problem. I think using "module" entry point is obsolete/deprecated with node.js.

I have tested this change with latest vitests and it seems to work correctly with this change. I will be using patch-package with graphql 16 for now, but it would be nice if this could land.

Also why was `private: true` set here?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,14 @@
   "version": "16.5.0",
   "description": "A Query Language and Runtime which can target any service.",
   "license": "MIT",
-  "private": true,
-  "main": "index",
-  "module": "index.mjs",
+  "type": "commonjs",
+  "main": "index.js",
+  "exports": {
+    ".": {
+      "require": "./index.js",
+      "import": "./index.mjs"
+    }
+  },
   "typesVersions": {
     ">=4.1.0": {
       "*": [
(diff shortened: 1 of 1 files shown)


capaj:
I'll revert that change with private then, thanks @n1ru4l

n1ru4l:
I also think your PR will not be accepted because adding an exports map breaks all the people that do "deep imports" today, thus it is considered a breaking change. 😁

capaj:
Sounds about right. It would be best to just revert back to pure CJS without introducing and exports map.

capaj:
I should have checked your links before responding :D
I will just use https://npm.im/graphql-esm from https://github.com/graphql/graphql-js/pull/3361

seems like it is an official package-might be worth adding to the docs.
Let's close this one, hopefully @pabloszx has got it

yaacovCR:
To be discussed at Thursday June 25 5pm UTC meeting https://github.com/graphql/graphql-js-wg/blob/main/agendas/2025/06-Jun/25-graphql-js-wg-june-2025.md

We invite everyone to consider joining if you are able!

PR #3681, opened 2022-07-25
Fix coerceInputValue test name

spawnia:


IvanGoncharov:
@spawnia Thanks for taking time to submit this PR.

Issue #3648, opened 2022-06-15
Big performance drop due to switching TS output from ES5 to ES6


IvanGoncharov:
I investigated the performance drop between v15 and v16 and it was caused by switching transpilation target from es5 to es6.
It's easy to reproduce with the following diff:
```diff
diff --git a/tsconfig.json b/tsconfig.json
index edb522f1..299b9309 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
   ],
   "compilerOptions": {
     "lib": ["es2020"],
-    "target": "es2020",
+    "target": "es5",
     "module": "commonjs",
     "moduleResolution": "node",
     "strict": true,
```
Note: some of the functionality stops working due to an error in this file (we can't extend Map in es5):
https://github.com/graphql/graphql-js/blob/main/src/jsutils/AccumulatorMap.ts

But even benchmarks that are working showing pretty big performance difference:
**`local` is ES5 and `HEAD` is ES2020**

<img width="613" alt="image" src="https://user-images.githubusercontent.com/8336157/173902860-401f93f3-f0ab-4e9f-807d-da73f149bdb3.png">

I don't think we need to switch back to ES5 since it has its own drawbacks but instead, we just need to figure out what is slowing us and implement that code changes manually in our code base.

JoviDeCroock:
As a reference there are a handful of features that are ES6 that are a lot faster when transpiled into ES5, might be worth scanning the code base for some of [these](http://incaseofstairs.com/six-speed/) and maybe set up some transformations for the published files or adjust the code.

marvinhagemeister:
Like @JoviDeCroock said my hunch is that ES2015+ features aren't as well optimized as ES5 ones. `for..of` is a good candidate as that always has to invoke the iterator protocol because the array prototype could've been messed with somewhere.

FYI: [The table](http://incaseofstairs.com/six-speed/) linked was last run in 2020 and might not represent current engine optimizations.

JoviDeCroock:
Quickly did a few manual passes to confirm the assumptions and we seem to be on the right track here

<img width="467" alt="The build schema from AST benchmark going from 96.61 ops/sec to 160ops/sec" src="https://user-images.githubusercontent.com/17125876/183102958-303568a1-8c40-4f20-87ba-492066ce25bf.png">

StefanFeederle:
[...]

PR #3687, opened 2022-08-05
ES2020 optimizations

JoviDeCroock:
This PR aims to explain what's going on for #3648 the main issues seem to be:

- using `for..of` in hot paths
- using array destructuring

If we find this to be sufficient we can find a way to add a build step or merge as-is with some lint-rules. This currently touches the following issues in hot-paths only.

There are more issues to be found. Other slowdowns could be due to raw ES6 classes, ....
EDIT: verifies yes, ES5 classes are a lot faster like [these](https://www.runpkg.com/?graphql@15.8.0/type/schema.mjs#96) but imho it's not worth writing that as opposed to the little change of avoiding `for..of` in hot paths 😅 

I have a suspicion that the custom extension of Map (AccumulatorMap) is also to blame as it lives on the hot-path

Preliminary results where local is ES5 and the other is this branch. (some benches don't run when I use ES5 in tsconfig)

```
⏱   Recreate a GraphQLSchema
  2 tests completed.

  local                                    x 1,774 ops/sec ±0.21% x 288 KB/op (25 runs sampled)
  a6a240321b7ae96622eb120b7c1fa5eaafc6f16f x 1,386 ops/sec ±0.26% x 453 KB/op (22 runs sampled)

⏱   Build Schema from AST
  2 tests completed.

  local                                    x 145 ops/sec ±0.11% x  3.6 MB/op (9 runs sampled)
  a6a240321b7ae96622eb120b7c1fa5eaafc6f16f x 166 ops/sec ±0.58% x 3.55 MB/op (11 runs sampled)

⏱   Build Schema from Introspection
  2 tests completed.

  local                                    x 169 ops/sec ±0.13% x 3.95 MB/op (15 runs sampled)
  a6a240321b7ae96622eb120b7c1fa5eaafc6f16f x 163 ops/sec ±0.23% x 3.98 MB/op (12 runs sampled)

⏱   Execute Introspection Query
  2 tests completed.

  local                                    x 94,695 ops/sec ±1.99% x 2.85 KB/op (23 runs sampled)
  a6a240321b7ae96622eb120b7c1fa5eaafc6f16f x  74.84 ops/sec ±0.60% x 29.2 MB/op (5 runs sampled)
```

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -266,7 +266,7 @@ export function buildExecutionContext(
   let operation: OperationDefinitionNode | undefined;
   const fragments: ObjMap<FragmentDefinitionNode> = Object.create(null);
   for (let i = 0; i < document.definitions.length; i++) {
-    const definition = document.definitions[i]
+    const definition = document.definitions[i];
     switch (definition.kind) {
       case Kind.OPERATION_DEFINITION:
         if (operationName == null) {
@@ -432,9 +432,9 @@ function executeFields(
   const results = Object.create(null);
   let containsPromise = false;
 
-  const fieldsEntries = Array.from(fields.entries())
+  const fieldsEntries = Array.from(fields.entries());
   for (let i = 0; i < fieldsEntries.length; i++) {
-    const { 0: responseName, 1: fieldNodes } = fieldsEntries[i]
+    const { 0: responseName, 1: fieldNodes } = fieldsEntries[i];
     const fieldPath = addPath(path, responseName, parentType.name);
     const result = executeField(
       exeContext,
@@ -1230,7 +1230,9 @@ function executeSubscription(
     rootType,
     operation.selectionSet,
   );
-  const {0:responseName, 1:fieldNodes} = Array.from(rootFields.entries())[0];
+  const { 0: responseName, 1: fieldNodes } = Array.from(
+    rootFields.entries(),
+  )[0];
   const fieldName = fieldNodes[0].name.value;
   const fieldDef = schema.getField(rootType, fieldName);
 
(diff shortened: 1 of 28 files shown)


src/execution/execute.ts
```diff
@@ -431,7 +432,9 @@ function executeFields(
   const results = Object.create(null);
   let containsPromise = false;
 
-  for (const [responseName, fieldNodes] of fields.entries()) {
+  const fieldsEntries = Array.from(fields.entries());
+  for (let i = 0; i < fieldsEntries.length; i++) {
+    const { 0: responseName, 1: fieldNodes } = fieldsEntries[i];
```
IvanGoncharov, line 437:
```suggestion
    const [responseName, fieldNodes] = fieldsEntries[i];
```
This version is less wierd

JoviDeCroock: That version invokes the iterator and is very slow though

IvanGoncharov:
@JoviDeCroock Thanks for PR and research.
Can you please rebase, since I added a few related improvements on `main`?
I also made an experiment to see, if we can keep modern syntax without a performance drop.
I published it as #3691, what do you think about it?

BTW, your branch is still faster in some tests but slower in others:
image
Maybe rebasing will fix this.

JoviDeCroock:
@IvanGoncharov #3691 optimizes some problematic ones, you might want to add a plugin like https://babeljs.io/docs/en/babel-plugin-transform-destructuring or the one from react-hooks because they transform `const [x, y] = useState` to `const { 0: x, 1: y }` there as well. I think that might be the last performance gap we have between our benches, A last point of optimization might be avoiding using `extends` on the `Map` built-in.

If you want I can also close this as your transform approach looks better to me, thank you!

PR #3693, opened 2022-08-12
Update deps

IvanGoncharov:


n1ru4l (approved):

package.json
```diff
@@ -85,12 +85,15 @@
...
+    "ts-node": "10.9.1",
+    "typedoc": "0.23.10",
     "typescript": "4.7.4",
     "url-loader": "4.1.1"
   },
   "publishConfig": {
     "tag": "alpha"
+  },
+  "dependencies": {
+    "npm-check": "^6.0.1"
```
yaacovCR, line 97: Should this be a dev dependency or not included at all or is this intentional?

yaacovCR (changes requested):

package.json
```diff
@@ -85,12 +85,15 @@
...
   }
```
saihaj, line 98:
```suggestion
  }
```

saihaj (changes requested):

package.json
IvanGoncharov: @yaacovCR Good, catch.
I just forgot to add `--global` to a command.

package.json
IvanGoncharov: @saihaj Thanks, good catch from you and @yaacovCR
Just forgot to add `--global`, fixed.

PR #3699, opened 2022-08-16
ci: add dependency review workflow

IvanGoncharov:
Motivation: increase security.
Discovered while looking into #3162


IvanGoncharov:
Also, it reduces pressure on maintainers to review PRs that add new dependencies.
We only have `devDependencies`, but they can still compromise developer machines (e.g., still GH or NPM keys) or compromise code inside the published NPM package.

PR #3720, opened 2022-08-31
Filter subsequent payloads when parent field is null

robrichard:
Implementing what was discussed in https://github.com/robrichard/defer-stream-wg/discussions/45

### DIFF
--- a/src/execution/__tests__/stream-test.ts
+++ b/src/execution/__tests__/stream-test.ts
@@ -58,7 +58,21 @@ const query = new GraphQLObjectType({
           scalarField: {
             type: GraphQLString,
           },
+          nonNullScalarField: {
+            type: new GraphQLNonNull(GraphQLString),
+          },
           nestedFriendList: { type: new GraphQLList(friendType) },
+          deeperNestedObject: {
+            type: new GraphQLObjectType({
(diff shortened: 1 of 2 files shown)


benjie:
I don't think this is the approach I would have taken. Instead:

1. When a nullable position becomes null because of an error, add this path to a new set on `exeContext` (let's call the new set `errorNulledPaths` since it's a list of the paths that were nulled out due to an error caught by a nullable type)
2. In `filterSubsequentPayloads`, find and cancel the subsequentPayloads whose path _starts with or matches_ one of the paths in `errorNulledPaths` - no need to traverse the response data

Assuming that the length of paths is negligible, I think this basic approach would scale at O(N*M) where N is the number of errors and M is the number of subsequent payloads, but I'm sure there are clever ways to optimize it better.

robrichard:
@benjie I significantly rewrote this. Now, as discussed in the WG today, I've added the logic to filter subsequent payloads at the point where the bubbled null is ultimately set. This no longer requires inspecting the payload response and should not add any overhead in the non-error case.

yaacovCR (approved):

benjie:
Definitely looks better; thanks @robrichard!

PR #3727, opened 2022-09-06
incrementalDelivery: original execute should not throw on new directives

yaacovCR:
original execute and executeSync should ignore `@defer` and `@stream` rather than throw.

extracted from #3726

### DIFF
--- a/src/execution/__tests__/defer-test.ts
+++ b/src/execution/__tests__/defer-test.ts
@@ -1,4 +1,3 @@
-import { expect } from 'chai';
 import { describe, it } from 'mocha';
 
 import { expectJSON } from '../../__testUtils__/expectJSON.js';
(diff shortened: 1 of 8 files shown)


robrichard:
Why is it preferable to ignore the directives rather than throw?

yaacovCR:
We are talking about operations to pass validation because the directives are in the schema. If for some reason the server has not actually enabled the incremental delivery behavior, it is not the fault of the client. I am just having a hard time figuring out why they should throw.

Basically, what I believe we are trying to do is enable experimental behavior. If the behavior is not implemented, it should behave just as before, no?

yaacovCR:
Alternative to #3722

robrichard:
@yaacovCR would it be more likely that someone is trying to get defer/stream to work but doesn't realize they are using the wrong functions/not passing the experimental flag? And in that case, an error message would be helpful to guide them to correct usage?

yaacovCR:
Suppose a server developer mistakenly thought that the alpha included defer/stream support without the use of an experimental function/flag.

So they upgraded GraphQL to alpha, sent in a query with defer, and disappointingly got just a single response.

I do think an error would help that developer.

But how likely is that? Upgrading a server to support incremental delivery requires preparing for execution to return an async iterable rather a single result... But the return type of the original execute just includes a single response -- so why would a developer think it would work?

Perhaps because the developer is using JavaScript and not TS and also does not have IDE support for types?

But we recently basically pulled support for JS #3642 ?

I also truly do not know what is more likely. I can imagine that a developer might by accident (or even on purpose?) ship to production a schema with the directives included, but not want to yet enable it in prod. That developer would expect that clients would continue to receive results.

I can imagine a different solution that might better help a developer. Rather than only failing queries with incremental results, we could modify the original execute to immediately throw if the defer/stream directives are in the schema, or at least print a warning. We could also modify the new function/flag to throw or print a warning if both directives are not in the schema.

robrichard:
>Suppose a server developer mistakenly thought that the alpha included defer/stream support without the use of an experimental function/flag.
>
>So they upgraded GraphQL to alpha, sent in a query with defer, and disappointingly got just a single response.
>
> I do think an error would help that developer.

Of course I have no data to back this up, but it does seem to me that would be the more likely case since we will have many people to set this up as this gets closer to inclusion in the spec.

> I can imagine that a developer might by accident (or even on purpose?) ship to production a schema with the directives included, but not want to yet enable it in prod. That developer would expect that clients would continue to receive results.

But if the schema is including these directives, then the spec says they should be supported, so I don't think we would want to cater to that behavior.

> We could also modify the new function/flag to throw or print a warning if both directives are not in the schema

The spec says defer & stream are optional, and one can be supported without the other so I don't think we should error if the directives are not in the schema. And ultimately we want to remove the experimental functions/flags
I don't think we would want to do this, since ultimately we would want to remove the experimental

yaacovCR:
Closing in favor of #3722.

PR #3728, opened 2022-09-07
incrementalDelivery: refactoring and streamlining

yaacovCR:
Each commit is meant to merged separately, i.e. "Rebase and Merge" rather than "Squash and Merge"

419fd37f: moves `executeStreamField` out of try block to better illustrate which code might throw
2991c79a: tiny refactor pushing the result of `handleFieldError` rather than explicit null, following convention elsewhere
71a8db4f: rename `fieldPath` to `itemPath` when dealing with items, as elsewhere
4df02ca7: skip unnecessary filtering when asyncIterator's next method throws, aligning stream to non-stream

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -1948,7 +1948,6 @@ async function executeStreamIteratorItem(
   } catch (rawError) {
     const error = locatedError(rawError, fieldNodes, pathToArray(itemPath));
     const value = handleFieldError(error, itemType, asyncPayloadRecord.errors);
-    filterSubsequentPayloads(exeContext, itemPath, asyncPayloadRecord);
     // don't continue if iterator throws
     return { done: true, value };
   }
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -1045,9 +1045,8 @@ async function completeAsyncIteratorValue(
         handleFieldError(error, itemType, errors);
       }
     } catch (rawError) {
-      completedResults.push(null);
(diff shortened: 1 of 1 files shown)


robrichard (approved):

src/execution/execute.ts
```diff
@@ -996,44 +996,35 @@ async function completeAsyncIteratorValue(
...
+        exeContext,
+        itemType,
+        fieldNodes,
+        info,
+        itemPath,
+        asyncPayloadRecord,
+      )
+    ) {
+      containsPromise = true;
+    }
```
IvanGoncharov, line 1027:
```suggestion
    containsPromise = completeListItemValue(
      iteration.value,
      completedResults,
      errors,
      exeContext,
      itemType,
      fieldNodes,
      info,
      itemPath,
      asyncPayloadRecord,
    )
```

IvanGoncharov: same in other place

IvanGoncharov (dismissed):

yaacovCR:
this will have to come after #3746 and #3747

src/execution/execute.ts
yaacovCR: These commits have been separated into #3752 -- but I don't think we want the suggested change, as we only want to store `containsPromise` if it's true. In particular, we do not want to overwrite an already `true` value with `false`.

yaacovCR:
`completeListItemValue` has been separated out into #3728 -- this is now ready for merge.

PR #3730, opened 2022-09-12
feat: allow defining symbol error extensions

n1ru4l:
Using symbol keys is a useful approach when composing error handlers as symbol keys are omitted when the error is finally stringified.

It allows middlewares to read extensions without having to assign them to `undefined` afterwards or using the `delete` keyword.


saihaj (approved):

benjie (approved):
Seems like a reasonable feature.

benjie:
Argh; we can't have it because typescript 4.3 doesn't support symbols so it fails to pass CI. I think I must not have checked that CI had passed. Rolling back.

PR #3732, opened 2022-09-13
incrementalDelivery: use single execute function with directives as flags 

yaacovCR:
The use of new experimental functionality triggered by directives does not need a separate runtime flag. The directives themselves are the flags and all operations passed to `execute` are assumed to be valid.

Using TS conditional types, when new generic `TMaybeIncremental` parameter is set to true, return type of `execute` will be either an ExecutionResult or an IncrementalExecutionResultMap containing an `initialResult` and a generator of `subsequentResults`. While incremental delivery is designated as experimental, `TMaybeIncremental` will default to `false`. When incremental delivery is merged into the spec, it may be useful to adjust the default to `true`.

### DIFF
--- a/src/execution/__tests__/defer-test.ts
+++ b/src/execution/__tests__/defer-test.ts
@@ -1,8 +1,6 @@
-import { expect } from 'chai';
 import { describe, it } from 'mocha';
 
 import { expectJSON } from '../../__testUtils__/expectJSON.js';
-import { expectPromise } from '../../__testUtils__/expectPromise.js';
 import { resolveOnNextTick } from '../../__testUtils__/resolveOnNextTick.js';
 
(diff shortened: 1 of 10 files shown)


yaacovCR:
My current preferred alternative to #3722 / #3727 / #3726
@robrichard @glasser @IvanGoncharov

yaacovCR:
Even if we don't take this solution in whole, we might want to extract out the removal of the unnecessary `singleResult` wrapper and just add that to #3722

yaacovCR:
> Even if we don't take this solution in whole, we might want to extract out the removal of the unnecessary `singleResult` wrapper and just add that to #3722

This portion has been extracted out to #3733

yaacovCR:
Closing in favor of #3722.

PR #3742, opened 2022-09-19
Remove defer/stream support from subscriptions

robrichard:
Discussion topic: https://github.com/robrichard/defer-stream-wg/discussions/55

This implements what was agreed on in the November 2022 (Secondary, EU) WG. There was consensus to move forward with disabling defer/stream on subscriptions via a field error. This also adds a new validation rule that prevents @defer or @stream from being used inside a subscription operation without an if argument that can be used to disable defer/stream.


src/execution/execute.ts
```diff
@@ -1478,90 +1483,15 @@ export const defaultFieldResolver: GraphQLFieldResolver<unknown, unknown> =
  *
  * This function does not support incremental delivery (`@defer` and `@stream`).
  * If an operation which would defer or stream data is executed with this
- * function, each `InitialIncrementalExecutionResult` and
- * `SubsequentIncrementalExecutionResult` in the result stream will be replaced
- * with an `ExecutionResult` with a single error stating that defer/stream is
- * not supported.  Use `experimentalSubscribeIncrementally` if you want to
- * support incremental delivery.
+ * function, the result stream will be replaced with an `ExecutionResult`
+ * with a single error stating that defer/stream is not supported.
```
yaacovCR, line 1487: I think that with the second commit, this needs to be tweaked a bit => as far as I can tell, field errors are thrown for each payload attempting to utilize incremental delivery, and so the returned `ExecutionResult` may contain data and potentially multiple errors, one of which being the aforementioned field error.

yaacovCR (commented):
Other than small comment above regarding comment, the implementation looks good to me.

More generally, I am also fine with leaving incremental delivery support with subscriptions for a potential later stage to help advance the initial release.

src/execution/execute.ts
robrichard: Thanks for catching. This comment is updated now.

yaacovCR (approved):

robrichard:
@IvanGoncharov this is ready for review, now implementing what was discussed at the November 2022 (Secondary, EU) WG. The [spec PR](https://github.com/graphql/graphql-spec/pull/742) has also been updated to reflect this.

src/validation/rules/DeferStreamDirectiveOnValidOperationsRule.ts
```diff
@@ -0,0 +1,82 @@
...
+      if (operation.operation === OperationTypeNode.SUBSCRIPTION) {
+        for (const fragment of context.getRecursivelyReferencedFragments(
+          operation,
+        )) {
+          fragmentsUsedOnSubscriptions.add(fragment.name.value);
+        }
+      }
+    },
+    Directive(node, _key, _parent, _path, ancestors) {
+      const definitionNode = ancestors[2];
```
robrichard, line 51: Is this the correct way to get the top level ancestor (either fragment definition or operation type)?

IvanGoncharov: Yes, this rule can be simplify but this could be done is subsequent PRs

PR #3748, opened 2022-09-22
Fix typo in the error message for max tokens

Cito:
Note, this also exists in the 16.x.x branch.


IvanGoncharov:
> Note, this also exists in the 16.x.x branch.

Thanks, @Cito if you have time pr for v16 would be great.

PR #3753, opened 2022-09-28
fix(incrementalDelivery): filtering should never affect the error source

yaacovCR:
The current asyncPayload should always be passed to the filtering function so that filtering will never affect the error source.


robrichard (approved):

src/execution/execute.ts
```diff
@@ -2052,8 +2056,8 @@ async function executeStreamIterator(
 
 function filterSubsequentPayloads(
   exeContext: ExecutionContext,
-  nullPath?: Path,
-  currentAsyncRecord?: AsyncPayloadRecord,
+  nullPath: Path | undefined,
+  currentAsyncRecord: AsyncPayloadRecord | undefined,
```
IvanGoncharov, line 2060: It looks like every call has both args, can you just make them non-optional?

IvanGoncharov (changes requested):

src/execution/execute.ts
yaacovCR: done

IvanGoncharov (approved):

src/execution/execute.ts
yaacovCR: The nullPath is always passed, and so we can remove the possibility of undefined there, but the currentAsyncRecord could be undefined if it is the initial response that is erroring, and so we can make it non-optional, but we cannot remove the possibility of undefined.

PR #3754, opened 2022-09-30
enhancement: remove extra ticks

yaacovCR:
Promise resolution order changes in some instances, resulting in different orders for some errors within the errors array, as well as in different values of hasNext within incremental delivery payloads.

This PR introduces an async `completePromisedValue` helper function rather than using a promise chain (see below links).

https://github.com/tc39/proposal-faster-promise-adoption
https://github.com/tc39/ecma262/issues/2770
https://github.com/tc39/ecma262/pull/2772
https://github.com/tc39/ecma262/pull/1250
https://v8.dev/blog/fast-async

Depends on #3793

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -715,7 +715,7 @@ function executeField(
     const result = resolveFn(source, args, contextValue, info);
 
     if (isPromise(result)) {
-      return completePromise(
+      return completePromisedValue(
         exeContext,
         returnType,
         fieldNodes,
@@ -912,7 +912,7 @@ function completeValue(
   );
 }
 
-async function completePromise(
+async function completePromisedValue(
   exeContext: ExecutionContext,
   returnType: GraphQLOutputType,
   fieldNodes: ReadonlyArray<FieldNode>,
@@ -1182,7 +1182,7 @@ function completeListItemValue(
 ): boolean {
   if (isPromise(item)) {
     completedResults.push(
-      completePromise(
+      completePromisedValue(
         exeContext,
         itemType,
         fieldNodes,
@@ -1911,7 +1911,7 @@ function executeStreamField(
     exeContext,
   });
   if (isPromise(item)) {
-    const completedItems = completePromise(
+    const completedItems = completePromisedValue(
       exeContext,
       itemType,
       fieldNodes,
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -715,25 +715,15 @@ function executeField(
     const result = resolveFn(source, args, contextValue, info);
 
     if (isPromise(result)) {
-      const completed = result.then((resolved) =>
-        completeValue(
-          exeContext,
-          returnType,
-          fieldNodes,
-          info,
-          path,
-          resolved,
-          asyncPayloadRecord,
-        ),
+      return completePromise(
+        exeContext,
+        returnType,
+        fieldNodes,
+        info,
+        path,
+        result,
+        asyncPayloadRecord,
       );
-      // Note: we don't rely on a `catch` method, but we do expect "thenable"
-      // to take a second callback for the error case.
-      return completed.then(undefined, (rawError) => {
-        const error = locatedError(rawError, fieldNodes, pathToArray(path));
-        const handledError = handleFieldError(error, returnType, errors);
-        filterSubsequentPayloads(exeContext, path, asyncPayloadRecord);
-        return handledError;
-      });
     }
 
     const completed = completeValue(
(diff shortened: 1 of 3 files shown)


src/execution/execute.ts
```diff
@@ -506,6 +508,8 @@ function buildPerEventExecutionContext(
   return {
     ...exeContext,
     rootValue: payload,
+    subsequentPayloads: new Set(),
```
yaacovCR, line 511: This bug fix needs to be in a separate commit

src/execution/execute.ts
```diff
@@ -2247,31 +2337,41 @@ class DeferredFragmentRecord {
...
+  streamContext: StreamContext;
   iterator: AsyncIterator<unknown> | undefined;
   isCompletedIterator?: boolean;
   isCompleted: boolean;
   _exeContext: ExecutionContext;
   _resolve?: (arg: PromiseOrValue<Array<unknown> | null>) => void;
   constructor(opts: {
     label: string | undefined;
-    path: Path | undefined;
+    path: Path;
```
yaacovCR, line 2384: This type narrowing could be in a separate commit.

src/execution/execute.ts
yaacovCR: Moved out of this PR.

src/execution/execute.ts
yaacovCR: Moved out of this PR.

yaacovCR:
This PR has been pruned to only include the changes the result in tick decreases -- and is now ready for review!

yaacovCR:
In terms of comparing the number of ticks, see the following, adapted from https://github.com/tc39/ecma262/issues/2770

```js
const resolved = Promise.resolve();

function dummyAsync(n) {
  return new Promise(resolve => resolve(n+1));
}

function returnPromiseWithThen() {
  return result.then(r => dummyAsync(r));
}

function returnPromiseWithThenAndCatch() {
  return result.then(r => dummyAsync(r)).then(undefined, () => 0);
}

async function asyncHelperWith2Awaits(p) {
  try {
    const result = await p;
    return await dummyAsync(result);
  } catch (err) {
    return 0;
  }
}

function returnAsyncWith2Awaits() {
  return asyncHelperWith2Awaits(result);
}

async function test(fn) {
  let done = false;
  let count = 0;
  fn().then(() => { done = true; });

  function counter() {
    if (done) {
      console.log(`${fn.name} took ${count} ticks`);
    } else {
      resolved.then(() => {
        count++;
        counter();
      });
    }
  }
  counter();
}

async function tests() {
  await resolved;
  await test(returnAsyncWith2Awaits);
  await test(returnPromiseWithThen);
  await test(returnPromiseWithThenAndCatch);
}

tests();
```

which emits on my instance of chrome:

```
returnAsyncWith2Awaits took 3 ticks
returnPromiseWithThen took 4 ticks
returnPromiseWithThenAndCatch took 5 ticks
```

Additional async helpers may be of use elsewhere; I have focused here so far on value completion.

NOTE:  The commits within this PR are designed to be added individually (rebased, rather than squashed).

yaacovCR:
in particular, the `completeValueCatchingErrors` helper could be improved!

yaacovCR:
I added [89521b3](https://github.com/graphql/graphql-js/pull/3754/commits/89521b3acffc3a87479a36a9fcb7ce5f44cdb691)
 to "optimize" `completeValueCatchingErrors`.

Benchmarking would presumably tell us whether the tick reduction is outweighted by the number of small helpers.

Awaiting the return of our friendly github-action-bot.  @IvanGoncharov

robrichard (approved):

yaacovCR:
Somewhat old write up: https://v8.dev/blog/fast-async but seems to still be true?

yaacovCR:
@github-actions run-benchmark

yaacovCR:
@github-actions run-benchmark

yaacovCR:
I removed the sync helper as it doesn't really help anything.

yaacovCR:
image

If I'm running this right, tests are a bit noisy (running on WSL on my laptop), but looks like a speed bump of 58%....

Which is good! Note that this comes entire from this PR rather than #3793 which is `HEAD~1`

dotansimha (approved):

yaacovCR:
#3796 formalizes the approach in this PR across the code base using eslint rules to prefer async functions over then/catch. It removes another extra tick.

src/execution/execute.ts
```diff
@@ -922,6 +912,41 @@ function completeValue(
   );
 }
 
+async function completePromise(
```
IvanGoncharov, line 915: The name is a little bit too generic.
@yaacovCR How about `completePromisedValue`?

IvanGoncharov (changes requested):
left, a name suggestion.
As far as every else, it is ready to merge.

src/execution/execute.ts
yaacovCR: Sounds great!

265e6873

yaacovCR:

yaacovCR:
[See edited comment above](https://github.com/graphql/graphql-js/pull/3754#issue-1392076280), the performance benefit with this PR does not have to in the end with faster promise adoption; it seems to have to do with subtracting a `.then(...)` call. We can probably close #3796 as the general case is not helpful.

yaacovCR:
This PR no longer results in a performance savings after #3886. Reverting this PR from the commit right before #3886 yields the following performance drop:

image

But with #3886, reverting this PR, actually leads to a performance improvement!

image

Why? .....

yaacovCR:
Above results were obtained with the help of `git bisect`

yaacovCR:
clearly this does not have to do with ticks and presumably has something to do with JIT changes ???

yaacovCR:
I have played around with [deopt-explorer](https://github.com/microsoft/deoptexplorer-vscode) trying to explore what optimization changes `completePromisedValue` might introduce and in what context.

There are definite changes in what is optimized/deoptimized with this functions. In terms of profiling, however, I could not see how this was tied to performance, because using deopt-explorer, it was never faster with this optimization, for reasons unknown.

These benchmarks are very synthetic => we use the same operation repeatedly, have a single field, and a single rootValue. My expectation is that v8's ability to optimize would be challenged with a mixture of complex operations, and root values. All of this is internal behavior, and reworking graphql-js in ways that make the code more complex to trick v8 into producing a set of more optimized functions for a synthetic test case that may not influence real world performance does not seem like the best idea.

So I am much more bearish on this PR from the get-go. My inclination even at this point is to revert it prior to the v17 release, especially as it no longer seems to produce the desired synthetic performance benefit.

PR #3756, opened 2022-10-05
polish: a few smaller refactors

yaacovCR:
1. prune try block within `completeAsyncIteratorValue` to better reflect what might throw
2. type result of iterator `.next()` calls
3. replaced `handleError` helper with `addError` with the new function also calling `locatedError`
4. introduce `handleRawError` helper which also includes filtering of async payloads
5. pass subsequentPayloads rather than entire exeContext

extracted from #3754

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -687,7 +687,6 @@ function executeField(
   path: Path,
   asyncPayloadRecord?: AsyncPayloadRecord,
 ): PromiseOrValue<unknown> {
-  const errors = asyncPayloadRecord?.errors ?? exeContext.errors;
   const fieldName = fieldNodes[0].name.value;
   const fieldDef = exeContext.schema.getField(parentType, fieldName);
   if (!fieldDef) {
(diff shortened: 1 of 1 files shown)


saihaj (approved):

src/execution/execute.ts
```diff
@@ -749,18 +748,18 @@ function executeField(
...
-        return handledError;
+        return null;
       });
     }
     return completed;
   } catch (rawError) {
-    const error = locatedError(rawError, fieldNodes, pathToArray(path));
-    const handledError = handleFieldError(error, returnType, errors);
+    const errors = asyncPayloadRecord?.errors ?? exeContext.errors;
+    addError(rawError, fieldNodes, returnType, path, errors);
```
IvanGoncharov, line 768: Not sure about this one, especially since `addError` throws located error.
But since it's not a public API, let's try and merge it.

Future suggestion: If we convert ExecutionContext into class it can have single method called `handleFieldError` that:
1. `locateError`
2. push error into `errors`
3. `filterSubsequentPayloads`

IvanGoncharov (approved):
Makes sense 👍 
@yaacovCR Thank you for extracting it 🙇

yaacovCR:
Probably makes sense to wait on this until #3815 is approved, which could require a spec change. Filtering is complex!

yaacovCR:
refactors 3 and 4 from this were merged in https://github.com/graphql/graphql-js/pull/3878

some of the remaining minor changes from this refactoring PR perhaps could be revisited, but I am closing this stale PR for now.

PR #3757, opened 2022-10-05
fix: re-initialize subsequentPayloads

yaacovCR:
`buildPerEventExecutionContext` must use a new rootValue, and re-initialize the execution "state" portion of the `ExecutionContext`


yaacovCR:
This bug could be pretty short-lived if #3742 is merged.

PR #3759, opened 2022-10-18
fix(github-actions-bot): add additional permissions

yaacovCR:
sub-workflows cannot be called from parent workflow without sufficient permissions

### DIFF
--- a/.github/workflows/github-actions-bot.yml
+++ b/.github/workflows/github-actions-bot.yml
@@ -89,6 +89,8 @@ jobs:
   cmd-publish-pr-on-npm:
     needs: [accept-cmd]
     if: needs.accept-cmd.outputs.cmd == 'publish-pr-on-npm'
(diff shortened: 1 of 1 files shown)


yaacovCR:

yaacovCR:
@IvanGoncharov I believe this change fixes the startup issue with the github-actions-bot.

I noticed that on this PR that itself changes the workflow, the unpatched workflow from main is being utilized. Is that the desired behavior?

See: https://github.com/graphql/graphql-js/actions/runs/3272036713/workflow

IvanGoncharov:
> I noticed that on this PR that itself changes the workflow, the unpatched workflow from main is being utilized. Is that the desired behavior?

@yaacovCR Yes, it is done for security reasons since we allow bot to do restricted stuff (e.g. posting comments, publishing to npm, etc.)

Issue #3629, opened 2022-06-06
`extendSchema()` doesn't set `GraphQLSchema.description` in some cases


sachindshinde:
`extendSchema()` doesn't set `GraphQLSchema.description` if the `schema` definition is in the input `GraphQLSchema` instead of the AST.

**Example**
Running this code:
```typescript
import { buildSchema, extendSchema, parse } from "graphql";

const extendedSchema = extendSchema(
  buildSchema(`
    "A schema description"
    schema {
      query: Foo
    }
    
    type Foo {
      foo: String
    }
  `),
  parse(`scalar Bar`),
);

console.log(`AST description: ${extendedSchema.astNode?.description?.value}`);
console.log(`GraphQLSchema description: ${extendedSchema.description}`);
```
will print this:
```
AST description: A schema description
GraphQLSchema description: undefined
```

PR #3776, opened 2022-11-13
extendSchema: preserve "description" and "extensions"

igrlk:
👋 

Closes https://github.com/graphql/graphql-js/issues/3629

### Current behavior:

When calling `extendSchema(schema, schemaModification)`, extended schema description is calculated as `schemaDef?.description?.value`, where
* `schemaDef` is a _nullable_ schema definition, that will not be null only if `schemaModification` declares the `type schema` (it happens only when we call `buildSchema` _or_ call `extendSchema(schema, 'extend type __Schema { ... }')`)
* `schemaDef?.description` is not undefined only when `extendSchema` is called from `buildSchema`, otherwise it's undefined because we don't allow a construction like:
```
"my awesome description" // error - graphql-js doesn't allow you to set the description when you extend schema
extend type __Schema { ... }
```

As a result, every `extendSchema` call besides the one in `buildSchema` will cause `description` to become `undefined`

### Proposed solution:
Always fallback to the original schema description: 
```description: schemaDef?.description?.value ?? schemaConfig.description```

It will allow us to make sure that we take a description from `schemaDef` when schema is being built and take original schema description in any other cases

### DIFF
--- a/src/utilities/__tests__/extendSchema-test.ts
+++ b/src/utilities/__tests__/extendSchema-test.ts
@@ -196,6 +196,27 @@ describe('extendSchema', () => {
     expect(extendedTwiceSchema.getType('ID')).to.equal(GraphQLID);
   });
 
+  it('copies original schema description to extended schema', () => {
+    const description = 'A schema description';
+
+    const extendedSchema = extendSchema(
+      buildSchema(`
+        "${description}"
+        schema {
+          query: Foo
+        }
+        
+        type Foo {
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
Hi @igrlk
Thank you for this PR, I also noticed that we have the exact same problem with `extensions` so I fixed it in the same PR.

igrlk:
Thank you, @IvanGoncharov!

PR #3778, opened 2022-11-17
Remove outdated instructions on experimental tag

robrichard:
I don't think we need any additional instructions since now the code is on main and published under the main alpha tag?


saihaj (approved):
I think now on v17 alpha it goes out as a default

PR #3784, opened 2022-11-22
extract out publisher

yaacovCR:
## Motivation:

1. separation of concerns
2. creating a base for publisher refactoring

## To elaborate on number 2:

Currently, emitting a stream utilizes the event loop to ensure that all stream items are emitted in order, with an extra microtask for each stream item. Utilizing the event loop in this way will be slower than doing so synchronously -- for cases that allow the latter.

In other words, if a late completing early stream item completes and many later stream items are then available for publishing; we currently waterfall through promise resolution for each stream item, rather than adding all the now available incremental results.

Extracting out our current publisher logic allows us to iterate on a proposal that potential has feature improvements such as a resolution of the above potential performance issue.


yaacovCR:
Now depends on #3787 which was extracted from #3786

yaacovCR:
Note that this PR depends on the "bug" fix (addressing an extra then) from #3787 and only the single initial commit from that PR changes tests. The remainder of the commits extract out the current publisher step by step, without changing any tests.

See https://github.com/graphql/graphql-js/pull/3784/files/773c90cd5d602bb4a15af814e753272bf40bf80e..189519a871afec0529cf77a540a19117e14564b1

yaacovCR:
This PR corresponds to a proposed PR on the existing ["incremental delivery" spec PR](https://github.com/graphql/graphql-spec/pull/742):

See https://github.com/robrichard/graphql-spec/pull/8

yaacovCR:
This has been rebased to longer include #3787. (The change within #3787 has been re-integrated into the follow-up to this PR, #3786.)

yaacovCR:
this was reworked and merged as #3903

PR #3786, opened 2022-11-24
introduce new Publisher for incremental delivery

yaacovCR:
Depends on #3784

The proposed new Publisher:

1. does not use the event loop to manage AsyncRecord dependencies
2. uses separate sets to store pending vs ready AsyncRecords
3. does not use `Promise.race`

## No event loop for managing AsyncRecord dependencies

The current Publisher wraps every AsyncRecord result in a promise that only resolves after the parent AsyncRecord resolves. If multiple items within a stream are released for publishing because their parent has just been published, the stream cannot be in its entirety until after all of these promises unwind.

The new Publisher keeps track of dependencies manually. When an AsyncRecord is pushed by the publisher, all of its dependent AsyncRecords are synchronously pushed, repeating as necessary, without using the event loop.

## Separate sets for pending vs ready AsyncRecords

The current publisher inspects all pending AsyncRecords whenever any of them resolves. All that are completed are added to the response. The new Publisher moves AsyncRecords from the pending set to the ready set as they are pushed, so that on each call to next, only the ready set must be iterated.

As a side-effect of this change, the incremental array is ordered by which items are ready for delivery first, and not by the initial document.
 This seems like a worthwhile tradeoff, and is still adherent to the spec, as far as I can tell.

## No `Promise.race`

The old Publisher uses `Promise.race` as the trigger to determine whether payloads are ready. The new Publisher uses a single triggering promise that is triggered whenever an AsyncRecord is pushed, and then reset. This may be beneficial as the implementation of `Promise.race` within V8 has a known memory leak for long-running promises. (see https://bugs.chromium.org/p/v8/issues/detail?id=9858). An alternative would be to utilize @brainkim 's memory-safe version detailed in that issue.


yaacovCR:
This PR depends on #3784 which extracts out the current publisher from the execution workflow without changing any tests.

The final commit (currently 9be87be) is the single commit unique to this PR that swaps out the old publisher for the new and demonstrates how the executor uses it differently and how batching of incremental payloads change.

1. How does publisher use by the executor change with the new publisher?

Currently, the executor adds a `promise` field to `AsyncPayloadRecord`s containing a promise that will resolve when the payload is ready. The promise would not resolve unless the `AsyncPayloadRecord`'s parent was also ready, using the event queue to manage the hierarchy.

The new Publisher introduces a new  `complete` method. The executor calls this method when appropriate to signal to the Publisher than an `AsyncPayloadRecord` is ready. The Publisher itself manages the parent hierarchy. The `complete` method checks to see whether an `AsyncPayloadRecord` can be pushed, and puts it on waiting queue as necessary; when the parent is pushed, the waiting queues are synchronously drained as possible.

2. How do tests change with the new publisher?

As a result of the fact that an `AsyncPayloadRecord`s can be completed synchronously, there are often fewer distinct payloads. In these cases, the `incremental` list within each payload contains a greater number of AsyncPayloadRecords.

yaacovCR:
This PR corresponds to a proposed PR on the existing ["incremental delivery" spec PR](https://github.com/graphql/graphql-spec/pull/742):

See https://github.com/robrichard/graphql-spec/pull/9

It turns out that the existing spec PR have introduced this asynchronous chaining into the spec as well. The associated spec PR follows the new implementation in removing it.

yaacovCR:
Failing tests for this PR after rebasing on main appear to have to do with bug identified in #3815. Apparently, the new Publisher is able to surface this issue with the current test suite after #3754 landed.

#3815 will need to be solved separately in main, and presumably this PR can be rebased on top of that.

yaacovCR:
this was reworked and merges as https://github.com/graphql/graphql-js/pull/3894

PR #3787, opened 2022-11-25
fix(incrementalDelivery): remove extra then from  AsyncRecord completion

yaacovCR:
extracted from #3786


yaacovCR:
Looking back on this. I'm not sure why I extracted this from #3786 -- I think the change that it entails a spec change just as much as the remainder of #3786 and should not be merged separately.

PR #3791, opened 2022-12-01
Allow execution to be cancelled

igrlk:
Hey everyone 👋 

This PR addresses https://github.com/graphql/graphql-js/issues/3764

### How it works
It extends `ExecutionArgs` with `signal?: AbortSignal`.
Then, during the execution, we check whether signal is present and aborted. If that's the case, we throw an error and fail the further execution of resolvers.

### Testing
To test it with a real graphql server (and to possibly give an implementation reference to the apollo-server team), I've made a [change to Mercurius](https://github.com/igrlk/mercurius/pull/1) so that it passes `abortSignal` to the `graphql.experimentExecuteIncrementally` and aborts the signal if request connection is closed.

I've also created a [minimal reproducible repo](https://github.com/igrlk/mercurius-graphql-execution-cancelling) that you can clone and test the execution cancelling on your machine 🔥 

### Feedback
Please let me know if that change makes sense to you.

@glasser that would be awesome if you can play around with apollo-server to see if this way of execution cancelling can be used there :) I deployed this PR to npm and you can install it by `npm install --save  graphql@canary-pr-3791`

@mcollina, I would love to hear your thoughts too!

Cheers! 🎩


igrlk:

igrlk:
@github-actions run-benchmark

igrlk:
@github-actions run-benchmark

src/execution/execute.ts
```diff
@@ -328,9 +333,29 @@ export function experimentalExecuteIncrementally(
   return executeImpl(exeContext);
 }
 
+function subscribeToAbortSignal(exeContext: ExecutionContext): {
+  unsubscribeFromAbortSignal: () => void;
+} {
+  const onAbort = () => {
```
glasser, line 339: On the other hand, I don't think you need any of this. The AbortSignal already has an `aborted` boolean property.  Since all we're doing is checking whether things are aborted inside completeValue, that can just check `aborted` directly, with no subscription needed. (Though if you take my suggestion from the top level of the review to always have an AbortController and hook up the user-provided one to it, then you'll still need the structure of subscribe/unsubscribe — it just won't be updating a boolean.)

src/execution/execute.ts
```diff
@@ -261,6 +265,7 @@ export interface ExecutionArgs {
   fieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
   typeResolver?: Maybe<GraphQLTypeResolver<any, any>>;
   subscribeFieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
+  signal?: AbortSignal;
```
glasser, line 268: ExecutionArgs is also the argument to `subscribe`/`experimentalSubscribeIncrementally`. So this PR allows you to specify `signal` to subscription APIs. I would expect that if I passed `signal` to a subscription API and then I aborted the signal at some point in the middle of a subscription (say, between source events), that it would also complete the subscription somehow, but I don't think your implementation does this.

Personally I think it would be reasonable for this PR to only work with non-subscription execution (and subscription support could be added later) — so my suggestion is for `experimentalSubscribeIncrementally` to throw if `signal` is provided, unless you're familiar with subscriptions and are motivated to make it work there too!

glasser, line 268: I think it would be reasonable to add this to the top-level `GraphQLArgs`/`graphql` function as well.

glasser (changes requested):
I'm not enough of an execution expert to know if the place where you're checking the signal is the best spot, but I'm really excited to see this!  Use cases include:

- Stopping execution when the client closes their request.
- Stopping execution after a request timeout.
- Stopping execution when it evaluates too many fields.

I do have one suggestion! There's an issue I've know about in graphql-js for a while which apparently I never filed as a bug, so I just did that: https://github.com/graphql/graphql-js/issues/3792

If we're going to add this AbortSignal support to graphql-js, I'd suggest doing something slightly differently, where graphql-js *always* creates a new AbortController which it passes to execution the way to do here.  If a signal is passed in then you do `passedInSignal.addEventListener('abort', () => abortController.abort(passedInSignal.reason))`. Finally, when execution has finished and there's a complete response to be passed back (the same places you're unsubscribing) it can abort the controller it created. This means that if there's any pieces of execution doing work that it doesn't know contributes to an already-rejected `Promise.all`, that work will stop recursing.

Another thought that doesn't go on a particular line: as a potential extension which doesn't need to be part of this PR, the signal could be made available to resolvers on `GraphQLResolveInfo`, so that resolvers can themselves abort their work when the operation is aborted.

src/execution/execute.ts
```diff
@@ -832,6 +867,14 @@ function completeValue(
   result: unknown,
   asyncPayloadRecord?: AsyncPayloadRecord,
 ): PromiseOrValue<unknown> {
+  if (exeContext.signal?.isAborted) {
```
glasser, line 870: Hmm, also: I think completeValue is probably the wrong place to put this, or at least if you do put this here you need to be a bit more careful. Because if `result` is an array of Promises, now you are "leaking" the Promises by not calling `catch` or putting them in a `Promise.all` or whatever, which seems incorrect... it's generally a bad idea to have unhandled Promises and can lead to weird logging and such.

I don't know the best place to put this check (and maybe it should go in multiple places) but putting it at the top of executeField (ie, *before* doing a resolver call, not midway through after) seems reasonable?

(Or you could leave it here but also be careful to find any `Promise`s in `result` and put an empty `catch` on them or something...)

glasser:
There's some CI failures. The integration test failures indicate that you probably either need to require some `@types/node` in packages that depend on `graphql`, or you can get the TS `AbortSignal` from some other package.

One of my suggestions involves the need to actually `new AbortController()` at runtime too, which raises the question of what implementation to use from being a TS issue to a runtime issue. I believe we need graphql-js (even execution, right?) to work in browsers and other non-Node environments, though I don't know how old the browsers we need to support are (and I don't know if we need to support environments *other* than browsers and Node). But in any case we still support Node v14 and the AbortController and AbortSignal classes aren't available as globals (without an experimental CLI flag) until v15, so you need some sort of polyfill anyway... so might as well use the polyfill for TS as well.

In Apollo Server we've been happy with https://www.npmjs.com/package/node-abort-controller but that's Node-specific; https://www.npmjs.com/package/abort-controller is fully portable I think.

Although I guess `graphql-js` has zero dependencies (is that an absolute requirement @IvanGoncharov ?) In that case the choices (assuming you like my suggestion that graphql-js should use its own AbortController to resolve #3792) are to only fix the #3792 bug in Node 16 and newer, or to copy a polyfill of AbortController into graphql-js.

yaacovCR:
Second that this is very exciting. My thoughts around adding an abort mechanism were more centered on #3792 ie not receiving an abortsignal, but rather passing it into resolvers.

I assumed that we would end up needing to create a separate AbortController for every nullable field, so that we could abort resolvers as soon as a null bubbled up. I hadn't thought of creating a single AbortController for everything and simply aborting after the result has been returned, ie @glasser's suggestion above. The down side is that we end up having to wait to cancel anything until we get the final result, while the upside is that it's easier to manage and may have a performance boost in the non-error case.

Following along here with interest!

glasser:
@yaacovCR yeah I think we could do a more complex thing to manage extraneous execution more precisely (perhaps by moving away from Promise.all to something more explicit) but that could be an iterative improvement later.

src/execution/execute.ts
igrlk: Would be happy to try implementing it for Subscriptions but I think it makes sense to reduce this PR scope to minimal.

I added error throwing in case a client tries to pass an abort signal to subscriptions execution, thanks for the suggestion!

igrlk: Not quite sure I understand the idea, can you explain why?

The way I understand it: clients (graphql servers) are calling `execute/experimentalExecuteIncrementally` and that's exactly the place where they will be passing an abort signal.

src/execution/execute.ts
igrlk: Decided to go with the option you suggested and create a new AbortController for each execution

src/execution/execute.ts
glasser: There's a function called `graphql` that you can use that wraps a bunch of stuff (parsing, validation, execution) into a single function call. So it might be reasonable for users of that API to be able to set a signal (it's just a matter of adding it to GraphQLArgs and threading it through to execute).

igrlk:
> There's some CI failures

@glasser, I've added dummy AbortSignal class in case of Node@14 - tests pass now!

src/execution/execute.ts
igrlk: Thanks. I added signal to the GraphQLArgs/graphql call

src/jsutils/AbortController.ts
```diff
@@ -0,0 +1,38 @@
...
+// We need to polyfill abort controller in case of Node@14 which doesn't have it
+/* c8 ignore start */
+export const AbortController: new () => IAbortController =
+  // eslint-disable-next-line no-undef
+  global.AbortController ||
+  // This is a dummy implementation that doesn't actually abort anything
+  class DummyAbortController implements IAbortController {
+    private _signal: IAbortSignal = {
+      aborted: false,
+      addEventListener: () => null,
```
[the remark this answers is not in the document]
igrlk: Are there any plans on dropping support for node@14? Would be cool to track it down somewhere so that people know when it's going to happen.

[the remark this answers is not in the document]
glasser: I don't know if graphql-js has any particular plans, but Node v14 does go out of LTS in April. (I'm not sure what non-Node/browser platforms graphql-js has to work on though?)

It might be reasonable to just only do the "cancel stuff when execution is over" if `global.AbortController` exists. It's a bit more annoying to do this partial support if we add an AbortSignal to GraphQLRequestInfo since it would be nice if the TS type of the field is non-optional — but we could have that field be optional for now (with a "if you know you're in an AbortController-in-global environment then you can `!` it") and non-optional once Node 14 support is dropped.

IvanGoncharov:
@igrlk Sorry for the delay in the review.
I like the idea of reusing the abort functionality of a fetch.
Before merging, I want to do a quick check to ensure that API matches the W3C standards.

Also, I want to try removing delays from a test since we don't use them anywhere else, and the performance of the tests is critical for [mutation testing](https://github.com/graphql/graphql-js/blob/main/.github/workflows/mutation-testing.yml).
If delays can't be easily removed, I'm ok to merge it as is, and I just want to try before merging.

That said, plan to check as much as possible and merge it tomorrow.

IvanGoncharov:
Update: I managed to remove `setTimeout` from tests but stuck on few other things including
>@glasser's question https://github.com/graphql/graphql-js/pull/3791#discussion_r1038543598 regarding where the check for aborted execution should happen - in completeValue or no and how we should handle it. Do you have an opinion on that?@glasser's question https://github.com/graphql/graphql-js/pull/3791#discussion_r1038543598 regarding where the check for aborted execution should happen - in completeValue or no and how we should handle it. Do you have an opinion on that?@glasser's question https://github.com/graphql/graphql-js/pull/3791#discussion_r1038543598 regarding where the check for aborted execution should happen - in completeValue or no and how we should handle it. Do you have an opinion on that?

Need to think about it more, will continue tomorrow morning.

IvanGoncharov:
@igrlk Sorry for the long delay. It's ended up being a more complicated task than I initially thought.
I made some changes:
1. Pass signal into resolvers
2. move the place where we interrupt execution before we call resolver per @glasser suggestion
3. Fix interruption of steam/defer
4. Implement it on subscriptions (need to fix some edge cases)
It failing some lints + need more testing, mainly steam/defer and subscription.
But all existing tests are passing, including the new test added by this PR.
Plan to finish it tomorrow but want to get early feedback on the changes.

src/execution/execute.ts
```diff
@@ -261,6 +262,7 @@ export interface ExecutionArgs {
   fieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
   typeResolver?: Maybe<GraphQLTypeResolver<any, any>>;
   subscribeFieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
+  signal?: AbortSignal | undefined;
```
igrlk, line 265: Is AbortSignal defined globally in node@14? I think that I intentionally created an interface for it since it was not defined there, but I might be mistaken, need to double check

src/execution/execute.ts
```diff
@@ -504,10 +514,37 @@ export function buildExecutionContext(
...
     subscribeFieldResolver: subscribeFieldResolver ?? defaultFieldResolver,
     subsequentPayloads: new Set(),
+    executionController: new ExecutionController(signal),
     errors: [],
   };
 }
 
+class ExecutionController {
+  /** For performance reason we can't use `signal.isAborted` so we cache it here */
+  isAborted: boolean = false;
```
igrlk, line 524: For more context, what are the performance implications you mention here?

src/execution/execute.ts
```diff
@@ -504,10 +514,37 @@ export function buildExecutionContext(
...
+
+  private readonly _passedInAbortSignal: AbortSignal | undefined;
+  private readonly _abortController: AbortController | undefined =
```
igrlk, line 527: Same as about `AbortSignal` - I believe it was causing an error in node@14 so I had to add the interfaces that match `AbortSignal` and `AbortController`

src/execution/execute.ts
```diff
@@ -504,10 +514,37 @@ export function buildExecutionContext(
...
+  constructor(signal?: AbortSignal) {
+    this._passedInAbortSignal = signal;
+    this._passedInAbortSignal?.addEventListener('abort', this._abortCB);
+  }
+
+  get signal(): AbortSignal | undefined {
+    return this._abortController?.signal;
+  }
+
+  abort(reason?: unknown) {
```
igrlk, line 539: Can `reason` be typed as a string here?

igrlk (commented):
Thanks a lot for making this huge change @IvanGoncharov! I left some comments - please let me know what you think.

I'm also happy to follow up on this and finalise it (linting and other things left before merging). But I'll only have time by the end of the week.

src/execution/execute.ts
IvanGoncharov: It's not, but TypeScript has typings for it.
Typescript knows what it is even on node@14, but it's only during type check in execution you can't call it on node@14 that's why I add `typeof AbortController === 'undefined'` check.

src/execution/execute.ts
IvanGoncharov: I will add a screenshot tomorrow.
But you can run `npm run benchmark`(locally) on top of my commit and see the difference.

src/execution/execute.ts
IvanGoncharov: I will check on node@14. Maybe my assumption about TS detecting this type on node@14 is wrong.

igrlk:

igrlk:
Screenshot 2023-02-10 at 18 12 44

AbortController and AbortSignal were not defined and failing the tests so I added the interfaces for them

igrlk:

igrlk:
@github-actions run-benchmark

igrlk:
@IvanGoncharov I tried to fix all the PR errors but there is one I'm not sure how to fix: C8 on Node 14 reports about missing coverage but there is no real missing coverage, c8 ignore also doesn't help. Do you have an idea of what it can be?

Besides that, I tested the latest version of the branch using [this repo](https://github.com/igrlk/mercurius-graphql-execution-cancelling) that just does simple scenario - 3 nested resolvers. The change works as expected (Although we now see error message in the console when execution stops, is it expected?)

Screenshot 2023-02-19 at 13 46 09

MamiStyle (approved):

igrlk:
Thanks, @ladal1
I've rebased the branch, fixed linters, but there are some failing tests. I don't have time to check it this week. Would you happen to have time to look into that?

ladal1:
Interesting, the tests were passing on my end when I copied it, thanks for the rebase, I'll take a look into those tests and try to make them pass ASAP

igrlk:
Thanks @ladal1 , I've invited you to be a collaborator in the fork

ladal1:
@github-actions run-benchmark

ladal1:
I'm fairly sure the interrupt is fixed, tests pass for me now, seems that either test edit of rebase mess caused the issues

igrlk:
We intentionally had `executionController.abort()` statements in all those places, though now I don't remember why exactly. Here is the code before rebase and resolving conflicts:
```
try {
    const result = executeOperation(exeContext);
    if (isPromise(result)) {
      return result.then(
        (data) => {
          const initialResult = buildResponse(data, exeContext.errors);
          if (exeContext.subsequentPayloads.size > 0) {
            return {
              initialResult: {
                ...initialResult,
                hasNext: true,
              },
              subsequentResults: yieldSubsequentPayloads(exeContext),
            };
          }

          exeContext.executionController.abort();
          return initialResult;
        },
        (error) => {
          exeContext.executionController.abort();
          exeContext.errors.push(error);
          return buildResponse(null, exeContext.errors);
        },
      );
    }

    const initialResult = buildResponse(result, exeContext.errors);
    if (exeContext.subsequentPayloads.size > 0) {
      return {
        initialResult: {
          ...initialResult,
          hasNext: true,
        },
        subsequentResults: yieldSubsequentPayloads(exeContext),
      };
    }

    exeContext.executionController.abort();
    return initialResult;
  } catch (error) {
    exeContext.executionController.abort();
    exeContext.errors.push(error);
    return buildResponse(null, exeContext.errors);
  }
```

igrlk:
Re-comitting from my account to run pr checks, @ladal1

ladal1:
@igrlk I thought they weren't intentional as they didn't appear in the copy I had, I'll try to take a look as to why they would be there and if it's possible to pass without removing them

ladal1:
Can't find why it should be there (and especially in all branches of the logic)

igrlk:

igrlk:
@ladal1 thanks for checking. I tried to reproduce cancellation in [this repo](https://github.com/igrlk/mercurius-graphql-execution-cancelling), but if previously it was working as expected (at least in the initial versions of this PR), now it seems to hang and not finish the execution at all + the execution is not being aborted

JoviDeCroock:
Implemented in https://github.com/graphql/graphql-js/pull/4250

PR #3793, opened 2022-12-06
polish: do not repeat isPromise check

yaacovCR:
If a field resolves asynchronously, then it will always complete asynchronously.

Adds additional test cases to complete code coverage.

### DIFF
--- a/src/execution/__tests__/stream-test.ts
+++ b/src/execution/__tests__/stream-test.ts
@@ -402,6 +402,54 @@ describe('Execute: stream directive', () => {
       },
     ]);
   });
+  it('Can stream a field that returns a list with nested promises', async () => {
+    const document = parse(`
+      query { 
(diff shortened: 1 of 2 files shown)


yaacovCR:
This has been extracted from #3754.

yaacovCR:
On my machine this gives only a nominal speed improvement, so any other speed improvement would have to be from  #3754 

image

yaacovCR:
> On my machine this gives only a nominal speed improvement, so any other speed improvement would have to be from #3754
> 
> image

I think I’m using the revs option incorrectly here, will update further if I get it right

yaacovCR:
The purpose of this PR is just to make the diff for #3754 smaller and to test to make sure where the speed improvements for #3754 are coming from. See https://github.com/graphql/graphql-js/pull/3754#issuecomment-1339961683

saihaj (approved):

PR #3795, opened 2022-12-08
Use explicit extension for package.json main field

ArnaudBarre:
This package can't be used inside a custom Vite plugin for a cjs project.

This is due to the fact that Vite reimplement a lot of node resolution logic to help pre-bundle the config file. This logic doesn't handle correctly extension-less main entrypoint which causes it to then resolve `index` to `index.mjs`.

I think it doesn't hurt to be have an explicit extension and would avoid yet another edge case in this complex part of Vite code.

Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "A Query Language and Runtime which can target any service.",
   "license": "MIT",
   "private": true,
-  "main": "index",
+  "main": "index.js",
   "module": "index.mjs",
   "typesVersions": {
     ">=4.1.0": {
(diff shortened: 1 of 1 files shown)


yaacovCR:
@ArnaudBarre this PR also adds `types` to be in `package.json` as well? That currently causes build scripts to error. Is `types` also a requirement for Vite?

ArnaudBarre:
No I did add `types because of the error of the first run: https://github.com/graphql/graphql-js/actions/runs/3652890851/jobs/6180925545#step:5:47

Not sure of adding types is the right fix, because the CI fails early in the second run

PR #3796, opened 2022-12-09
add eslint rules for promises

yaacovCR:
prefer-await-over-promises helps avoid extra ticks, formalizing the optimization from #3754 across the board

various additional rules add additional best practices with regard to promise use with at most trivial code changes.


yaacovCR:
This ends up being a "bug fix" because it does remove an extra tick and so is observable.

.eslintrc.yml
```diff
@@ -196,7 +196,7 @@ rules:
   promise/no-return-wrap: error
   promise/param-names: off
   promise/prefer-await-to-callbacks: off
-  promise/prefer-await-to-then: off
+  promise/prefer-await-to-then: error
```
IvanGoncharov, line 199: I'm afraid I have to disagree with this one.
It creates more verbose code that is harder to read, especially in the executor.
Can we do optimization using just `then`?

IvanGoncharov (changes requested):

.eslintrc.yml
yaacovCR: nope, then is broken

package.json
```diff
@@ -77,6 +77,7 @@
     "eslint-plugin-import": "2.26.0",
     "eslint-plugin-internal-rules": "file:./resources/eslint-internal-rules",
     "eslint-plugin-node": "11.1.0",
+    "eslint-plugin-promise": "^6.1.1",
```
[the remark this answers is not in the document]
yaacovCR: 2c916a43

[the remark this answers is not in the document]
yaacovCR: i mean 8a624719

.eslintrc.yml
yaacovCR: but instead of using an immediately invoked anonymous arrow function, we could use arrow functions stored in variables. or even named functions, defined even on the top level; if the functions are defined on the top level, we would have to pass parameters rather than rely on the closure. I don't have a preference per se.

See [this comment](https://github.com/graphql/graphql-js/pull/3754#issue-1392076280) for links on how `then` is comparatively slower than async functions. I don't know how we can do optimizations and still use `then`. The entire point of #3754 and this PR is to avoid then as until it is fixed, it is too easy to unnecessarily add extra ticks.

yaacovCR: I settled on introducing new helpers `after`, `tryAfter`, `catchAfter` rather than using `.then` or immediately invoked anonymous arrow functions.

This reduces the diff considerably and I think addresses the readability concern.

We can also use these helpers to get rid of completePromisedValue, which I am about to do.

yaacovCR: Actually, probably better to do in a separate PR so that the diff is cleaner in `execute.ts`.

yaacovCR:
#3801 extracted from this PR => with this PR now rebased on top of #3801

.eslintrc.yml
yaacovCR: Ready for re-review.

yaacovCR:
For simplicity, I have dropped the specialized helpers, provider just a single `after` helper than works pretty similar to `.then`.

yaacovCR:
Closing this, [see edited comment on #3754](https://github.com/graphql/graphql-js/pull/3754#issue-1392076280).

The real world implications of the extra tick within `.then(...)` as opposed to async functions is not clear. It does still seem prudent to consolidate `.then(...)` calls when possible, as extra `.then(...)` calls seem to have been the reason for the performance benefit of #3754.

PR #3798, opened 2022-12-13
remove extra ticks from executeStreamField

yaacovCR:
...by using async helper functions.


yaacovCR:
updated, no longer depends on #3796

PR #3802, opened 2022-12-22
fix: return correct originalError after normalize for deprecated GraphQLError

marko-zub:
In `graphql-js v16` property becomes nested `error.originalError.originalError` in `toNormalizedOptions`, but should be just `error.originalError`.

Added condition to check if `error` instance of `GraphQLError` then just return `args[4].originalError`.

Issue happens in `executeFieldsSerially`.

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -45,12 +45,17 @@ function toNormalizedOptions(
 ): GraphQLErrorOptions {
   const firstArg = args[0];
   if (firstArg == null || 'kind' in firstArg || 'length' in firstArg) {
+    const originalError =
+      args[4] instanceof GraphQLError && args[4].originalError
+        ? args[4].originalError
+        : args[4];
+
     return {
       nodes: firstArg,
       source: args[1],
       positions: args[2],
       path: args[3],
-      originalError: args[4],
+      originalError,
       extensions: args[5],
     };
   }
--- a/src/error/__tests__/GraphQLError-test.ts
+++ b/src/error/__tests__/GraphQLError-test.ts
@@ -71,6 +71,30 @@ describe('GraphQLError', () => {
     });
   });
 
+  it('uses correct originalError for deprecated declaration of GraphQLError', () => {
+    const originalError = new Error('original');
+    const graphQLError = new GraphQLError('msg', {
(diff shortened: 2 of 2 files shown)


marko-zub:
@yaacovCR please trigger github workflow.
Lint errors has been fixed and added test.
Thanks

marko-zub:
@yaacovCR 
What next steps to merge current PR?

yaacovCR:
Hi @marko-zub --

As far as I follow here, this PR is asserting that when the GraphQLError constructor is supplied an object of type `GraphQLError` as the `originalError`, the new GraphQLError object's `originalError` property should point to the `originalError` within the supplied `GraphQLError` property.

Whereas the existing behavior (as far as I can tell) in both v15, v16, v17-alpha all assume that the new GraphQLError object's `originalError` property should point to the intermediate GraphQLError object which will have its own `originalError` property.

I suppose the question is whether the interpretation of the `originalError` should be more like `rootError` (this PR) or the `immediatelyCausativeError` (existing behavior).

Considering that the latter option preserves more data, I guess I am in favor. It is also similar to the new `cause` property for `Error` objects. In fact, perhaps it could eventually be superseded by this property.

Links:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#differentiate_between_similar_errors
https://github.com/tc39/proposal-error-cause

But from your text above, it seems like you believe there is something that changed with v16 or something v16-specific that requires fixing/changing?

marko-zub:
Issue specific for `v16`
we rely on `error.originalError` property (in `v15` it works correctly), but in `v16` after execution [`executeFieldsSerially`](https://github.com/graphql/graphql-js/blob/16.x.x/src/execution/execute.ts#L382) and pass `GraphQLError` we receive nested `error.originalError.originalError` it's seems that reason is in [`toNormalizedOptions`](https://github.com/graphql/graphql-js/blob/16.x.x/src/execution/execute.ts#L382)

yaacovCR:
I don’t yet see the code change resulting in a change of behavior. Are you able to set up a small reproduction that works in version 15 and brakes with version 16?

yaacovCR:
Closing for now, please feel free to reopen with more information!

PR #3806, opened 2022-12-27
polish: remove misleading comment

yaacovCR:
We could not demonstrate via actual benchmarking that the tick reduction from awaiting a promise prior to returning it translates into real-world performance gains. See #3796.

In any case, the promise must be awaited (irrespective of performance) to allow the catch block to handle rejection errors. Simply returning `completed` would result in test failures.


saihaj (approved):

yaacovCR:
To be clear, I was the author of the misleading comment in question. :)

PR #3808, opened 2022-12-30
Add coordinate field to schema element definitions

yaacovCR:
[#3145 rebased on main](https://github.com/graphql/graphql-js/pull/3145).

depends on #3044

* Changes `GraphQLSchemaElement` interface into a base class which defines a `.coordinate` property and `toString`/`toJSON` methods.
* Uses the base class to types, fields, arguments, input fields, enum values, and directives.
* Uses this in validation error printing string templates.
* Exports the now finalized GraphQLSchemaElement, GraphQLField, GraphQLArgument, GraphQLInputField, GraphQLEnumValue classes.

This differs from the original PR [#3145](https://github.com/graphql/graphql-js/issues/3145) in that GraphQLField, GraphQLArgument, GraphQLInputField, GraphQLEnumValue constructors now take the parent type/field itself as the first argument instead of the parent coordinate. This obviates any need to validate the parent coordinate.

@leebyron comments from original PR:

> * Defines a `GraphQLSchemaElement` base class which defines a `.coordinate` property and `toString`/`toJSON` methods.
> * Adds base class to types, fields, arguments, input fields, enum values, and directives.
> * Uses this in validation error printing string templates.
> 
> Results in a net-reduction of code, minor simplification of definition.ts, and simplification/standardization of error messages

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -34,24 +34,23 @@ export type { GraphQLArgs } from './graphql.js';
 export { graphql, graphqlSync } from './graphql.js';
 
 // Create and operate on GraphQL type definitions and schema.
-export type {
-  GraphQLField,
-  GraphQLArgument,
-  GraphQLEnumValue,
-  GraphQLInputField,
-} from './type/index.js';
 export {
   resolveObjMapThunk,
   resolveReadonlyArrayThunk,
   // Definitions
   GraphQLSchema,
   GraphQLDirective,
+  GraphQLSchemaElement,
   GraphQLScalarType,
   GraphQLObjectType,
   GraphQLInterfaceType,
   GraphQLUnionType,
   GraphQLEnumType,
   GraphQLInputObjectType,
+  GraphQLField,
+  GraphQLArgument,
+  GraphQLEnumValue,
+  GraphQLInputField,
   GraphQLList,
   GraphQLNonNull,
   // Standard GraphQL Scalars
--- a/src/type/__tests__/definition-test.ts
+++ b/src/type/__tests__/definition-test.ts
@@ -329,6 +329,7 @@ describe('Type System: Objects', () => {
       }),
     });
     expect(objType.getFields().f).to.deep.include({
+      coordinate: 'SomeObject.f',
       parentType: objType,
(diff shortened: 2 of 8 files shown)


yaacovCR:
@IvanGoncharov @leebyron 

While rebasing/reviewing this, I noticed that the new `.toConfig()` methods return actual `*Config` types rather than `*NormalizedConfig` types.

I initially thought to just add them in, but then I also noticed that some of the existing `*NormalizedConfig` do not carry all of the type parameters from the original `*Config` types, and I was unsure why and therefore unsure of how to proceed with the potentially generic `*NormalizedConfig` types.

Questions:
1. Must the new `.toConfig()` methods return `*NormalizedConfig` types?
2. What is the right approach to type parameters with `*NormalizedConfig` types?

src/index.ts
```diff
@@ -34,24 +34,23 @@ export type { GraphQLArgs } from './graphql.js';
...
-  GraphQLEnumValue,
-  GraphQLInputField,
-} from './type/index.js';
 export {
   resolveObjMapThunk,
   resolveReadonlyArrayThunk,
   // Definitions
   GraphQLSchema,
   GraphQLDirective,
+  GraphQLSchemaElement,
```
martinbonnin, line 43: Answering here from https://github.com/graphql/graphql-js/pull/3044#issuecomment-2958278527. I missed that PR during the recent coordinates dicussion. Do we need `GraphQLSchemaElement`? 

I'm fine if there's no "bidirectionality" between a coordinate and a schema element. We could even have `GraphQLSchemaElement` without the `coordinate` property and that would probably be fine for me too. In other words, I'm ok with coordinates being "1-way". 

Did we have any more context about this? Are there use cases out there that would benefit from `GraphQLField` having a `coordinate` property for an example?

yaacovCR: We can ask @leebyron for more context, I think the original idea was to use the coordinate concept as an internal convenience to standardize our error messages. In #4288 we basically took the concept with their standardized error messages without the coordinate field itself, so that's why this rebased PR doesn't have a large error message diff and #3145 does.

PR #3810, opened 2022-12-31
Preserve defaultValue literals

### LINKED ISSUE #3051: Default value declared via SDL may not introspect correctly

yaacovCR:
[#3074 rebased on main](https://github.com/graphql/graphql-js/pull/3074).

Depends on #3809

@leebyron comments from original PR (edited, hopefully correctly):

> Fixes #3051
> 
> This change solves the problem of default values defined via SDL not always resolving correctly through introspection by preserving the original GraphQL literal in the schema definition. This changes argument and input field definitions `defaultValue` field from just the "value" to a new `GraphQLDefaultValueUsage` type which contains either (EDIT: but not both!) "value" and "literal" fields.
> 
> Here is the flow for how a default value defined in an SDL would be converted into a functional schema and back to an SDL:
> 
> **Before this change:**
> 
> ```
> (SDL) --parse-> (AST) --coerceInputLiteral--> (defaultValue config) --valueToAST--> (AST) --print --> (SDL)
> ```
> 
> `coerceInputLiteral` performs coercion which is a one-way function, and `valueToAST` is unsafe and set to be deprecated in #3049.
> 
> **After this change:**
> 
> ```
> (SDL) --parse-> (defaultValue literal config) --print --> (SDL)
> ```

### DIFF
--- a/src/execution/getVariableSignature.ts
+++ b/src/execution/getVariableSignature.ts
@@ -4,9 +4,12 @@ import type { VariableDefinitionNode } from '../language/ast.js';
 import { print } from '../language/printer.js';
 
 import { isInputType } from '../type/definition.js';
-import type { GraphQLInputType, GraphQLSchema } from '../type/index.js';
+import type {
+  GraphQLDefaultValueUsage,
+  GraphQLInputType,
+  GraphQLSchema,
+} from '../type/index.js';
 
-import { coerceInputLiteral } from '../utilities/coerceInputValue.js';
 import { typeFromAST } from '../utilities/typeFromAST.js';
 
 /**
@@ -18,7 +21,7 @@ import { typeFromAST } from '../utilities/typeFromAST.js';
 export interface GraphQLVariableSignature {
   name: string;
   type: GraphQLInputType;
-  defaultValue: unknown;
+  defaultValue: GraphQLDefaultValueUsage | undefined;
 }
 
 export function getVariableSignature(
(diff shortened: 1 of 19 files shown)


JoviDeCroock (approved):

yaacovCR:
Just changing the label on this. This is a bug fix, but it also changes the type/structure of the `defaultValue` property, and so it is also a breaking change.

PR #3813, opened 2023-01-01
Input Value Validation

yaacovCR:
[#3086 rebased on main](https://github.com/graphql/graphql-js/pull/3086).

Depends on #3812 

@leebyron comments from original PR:

> Factors out input validation to reusable functions:
> 
> * Introduces `validateInputLiteral` by extracting this behavior from `ValuesOfCorrectTypeRule`.
> * Introduces `validateInputValue` by extracting this behavior from `coerceInputValue`
> * Simplifies `coerceInputValue` to return early on validation error
> * Unifies error reporting between `validateInputValue` and `validateInputLiteral`, causing some error message strings to change, but error data (eg locations) are preserved.
> 
> These two parallel functions will be used to validate default values
> 
> Potentially breaking if you rely on the existing behavior of `coerceInputValue` to call a callback function, as the call signature has changed. GraphQL behavior should not change, though error messages are now slightly different.

Note: also breaking if you rely on the default callback function to throw. Grossly similar behavior is available with `validateInputValue()`.

### DIFF
--- a/src/execution/__tests__/nonnull-test.ts
+++ b/src/execution/__tests__/nonnull-test.ts
@@ -647,7 +647,7 @@ describe('Execute: handles non-nullable types', () => {
         errors: [
           {
             message:
-              'Argument "cannotBeNull" of non-null type "String!" must not be null.',
+              'Argument "cannotBeNull" has invalid value: Expected value of non-null type "String!" not to be null.',
             locations: [{ line: 3, column: 42 }],
             path: ['withNonNullArg'],
           },
@@ -677,7 +677,7 @@ describe('Execute: handles non-nullable types', () => {
         errors: [
           {
             message:
-              'Argument "cannotBeNull" of required type "String!" was provided the variable "$testVar" which was not provided a runtime value.',
+              'Argument "cannotBeNull" has invalid value: Expected variable "$testVar" provided to type "String!" to provide a runtime value.',
             locations: [{ line: 3, column: 42 }],
             path: ['withNonNullArg'],
           },
@@ -705,7 +705,7 @@ describe('Execute: handles non-nullable types', () => {
         errors: [
           {
             message:
-              'Argument "cannotBeNull" of non-null type "String!" must not be null.',
+              'Argument "cannotBeNull" has invalid value: Expected variable "$testVar" provided to non-null type "String!" not to be null.',
             locations: [{ line: 3, column: 43 }],
             path: ['withNonNullArg'],
           },
--- a/src/execution/__tests__/oneof-test.ts
+++ b/src/execution/__tests__/oneof-test.ts
@@ -30,7 +30,12 @@ function executeQuery(
   rootValue: unknown,
   variableValues?: { [variable: string]: unknown },
 ): ExecutionResult | Promise<ExecutionResult> {
-  return execute({ schema, document: parse(query), rootValue, variableValues });
+  return execute({
+    schema,
+    document: parse(query, { experimentalFragmentArguments: true }),
+    rootValue,
+    variableValues,
+  });
 }
 
 describe('Execute: Handles OneOf Input Objects', () => {
@@ -83,7 +88,7 @@ describe('Execute: Handles OneOf Input Objects', () => {
             message:
               // This type of error would be caught at validation-time
               // hence the vague error message here.
-              'Argument "input" of non-null type "TestInputObject!" must not be null.',
+              'Argument "input" has invalid value: Expected variable "$input" provided to type "TestInputObject!" to provide a runtime value.',
             path: ['test'],
           },
         ],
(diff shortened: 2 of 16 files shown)


yaacovCR:
@leebyron @erikkessler1 @benjie et al

I hit a wrinkle rebasing this on main considering the addition of `@oneOf`.

[The spec proposal](https://github.com/graphql/graphql-spec/pull/825/files) for `@oneOf` appropriately adds the following:

>     - If {value} is a variable:
>       - Let {variableName} be the name of {variable}.
>       - Let {variableDefinition} be the {VariableDefinition} named
>         {variableName} defined within {operation}.
>       - Let {variableType} be the expected type of {variableDefinition}.
>       - {variableType} must be a non-null type.

This validation logic was implemented in https://github.com/graphql/graphql-js/pull/3513 within the `ValuesOfCorrectTypeRule`.

This `Input Validation PR` extracts the logic for `ValuesOfCorrectTypeRule` into a separate function `validateInputLiteral` with TS signature:

```ts
/**
 * Validate that the provided input literal is allowed for this type, collecting
 * all errors via a callback function.
 *
 * If variable values are not provided, the literal is validated statically
 * (not assuming that those variables are missing runtime values).
 */
export function validateInputLiteral(
  valueNode: ValueNode,
  type: GraphQLInputType,
  variables: Maybe<VariableValues>,
  onError: (error: GraphQLError, path: ReadonlyArray<string | number>) => void,
  path?: Path,
): void
```

To implement the additional validation for OneOf Input Types mentioned above , we need to pass the operation's variable definitions to `validateInputLiteral`.

It might seem that this is very doable, as because of #3811, further up in this PR stack, `variables` is now of new type `VariableValues` that preserves the variable source -- and its definition. However, `variables` is not supplied by the implementation of `ValuesForCorrectType`, and in fact the presence of absence of `variables` is what changes the behavior above in terms of the difference between static or runtime validation referenced in the comment above.

Validation for `@oneOf` seems to be the first case where we can statically check something about the variable definition without actually having a variable value.

We have some options!

1. We can validate the variable definition when used with `@oneOf` at runtime (Ed: I don't think this is a great option).
2. We can collect and pass the `variableDefinitions` separately to `validateInputLiteral` as an additional parameter.
3. We can reorganize the `VariableValues` interface such that the definitions are stored as a separate key rather than within the sources themselves, and then pass a slender `Static` version of this interface.

Current:

```ts
export interface VariableValues {
  readonly sources: ReadOnlyObjMap<VariableValueSource>;
  readonly coerced: ReadOnlyObjMap<unknown>;
}

interface VariableValueSource {
  readonly variable: VariableDefinitionNode;
  readonly type: GraphQLInputType;
  readonly value: unknown;
}
```

Potential Option -- I'm sure ridden with TS errors:

```ts
export type VariableValues = StaticVariableValues | DynamicVariableValues;

export interface StaticVariableValues {
  readonly variables: ReadOnlyObjMap<VariableValueVariable>;
  sources?: never;
  coerced?: never;
}

export interface DynamicVariableValues {
  readonly variables: ReadOnlyObjMap<VariableValueVariable>;
  readonly sources: ReadOnlyObjMap<unknown>;
  readonly coerced: ReadOnlyObjMap<unknown>;
}

interface VariableValueVariable {
  readonly variable: VariableDefinitionNode;
  readonly type: GraphQLInputType;
}
```

All input (pun-intended) welcome!

yaacovCR:
Actually, we don't have the types when just looking at an operation either. Maybe:

```ts
export interface VariableValues {
  readonly definitions: ReadOnlyObjMap<VariableDefinitionNode>;
  readonly values?: ReadOnlyObjMap<VariableValueValues> | undefined;
}

export interface VariableValueValues {
  readonly sources: ReadOnlyObjMap<VariableValueSource>;
  readonly coerced: ReadOnlyObjMap<unknown>;
}

interface VariableValueSource {
  readonly type: GraphQLInputType;
  readonly value: unknown;
}
```

yaacovCR:
> Validation for `@oneOf` seems to be the first case where we can statically check something about the variable definition without actually having a variable value.

Correction: this is definitely not the case, there is a wholly separate validation rule specifying that variables should be of the correct type. My current suggestion actually is to pull the variable validation from one-of altogether, and let whether we allow nullable variables in non-nullable positions re: oneOf be decided by the general rule.

See: https://github.com/graphql/graphql-spec/pull/825/files#r1538875839

acao:
is there any way I can help with this PR? is it awaiting further spec deliberation? the rules for `oneOf` we will use for validation in `graphiql` and the various modes and the LSP

yaacovCR:
I have separated out my oneOf concerns to #4194 and $4195. This is ready for review!

yaacovCR:
Rereading the above, and the linked PRs refreshed my memory. This PR relies on #4194 to perform variable validation, moving it to `VariablesInAllowedPositionRule`, a fourth option besides the one above, and one that has independent appeal for me.

Hopefully, #4194 will be discussed at an upcoming GraphQL WG meeting as it moves closer to Stage 3. If #4194 is not accepted as part of this process, this PR will require greater refiguring.

benjie:
Hi @yaacovCR; are a lot of the comments above dealt with now via:

- https://github.com/graphql/graphql-js/pull/4194
- https://github.com/graphql/graphql-spec/pull/1118
- https://github.com/benjie/graphql-spec/pull/1

If so, are there remaining complexities?

yaacovCR:
> If so, are there remaining complexities?

I don't think so! Just need some approvals :)

JoviDeCroock (approved):

PR #3815, opened 2023-01-04
fix(incremental): abort async iteration when stream errors or stream is filtered

yaacovCR:
Currently:

When list item completion fails for a stream with a non-nullable list:
1. the entire list must be nulled within the given AsyncPayloadRecord
2. any other pending AsyncPayloadRecords must be filtered
3. async iteration powering the stream must be cancelled

Currently, the third objective is accomplished by way of the second; during AsyncPayloadRecord filtering, if a stream record is filtered and has an associated asyncIterator, its return() method is called, which _should_ end the stream.

This can go wrong in a few ways:
A: The return() method may not exist; by specification, the return() method exists for the caller to notify the callee that the caller no longer intends to call next(), allowing for early cleanup. The method is optional, however, and so should not be relied on.
B: The return method, even if it exists, may not be set up to block any next() calls while it operates. Async generators have next and return methods that always settle in call order, but async iterables do not.

This PR adds tests addressing these scenarios and fixes the test failures.

### DIFF
--- a/src/execution/__tests__/stream-test.ts
+++ b/src/execution/__tests__/stream-test.ts
@@ -1344,25 +1344,19 @@ describe('Execute: stream directive', () => {
         } /* c8 ignore stop */,
       },
     });
-    expectJSON(result).toDeepEqual([
-      {
-        errors: [
-          {
-            message:
-              'Cannot return null for non-nullable field NestedObject.nonNullScalarField.',
-            locations: [{ line: 4, column: 11 }],
-            path: ['nestedObject', 'nonNullScalarField'],
-          },
-        ],
-        data: {
-          nestedObject: null,
+    expectJSON(result).toDeepEqual({
+      errors: [
+        {
+          message:
+            'Cannot return null for non-nullable field NestedObject.nonNullScalarField.',
+          locations: [{ line: 4, column: 11 }],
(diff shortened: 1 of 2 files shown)


yaacovCR:
@robrichard 

one potential avenue for a fix is to rename our existing StreamRecords => StreamItemRecords, making nomenclature space available for introduction of a distinct StreamRecord to represent the stream itself, which can be cancelled/filtered independently. During async iteration, perhaps we can check if the individual StreamItemRecord's StreamRecord has been cancelled.

this does mean that filtering has some run-time cost even in the non-presence of errors, but it seems minimal...

yaacovCR:
@robrichard @graphql/graphql-js-reviewers tests fixed by adding a context object scoped to each stream that can be used to check/set error status of the entire stream.

yaacovCR:
Woops, initial approach worked, but was not robust, i.e. didn't end up fixing similar problem with the new Publisher.

The wrinkle is that errors outside of a stream that would lead to the entire stream being filtered should also abort async iteration. This can be managed by having a set of running streams handled by the publisher/dispatcher (currently simply stored on `ExecutionContext` with a path. Filtering now runs through all of those streams and if the path matches, the stream is removed from the set.

yaacovCR:
closing in favor of #3886

PR #3818, opened 2023-01-06
Expand @deprecated to Object types

fotoetienne:
See [RFC](https://github.com/graphql/graphql-spec/pull/997)

### DIFF
--- a/src/type/__tests__/definition-test.ts
+++ b/src/type/__tests__/definition-test.ts
@@ -188,6 +188,7 @@ describe('Type System: Objects', () => {
       extensions: {},
       astNode: undefined,
       extensionASTNodes: [],
+      deprecationReason: undefined,
     });
   });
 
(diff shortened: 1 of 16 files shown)


IvanGoncharov (changes requested):
Please update this rule: https://github.com/graphql/graphql-js/blob/main/src/validation/rules/custom/NoDeprecatedCustomRule.ts
and also add validation preventing the usage of the deprecated type directly here: https://github.com/graphql/graphql-js/blob/main/src/type/validate.ts

fotoetienne:
Rebased onto latest next and addressed the review feedback:

  Schema validation (validate.ts):
  - Non-deprecated fields that return a deprecated object type now produce a validation error, ensuring the schema remains valid when deprecated types are filtered from introspection.

  Query validation (NoDeprecatedCustomRule.ts):
  - InlineFragment — reports usage of deprecated types via ... on DeprecatedType
  - FragmentDefinition — reports usage of deprecated types via fragment Foo on DeprecatedType

  All existing tests pass, new tests added for both validation rules
  
   Resolves https://github.com/graphql/graphql-wg/issues/1353

glen-84:
We've implemented opt-in support for object deprecation in Hot Chocolate and Fusion -- what is preventing this from moving forward? Could it land behind an experimental option?

PR #3828, opened 2023-01-15
deduplicate leaf fields already sent in initial payload

yaacovCR:
extacted from #3820


yaacovCR:
closed in favor of #3844

PR #3830, opened 2023-01-17
feat(incremental): do not initiate multiple streams at the same path

yaacovCR:
This builds upon the defer merging work in #3820.

The following operation:
```graphql
query { 
  friendList @stream(initialCount: 2) {
    name
    id
  }
  ... @defer {
    friendList @stream(initialCount: 2) {
      name
      id
    }
  }
}
```

will now only result in a single stream.

Corresponding spec change is at: https://github.com/robrichard/graphql-spec/pull/12


yaacovCR:
closing this in favor of #3862

PR #3832, opened 2023-01-19
internal: fix inlineInvariant for cjs builds that require filenames with explicit extensions

yaacovCR:
Current build script rewrites invariant to avoid function call if invariant true.

However, the rewriting should take into account the import name as rewritten (by TS?):

Current output for cjs is:

https://github.com/graphql/graphql-js/blob/ba59ddd4afbe1129616ddb1ccbc79d2cd8dbf615/execution/execute.js#L14

while the call is still:

https://github.com/graphql/graphql-js/blob/ba59ddd4afbe1129616ddb1ccbc79d2cd8dbf615/execution/execute.js#L723-L724
yielding an error of the sort: `invariant is not defined`

### DIFF
--- a/integrationTests/node/index.cjs
+++ b/integrationTests/node/index.cjs
@@ -29,6 +29,13 @@ assert.deepStrictEqual(result, {
   },
 });
 
+/**
+ * The below test triggers a call `invariant` method during execution (by
+ * passing a negative number to the `initialCount` parameter on the `@stream`
+ * directive). This ensures that the `inlineInvariant` method called by our
+ * build script works correctly.
+ **/
+
 const experimentalSchema = buildSchema(`
   directive @stream(initialCount: Int!) on FIELD
 
--- a/integrationTests/node/index.cjs
+++ b/integrationTests/node/index.cjs
@@ -1,7 +1,11 @@
 const assert = require('assert');
 const { readFileSync } = require('fs');
 
-const { graphqlSync } = require('graphql');
+const {
+  experimentalExecuteIncrementally,
+  graphqlSync,
+  parse,
+} = require('graphql');
 const { buildSchema } = require('graphql/utilities');
 const { version } = require('graphql/version');
 
@@ -12,7 +16,7 @@ assert.deepStrictEqual(
 
 const schema = buildSchema('type Query { hello: String }');
 
-const result = graphqlSync({
+let result = graphqlSync({
   schema,
   source: '{ hello }',
   rootValue: { hello: 'world' },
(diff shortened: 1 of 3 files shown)


yaacovCR:
Current build script rewrites invariant to avoid function call if invariant true.

However, the rewriting should take into account the import name as rewritten (by TS?):

Current output for cjs is:

```
const invariant_js_1 = require('../jsutils/invariant.js');
```

while the call is still:

```
  typeof stream.initialCount === 'number' ||
    invariant(false, 'initialCount must be a number');
```

yielding an error of the sort: `invariant is not defined`

yaacovCR:
Strangely, the new test passes here, but fails for me in a local project where I have npm point to the npm branch ie the latest source from main

gotta investigate further

yaacovCR:
> Strangely, the new test passes here, but fails for me in a local project where I have npm point to the npm branch ie the latest source from main
> 
> gotta investigate further

finally reproduced this appropriately! phew.

yaacovCR:
Bug fixed, ready for review.

integrationTests/node/index.cjs
```diff
@@ -24,3 +28,20 @@ assert.deepStrictEqual(result, {
...
+`);
+
+result = experimentalExecuteIncrementally({
+  schema: experimentalSchema,
+  document: parse('{ greetings @stream(initialCount: -1) }'),
+  rootValue: { greetings: ['hi', 'hello'] },
+});
+
+assert(result.errors?.[0] !== undefined);
+assert(!result.errors[0].message.includes('is not defined'));
```
IvanGoncharov, line 47: can you please add a block comment, explaining what is tested here?

resources/inline-invariant.ts
```diff
@@ -32,11 +32,10 @@ export function inlineInvariant(context: ts.TransformationContext) {
...
             ts.SyntaxKind.BarBarToken,
-            factory.createCallExpression(
-              factory.createIdentifier(funcName),
-              undefined,
-              [factory.createFalse(), ...otherArgs],
-            ),
+            factory.createCallExpression(expression, undefined, [
+              factory.createFalse(),
+              ...otherArgs,
+            ]),
```
IvanGoncharov, line 38: smart 🧠 👏

IvanGoncharov (changes requested):
Looks good!
We just need a comment documenting what we are testing in both files.
Otherwise ready to merge

integrationTests/node/index.cjs
IvanGoncharov: Just to clarify, I mean block comments before adding a sub-test.
In the future, if we add more of them it makes sense to have separate files, but it will make `test.js` more complicated.

yaacovCR: Done! 8700dda

yaacovCR:
> Looks good! We just need a comment documenting what we are testing in both files. Otherwise ready to merge

Ready for re-review.

yaacovCR:
> Looks good! We just need a comment documenting what we are testing in both files. Otherwise ready to merge

Ready for review -- checks should pass once #3849 is merged.

IvanGoncharov:
@yaacovCR I rebased it on top of #3849 and merged it.

PR #3833, opened 2023-01-19
Ignore meta-files from Sapling SCM

mjmahone:
[Sapling](https://sapling-scm.com/) is a new incantation of Mercurial, designed to work with git repositories, while supporting Mercurial's preference for stacked commits and stacked rebases as a first-class operation.

This ensures our tooling ignores `/.sl` directories produced from people who make PRs using Sapling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /.eslintcache
 /.cspellcache
 /.docusaurus
+/.sl
 /.stryker-tmp
 /node_modules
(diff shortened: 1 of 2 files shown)


mjmahone:
This is implemented upstream in prettier https://github.com/prettier/prettier/pull/14206

So I'll close this as, when we upgrade prettier after the next prettier release, we'll get this functionality for free!

PR #3834, opened 2023-01-19
Package lock update

mjmahone:
Every time I run `npm install` on a new `graphql-js` repo, across multiple computer setups, I'm seeing a `package-lock.json` update that I need to ignore within my ocmmit.

I suspect other people are seeing this too, so let's just upgrade it.


mjmahone:
Something is funky with my npm setup vs. the setup used in CI.

PR #3835, opened 2023-01-19
Fragment Arguments Reference Implementation

mjmahone:
This implements Fragment Arguments, as described by the spec changes in https://github.com/graphql/graphql-spec/pull/1010.

We have:
- Updates AST to support new syntax of arguments on fragment spreads, and updating our legacy "experimental" fragment variables definition syntax to be fully supported as a way of defining the names and types for the arguments passed to fragment spreads.
- Updated Validation (including Fields Can Merge)
- Updated the Executor.

I am creating a clean commit rather than updating https://github.com/graphql/graphql-js/pull/3152, as the discussion from the previous PR was confusing newcomers with old historical details that we've decided to abandon.

Please use this PR for reviewing and verifying the *accuracy* of the PR, with inline comments when possible.

**For discussions about Fragment Arguments as an idea, missing features or open questions, please open or respond to a thread in the graphql-wg discussion space: https://github.com/graphql/graphql-wg/discussions/1239**

### DIFF
--- a/src/execution/collectFields.ts
+++ b/src/execution/collectFields.ts
@@ -22,6 +22,8 @@ import {
 } from '../type/directives.js';
 import type { GraphQLSchema } from '../type/schema.js';
 
+import { keyForFragmentSpread } from '../utilities/keyForFragmentSpread.js';
+import { substituteFragmentArguments } from '../utilities/substituteFragmentArguments.js';
 import { typeFromAST } from '../utilities/typeFromAST.js';
 
 import { getDirectiveValues } from './values.js';
@@ -124,7 +126,7 @@ function collectFieldsImpl(
   selectionSet: SelectionSetNode,
   fields: AccumulatorMap<string, FieldNode>,
   patches: Array<PatchFields>,
-  visitedFragmentNames: Set<string>,
+  visitedFragmentKeys: Set<string>,
 ): void {
   for (const selection of selectionSet.selections) {
     switch (selection.kind) {
@@ -156,7 +158,7 @@ function collectFieldsImpl(
             selection.selectionSet,
             patchFields,
             patches,
-            visitedFragmentNames,
+            visitedFragmentKeys,
           );
           patches.push({
             label: defer.label,
@@ -172,24 +174,24 @@ function collectFieldsImpl(
             selection.selectionSet,
             fields,
             patches,
-            visitedFragmentNames,
+            visitedFragmentKeys,
           );
         }
         break;
       }
       case Kind.FRAGMENT_SPREAD: {
-        const fragName = selection.name.value;
+        const fragmentKey = keyForFragmentSpread(selection);
 
         if (!shouldIncludeNode(variableValues, selection)) {
           continue;
         }
 
         const defer = getDeferValues(operation, variableValues, selection);
-        if (visitedFragmentNames.has(fragName) && !defer) {
+        if (visitedFragmentKeys.has(fragmentKey) && !defer) {
           continue;
         }
 
-        const fragment = fragments[fragName];
+        const fragment = fragments[selection.name.value];
         if (
           !fragment ||
           !doesFragmentConditionMatch(schema, fragment, runtimeType)
@@ -198,9 +200,13 @@ function collectFieldsImpl(
         }
 
         if (!defer) {
-          visitedFragmentNames.add(fragName);
+          visitedFragmentKeys.add(fragmentKey);
         }
 
+        const fragmentSelectionSet = substituteFragmentArguments(
+          fragment,
+          selection,
+        );
         if (defer) {
           const patchFields = new AccumulatorMap<string, FieldNode>();
           collectFieldsImpl(
@@ -209,10 +215,10 @@ function collectFieldsImpl(
             variableValues,
             operation,
             runtimeType,
-            fragment.selectionSet,
+            fragmentSelectionSet,
             patchFields,
             patches,
-            visitedFragmentNames,
+            visitedFragmentKeys,
           );
           patches.push({
             label: defer.label,
@@ -225,10 +231,10 @@ function collectFieldsImpl(
             variableValues,
             operation,
             runtimeType,
-            fragment.selectionSet,
+            fragmentSelectionSet,
             fields,
             patches,
-            visitedFragmentNames,
+            visitedFragmentKeys,
           );
         }
         break;
--- a/src/utilities/TypeInfo.ts
+++ b/src/utilities/TypeInfo.ts
@@ -1,6 +1,12 @@
 import type { Maybe } from '../jsutils/Maybe.js';
+import type { ObjMap } from '../jsutils/ObjMap.js';
 
-import type { ASTNode, FieldNode } from '../language/ast.js';
+import type {
+  ASTNode,
+  FieldNode,
+  FragmentDefinitionNode,
+  FragmentSpreadNode,
+} from '../language/ast.js';
 import { isNode } from '../language/ast.js';
 import { Kind } from '../language/kinds.js';
 import type { ASTVisitor } from '../language/visitor.js';
@@ -31,6 +37,7 @@ import type { GraphQLDirective } from '../type/directives.js';
 import type { GraphQLSchema } from '../type/schema.js';
 
 import { typeFromAST } from './typeFromAST.js';
+import { valueFromAST } from './valueFromAST.js';
 
 /**
  * TypeInfo is a utility class which, given a GraphQL schema, can keep track
@@ -39,6 +46,7 @@ import { typeFromAST } from './typeFromAST.js';
  */
 export class TypeInfo {
   private _schema: GraphQLSchema;
+
   private _typeStack: Array<Maybe<GraphQLOutputType>>;
   private _parentTypeStack: Array<Maybe<GraphQLCompositeType>>;
   private _inputTypeStack: Array<Maybe<GraphQLInputType>>;
@@ -47,6 +55,8 @@ export class TypeInfo {
   private _directive: Maybe<GraphQLDirective>;
   private _argument: Maybe<GraphQLArgument>;
   private _enumValue: Maybe<GraphQLEnumValue>;
+  private _fragmentSpread: Maybe<FragmentSpreadNode>;
+  private _fragmentDefinitions: ObjMap<FragmentDefinitionNode>;
   private _getFieldDef: GetFieldDefFn;
 
   constructor(
@@ -69,6 +79,8 @@ export class TypeInfo {
     this._directive = null;
     this._argument = null;
     this._enumValue = null;
+    this._fragmentSpread = null;
+    this._fragmentDefinitions = Object.create(null);
     this._getFieldDef = getFieldDefFn ?? getFieldDef;
     if (initialType) {
       if (isInputType(initialType)) {
@@ -142,6 +154,17 @@ export class TypeInfo {
     // checked before continuing since TypeInfo is used as part of validation
     // which occurs before guarantees of schema and document validity.
     switch (node.kind) {
+      case Kind.DOCUMENT: {
+        // A document's fragment definitions are type signatures
+        // referenced via fragment spreads. Ensure we can use definitions
+        // before visiting their call sites.
+        for (const astNode of node.definitions) {
+          if (astNode.kind === Kind.FRAGMENT_DEFINITION) {
+            this._fragmentDefinitions[astNode.name.value] = astNode;
+          }
+        }
+        break;
+      }
       case Kind.SELECTION_SET: {
         const namedType: unknown = getNamedType(this.getType());
         this._parentTypeStack.push(
@@ -180,6 +203,10 @@ export class TypeInfo {
         this._typeStack.push(isOutputType(outputType) ? outputType : undefined);
         break;
       }
+      case Kind.FRAGMENT_SPREAD: {
+        this._fragmentSpread = node;
+        break;
+      }
       case Kind.VARIABLE_DEFINITION: {
         const inputType: unknown = typeFromAST(schema, node.type);
         this._inputTypeStack.push(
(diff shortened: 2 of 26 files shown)


mjmahone:
@github-actions run-benchmark

src/language/parser.ts
```diff
@@ -91,23 +91,6 @@ export interface ParseOptions {
...
-   *
-   * The syntax is identical to normal, query-defined variables. For example:
-   *
-   * ```graphql
-   * fragment A($var: Boolean = false) on T {
-   *   ...
-   * }
-   * ```
-   */
-  allowLegacyFragmentVariables?: boolean | undefined;
```
mjmahone, line 109: Will update this to `enableExperimentalFragmentArguments?: boolean | undefined` and use this so we can turn on fragment arguments within *tooling* prior to *servers* supporting them, which should unblock *Relay* fully supporting the syntax (even in open source) and using compiler transforms to remap fragments with arguments to duplicated fragment definitions.

Then we'll be able to add a flag to Relay's compiler on whether the *server* supports fragment spreads, and then when we finally merge the spec we'll be able to remove the flags entirely.

src/utilities/substituteFragmentArguments.ts
```diff
@@ -0,0 +1,78 @@
...
+      //
+      // To remove this hack, we need to either:
+      //    - include fragment argument scope when evaluating fields
+      //    - make unset fragment arguments invalid
+      // Requiring the spread to pass all non-default-defined arguments is nice,
+      // but makes field argument default values impossible to use.
+      substitutions[argumentName] = {
+        kind: Kind.VARIABLE,
+        name: { kind: Kind.NAME, value: '__UNSET' },
+      };
```
leebyron, line 74: > ```
> // To remove this hack, we need to either:
> //   - include fragment argument scope when evaluating fields
> ```

I think probably yes, that. Is that too crazy? This hack is not ideal.

Also, `query ($__UNSET: String = "SomeDefault") {...` is valid and while most query authors wouldn't write this, this seems like some kind of security vector or something

src/validation/ValidationContext.ts
```diff
@@ -199,16 +202,24 @@ export class ValidationContext extends ASTValidationContext {
     let usages = this._variableUsages.get(node);
     if (!usages) {
       const newUsages: Array<VariableUsage> = [];
-      const typeInfo = new TypeInfo(this._schema);
+      const typeInfo = this._typeInfo;
+      const localVariableDefinitions =
```
leebyron, line 206: `fragmentVariableDefinitions`

src/validation/__tests__/KnownDirectivesRule-test.ts
```diff
@@ -219,9 +222,13 @@ describe('Validate: Known directives', () => {
         message: 'Directive "@onQuery" may not be used on FIELD.',
         locations: [{ column: 19, line: 16 }],
       },
+      {
+        message: 'Directive "@onField" may not be used on VARIABLE_DEFINITION.',
```
leebyron, line 226: Not blocking for this PR, but we may want to consider this directive location name being different - maybe to `FRAGMENT_VARIABLE_DEFINITION`?

src/validation/rules/NoUnusedVariablesRule.ts
```diff
@@ -17,7 +17,10 @@ export function NoUnusedVariablesRule(context: ValidationContext): ASTVisitor {
     OperationDefinition(operation) {
       const usages = context.getRecursiveVariableUsages(operation);
       const variableNameUsed = new Set<string>(
-        usages.map(({ node }) => node.name.value),
+        usages
+          // Skip variables used as fragment arguments
+          .filter(({ fragmentVarDef }) => !fragmentVarDef)
+          .map(({ node }) => node.name.value),
       );
```
leebyron, line 24: I noticed these get skipped over in NoUndefinedFragmentVariableRule, maybe the better place to edit is changing `getRecursiveVariableUsages()` to something like `getOperationVariableUsages()`?

leebyron (commented):
This is really exciting! Nuanced behavior here that's found a sweet spot of value.

I dropped in a mix of tactical and high level feedback.

The one overarching piece of feedback is to be careful with naming things to get consistency. This adds both "fragment variable definitions" and "fragment spread arguments". I don't have a strong opinion about what we call the whole RFC'd feature, but we should be consistent about calling the things you provide to fragment spreads "arguments" and the things you define at the top of a fragment "variable definitions". I noticed some swapping of this terminology in a confusing way, as well as a few others like "local" vs "fragment".

src/utilities/keyForFragmentSpread.ts
```diff
@@ -0,0 +1,23 @@
...
+  const fragmentName = fragmentSpread.name.value;
+  const fragmentArguments = fragmentSpread.arguments;
+  if (fragmentArguments == null || fragmentArguments.length === 0) {
+    return fragmentName;
+  }
+
+  const printedArguments: Array<string> = fragmentArguments
+    .map(print)
+    .sort((a, b) => a.localeCompare(b));
+  return fragmentName + '(' + printedArguments.join(',') + ')';
```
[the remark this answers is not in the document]
mjmahone: > Should this just be strictly disallowed in the first place similar to how repeated fields need to have equivalent arguments?

Yes, IMO: we should be treating fragment spreads with the same name *exactly like* fields with the same name that need to be able to merge. If we don't it greatly increases the complexity of field merging validation, and likely of the executor.

If you're only changing a variable that isn't actually used within one spread's context, technically the executor ought to be able to handle it. But I think the devex is worse to allow that case than to just make it invalid.

> If so, then we'd need to know the result of fragmentArgumentSubstitutions as an input to determining this key. ...AorB discussion...

This is a good test case to add for OverlappingFieldsCanBeMergedRule, and it should fail (but likely doesn't as-is, great catch!). I think we might need to add a local-variable-scope indicator to the key for validation (though at execution time we should always have actualized values). Good call that the keying function can't *quite* be shared as-is.

src/utilities/substituteFragmentArguments.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * NOTE: fragment arguments are scoped to the fragment they're defined on.
+ * Therefore, after we apply the passed-in arguments, all remaining variables
+ * must be either operation defined variables or explicitly unset.
+ */
+export function substituteFragmentArguments(
+  def: FragmentDefinitionNode,
+  fragmentSpread: FragmentSpreadNode,
+): SelectionSetNode {
+  const argumentDefinitions = def.variableDefinitions;
```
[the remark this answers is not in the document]
mjmahone: Agree it's not super clean. I likely need to do another pass but there is some logic here: `argumentDefinitions` are the definitions for the arguments we need to look up based on the spread, *not* the variable definitions that the spread currently has access to.

src/utilities/substituteFragmentArguments.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+  const argumentValues = fragmentArgumentSubstitutions(
+    argumentDefinitions,
+    fragmentSpread.arguments,
+  );
+  return visit(def.selectionSet, {
+    Variable(node) {
+      return argumentValues[node.name.value];
+    },
+  });
```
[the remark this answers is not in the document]
mjmahone: I flip-flopped. The *current* spec draft uses the replacement algorithm because adding actual scope infected the algorithms a lot more. The problem is you need to basically replace `type VariableDefinitions = Map<string, [VariableDefinition, VariableValue]>` everywhere with `type VariableScope = { localVariables: VariableDefinitions, operationVariables: VariableDefinitions }`. It's certainly worth trying to get the scoped version out before shipping the spec change, but it's not the *smallest* change possible, which is likely to lead to some mistakes.

Whereas if we can get all the test cases described, working and solidified with the smaller-change replacement option, we'll have a lot of confidence in switching to scoped variables.

src/utilities/substituteFragmentArguments.ts
mjmahone: Oh I should just replace it with `$0__UNSET` and then it's not valid :)

> Is that too crazy?

I spent a week or so of full-time work trying to get the spec algorithms and graphql-js to have full scope. It was just a much more pervasive change. It's not too crazy, but the replacement strategy is certainly a *smaller* change for both.

src/validation/__tests__/KnownDirectivesRule-test.ts
mjmahone: Yeah it's weird: if they're their own AST node then absolutely. If not... maybe? I think probably?

This is a small change and another thing I flip-flopped on.

src/validation/rules/NoUnusedVariablesRule.ts
mjmahone: If we're OK making a breaking change. But Meta has custom validation rules that use `getRecursiveVariableUsages` so I imagine other people likely do too?

mjmahone (commented):
Given our discussion at the APAC WG: if we're going to have a flag, I *think* it makes sense to split this PR into a series of dependent PRs. We can probably also pair graphql-spec PRs with each:
- AST + parse + print changes (I don't think there's *much* controversy on syntax, maybe just directive location being the one thing)
  - Validation rule changes (a little more controversial, specifically with how shadowing must work)
  - Executor changes

This sequencing would help us unblock the tooling changes we need while still making forward progress on the gnarlier reviews and choices.

src/validation/rules/NoUnusedVariablesRule.ts
leebyron: Or leave the existing method behind (optionally deprecate it if its not used in core anymore?) and introduce this new one?

src/validation/__tests__/KnownDirectivesRule-test.ts
leebyron: DirectiveLocations and AST nodes are not a perfect 1:1

See `getDirectiveLocationForASTPath()` for the mapping, there's already a few bits of logic there to consider extending in a similar way.

I think keeping the AST minimally changed is great, so it makes sense to me to have the `VARIABLE_DEFINITION` AST node map to one of two directive locations based on the parent AST node being a fragment or not.

src/utilities/substituteFragmentArguments.ts
leebyron: fair enough. That makes me think we should call this "Fragment Templates" and mirror some of the C++ nomenclature. I don't think I like it, but it's closer to what we're describing :)

would an equivalent behavior be to also visit each InputValue on leave and if its unset to remove it? In other words - rather than putting in a hacked variable AST node, leave the AST that would be the equivalent to an unset var

src/utilities/substituteFragmentArguments.ts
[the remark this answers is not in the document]
leebyron: Fair enough. That makes me think we should call this "Fragment Templates" and mirror some of the C++ nomenclature. I don't think I like it, but it's closer to what we're describing :)

src/utilities/substituteFragmentArguments.ts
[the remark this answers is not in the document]
leebyron: I don't know that I follow that logic exactly for this specific use. In our existing nomenclature, arguments are the input values you provide (to a field, directive, now frag spread) and variables are the placeholders a value will go to.

I do get the logic that "variable definition" does some double-duty here. It defines the variables within fragment but it also defines the arguments allowed to that fragment's spreads. For consistency we should choose one naming convention and use comments when that could cause confusion. Since the AST node is "variable definition" - I think we should stick with that.

In this particular case you're actually substituting/replacing the variables within a fragment with values which were provided via fragment spread arguments. We're not substituting/replacing any fragment spread arguments themselves.

The first step here `fragmentArgumentSubstitutions` seems to me to be logically taking input of "variable definitions" and "fragment spread arguments" and returning output of "variable values"


---

Tl;dr - naming things is hard

src/utilities/keyForFragmentSpread.ts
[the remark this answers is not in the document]
leebyron: Oh! I really like this. It's constraining, but in a way that's far more simplifying than restricting. I think if you apply this then you don't need "fragment key" as a concept at all!

I imagine the validation rule either akin to  or part of `OverlappingFieldsCanBeMergedRule` which ensures any two fragment spreads of the same name in the same scope would always produce exactly the same set of fragment variable values (aka: always has the same source of values). I think most of the logic you added is approaching that, but I had just misunderstood that.

In that case, you can assume that for a valid query you only need to spread the first instance of any fragment with a given name since the above rule guarantees that any fragment you encounter later with the same name would provide identical results.

Because you can assume that, you should be able to just keep the existing logic of using the fragment name instead of computing a key during `collectFields`

src/utilities/TypeInfo.ts
```diff
@@ -190,15 +217,50 @@ export class TypeInfo {
...
+          if (fragVarDef) {
+            const fragVarType = typeFromAST(schema, fragVarDef.type);
+            if (isInputType(fragVarType)) {
+              const fragVarDefault = fragVarDef.defaultValue
+                ? valueFromAST(fragVarDef.defaultValue, fragVarType)
+                : undefined;
+
+              // Minor hack: transform the FragmentArgDef
+              // into a schema Argument definition, to
+              // enable visiting identically to field/directive args
```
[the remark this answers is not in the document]
leebyron: Agreed.

It might be a bit more refactoring in the other methods that produce the expected type, but I'm curious if we can follow your suggestion where a "Signature" is just the union of ArgumentDef and VariableDef? That way we can avoid the risk of some subtle bug introduced by a virtual AST element here and keep the type of `argDef` as is, while still providing the correct `argType`?

In other words, the logic below adding to the "Default value" and "input type" stack is correct, but what needs to change to set _argument to undefined or to add a generalized "signature" getter?

src/execution/collectFields.ts
```diff
@@ -172,24 +174,24 @@ function collectFieldsImpl(
...
             patches,
-            visitedFragmentNames,
+            visitedFragmentKeys,
           );
         }
         break;
       }
       case Kind.FRAGMENT_SPREAD: {
-        const fragName = selection.name.value;
+        const fragmentKey = keyForFragmentSpread(selection);
```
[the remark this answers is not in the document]
leebyron: +100 - I got too far into solutioning here. See the other thread about validation which might side-step all of this by removing the need for it in the first place

leebyron:
> Given our discussion at the APAC WG: if we're going to have a flag, I think it makes sense to split this PR into a series of dependent PRs.

@mjmahone I support this. I think we need exactly two:

1. Adds the Parse/Print/AST along with the experimental flag. There's an opportunity to get DirectiveLocations, and any other language-specific stuff right in this PR. Should be mostly uncontroversial since the language decisions seem agreed and the flag protects the ability to go back and correct if necessary.
2. Validation and Execution. Too hard to untangle these where behavior of one needs to influence the other, and iterating here is going to meaningfully impact spec text.

yaacovCR:
Closing in favor of #4015 and maybe #4159

PR #3837, opened 2023-01-29
fix: invalid originalError propagation in custom scalars

stenreijers:
Errors thrown in the `parseValue` function for custom scalars are not propagated correctly using the `originalError` variable of the GraphQLError on invalid input. As a result, error codes from the `extensions.code` are not propagated correctly.

The fix is simple: Replace `error.orginalError` with `error`, since `error.orginalError` does not exist yet.
see:
https://github.com/graphql/graphql-js/blob/8be83d8d528991aed04ca17434b7e26e56f649cf/src/execution/values.ts#L134

I have added test-cases to avoid this problem in the future.

PR #3838, opened 2023-02-02
Fix/invalid error propagation custom scalars (backport for 16.x.x)

stenreijers:
This is the same fix for invalid error propagation on custom scalars, but now backported for 16.x.x

see:
https://github.com/graphql/graphql-js/pull/3837

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7712,13 +7712,8 @@
       "dev": true
     },
     "node_modules/eslint-plugin-internal-rules": {
-      "name": "eslint-plugin-graphql-internal",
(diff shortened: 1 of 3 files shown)


stenreijers:
I had to do two commits, because I run locally with a newer version of NPM, so the second commit is to restore the original package-lock.json such that github can do its checks.

stenreijers:
Any ETA on this backport?

IvanGoncharov:
@stenreijers Sorry for the delay. 
Looking at it right now.

PR #3839, opened 2023-02-03
 Fix ambiguity around when schema definition may be omitted

benjie:
This PR implements the tests and fixes necessary for [Spec RFC #987](https://github.com/graphql/graphql-spec/pull/987).

## The main changes

This PR is made of three main commits:

1. Test printing a schema that has `Query`, `Mutation` and `Virus` types, but only supports `query` operations (via the `Query` type) and does _not_ support `mutation` operations.
2. Test parsing this same schema text, and assert that the schema does not have a mutation type.
3. Fix the printing of the schema.

## Unnecessary test?

The fourth commit addresses the following failing test that I didn't know what to do about:

https://github.com/graphql/graphql-js/blob/76e47fcd7d432f515d5bb99e95af1851148a0c54/src/utilities/__tests__/buildClientSchema-test.ts#L65-L81

I don't understand the reasoning behind this test. It seems to me that the schema output from this should instead be:

```graphql
schema {}

type Query {
  foo: String
}
```

since the schema has no operation types (but uses the default names); however this is unparseable (because `{}` is unparseable). Perhaps @IvanGoncharov or @leebyron can shed some light on the reasoning behind this test? I've worked around it by special casing it, but I'd prefer to remove the special case code and also remove the test case.

### DIFF
--- a/src/__testUtils__/viralSDL.ts
+++ b/src/__testUtils__/viralSDL.ts
@@ -1,0 +1,19 @@
+export const viralSDL = `\
+schema {
+  query: Query
+}
+
+type Query {
+  viruses: [Virus!]
+}
+
+type Virus {
+  name: String!
+  knownMutations: [Mutation!]!
+}
+
+type Mutation {
+  name: String!
+  geneSequence: String!
+}\
+`;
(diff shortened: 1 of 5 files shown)


benjie:
cc @graphql/graphql-js-reviewers

benjie:
Going to close/reopen to see if Netlify retries 🤷‍♂️

spawnia:
Works nicely when ported to PHP: https://github.com/webonyx/graphql-php/pull/1303.

PR #3842, opened 2023-02-05
incremental delivery: deduplicates all previously sent values

yaacovCR:
This implementation is meant as a straw man, demonstrating that total deduplication is possible.

@IvanGoncharov @robrichard et al

### Strategy

1. A path of undefined has been replaced by a path of type `Root` so that we can use a WeakMap on the empty path.
2. `promiseForObject` now mutates the original object so that we can stably associate every value in a response to the execution path. This function probably be renamed.
3. The initial result tree is traversed, marking which values have been sent.
4. Immediately prior to inclusion of any incremental payload, we iterate through the payload's result tree to exclude/mask:
    - any leaf that has already been sent
    - any value that does not include any deeper unsent leaves

### Notes

- We could mask values present in the initial result using a different method, see #3828 
- We need additional tests for this type of masking for more complicated cases involving lists/streams.

### Open Questions:

1. Is it possible to associate a large bidirectional map of potential assignments of future leaf values and their paths with each asynchronous payload, such that as each incremental payload is built from the leaves, we could remove the potential assignments from all other payloads, resulting in payload tree creation after de-deduplication.
2. How would error handling work in this approach?
3. Would it be more performant?


yaacovCR:
note: tests pass except for coverage.

yaacovCR:
coverage fixed

yaacovCR:
Also note that the current implementation relies heavily on WeakMaps that may be difficult to translate to other languages.

yaacovCR:
closed in favor of #3844

Issue #3840, opened 2023-02-03
debug failing netlify CI 


saihaj:
On https://github.com/graphql/graphql-js/pull/3839 the neltify CI started to fail for some reason. Code changed in that PR has nothing to do with the netlify build. Here are the logs https://gist.github.com/benjie/9a95fc4d8e2c3aafde0cf47c34dec07f

PR #3843, opened 2023-02-06
internal: fix website build error

yaacovCR:
#3840


yaacovCR:
@saihaj @benjie 

This seems to fix the problem with build script on netlify missing the internal eslint rules => they were not previously being copied to the temporary folder.

Why then was the build script ever able to complete successfully? I am not sure. Perhaps somehow the package was in the cache on some earlier run and the cache expired?

saihaj (approved):
Ideally we shouldn't even need to copy any eslint things. We just run build of docusaurus and output things out but @IvanGoncharov would know why we decided to do this 🤔

benjie (approved):
Seems harmless enough... Thanks Yaacov!

PR #3849, opened 2023-02-16
ESLint: use the `--rulesdir` option for internal rules.

IvanGoncharov:
Motivation: We need to have a way to have internal ESLint rules. ESLint allows you to import rules in two ways:
1. As an npm package, it is problematic for internal rules since we don't want to publish it on NPM. So in the past, we used NPM-specific workarounds to treat a directory as a package.
2. `--rulesdir` CLI option, the problem with it is deprecated.

ESLint team is planning to deprecate the current config format and switch to: https://eslint.org/docs/latest/use/configure/configuration-files-new This format forces you to import plugins manually, and it entirely resolves our problem with internal rules. But it requires that all plugins that we use to support this format.

I propose to temporarily switch to `--rulesdir` and start working on changing to the new format since the entire config format will be deprecated anyway. This switch will unblock support for node@19 in CI and local development and unblock #3848.


yaacovCR:
@IvanGoncharov that sounds great to me.

We can also update the package.json and README files within resources/eslint-internal-rules

We still need the package.json to mark the file as CJS, but we don't really need much in there, we can even take out the package name.

The README should be updated to reflect that the need for the package.json is just to mark the files as cjs (which won't be necessary with the new config format!) and updated as well in that you no longer need to run `npm install` to pick up changes.

saihaj (approved):

yaacovCR (approved):
Looks good, we could update the package.json and README as I commented separately, but that could be a separate PR.

IvanGoncharov:
Thanks, @yaacovCR. 
I implemented your suggestions before merging.

PR #3856, opened 2023-02-26
Use `replaceAll` instead of RegExp with global flag

IvanGoncharov:
Motivation: since we droped support for node14 we can finally use this function.


saihaj:
@github-actions run-benchmark

saihaj (approved):

B2o5T:
@IvanGoncharov same, you can enable rule https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-replace-all.md as I made in graphiql https://github.com/graphql/graphiql/pull/3043 :)

PR #3860, opened 2023-03-09
Make print() break long List and Object Values over multiple line

dylanowen:
This builds upon #2797 to add multiline support for ObjectValues and ListValues.
Example:
```graphql
{
  trip(
    obj: {wheelchair: false, smallObj: {a: 1}, largeObj: {wheelchair: false, smallObj: {a: 1}, arriveBy: false, includePlannedCancellations: true, transitDistanceReluctance: 2000, anotherLongFieldName: "Lots and lots and lots and lots of text"}, arriveBy: false, includePlannedCancellations: true, transitDistanceReluctance: 2000, anotherLongFieldName: "Lots and lots and lots and lots of text"}
  ) {
    dateTime
  }
  another(
    list: [["small array", "small", "small"], ["Lots and lots and lots and lots of text", "Lots and lots and lots and lots of text", "Lots and lots and lots and lots of text"]]
  ) {
    dateTime
  }
}
```
is now
```graphql
{
  trip(
    obj: {
      wheelchair: false,
      smallObj: { a: 1 },
      largeObj: {
        wheelchair: false,
        smallObj: { a: 1 },
        arriveBy: false,
        includePlannedCancellations: true,
        transitDistanceReluctance: 2000,
        anotherLongFieldName: "Lots and lots and lots and lots of text"
      },
      arriveBy: false,
      includePlannedCancellations: true,
      transitDistanceReluctance: 2000,
      anotherLongFieldName: "Lots and lots and lots and lots of text"
    }
  ) {
    dateTime
  }
  another(
    list: [
      ["small array", "small", "small"],
      [
        "Lots and lots and lots and lots of text",
        "Lots and lots and lots and lots of text",
        "Lots and lots and lots and lots of text"
      ]
    ]
  ) {
    dateTime
  }
}
```

I agree that https://github.com/graphql/graphql-js/pull/2797#issuecomment-691634068 would be a better long term approach but rewriting the printer to support something like the prettier document builder with softlines and length aware line breaks is a much larger effort. This change will allow lines greater than 80 but it at least gives us a reasonable default for when to break out these larger lines.

### DIFF
--- a/src/language/__tests__/printer-test.ts
+++ b/src/language/__tests__/printer-test.ts
@@ -110,6 +110,64 @@ describe('Printer: Query document', () => {
     `);
   });
 
+  it('puts large object values on multiple lines if line is long (> 80 chars)', () => {
+    const printed = print(
+      parse(
+        '{trip(obj:{wheelchair:false,smallObj:{a: 1},largeObj:{wheelchair:false,smallObj:{a: 1},arriveBy:false,includePlannedCancellations:true,transitDistanceReluctance:2000,anotherLongFieldName:"Lots and lots and lots and lots of text"},arriveBy:false,includePlannedCancellations:true,transitDistanceReluctance:2000,anotherLongFieldName:"Lots and lots and lots and lots of text"}){dateTime}}',
+      ),
+    );
+
+    expect(printed).to.equal(dedent`
+      {
+        trip(
+          obj: {
+            wheelchair: false,
+            smallObj: { a: 1 },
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@dylanowen Thanks for PR!
I changed it to match both spec examples and prettier output (removed commas)
Also I changed a code to match existing solution for args from #2797

dylanowen:
@IvanGoncharov Thanks for fixing this up and getting it in!

PR #3867, opened 2023-03-27
Enable `noUncheckedIndexedAccess`

louisscruz:
I'm interested in possibly contributing to `graphql` in the future. I noticed that `noUncheckedIndexedAccess` is not yet enabled and that there was an associated `FIXME` comment in the configuration. Because it would make contributions safer in the future, and because I wanted a kind of tour of the code in this package, I turned on `noUncheckedIndexedAccess` and fixed all of the violations.

The vast majority of violations were fixed with either `assert` or `invariant`. In some cases, fixes were simple in that they only required different iterating mechanisms to be done in a more type safe way.

General comments/questions:

* Could any of the code in this PR break in some environments? I'm not sure which browsers/Node versions are supported here. I think the most advanced API I used was `Object.entries`, but that seems to be compatible back to Node 7. Any reviews with this in mind would be good.

### DIFF
--- a/resources/benchmark.ts
+++ b/resources/benchmark.ts
@@ -270,9 +270,9 @@ function runBenchmark(
   benchmark: string,
   benchmarkProjects: ReadonlyArray<BenchmarkProject>,
 ) {
-  const results = [];
-  for (let i = 0; i < benchmarkProjects.length; ++i) {
-    const { revision, projectPath } = benchmarkProjects[i];
+  const results: Array<BenchmarkComputedStats> = [];
+
+  benchmarkProjects.forEach(({ revision, projectPath }, i) => {
     const modulePath = path.join(projectPath, benchmark);
 
     if (i === 0) {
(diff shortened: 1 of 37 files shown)


IvanGoncharov:
@louisscruz Thanks for PR 👍 
I quickly look through it and I will try to give you a partial review tomorrow.

louisscruz:
@IvanGoncharov Just giving a ping on this.

I just rebased to address a conflict. I noticed the use of delegating multiple checks to single `invariant` calls. I can make a pass to do that in this PR if you'd like. Seems nicer.

yaacovCR:
I guess we were not ready to be that strict yet.

I have been reviewing old PRs, and am just noting that https://github.com/graphql/graphql-js/pull/3504 might be a useful addition to this PR, but I think in general this PR would require careful review to ensure that it does not introduce unnecessary runtime checks that can be avoided in similar ways to #3504

louisscruz:
If it'd still be welcome to make this change, I'd be happy to get it back up to a working state. And yeah, happy to remove any runtime checks that can be avoided through other means.

PR #3862, opened 2023-03-12
incremental delivery without branching

yaacovCR:
this version of incremental delivery works without branching, solving the issue raised by @Keweiqu 

see discussion here: https://github.com/robrichard/defer-stream-wg/discussions/65

UPDATE: this is now ready for review

** huge thanks to @urigo and @dotansimha of [the guild](https://the-guild.dev/) for sponsoring my open-source work on this. **

PR #3885, opened 2023-04-18
remove unnecessary duplicated fields from defer tests

yaacovCR:
extracted from #3862


yaacovCR:
Motivation: these changes may reduce the test diff if we add duplication; in either case, the tests as they are do not require field duplication.

robrichard (approved):

PR #3886, opened 2023-04-24
RFC: incremental delivery: without branching, with deduplication

yaacovCR:
This in the main PR implements the major ideas from the latest incremental delivery proposal, https://github.com/graphql/defer-stream-wg/discussions/69 which is centered around the idea of deduplication of execution and delivery.

This PR is part of the following stack:

#3886 = main PR (this PR!) switching to deduplication of execution/delivery
#3897 = adds pending
#3911 = adds helpers to resolvers to further delay execution of deferred fields or manage concurrency
#3895 = consolidates payloads

TODO:
[x] convert existing tests
[x] error handling
[x] filter late nulls correctly
[x] defer/streams interactions
[x] completed
[x] pending
[ ] ids and subpaths instead of paths
[ ] even more tests?

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -18,6 +18,7 @@ import { locatedError } from '../error/locatedError.js';
 
 import type {
   DocumentNode,
+  FieldNode,
   FragmentDefinitionNode,
   OperationDefinitionNode,
 } from '../language/ast.js';
@@ -46,20 +47,31 @@ import { GraphQLStreamDirective } from '../type/directives.js';
 import type { GraphQLSchema } from '../type/schema.js';
 import { assertValidSchema } from '../type/validate.js';
 
-import type { FieldGroup, GroupedFieldSet } from './collectFields.js';
+import type {
+  DeferUsage,
+  DeferUsageSet,
+  FieldGroup,
+  GroupedFieldSet,
+  GroupedFieldSetDetails,
+} from './collectFields.js';
 import {
   collectFields,
   collectSubfields as _collectSubfields,
+  NON_DEFERRED_TARGET_SET,
 } from './collectFields.js';
 import type {
   ExecutionResult,
   ExperimentalIncrementalExecutionResults,
   IncrementalDataRecord,
+} from './IncrementalPublisher.js';
+import {
+  DeferredFragmentRecord,
+  DeferredGroupedFieldSetRecord,
+  IncrementalPublisher,
   InitialResultRecord,
   StreamItemsRecord,
-  SubsequentDataRecord,
+  StreamRecord,
 } from './IncrementalPublisher.js';
-import { IncrementalPublisher } from './IncrementalPublisher.js';
 import { mapAsyncIterable } from './mapAsyncIterable.js';
 import {
   getArgumentValues,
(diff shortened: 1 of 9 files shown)


src/execution/execute.ts
```diff
@@ -2327,10 +2330,10 @@ class StreamRecord {
...
 }
 
-type AsyncPayloadRecord = DeferredFragmentRecord | StreamRecord;
+type IncrementalDataRecord = DeferredFragmentRecord | StreamRecord;
 
-function isStreamPayload(
-  asyncPayload: AsyncPayloadRecord,
-): asyncPayload is StreamRecord {
-  return asyncPayload.type === 'stream';
+function isStreamResult(
```
robrichard, line 2755: Maybe call this `isStreamRecord`?

yaacovCR: incorporated, thanks!

yaacovCR:
This week's update:

I have pushed a few fixes to the spec and impl PRs ==> these revolve mostly around when it's actually appropriate to figure whether a deferred fragment is "empty," we now do so lazily rather than eagerly to avoid some edge cases.

This serendipitously also simplifies the `Executor`==> `Publisher` event/messaging model; previously we had a funny hook/event/message from the `Executor` that was triggered/sent for every grouped field set executed on an object field. If named appropriately, its weirdness would really have stood out, something like: `Hit_Checkpoint_Where_We_Have_Initialized_Defers_at_a_Given_Field_Level`

Now the only events/messages from the `Executor` to the `Publisher` center around what has conclusively happened to a given `IncrementalDataRecord`:

(A). `DeferredGroupedFieldSet_Has_Completed`
(B). `DeferredGroupedFieldSet_Has_Errored`
(C). `StreamItemsRecord_Has_Completed`
(D). `StreamItemsRecord_Has_Completed_As_Closing_Iterator_Without_Data`
(E). `StreamItemsRecord_Has_Errored`

or what has happened at a given path of execution, i.e.:

(F). `Path_Has_Errored_Either_In_Initial_Result_Or_Within_IncrementalDataRecord`

No hidden knowledge required.

For those interested:

The birds eye flow within the implementation/spec:

1. The `Executor` spins up a `Publisher` and initiates execution of the initial result.
2. As directed by any `@defer` and `@stream` directives it encounters, the executor requests the publisher to generate `IncrementalDataRecord`s and semi-concurrently initiates execution of the associated grouped field sets for deferred fragments and completion of the given items for streams.
3. The `Executor` reports the  6 events (A)-(F) to the `Publisher` [which may or may not lead to filtering of those `IncrementalDataRecord`s from the final stream].
4. When the initial result completes, the `Executor` notifies the  `Publisher`.
5. The `Executor` then checks with the `Publisher` whether any subsequent results are pending, and, if so, returns a subscription to the stream from the `Publisher`. Otherwise, it just returns the initial result.

The `Publisher` tracks all the state necessary to make sure that the above works, including whether results are ordered correctly, filtered when appropriate, etc, etc.

In fact, because the `Executor` doesn't itself mutate anything about the `IncrementalDataRecord`s it receives, _it doesn't even actually need to receive the `IncrementalDataRecord` at all_ in step (2) above. It could simpy receive an `id` that the `Publisher` knows points to such a record, so that events (A)-(F) can reference that `id` in further communications to the `Publisher`.

yaacovCR:
Merging to implement https://github.com/graphql/defer-stream-wg/discussions/69

PR #3893, opened 2023-05-09
doc: add note to `printSchema` about the missing directives

YassinEldeeb:
This adds jsDoc to the `printSchema` to raise awareness that it won't include directives like the `@oneOf` directive.

If the user intended to have those directives then use the [`printSchemaWithDirectives`](https://the-guild.dev/graphql/tools/docs/api/modules/utils_src#printschemawithdirectives) instead.

### DIFF
--- a/src/utilities/printSchema.ts
+++ b/src/utilities/printSchema.ts
@@ -36,6 +36,18 @@ import type { GraphQLSchema } from '../type/schema.js';
 
 import { astFromValue } from './astFromValue.js';
 
+/**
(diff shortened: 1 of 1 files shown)


benjie:
> This adds jsDoc to the printSchema to raise awareness that it won't include directives like the @oneOf directive.

That example is not accurate; the `@oneOf` directive should be output by `printSchema` in the same way that a `@deprecated` directive should. However it is currently true for directives that are not built in.

PR #3894, opened 2023-05-13
introduce new IncrementalPublisher class

yaacovCR:
extracted from #3886

depends on #3903

more refactors from the without-branching branch, a bit more fundamental than #3891

[set as patch release because it does have an observable effect on the number of payloads, see below]

= iterate only through completed items
= remove extra ticks by making the publisher manage changes to its state synchronously.
= use children array instead of promises to manage hierarchy
= have IncrementalPublisher instantiate new IncrementalDataRecords

= The new publisher sometimes cause an empty `{ hasNext: false }` to be emitted. In particular, because the publisher is faster than it was, it may emit a stream result before the stream's asynchronous iterator has completed.
= The new publisher may sometimes reduce the number of `{ hasNext: false }` records that are emitted. For example, when errors on the initial result filter all subsequent results, this now happens synchronously, and so the publisher knows immediately that there are no subsequent results, such that there is no need for an empty final payload.

### DIFF
--- a/src/execution/IncrementalPublisher.ts
+++ b/src/execution/IncrementalPublisher.ts
@@ -1,7 +1,6 @@
 import type { ObjMap } from '../jsutils/ObjMap.js';
 import type { Path } from '../jsutils/Path.js';
 import { pathToArray } from '../jsutils/Path.js';
-import type { PromiseOrValue } from '../jsutils/PromiseOrValue.js';
 import { promiseWithResolvers } from '../jsutils/promiseWithResolvers.js';
 
 import type {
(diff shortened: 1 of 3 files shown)


src/execution/IncrementalPublisher.ts
```diff
@@ -85,237 +84,401 @@ export type FormattedIncrementalResult<
...
+    isCompleted: boolean;
+  };
 
-  async function next(): Promise<
-    IteratorResult<SubsequentIncrementalExecutionResult, void>
-  > {
-    if (isDone) {
-      return { value: undefined, done: true };
-    }
+  _released: Set<IncrementalDataRecord>;
```
IvanGoncharov, line 116: We also mark them in as `private` in TS, see other classes.

src/execution/IncrementalPublisher.ts
```diff
@@ -85,237 +84,401 @@ export type FormattedIncrementalResult<
...
+      isCompleted: false,
+    };
+    this._released = new Set();
+    this._pending = new Set();
+    this._reset();
+  }
 
-    const incremental = getCompletedIncrementalResults(subsequentPayloads);
-    const hasNext = subsequentPayloads.size > 0;
+  _trigger() {
```
IvanGoncharov, line 133: Methods also should be `private`.
BTW. If you know how to configure ESLint to check this it will be great.

IvanGoncharov (dismissed):
Sadly, I'm not in the context of this code :(
So I think @robrichard is the most knowledgeable person on this topic, so I he approves, let's merge this one.

P.S. From a general code quality standpoint, it's good after you add missing `private` attributes.

src/execution/IncrementalPublisher.ts
yaacovCR: sounds good [8ea9e99](https://github.com/graphql/graphql-js/pull/3894/commits/8ea9e990db37de4d631a8a63516425ce7df3bd00)

src/execution/IncrementalPublisher.ts
yaacovCR: not sure about how to configure eslint for this...

...but, all good for this file [8ea9e99](https://github.com/graphql/graphql-js/pull/3894/commits/8ea9e990db37de4d631a8a63516425ce7df3bd00)

yaacovCR:
> P.S. From a general code quality standpoint, it's good after you add missing `private` attributes.

done! [8ea9e99](https://github.com/graphql/graphql-js/pull/3894/commits/8ea9e990db37de4d631a8a63516425ce7df3bd00)

robrichard (approved):

PR #3895, opened 2023-05-14
incremental delivery: inline available payloads

yaacovCR:
depends on #3911

emit all completed payloads in a dependency tree within the same batch, by including subsequent deferred grouped field sets and stream items into their parents.

results that are ready are not only batched, they are wholly consolidated/inlined

as a notable excepetion, the initial result still does not include any defers, even if they are completed! although early execution still does kick off.  this may need revisiting!

### DIFF
--- a/src/execution/IncrementalPublisher.ts
+++ b/src/execution/IncrementalPublisher.ts
@@ -615,12 +615,16 @@ export class IncrementalPublisher {
       }
 
       this._introduce(subsequentResultRecord);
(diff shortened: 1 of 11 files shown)


yaacovCR:
Closing this old PR; the batching logic was merged separately in the rewrite in https://github.com/graphql/graphql-js/pull/4026. The aggressive inlining probably is not super-appropriate for the reference implementation, but could be considered after we actually get the spec changes released.

PR #3902, opened 2023-06-01
introduce promiseWithResolvers helper

robrichard:
Promise.withResolvers is currently a [stage 2 proposal for JS](https://github.com/tc39/proposal-promise-with-resolvers). This adds an equivalent jsutil to standardize usage.


yaacovCR (approved):
this looks great.

i wonder if in terms of the ECMA construct if it wouldn't be _even_ better to have different versions of this; if the caller doesn't need access to reject, maybe a different method could only return resolve and then it can be known that the promise will never throw....

but that is not for us I suppose :)

yaacovCR (approved):

PR #3904, opened 2023-06-02
add tests with regard to duplication

yaacovCR:
to minimize later diffs, we can watch how the tests change, rather than simple introducing tons of new tests

depends on https://github.com/graphql/graphql-js/pull/3930

### DIFF
--- a/src/execution/__tests__/defer-test.ts
+++ b/src/execution/__tests__/defer-test.ts
@@ -37,7 +37,86 @@ const friends = [
   { name: 'C-3PO', id: 4 },
 ];
 
-const hero = { name: 'Luke', id: 1, friends };
+const deeperObject = new GraphQLObjectType({
+  fields: {
+    foo: { type: GraphQLString, resolve: () => 'foo' },
+    bar: { type: GraphQLString, resolve: () => 'bar' },
+    baz: { type: GraphQLString, resolve: () => 'baz' },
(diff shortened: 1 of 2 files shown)


yaacovCR:
depends on #3894

suenalaba (approved):

src/execution/__tests__/defer-test.ts
```diff
@@ -37,7 +37,86 @@ const friends = [
   { name: 'C-3PO', id: 4 },
 ];
 
-const hero = { name: 'Luke', id: 1, friends };
+const deeperObject = new GraphQLObjectType({
+  fields: {
+    foo: { type: GraphQLString, resolve: () => 'foo' },
+    bar: { type: GraphQLString, resolve: () => 'bar' },
+    baz: { type: GraphQLString, resolve: () => 'baz' },
+    bak: { type: GraphQLString, resolve: () => 'bak' },
```
robrichard, line 45: I think in general it would be better for the schema to be more generic and pass in the resolver values we want in the tests. That's what's happening in the existing fields. But, I don't think that should hold up this PR.

For example, you wouldn't need both `c.slowNonNullErrorField` and `c.nonNullErrorField` in the schema if you could pass the resolver value from the test

yaacovCR: done in 233a96e8

PR #3910, opened 2023-06-09
getIntrospectionQuery should select kind for all types

### LINKED ISSUE #3909: getIntrospectionQuery result does not match the IntrospectionQuery type
https://github.com/graphql/graphql-js/blob/04e948bbc4972ddeb61443fab540e03fdff457b4/src/utilities/getIntrospectionQuery.ts#L70-L72

The query needs to select `kind` for each of these in order for the result to match the defined `IntrospectionQuery` type

https://github.com/graphql/graphql-js/blob/04e948bbc4972ddeb61443fab540e03fdff457b4/src/utilities/getIntrospectionQuery.ts#L172-L178

https://github.com/graphql/graphql-js/blob/04e948bbc4972ddeb61443fab540e03fdff457b4/src/utilities/getIntrospectionQuery.ts#L293-L298

jonapgar-groupby:
Prior to this change, the result when using the `getIntrospectionQuery` does not match the declared `IntrospectionQuery` type.

Fixes #3909

### DIFF
--- a/src/utilities/getIntrospectionQuery.ts
+++ b/src/utilities/getIntrospectionQuery.ts
@@ -67,9 +67,9 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
     query IntrospectionQuery {
       __schema {
         ${schemaDescription}
-        queryType { name }
-        mutationType { name }
-        subscriptionType { name }
+        queryType { name kind }
+        mutationType { name kind }
+        subscriptionType { name kind }
         types {
           ...FullType
         }


yaacovCR:
I think this is a legitimate bug. I guess the current behavior is working as is because the root types in a valid schema are always object types, but then we should have a different typing.

I am fine with either changing the typing on the root types to something specific where the `Kind` is not present, or this fix in asking for the Kind, preferring this fix, as in theory -- I think -- it is possible to introspect an invalid schema.

To proceed we will need to:

1. Get the CLA signed.
2. Rebase against v16.x.x or main (and after merge, we can port to the other branch).

yaacovCR:
@jonapgar-groupby

I know this is from a while back, thank you for your contribution and apologies for the delay!

We are currently trying to clean up our PR backlog, let us know if you are still interested in working on this!

JoviDeCroock:
Fixed in #4222

PR #3915, opened 2023-06-21
suggestion: alternative check for multiple graphql packages

phryneas:
This could in the long term enhance early detection if users bundle multiple versions of the `graphql` package, and (in a future version of `graphql`) retire the current check that needs inside of `instanceOf` and thus needs to be restricted to development node, since it slows down the function.

### DIFF
--- a/integrationTests/webpack/package.json
+++ b/integrationTests/webpack/package.json
@@ -3,7 +3,7 @@
   "description": "graphql-js should be compatible with Webpack",
   "type": "module",
   "scripts": {
-    "test": "webpack && node test.js"
+    "test": "webpack && node test.js && node test-esm.js"
   },
   "dependencies": {
     "graphql": "file:../graphql.tgz",
--- a/integrationTests/webpack/test-esm.js
+++ b/integrationTests/webpack/test-esm.js
@@ -1,14 +1,14 @@
 import assert from 'assert';
 
 /* eslint-disable n/no-missing-import */
-import cjs from './dist/main-cjs.cjs';
+import mjs from './dist/main-mjs.cjs';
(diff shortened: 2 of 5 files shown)


yaacovCR:
This is really interesting! Is there the possibility that this side effect might get pruned by tree shaking? Might we need to list it formally as a side effect?

Note also that the current check emits a different instanceof function entirely so it also runs only once.

phryneas:
@yaacovCR something like this shouldn't get tree-shaken, as it's not a declaration, but a method-call, which bundlers will assume as "with side-effects", unless you add a `/** PURE */` annotation.

And even if it gets tree-shaken: nothing will break, it will just not do the check.

That said, this approach has the drawback that it will only work if all `graphql` versions that take part use it. If we'd ship this in v17, but not v15 and v16, it would only warn on different v17 versions, but wouldn't trigger with v16 and v17 in parallel.

yaacovCR:
Does it matter that we have sideEffects set to false?

Does it depend on the bundles?

https://github.com/evanw/esbuild/issues/1241#issuecomment-831755947

phryneas:
`sideEffects` should stay `false` either way - while this is a side effect, it's not a very observable one or one that has an impact on anything than this package.

With `sideEffects` `true`, no tree-shaking will happen at all.

Beyond that, it *is* up to the bundler, but as far as I'm aware I haven't seen a bundler that has removed a top-level function call in a file that was imported for some other reasons.

The examples you linked to are not dropped because they are function calls (some bundlers mark `console.log` as pure, but most don't) that cannot be determined to be 100% pure.

Tree-shaking usually either drops full files (if none of the exports are used), or export definitions that are pure, either by annotation or by heuristics.

PR #3936, opened 2023-07-12
Add Developer best practices in starWarsSchema.ts

dhrumitpatel48:
Added Developer best practices in starWarsSchema.ts

Ideally, the switch-case statement should have default keyword at the end. 

In example, there was no default keyword. So as a part of developer best practices, added the default keyword and return the first switch-case value. So that if anyone enters the wrong value in the switch case, the code will not break.


IvanGoncharov:
I don't think it's applicable in this particular case.

PR #3937, opened 2023-07-17
Improve description for `@oneOf` directive

spawnia:
Follow up from https://github.com/graphql/graphql-js/pull/3513.

I think it is fine for the spec text in https://github.com/graphql/graphql-spec/pull/825 to refer to `OneOf Input Object`, as the spec explains what that is in another section. In a directive definition, that explanation by itself does not make much sense without prior knowledge.

Channeling the RFC champion @benjie.

### DIFF
--- a/src/type/directives.ts
+++ b/src/type/directives.ts
@@ -253,7 +253,7 @@ export const GraphQLSpecifiedByDirective: GraphQLDirective =
 export const GraphQLOneOfDirective: GraphQLDirective = new GraphQLDirective({
   name: 'oneOf',
   description:
-    'Indicates that exactly one field must be supplied and that field must not be `null`.',
+    'Indicates that exactly one field must be supplied and not `null`.',
   locations: [DirectiveLocation.INPUT_OBJECT],
   args: {},
 });
--- a/src/type/directives.ts
+++ b/src/type/directives.ts
@@ -253,7 +253,7 @@ export const GraphQLSpecifiedByDirective: GraphQLDirective =
 export const GraphQLOneOfDirective: GraphQLDirective = new GraphQLDirective({
   name: 'oneOf',
   description:
-    'Indicates that exactly one field must be supplied and not `null`.',
+    'Indicates exactly one field must be supplied and this field must not be `null`.',
   locations: [DirectiveLocation.INPUT_OBJECT],
   args: {},
 });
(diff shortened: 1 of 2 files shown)


benjie (approved):

src/type/directives.ts
```diff
@@ -248,11 +248,12 @@ export const GraphQLSpecifiedByDirective: GraphQLDirective =
...
 
 /**
- * Used to declare an Input Object as a OneOf Input Objects.
+ * Used to indicate an Input Object is a OneOf Input Object.
  */
 export const GraphQLOneOfDirective: GraphQLDirective = new GraphQLDirective({
   name: 'oneOf',
-  description: 'Indicates an Input Object is a OneOf Input Object.',
+  description:
+    'Indicates that exactly one field must be supplied and that field must not be `null`.',
```
yaacovCR, line 256:
```suggestion
    'Indicates that exactly one field must be supplied and that the field must not be `null`.',
```

src/utilities/__tests__/printSchema-test.ts
```diff
@@ -795,7 +795,9 @@ describe('Type System Printer', () => {
         url: String!
       ) on SCALAR
 
-      """Indicates an Input Object is a OneOf Input Object."""
+      """
+      Indicates that exactly one field must be supplied and that field must not be \`null\`.
```
yaacovCR, line 799: change to match above

yaacovCR (commented):
does it work better with the "the" article?

src/type/directives.ts
benjie: I don't think so. Arguably "and that _that field_ must not be `null`" would work, but generally it's best to remove redundant "that"'s. I guess `and this` would also work:
```suggestion
    'Indicates that exactly one field must be supplied and that this field must not be `null`.',
```

benjie:
> Indicates that exactly one field must be supplied and not \`null\`.

The current wording (after the latest edits) is acceptable, but it has the bracketing problem: `(exactly one field must be supplied) and (not null)` versus `exactly one field must be (supplied and not null)`. The latter interpretation is incorrect: `{a: null, b: 2}` supplies two fields, but has exactly one field that is both "supplied and not null".

The original wording was clearer in my opinion. That said, happy to see it merged either way.

IvanGoncharov:
@spawnia Can you please address @benjie last comment.
Otherwise ready to be merged 👍

benjie (approved):

benjie:

benjie:
Example of using this can be found here: https://github.com/graphql/graphql-spec/pull/825#issuecomment-1697137074

PR #3948, opened 2023-08-10
Add support for fourfold nested lists in introspection

### LINKED ISSUE #2643: Decorated type deeper than introspection query. 4d array
It does not seem like this library supports 4d arrays.

I have this type defined on my server

```graphql
type MultiPolygon {
  type: GeometryType!
  coordinates: [[[[Float!]!]!]!]!
  bbox: [Float!]
}
```

When I try to load the graphql playground I get this error:

```
Uncaught Error: Decorated type deeper than introspection query.
    at m (buildClientSchema.mjs:69)
    at m (buildClientSchema.mjs:82)
    at m (buildClientSchema.mjs:72)
    at m (buildClientSchema.mjs:82)
    at m (buildClientSchema.mjs:72)
    at m (buildClientSchema.mjs:82)
    at m (buildClientSchema.mjs:72)
    at m (buildClientSchema.mjs:82)
    at w (buildClientSchema.mjs:251)
    at keyValMap.mjs:20
```

This error only occurs when I change the coordinates field from 3d to 4d.

I'm wondering if this is a bug that is fixable or is there a workaround?

gschulze:
This PR adds support for introspecting four-fold nested lists, which in turn allows proper representation of GeoJSON multipolygons (https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.7).

Closes #2643

### DIFF
--- a/src/utilities/__tests__/buildClientSchema-test.ts
+++ b/src/utilities/__tests__/buildClientSchema-test.ts
@@ -877,10 +877,10 @@ describe('Type System: build schema from introspection', () => {
   });
 
   describe('very deep decorators are not supported', () => {
-    it('fails on very deep (> 7 levels) lists', () => {
+    it('fails on very deep (> 8 levels) lists', () => {
       const schema = buildSchema(`
         type Query {
-          foo: [[[[[[[[String]]]]]]]]
+          foo: [[[[[[[[[[String]]]]]]]]]]
         }
       `);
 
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@gschulze Thank you for explaining why you need this.
Definitely worth merging, if you need it in the v16 line please submit a PR to the `16.x.x` branch.

P.S. We definitely need to add something to the spec to properly solve this problem 🤔

IvanGoncharov (approved):

PR #3950, opened 2023-08-11
Support fourfold nested lists

gschulze:
This PR adds support for introspecting four-fold nested lists, which in turn allows proper representation of GeoJSON multipolygons (https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.7).

See https://github.com/graphql/graphql-js/issues/2643

### DIFF
--- a/src/utilities/__tests__/buildClientSchema-test.ts
+++ b/src/utilities/__tests__/buildClientSchema-test.ts
@@ -879,10 +879,10 @@ describe('Type System: build schema from introspection', () => {
   });
 
   describe('very deep decorators are not supported', () => {
-    it('fails on very deep (> 7 levels) lists', () => {
+    it('fails on very deep (> 8 levels) lists', () => {
       const schema = buildSchema(`
         type Query {
-          foo: [[[[[[[[String]]]]]]]]
+          foo: [[[[[[[[[[String]]]]]]]]]]
         }
       `);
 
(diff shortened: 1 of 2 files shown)


alessdiimperio:
what version of 16 supports introspecting 4d arrays? I am getting the following error

Error: JavaScriptError: Error-Decorated type deeper than introspection query.-u@

https://github.com/graphql/graphql-js/pull/3948 was added to version 17 alpha 3

are there any versions lower than 17 that has this support for introspection query

JoviDeCroock:
@alessdiimperio this should be in 16.8.1

alessdiimperio:
🤔 im on 16.8.1 and it builds but i get the error when using the introspection query on playground or using the codegen-cli for iOS
it seems as though it supports nested 4d arrays however it doesn't allow me to query for introspection

alessdiimperio:
also it's stated as being on version 17.0.0-alpha 3...
I did think it sounded suspect that support for it in the schema was included in 16.8 but not for introspection. however i can compile fine but not introspect and thus thought it was so. Are you certain that one can introspect 4d arrays in 16.8?

JoviDeCroock:
This PR can be seen in the release notes https://github.com/graphql/graphql-js/releases/tag/v16.8.0

alessdiimperio:
Alright, 🤔 why is this then happening to me while on 16.8.1. Very strange, i'll try and see if we are doing some funny business somewhere. cheers.

PR #3954, opened 2023-08-15
Fix win compat

sakesun:
This change is for developer using Windows OS to be able to run yarn test.
- use url.fileURLToPath instead of URL.pathname
- use double quote instead on single quote for cspell command


IvanGoncharov (changes requested):
@sakesun Can you please split either docs changes or Windows compatibility changes into separate PR?
Otherwise ready to merge

PR #3958, opened 2023-08-24
OverlappingFieldsCanBeMergedRule: Fix performance degradation

### LINKED ISSUE #3955: Resource exhaustion exploit when parsing queries
Re-posting here as advised in https://github.com/apollographql/apollo-server/issues/7688. Without digging into source code I'm not exactly sure which part of graphql-js is causing this.

### Issue Description

We've identified a potential resource exhaustion vector which has a significant impact on our CPU and response times. We're currently experiencing this on AS 3 however we're able to reproduce on AS 4.

We have a [custom query complexity calc plugin](https://github.com/tadhglewis/apollo-koa-minimal/blob/main/index.mjs#L33) which is intended to handle this scenario and others however it seems before we get to `didResolveOperation` event/stage, there is some processing that takes a long time.

I'm assuming this is an underlying dependancy in Apollo however I'm not 100% sure so any ideas or suggestions on how we can mitigate this would be great.

### Link to Reproduction

https://github.com/tadhglewis/apollo-koa-minimal

### Reproduction Steps

1. `pnpm install && pnpm start`
2. Run
```zsh
curl \
--data "{\"query\":\"{ $(python3 -c "print('%s' % ('__typename ' * 1000))")}\"}" \
--header 'Content-Type: application/json' \
--include \
--request POST \
https://example.com/graphql
```
This will take ~2.5s and increase to ~21s if you change the number of `__typename` to `3000`

Screenshot 2023-08-16 at 10 33 30 pm

Notes
- On subsequent requests there seems to be some caching and response time goes to ~10ms however if you change the number of `__typename` by one it will bypass this 
- This isn't limited to `__typename`, it can also be hit when using `hello` and I imagine introspection as well

AaronMoat:
Update(from @IvanGoncharov): In addition to reversing #3457 this PR improves performance further, see below discussion.

Resolves https://github.com/graphql/graphql-js/issues/3955 (at least greatly mitigates it) 

Effectively reverts https://github.com/graphql/graphql-js/pull/3457 which introduces a performance degradation found in https://github.com/graphql/graphql-js/issues/3955. This pull request was identified in a git bisect. The performance issue results in the potential for DOS attacks. 

I think there are potential performance increases still available in this file, but this is a simple enough fix for the immediate problem. 

The following query:

```
const maliciousQuery = `{ ${'hello '.repeat(2000)}}`;
```

now takes 564ms on my machine, and on main takes 12s.

### DIFF
--- a/benchmark/repeated-fields-benchmark.js
+++ b/benchmark/repeated-fields-benchmark.js
@@ -1,0 +1,18 @@
+import { graphqlSync } from 'graphql/graphql.js';
+import { buildSchema } from 'graphql/utilities/buildASTSchema.js';
+
+import { bigSchemaSDL } from './fixtures.js';
+
+const schema = buildSchema(bigSchemaSDL, { assumeValid: true });
+const query = `{ ${'__typename '.repeat(2000)}}`;
+
+export const benchmark = {
+  name: 'Many repeated fields',
+  count: 50,
+  measure() {
+    graphqlSync({
+      schema,
+      source: query,
+    });
+  },
+};
--- a/benchmark/repeated-fields-benchmark.js
+++ b/benchmark/repeated-fields-benchmark.js
@@ -1,18 +1,13 @@
 import { graphqlSync } from 'graphql/graphql.js';
 import { buildSchema } from 'graphql/utilities/buildASTSchema.js';
 
-import { bigSchemaSDL } from './fixtures.js';
-
-const schema = buildSchema(bigSchemaSDL, { assumeValid: true });
-const query = `{ ${'__typename '.repeat(2000)}}`;
+const schema = buildSchema('type Query { hello: String! }');
+const source = `{ ${'hello '.repeat(250)}}`;
 
 export const benchmark = {
   name: 'Many repeated fields',
-  count: 50,
+  count: 5,
   measure() {
-    graphqlSync({
-      schema,
-      source: query,
-    });
+    graphqlSync({ schema, source });
   },
 };
--- a/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
+++ b/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
@@ -179,6 +179,24 @@ describe('Validate: Overlapping fields can be merged', () => {
     ]);
   });
 
+  it('different stream directive extra argument', () => {
+    expectErrors(`
+      fragment conflictingArgs on Dog {
+        name @stream(label: "streamLabel", initialCount: 1)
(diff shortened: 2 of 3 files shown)


AaronMoat:
Hi @graphql/graphql-js-reviewers (I can't seem to ping this list - so @martijnwalraven @Cito @yaacovCR @IvanGoncharov @saihaj) any chance we could get some eyes on this? 🙏

tadhglewis:
I can create a issue as this can be done in a separate PR but it's somewhat concerning this issue this regression was possible in the first place. Would be great to have some perf tests in place.

yaacovCR:
@github-actions run-benchmark

would love to get @IvanGoncharov feedback on this

AaronMoat:
@github-actions run-benchmark

yaacovCR:
If I’m interpreting the benchmark results, I am not seeing a performance improvement with this fix…

AaronMoat:
I don't think this specific scenario is in the benchmarks.

yaacovCR:
> I don't think this specific scenario is in the benchmarks.

I guess what I’m wondering is (1) where the crossover point is, as performance seems better on main with our current tests but worse in your above scenario, and (2) whether the better mitigation is to use the maxTokens option when parsing?

https://github.com/graphql/graphql-js/blob/b12dcffe83098922dcc6c0ec94eb6fc032bd9772/src/language/parser.ts#L92

That’s just my own 2 cents, would really love @IvanGoncharov  take

AaronMoat:
The note on maxTokens is an interesting one, noting it's linear. I'd argue this is nonlinear, even with this change.

IvanGoncharov:
@AaronMoat Can you please add a benchmark to this PR otherwise this problem can reappear in the future.

AaronMoat:
@github-actions run-benchmark

AaronMoat:
@github-actions run-benchmark

AaronMoat:
Unsure what's going on with the benchmarks; it seems to have not even got up to the new one this time.

IvanGoncharov:
@AaronMoat Sorry for delay, I spend some time playing with this code and came up with [2991112](https://github.com/graphql/graphql-js/pull/3958/commits/2991112867dbe25937c330806864e133951a57e7) it further improves performance:
image

It reduces memory usage by ~40%.

IvanGoncharov:
@AaronMoat If you need this fix in `v16`, please a create separate PR for:
https://github.com/graphql/graphql-js/tree/16.x.x

AaronMoat:
Thanks @IvanGoncharov for taking that and running with it -- raised https://github.com/graphql/graphql-js/pull/3967 for 16.x.x

PR #3960, opened 2023-08-25
incremental: utilize `id` and `subPath` rather than `path` and `label`

yaacovCR:
to match new format at https://github.com/graphql/defer-stream-wg/discussions/69

### DIFF
--- a/src/execution/IncrementalPublisher.ts
+++ b/src/execution/IncrementalPublisher.ts
@@ -100,7 +100,8 @@ export interface IncrementalDeferResult<
 > {
   errors?: ReadonlyArray<GraphQLError>;
   data: TData;
-  path: ReadonlyArray<string | number>;
+  id: string;
+  subPath?: ReadonlyArray<string | number>;
   extensions?: TExtensions;
 }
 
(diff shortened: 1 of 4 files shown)


src/execution/IncrementalPublisher.ts
```diff
@@ -554,7 +566,8 @@ export class IncrementalPublisher {
         }
         const incrementalResult: IncrementalStreamResult = {
           items: subsequentResultRecord.items,
-          path: subsequentResultRecord.streamRecord.path,
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          id: subsequentResultRecord.streamRecord.id!,
```
robrichard, line 570: why do you need a non-null assertion?

yaacovCR: > why do you need a non-null assertion?

An `id` is assigned only when the pending notification is sent, so that a client cannot determine whether the server has executed early from the id (in a situation where the `id` is simply a sequence of numbers like we are using here).

By strict TS, there is therefore the possibility that the `id` is undefined, because although we know that the `id` is always defined if the record has completed, because the pending notification must have been sent and an `id` assigned, TS does not know that.

robrichard (approved):

PR #3964, opened 2023-09-01
fix for development on Windows

sakesun:
- use url.fileURLToPath instead of URL.pathname
- use double quote instead on single quote for shell command


sakesun:
@IvanGoncharov  Don't know how to fix last PR, so I just close the previous and open this one. Hope that's alright.

PR #3969, opened 2023-09-10
Add support for @oneOf directives in printSchema

### LINKED ISSUE #3968: Printing the @oneOf directive
It looks like currently the `printSchema` function doesn't have any handling for `@oneOf`.

I'm not sure if this is intentional or not, because I know printing directives has historically been a controversial topic, but my expectation would be that for directives that are part of the spec and have a corresponding option in the options when creating the type/field in code (eg `deprecationReason`/`@deprecated` ) would be printed when using  `printSchema`.

I'm happy to make a PR to add this, and it looks like it should be a pretty simple fix handled here: https://github.com/graphql/graphql-js/blob/main/src/utilities/printSchema.ts#L207C5-L207C5

hayes:
fixes #3968

### DIFF
--- a/src/utilities/__tests__/printSchema-test.ts
+++ b/src/utilities/__tests__/printSchema-test.ts
@@ -509,6 +509,23 @@ describe('Type System Printer', () => {
     `);
   });
 
+  it('Print Input Type with @oneOf directive', () => {
+    const InputType = new GraphQLInputObjectType({
+      name: 'InputType',
+      isOneOf: true,
+      fields: {
+        int: { type: GraphQLInt },
+      },
+    });
+
+    const schema = new GraphQLSchema({ types: [InputType] });
+    expectPrintedSchema(schema).to.equal(dedent`
+      input InputType @oneOf {
+        int: Int
+      }
+    `);
+  });
+
   it('Custom Scalar', () => {
     const OddType = new GraphQLScalarType({ name: 'Odd' });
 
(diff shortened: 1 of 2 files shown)


src/utilities/__tests__/printSchema-test.ts
```diff
@@ -663,7 +680,7 @@ describe('Type System Printer', () => {
       schema {
         query: Query
       }
-      
+
```
IvanGoncharov, line 683: Thanks for spotting it!
I applied global fix, here: https://github.com/graphql/graphql-js/pull/3970

src/utilities/printSchema.ts
```diff
@@ -204,7 +204,12 @@ function printInputObject(type: GraphQLInputObjectType): string {
   const fields = Object.values(type.getFields()).map(
     (f, i) => printDescription(f, '  ', !i) + '  ' + printInputValue(f),
   );
-  return printDescription(type) + `input ${type.name}` + printBlock(fields);
+  return (
+    printDescription(type) +
+    `input ${type.name}` +
+    printOneOf(type.isOneOf) +
```
IvanGoncharov, line 210: nit: It's overkill to create separate function, ternary operator would be enough here: `type.isOneOf ? ' @oneOf' : ''`.

hayes: makes sense, was just trying to match the existing pattern for the deprecated directive (which is slightly more complex).  Updated!

hayes:
anything else you need from my end to get this merged?

benjie (approved):
Thanks for spotting this omission @hayes; this is critical!

benjie:

benjie:
@graphql/graphql-js-reviewers This seems like an easy merge now all the feedback has been addressed? :raised_hands:

yaacovCR (approved):

saihaj (approved):

PR #3967, opened 2023-09-05
OverlappingFieldsCanBeMergedRule: Fix performance degradation

AaronMoat:
Applies fixes from https://github.com/graphql/graphql-js/pull/3958 into 16.x.x, fewer changes as no stream support in 16.

PR #3972, opened 2023-09-19
Release 16.8.1 to address resource exhaustion vulnerability

AaronMoat:
As per https://github.com/graphql/graphql-js/issues/3955#issuecomment-1719209803 & https://github.com/graphql/graphql-js/pull/3967#issuecomment-1713484642 it would be great to get this addressed and released.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "graphql",
-  "version": "16.8.0",
+  "version": "16.8.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "graphql",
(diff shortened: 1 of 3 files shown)


tadhglewis (approved):
Are we able to get this in asap? This impacts all consumers unless they have some kind of mitigation in place (which Apollo doesn't do by itself) which I found to be the case about <30% of the time when testing a bunch of different companies

Q: should this be disclosed as a vulnerability? it's already public but consumers should upgrade asap

IvanGoncharov:
@AaronMoat Thanks for the ping, released https://github.com/graphql/graphql-js/releases/tag/v16.8.1

PR #3982, opened 2023-11-06
Simplify collectFields for @defer & @stream

robrichard:


yaacovCR:
Hey @robrichard!

Could we take a look at this rebased on https://github.com/graphql/graphql-js/pull/3984 after that is reviewed/merged => ideally there would be no/minimal test changes at that point...

src/execution/execute.ts
```diff
@@ -1890,9 +1932,10 @@ function executeDeferredGroupedFieldSets(
   path: Path | undefined,
   newDeferredGroupedFieldSetRecords: ReadonlyArray<DeferredGroupedFieldSetRecord>,
   deferMap: ReadonlyMap<DeferUsage, DeferredFragmentRecord>,
+  parentIncrementalDataRecord: IncrementalDataRecord,
 ): void {
   for (const deferredGroupedFieldSetRecord of newDeferredGroupedFieldSetRecords) {
-    if (deferredGroupedFieldSetRecord.shouldInitiateDefer) {
+    if (!('deferredFragmentRecords' in parentIncrementalDataRecord)) {
```
yaacovCR, line 1938: I think this new logic will never defer a new grouped field set if it was spawned by an existing deferred grouped field set, but I think we could have a situation where a defer spawns another defer in a child selection set and must actually be deferred, like:

```graphql
{
  ... @defer {
    field {
      ... @defer {
        subField
      }
    }
  }
}
```

robrichard: If you don't fall in this condition, `executeDeferredGroupedFieldSet` is still called, without being wrapped in `Promise.resolve().then(...)`.

robrichard: I'm not sure if this condition lines up exactly with `shouldInitiateDefer` in the current implementation

robrichard: @yaacovCR this is updated now and the logic should match the current implementation

yaacovCR: i will take another look

src/execution/execute.ts
```diff
@@ -1491,87 +1491,134 @@ function deferredFragmentRecordFromDeferUsage(
...
+    if (
+      !everyNodeIsDeferred ||
+      (parentDeferUsageSet.size !== 0 &&
+        setIsSubsetOf(parentDeferUsageSet, deferUsageSet))
+    ) {
+      currentGroup.set(key, fieldDetails);
+    } else {
+      let deferredGroup = deferredGroups.find((group) =>
+        isSameSet(group.deferUsageSet, deferUsageSet),
+      );
```
yaacovCR, line 1549: because masking here is done differently, this find is not necessarily stable, consider the operation:

```graphql
query HeroNameQuery {
        ... @defer(label: "DeferID") {
          hero {
            id
          }
        }
        ... @defer(label: "DeferName") {
          hero {
            name
            lastName
            ... @defer {
              lastName
            }
          }
        }
      }
```

`hero` is used in two parallel defers, with the `id` subfield present in `DeferID` and both `name` and `lastName` present in `DeferName`. The subfields of `hero` should therefore yield two new grouped field sets, but according to this algorithm, it would yield three, as the additional superfluous defer would still be counted and so the `deferUsageSet` for `lastName` would actually be different for `name`.

yaacovCR (changes requested):

src/execution/execute.ts
yaacovCR: to fix this, you may need to bring back `ancestors` on `DeferUsage`

PR #3983, opened 2023-11-07
fix: cancels deferred fields when overlapping deferred results exhibits null bubbling

yaacovCR:
Given the following operation:

```graphql
      query {
        ... @defer {
          hero {
            nonNullName
            id
          }
        }
        ... @defer {
          hero {
            nonNullName
            name
          }
        }
      }
```

If `nonNullName` returns `null`, the separate incremental entry with that overlapping field will error, and all the "sibling" incremental errors should also be filtered.


yaacovCR:
this additional filtering test case is fixed by the rewrite in #4026 with the test case included => it could be fixed separately prior to #4026, but I do not see the need.

PR #3986, opened 2023-11-09
Reject deprecated fields when interface field is not deprecated

benjie:
This PR implements GraphQL RFC https://github.com/graphql/graphql-spec/pull/1053; please see that PR for details.


saihaj (approved):

src/type/validate.ts
```diff
@@ -433,6 +433,21 @@ function validateTypeImplementsInterface(
         );
       }
     }
+
+    // Asserts that field is not deprecated unless interface field is
+    if (
+      ifaceField.deprecationReason == null &&
+      typeField.deprecationReason != null
```
leebyron, line 440: super nit: flip the order of these two so it reads the same as the spec? "if the type's field is deprecated, then the interface's field must be deprecated"

leebyron (approved):

JoviDeCroock (approved):

src/type/validate.ts
```diff
@@ -573,6 +573,21 @@ function validateTypeImplementsInterface(
...
       }
     }
+
+    // Asserts that field is not deprecated unless interface field is
+    if (
+      typeField.deprecationReason != null &&
+      ifaceField.deprecationReason == null
+    ) {
+      context.reportError(
+        `Interface field ${iface.name}.${fieldName} is not deprecated, so ` +
```
JoviDeCroock, line 583:
```suggestion
        `Interface field ${iface.name}.${ifaceField.name} is not deprecated, so ` +
```

src/type/validate.ts
```diff
@@ -573,6 +573,21 @@ function validateTypeImplementsInterface(
...
+          `implementation field ${type.name}.${fieldName} must not be deprecated.`,
```
JoviDeCroock, line 584:
```suggestion
          `implementation field ${type.name}.${ifaceField.name} must not be deprecated.`,
```

JoviDeCroock (approved):

PR #3987, opened 2023-11-09
fix(incremental): fix logic around selecting id/subPath

yaacovCR:
current loop assumes that the first analyzed DeferredFragmentRecord has been released as pending and thereofre has an `id`, which happens to be the case in this scenario, but is not reliable, uncovered by #3982.

Demonstrates the peril of `!` which might point to requiring additional types, but we can leave that for a different PR.

### DIFF
--- a/src/execution/IncrementalPublisher.ts
+++ b/src/execution/IncrementalPublisher.ts
@@ -609,27 +609,31 @@ export class IncrementalPublisher {
     deferredGroupedFieldSetRecord: DeferredGroupedFieldSetRecord,
   ): IncrementalDeferResult {
     const { data, deferredFragmentRecords } = deferredGroupedFieldSetRecord;
-    let maxLength = deferredFragmentRecords[0].path.length;
-    let maxIndex = 0;
-    for (let i = 1; i < deferredFragmentRecords.length; i++) {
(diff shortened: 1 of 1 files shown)


src/execution/IncrementalPublisher.ts
```diff
@@ -609,27 +609,31 @@ export class IncrementalPublisher {
...
+    for (const deferredFragmentRecord of deferredFragmentRecords) {
+      if (deferredFragmentRecord.id === undefined) {
+        continue;
+      }
       const length = deferredFragmentRecord.path.length;
-      if (length > maxLength) {
+      if (maxLength === undefined || length > maxLength) {
         maxLength = length;
-        maxIndex = i;
+        recordWithLongestPath = deferredFragmentRecord;
```
robrichard, line 621: You can remove one `!` by declaring `id` before the loop and assigning `id` inside this condition.

robrichard (approved):

src/execution/IncrementalPublisher.ts
yaacovCR: awesome, thanks

[969f417](https://github.com/graphql/graphql-js/pull/3987/commits/969f4173dd7b241c885c5920fd4e7b23961e1e0a)

PR #3992, opened 2023-11-22
Use node v20 for CI

robrichard:


package.json
```diff
@@ -30,10 +30,10 @@
   },
   "scripts": {
     "preversion": ". ./resources/checkgit.sh && npm ci --ignore-scripts",
-    "version": "ts-node resources/gen-version.ts && npm test && git add src/version.ts",
+    "version": "node --loader ts-node/esm resources/gen-version.ts && npm test && git add src/version.ts",
```
robrichard, line 33: See https://github.com/TypeStrong/ts-node/issues/1997

IvanGoncharov (approved):
Looks, good!
One problem is that it breaks, CI job that diffs packages.
If you can't quickly fix it, then just disable it since it's critical to dev workflow.

robrichard:
@IvanGoncharov i think the diff check breaks because it’s loading the prior commit which does not have the fixes in the pr. 

So I think once this is merged it will start working again

PR #3994, opened 2023-12-08
simplify CollectFields for `@defer` and `@stream`

yaacovCR:
minimizes changes to CollectFields a la #3982 but still creates a single memoized incremental field plan per list item.

@robrichard

### DIFF
--- a/src/execution/buildFieldPlan.ts
+++ b/src/execution/buildFieldPlan.ts
@@ -1,12 +1,7 @@
 import { getBySet } from '../jsutils/getBySet.js';
 import { isSameSet } from '../jsutils/isSameSet.js';
 
-import type { FieldDetails } from './collectFields.js';
-
-export interface DeferUsage {
-  label: string | undefined;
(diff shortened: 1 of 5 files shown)


src/execution/buildFieldPlan.ts
```diff
@@ -0,0 +1,152 @@
+import { getBySet } from '../jsutils/getBySet.js';
+import { isSameSet } from '../jsutils/isSameSet.js';
+
+import type { FieldDetails } from './collectFields.js';
+
+export interface DeferUsage {
```
robrichard, line 6: This same type is also defined in `collectFields.ts`?

yaacovCR: cff34775 thanks!

yaacovCR:
Updated https://github.com/graphql/graphql-spec/pull/1052 to reflect simplification from this PR.

robrichard (approved):

PR #3995, opened 2023-12-10
Add reference types

fersilva16:
This pull request introduces string references for GraphQL types, which aims to solve the problem with circular dependency between types in multiple files. Here's an usage example:

```ts
type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull('User')))
```

This PR has only the implementation for output types for now.

I've also made a RFC repository with a patch and pratical examples for testing: https://github.com/fersilva16/graphql-refs-rfc

Contributions and feedback are welcome.

### DIFF
--- a/src/__tests__/starWarsData.ts
+++ b/src/__tests__/starWarsData.ts
@@ -27,6 +27,22 @@ export interface Droid {
   primaryFunction: string;
 }
 
+export interface Jedi {
+  type: 'Jedi';
+  id: string;
+  human: string;
+  lightsaberColor: string;
+  enemies: ReadonlyArray<string>;
+}
+
+export interface Sith {
+  type: 'Sith';
+  id: string;
+  human: string;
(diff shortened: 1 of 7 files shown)


noghartt (approved):

samueldurantes (approved):

daniloab:
How could we move forward with this pr?

yaacovCR:
This was discussed at the GraphQL JS WG meeting on [March 27](https://github.com/graphql/graphql-js-wg/blob/main/agendas/2024/03-Mar/27-graphql-js-wg-march-2024.md) per the notes.

Summarizing from my recollection, I think this PR was felt to be a reasonable alternative, but the question was how is it better than the current option of using thunks, and balancing the desire for flexibility vs. reducing maintenance burden by having a canonical way of doing things.

I am going to close this for now, feel free to reopen as needed.

PR #3997, opened 2023-12-18
Add test for consolidating grouped field sets properly into deferred fragments

yaacovCR:
Group field sets should be properly consolidated when some of the fields in a sibling defer are masked by parent fields.


yaacovCR:
Extracted from #3993 @robrichard

robrichard (approved):

PR #3998, opened 2023-12-19
create new deferred grouped field set records lazily

yaacovCR:
Before https://github.com/graphql/graphql-js/pull/3984, we could not do this because of problems with empty defers. Now, it's a safe optimization.

Consider:

```graphql
      query {
        alwaysThrows
        ... @defer(name: "WillBeEmpty") {
          someField
        }
      }
```

Before the current change, when `alwaysThrows` throws, the deferred grouped field set for `someField` would be found to be in the "blast radius", and deferred fragment `WillBeEmpty` would be filtered. This would yield:

```json
    {
      "data": {
        "alwaysThrows": null,
      },
      "errors": [ ... ]
    }
```

Prior to https://github.com/graphql/graphql-js/pull/3984, if we had made the current change and created the deferred grouped field set for `someField` lazily, `someField` would not be found to be in the "blast radius" -- because it would not have been created yet -- and `WillBeEmpty` would not have been filtered.

However, with https://github.com/graphql/graphql-js/pull/3984, `WillBeEmpty` is not sent because all empty deferred fragments are never sent and do not require filtering. This is now a safe change.

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -426,13 +426,6 @@ function executeOperation(
 
   const path = undefined;
 
-  const newDeferredGroupedFieldSetRecords = addNewDeferredGroupedFieldSets(
-    incrementalPublisher,
-    newGroupedFieldSetDetailsMap,
-    newDeferMap,
-    path,
-  );
-
   let result;
   switch (operation.operation) {
     case OperationTypeNode.QUERY:
@@ -476,7 +469,7 @@ function executeOperation(
     rootType,
     rootValue,
     path,
-    newDeferredGroupedFieldSetRecords,
+    newGroupedFieldSetDetailsMap,
     newDeferMap,
   );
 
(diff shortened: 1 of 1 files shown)


yaacovCR:
Note that there is a tweak/fix for #3984 in #3993 for subsequent results.

yaacovCR:
Actually, wait => this is still not safe!

What if a deferred fragment that should be filtered is never filtered, but it wasn't otherwise empty?

Consider:

```graphql
      query {
        someRootField {
          alwaysThrows
        }
        ... @defer(name: "WillNotBeEmpty") {
          someRootField {
            someNestedField
          }
          someOtherRootField
        }
      }
```

By the time `alwaysThrows` throws, a deferred grouped field set could be created for `someOtherRootField` (if `someRootField` itself is async, for example) and so deferred fragment `WillNotBeEmpty` will not be empty. But it cannot be delivered, because it should be found to be in the blast radius if `alwaysThrows` by `someRootField`.

yaacovCR:
TODO: add test for above

Issue #4001, opened 2023-12-20
Errors thrown when iterating over subscription source event streams (AsyncIterables) should be caught


aseemk:
Hi there. We're using `graphql-js` and serving subscriptions over WebSocket via [`graphql-ws`](https://github.com/enisdenjo/graphql-ws) (as recommended by Apollo for both server and client).

### Our usage/context

In our subscriptions' `subscribe` methods, we always return an `AsyncIterable` pretty much right away. We typically do this either by defining our methods via async generator functions (`async function*`), or by calling [`graphql-redis-subscriptions`](https://github.com/davidyaha/graphql-redis-subscriptions)'s `asyncIterator` method. Our `subscribe` methods effectively never throw an error just providing an `AsyncIterable`.

However, we occasionally hit errors actually streaming subscription _events_, when `graphql-js` calls our `AsyncIterable`'s `next()` method. E.g. Redis could be momentarily down, or an upstream producer/generator could fail/throw. So we sometimes `throw` errors during iteration. And importantly, this can happen _mid_-stream.

### Problem

**`graphql-js` does not try/catch/handle errors when iterating over an `AsyncIterable`:**

https://github.com/graphql/graphql-js/blob/2aedf25e157d1d1c8fdfeaa4c0d2f3d9d3457dba/src/execution/mapAsyncIterable.ts#L38-L40

There's even a test case today that explicitly expects _these_ errors to be re-thrown:

https://github.com/graphql/graphql-js/blob/8a95335f545024c09abfa0f07cc326f73a0e466f/src/execution/__tests__/subscribe-test.ts#L1043-L1047

`graphql-ws` doesn't try/catch/handle errors thrown during iteration either:

https://github.com/enisdenjo/graphql-ws/blob/e4a75cc59012cad019fa3711287073a4aef9ed05/src/server.ts#L813-L815

As a result, when occasional errors happen like this, **the entire underlying WebSocket connection is closed.**

This is obviously not good! 😅 This interrupts every other subscription the client may be subscribed to at that moment, adds reconnection overhead, drops events, etc. And if we're experiencing some downtime on a specific subscription/source stream, this'll result in repeat disconnect-reconnect thrash, because the client also has no signal on _which_ subscription has failed!!

### Inconsistency

[...]

PR #4004, opened 2023-12-31
Fix:Errors thrown when iterating over subscription source event streams (AsyncIterables) should be caught.

xonx4l:
issue-:#4001


yaacovCR:
Overall, +1 on the change, I think it's a great feature. I wonder if we can streamline a bit how the stream errors are caught. Commenting more on the linked issue.

philippemiguet:
We also need this change: we are using AsyncGenerator to fight EventLoopLag using GraphQL v17-alpha and throwing in an AsyncGenerator is crashing the NodeJS process and our pods (Scalar validation error for instance).

PR #3859, opened 2023-03-03
BugFix: Nested List Input Coercion not matching GraphQL spec 

abhinand-c:
This PR fixes the issue mentioned in https://github.com/graphql/graphql-js/issues/3858 to align input coercion of nested list with the [GraphQL spec](https://spec.graphql.org/October2021/#sel-HAHjBJLDLEBAAADFAb0le).

#### Changes (Similar to https://github.com/graphql-python/graphql-core/pull/194)
- Adds check to validate and raises error for the condition
- Updated tests to reflect the new logic, and validate error

PR #4006, opened 2024-01-05
Preserve @deprecated when reason = null

captbaritone:
I noticed an issue where creating a `GraphQLSchema` from an SDL and then printing it again results in dropping a `@deprecated` directive if the reason is explicitly set to `null`.

In `GraphQLSchema` we don't store the existence of a deprecated directive and its reason separately. We simply store the `deprecatedReason`. Because `reason` has a default, this is mostly fine since even if you don't provide a reason string, you still end up with one. However, the oddities of defaults in GraphQL means that you can pass an _explicit_ `null` as your reason.

In this case you end up with `deprecatedReason: null` in your `GraphQLSchema` which we currently treat as the same as `undefined` when printing, meaning the `@deprecated(reason: null)` gets dropped.

I open this PR mostly to raise this issue rather than to advocate for this specific solution. I recognize that interfaces like `GraphQLFieldConfig` are part of the public API, and thus this might mean the change is technically a breaking change, since there are likely places where people assumed they could set `deprecatedReason: null` to indicate the field was not deprecated.

Another (even more breaking) option would be to decide that the reason argument to `@deprecated` should be non-optional, and thus allow us to always assume that every deprecated field has a non-nullable reason.


captbaritone:
Here's a minimal repro of the issue:

```javascript
import { printSchema, buildSchema } from "graphql";

const SDL = `
type Query {
  hello: String @deprecated(reason: null)
}`;

console.log(printSchema(buildSchema(SDL)));

/* Outputs:
type Query {
  hello: String
}
*/
```

JoviDeCroock (approved):
This looks like a valid change to me as the value is explicitly set to be nullable, and is hinted at at in the description that the reason is _usually_ there

https://github.com/graphql/graphql-js/blob/9c90a23dd430ba7b9db3d566b084e9f66aded346/src/type/directives.ts#L224-L231

yaacovCR:
My guess is that this is an artifact from the fact that originally explicit nulls were not allowed (or possibly the default superceded them?) -- I am not personally aware of the history, just reading a bit from https://github.com/graphql/graphql-spec/pull/418 -- and that when those changes were made, the directive definition should have been updated from:

```graphql
directive @deprecated(
  reason: String = "No longer supported"
) on FIELD_DEFINITION | ENUM_VALUE
```

to

```graphql
directive @deprecated(
  reason: String! = "No longer supported"
) on FIELD_DEFINITION | ENUM_VALUE
```

I do wonder then whether a "bug fix" should be issued here in terms of the implementation or in terms of the spec. [Reminds me of https://github.com/graphql/graphql-js/pull/3859]

My guess is that the best course of action is to bring this up at a WG meeting!

Nice find!

benjie (approved):
We should make this argument non-nullable in the spec. Nevertheless, I approve of this change.

yaacovCR:
Thanks for bringing this up.

Looks like in terms of the solution, this was superseded by:

- #4299

Closing, feel free to reopen if I've misunderstood or there is more to discuss!

PR #3847, opened 2023-02-09
Fragment Arguments: Only Syntax Changes

mjmahone:
This pulls in requested changes from https://github.com/graphql/graphql-js/pull/3835 and splits the PR into syntax changes and the derivatives, so that we can focus most of our review effort on the gnarlier algorithmic and validation changes that are happening on top of the syntax updates.

PR #4015, opened 2024-01-26
feat: add experimental support for parsing fragment arguments

JoviDeCroock:
This is a rebase of https://github.com/graphql/graphql-js/pull/3847

I will follow up with a PR for

- New Validation rules
- Validation changes
- Execution changes


mjmahone (commented):
Thank you for rebasing the syntax changes! This looks correct to me.

I'm not sure whether we need the full stack of changes up before merging any of the changes: I'd be surprised if the syntax changes need to be modified due to the validation or execution changes.

This update is probably worth bringing to the graphql-wg this week: https://github.com/graphql/graphql-wg/blob/main/agendas/2024/02-Feb/01-wg-primary.md

JoviDeCroock:
@mjmahone worth bringing up that the PR is up, or should that be for the graphql-js wg? Not entirely sure 😅

mjmahone:
> worth bringing up that the PR is up, or should that be for the graphql-js wg? 

I think it's worth adding a 5-10 minute dicussion item to the WG meeting this thursday https://github.com/graphql/graphql-wg/blob/main/agendas/2024/02-Feb/01-wg-primary.md

mjmahone:
> The implementation in this branch passes tests but uses UNSET. However, there's an alternative here: https://github.com/JoviDeCroock/graphql-js/commit/1ee3f42d6101e9755a7103bca3316006ba6ec901, which has some updated tests.

I don't think that is true anymore? I think @JoviDeCroock removed all the UNSET hacks from *this* PR (they might still exist in the executor changes). It also looks like [graphql-js #4015](https://github.com/JoviDeCroock/graphql-js/pull/2) no longer uses the UNSET hack.

We should definitely ship the graphql-js changes that most closely match the changes to the spec itself, and the UNSET hack *does not* match the spec changes from [Spec #1081 ](https://github.com/graphql/graphql-spec/pull/1081), but this looks like the parser changes are in a reasonably good state right now to my eyes!

JoviDeCroock:
@mjmahone good spot, I forgot to update this description thank you for spotting

JoviDeCroock:
@graphql/graphql-js-reviewers is there anything missing to move this and the linked PR's forward?

nandorojo:
So excited for this! Let me know if you want any help testing it in a real app.

JoviDeCroock:
Would this PR be better served being aimed at the 16.x.x branch? Not sure how the release schedule of this would work

src/language/parser.ts
```diff
@@ -576,29 +592,20 @@ export class Parser {
...
-        selectionSet: this.parseSelectionSet(),
-      });
-    }
     return this.node<FragmentDefinitionNode>(start, {
       kind: Kind.FRAGMENT_DEFINITION,
       name: this.parseFragmentName(),
+      variableDefinitions:
+        this._options.experimentalFragmentArguments === true
+          ? this.parseVariableDefinitions()
+          : undefined,
```
yaacovCR, line 608: do we want the key to still show up (as undefined) when the experimental flag is turned off? or is the current behavior (for legacy frag args) better in that the key is skipped entirely?

JoviDeCroock: I think undefined shouldn't be a problem but can always undo, don't mind that

src/language/printer.ts
```diff
@@ -392,3 +391,15 @@ function hasMultilineItems(maybeArray: Maybe<ReadonlyArray<string>>): boolean {
   /* c8 ignore next */
   return maybeArray?.some((str) => str.includes('\n')) ?? false;
 }
+
+function wrappedLineAndArgs(
+  prefix: string,
```
yaacovCR, line 396: Great that this is extracted out to re-use logic.

yaacovCR (approved):

yaacovCR:
Motivation is only to preserve backwards compatibility for those not wishing to upgrade their visitors while fragment arguments are experimental. So another alternative is something in between, introduce a new Node now, maybe something like `Kind.EXPERIMENTAL_FRAGMENT_ARGUMENT` and then when fragment arguments are released, we would remove the EXPERIMENTAL and force everyone to upgrade their existing stuff (and force early adopters to integrate any custom visitors for both types of arguments).

saihaj (approved):

PR #4017, opened 2024-02-06
fix(jsutils): use optional chaining to prevent crash in browser

michaelbirchler:
## Context

We use [contentful/live-preview](https://github.com/contentful/live-preview). The library has `graphql-js` as a dependency. 

## Problem

If the DOM has an element with `id="process"` eg. `<div id="process"></div>` than the current implementation of `utils/instanceOf` would fail.

### Current implementation:
```js
globalThis.process != null && globalThis.process.env.NODE_ENV === 'production'
```
with this implementation `globalThis.process` will return the DOM element:
image

Because `globalThis.process is true`, the condition `globalThis.process.env.NODE_ENV === 'production'` gets executed and will result with an javascript error:
image

### DIFF
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -9,7 +9,7 @@ import { inspect } from './inspect.js';
 export const instanceOf: (value: unknown, constructor: Constructor) => boolean =
   /* c8 ignore next 6 */
   // FIXME: https://github.com/graphql/graphql-js/issues/2317
-  globalThis.process != null && globalThis.process.env.NODE_ENV === 'production'
+  globalThis?.process?.env?.NODE_ENV === 'production'
     ? function instanceOf(value: unknown, constructor: Constructor): boolean {
         return value instanceof constructor;
       }


src/jsutils/instanceOf.ts
```diff
@@ -9,7 +9,7 @@ import { inspect } from './inspect.js';
 export const instanceOf: (value: unknown, constructor: Constructor) => boolean =
   /* c8 ignore next 6 */
   // FIXME: https://github.com/graphql/graphql-js/issues/2317
-  globalThis.process != null && globalThis.process.env.NODE_ENV === 'production'
+  globalThis?.process?.env?.NODE_ENV === 'production'
```
JoviDeCroock, line 12: This will be more inline with the rules of the project

```suggestion
  globalThis.process != null && globalThis.process.env != null && globalThis.process.env.NODE_ENV === 'production'
```

yaacovCR:
This is fixed in v17 which no longer uses environment variables such as `NODE_ENV`, but instead uses the `development` condition within the exports map.

PR #4018, opened 2024-02-08
Enable passing values configuration to GraphQLEnumType as a thunk

benjie:
Many of the type configuration objects in GraphQL-js have properties that accept thunks:

- GraphQLObjectTypeConfig.fields
- GraphQLObjectTypeConfig.interfaces
- GraphQLInterfaceTypeConfig.fields
- GraphQLInterfaceTypeConfig.interfaces
- GraphQLInputObjectTypeConfig.fields
- GraphQLUnionTypeConfig.types

These are because their definitions may contain references to other types which may form cycles, and thus a thunk is required to enable their definition.

This PR allows setting `GraphQLEnumTypeConfig.values` to be a thunk. At first, this might seem unnecessary (and, well, it has been this way for 9 years so you can be forgiven for thinking that) but it can be quite useful when building enums that reference types. For example: you might have a field that returns a union (`Person.pets`), and you might want to add an argument that allows you to limit the returned types to only certain types (`Person.pets(only:)`):

```graphql
type Person {
  name: String
  pets(only: [AnimalType!]): [Animal]
}
enum AnimalType {
  Cat
  Dog
  Hamster
  Mouse
}
union AnimalType = Cat | Dog | Hamster | Mouse
type Cat { name: String, owner: Person, numberOfLives: Int }
type Dog implements Animal { name: String, owner: Person, wagsTail: Booean }
type Hamster implements Animal { ... }
type Mouse implements Animal { ... }
type Query {
  currentPerson: Person
}
```

```graphql
query {
  currentPerson {
    pets(only: [Cat, Dog]) {
      ... on Cat {
        name
        numberOfLives
      }
      ... on Dog {
        name
        wagsTail
      }
    }
  }
}
```

For this it can be useful for the `AnimalType` type to enumerate the types in the `Animal` union, but those types contain references back to `AnimalType` -> there's a cycle.

The thunk approach outlined in this PR would solve the cycle. There are workarounds (i.e. manually listing out the types) but they're not as ergonomic as (and are more error-prone than) resolving the values directly from the union.

### DIFF
--- a/src/type/__tests__/enumType-test.ts
+++ b/src/type/__tests__/enumType-test.ts
@@ -31,6 +31,14 @@ const ComplexEnum = new GraphQLEnumType({
   },
 });
 
+const ThunkValuesEnum = new GraphQLEnumType({
+  name: 'ThunkValues',
+  values: () => ({
+    A: { value: 'a' },
+    B: { value: 'b' },
+  }),
+});
+
 const QueryType = new GraphQLObjectType({
   name: 'Query',
   fields: {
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
That's a great find!

PR #4019, opened 2024-02-12
Use Optional Chaining to `globalThis.process.env.NODE_ENV` Retrieval

johnfabre:
This PR introduces optional chaining when accessing `globalThis.process.env.NODE_ENV` in the graphql-js codebase. 

The current implementation assumes that `globalThis.process.env` is always defined, which may not be the case in all environments. This assumption leads to a runtime error when `NODE_ENV` is accessed on an undefined `process.env`.

By adding optional chaining (`?.`), we ensure that `NODE_ENV` is only accessed if `globalThis.process` and `process.env` are defined, thereby preventing potential runtime errors.

This change improves the robustness of the code and makes it more compatible with various JavaScript environments.

### DIFF
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -9,7 +9,7 @@ import { inspect } from './inspect.js';
 export const instanceOf: (value: unknown, constructor: Constructor) => boolean =
   /* c8 ignore next 6 */
   // FIXME: https://github.com/graphql/graphql-js/issues/2317
-  globalThis.process != null && globalThis.process.env.NODE_ENV === 'production'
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
This is a duplicate of https://github.com/graphql/graphql-js/pull/4017

johnfabre:
added myself to CLA

> This is a duplicate of #4017

used the code suggested in here
https://github.com/graphql/graphql-js/pull/4017/files#r1483265402

JoviDeCroock:
Fixed with https://github.com/graphql/graphql-js/pull/4022

PR #4021, opened 2024-02-13
fix#3978

### LINKED ISSUE #3978: Error after upgrading from 16.6 to 16.8.1 with globalThis.process.env.NODE_ENV
Running a project with Vue3 / Quasar2 / Vite in development mode fails after update.

```
// node_modules/graphql/jsutils/instanceOf.mjs
globalThis.process && globalThis.process.env.NODE_ENV === 'production'
```

is compiled to

```
var instanceOf = globalThis.process && globalThis."development" === "production" ? function instanceOf2(value, constructor) { ... }
```

Changing

```
globalThis.process && globalThis.process.env.NODE_ENV === 'production'
```

to

```
globalThis.process && globalThis.process.env['NODE_ENV'] === 'production'
```

solves the problem for me.

mobsean:
fix #3978

### DIFF
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -9,7 +9,7 @@ import { inspect } from './inspect.js';
 export const instanceOf: (value: unknown, constructor: Constructor) => boolean =
   /* c8 ignore next 6 */
   // FIXME: https://github.com/graphql/graphql-js/issues/2317
-  globalThis.process != null && globalThis[process.env.NODE_ENV] === 'production'
+  globalThis.process != null && globalThis['development' | 'production'] === 'production'
     ? function instanceOf(value: unknown, constructor: Constructor): boolean {
         return value instanceof constructor;
       }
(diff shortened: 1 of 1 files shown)


src/jsutils/instanceOf.ts
```diff
@@ -9,7 +9,7 @@ import { inspect } from './inspect.js';
 export const instanceOf: (value: unknown, constructor: Constructor) => boolean =
   /* c8 ignore next 6 */
   // FIXME: https://github.com/graphql/graphql-js/issues/2317
-  globalThis.process != null && globalThis.process.env.NODE_ENV === 'production'
+  globalThis.process != null && globalThis[process.env.NODE_ENV] === 'production'
```
JoviDeCroock, line 12: This might "fix" the issue but introduces another one, this will get replaced with `globalThis['development' | 'production']` rather than the intended way where we analyse the NODE_ENV

mobsean: changed my PR

JoviDeCroock: That's not what I meant 😅 more that this is breaking off of the existing behavior as described in https://github.com/graphql/graphql-js/pull/3927

Starting a thread on Discord atm how we should go about this

JoviDeCroock:
Fixed with https://github.com/graphql/graphql-js/pull/4022

Issue #3934, opened 2023-07-11
Unexpected string (at @apollo_client.js?v=:1230:51) in browser


bricekams:
# Error in dev tools

![Screenshot from 2023-07-11 12-54-08](https://github.com/graphql/graphql-js/assets/120424696/4383609c-ce94-4a0a-92f9-2dbc108c67d3)

## Dependencies 
![Screenshot from 2023-07-11 12-57-48](https://github.com/graphql/graphql-js/assets/120424696/bc6aebb5-9acd-4eba-89cb-64327f4f8032)

## Environments:
- node: v18.16.0
- npm: 9.5.1
- yarn: 1.22.19
- pnpm: 8.3.1

JoviDeCroock:
This is fixed in https://github.com/graphql/graphql-js/pull/4022 and published in GraphQL 16.8.2, you can find the instructions for various bundlers [here](https://github.com/graphql/graphql-js/blob/16.x.x/website/docs/tutorials/going-to-production.md) on how to remove it in production.

Issue #3928, opened 2023-06-27
rollup-plugin-inject: failed to parse graphql/jsutils/instanceOf.mjs


iBobik:
When this library is used in the Nuxt 3 project and compiled for Cloudflare it fails like this:

```
ℹ Building Nitro Server (preset: cloudflare-pages)                  nitro 19:16:00
(inject plugin) rollup-plugin-inject: failed to parse /home/projects/github-twqhau-upewai/node_modules/graphql/jsutils/instanceOf.mjs. Consider restricting the plugin to particular files via options.include

[nitro 19:16:14]  ERROR  RollupError: Unexpected token (Note that you need plugins to import files that are not JavaScript)


10:   /* c8 ignore next 6 */
11:   // FIXME: https://github.com/graphql/graphql-js/issues/2317
12:   globalThis.process && globalThis.process.env.NODE_ENV === 'production'
                                       ^
13:     ? function instanceOf(value, constructor) {
14:         return value instanceof constructor;


[19:16:14]  ERROR  Unexpected token (Note that you need plugins to import files that are not JavaScript)

  at error (<anonymous>)
  at setSource (<anonymous>)
  at  (<anonymous>)

```

Happens in version 16.7.1 and 16.7.0 but works with 16.6.0 and 17.0.0-alpha.2.

Repro: https://stackblitz.com/edit/github-twqhau-upewai
Start with: `NITRO_PRESET=cloudflare-pages npm run build`

Aietes:
This is tracked as an issue in Nuxt and nitro: nuxt/nuxt#21768. The module tries to access `globalThis.process`, which makes the Cloudflare pages deployment fail. You can't access `process` in Cloudflare, as it's a node-only API. Technically it should be poyfilled by unenv, but there seems to be an issue preventing this. As a work-around, in our project we are currently patching the respective code in the graphql module.

```bash
pnpm patch graphql
code <generated_folder>
```
In `node_modules/graphql/jsutils/instanceOf.mjs` remove the code that tries to access `process`.

```ts
export const instanceOf = function instanceOf(value, constructor) {
  return value instanceof constructor
}
```
Apply the patch.

```bash
pnpm patch-commit <generated_folder>
```

The resulting patch should appear in package.json to be applied during deployment on Cloudflare.

```json
  "pnpm": {
    "patchedDependencies": {
      "graphql@16.7.1": "patches/graphql@16.7.1.patch"
    }
  }
```

utsavkapoor:
[...]

Issue #3918, opened 2023-06-21
Regression: Importing the package fails in a brand new Vite project


srmagura:
Minimal repro: https://github.com/srmagura/graphql-repro

The project was created with:

```shell
npm create-vite@latest graphql-repro
```

Then I chose "Vanilla" and "TypeScript" at the prompts.

Install the `graphql` npm package and add the import

```ts
import * as graphql from "graphql";
```

It produces an error like:

```ts
Uncaught SyntaxError: Unexpected string (at graphql.js?v=4b27e75c:1248:113)
```

If you click on the error in the console, you'll see that `process.env.NODE_ENV` has been replaced with "development".

![image](https://github.com/graphql/graphql-js/assets/801549/e588221f-afbc-4e1c-a6c8-97587e2ae828)

Let me know if you think this is a Vite issue instead, and I will report it there.

airhorns:
This got us too -- I think the regexp here is not working quite right: https://github.com/sun0day/vite/blob/5c6c0d381755cf7bd46854937e7dd3e2aea323f3/packages/vite/src/node/plugins/clientInjections.ts#L68-L72

bburr:
Just adding more info - the `$` in `_globalThis$process` seen here is what's not accounted for in the regexp - https://diff.intrinsic.com/graphql/16.6.0/16.7.0#file-jsutils__instanceOf.mjs

So `_globalThis$process.env.NODE_ENV` gets changed to `_globalThis$"development"` by Vite

IvanGoncharov:
@srmagura It's an issue with Vite internals, but I added a workaround for it in `16.7.1`
Can you please confirm that it is fixed?

KangMiSun17:
@IvanGoncharov I was having the same problem and it was solved. Thank you.

LrMiLoef:
i still have the problem, using the `16.7.1` and vite `2.9.16`.  
when i undo your fix, it's working.

lemonmade:
@IvanGoncharov I believe that the issue is still present, even with your change applied:

<img width="1185" alt="Screenshot 2023-06-23 at 2 46 14 PM" src="https://github.com/graphql/graphql-js/assets/3012583/ee6763d5-7d41-4026-8fef-e3652b791675">

[...]

PR #4022, opened 2024-02-14
fix: remove `globalThis` check and align with what bundlers can accept

JoviDeCroock:
As surfaced in [Discord](https://discord.com/channels/625400653321076807/862957336082645006/1206980831915282532) this currently is a breaking change in the 16.x.x release line which is preventing folks from upgrading towards a security fix. This PR should result in a patch release on the 16 release line.

This change was originally introduced to support CFW and browser environments which should still be supported with the `typeof` check CC @n1ru4l

This also adds a check whether `.env` is present as in the DOM using `id="process"` defines that as a global which we don't want to access on accident. as shown in https://github.com/graphql/graphql-js/pull/4017

Bundles also target `process.env.NODE_ENV` specifically which fails when it replaces `globalThis.process.env.NODE_ENV` as this becomes `globalThis."production"` which is invalid syntax.

Fixes https://github.com/graphql/graphql-js/issues/3978
Fixes https://github.com/graphql/graphql-js/issues/3918
Fixes https://github.com/graphql/graphql-js/issues/3928
Fixes https://github.com/graphql/graphql-js/issues/3758
Fixes https://github.com/graphql/graphql-js/issues/3934

This purposefully does not account for https://github.com/graphql/graphql-js/issues/3925 as we can't address this without breaking CF/plain browsers so the small byte-size increase will be expected for bundled browser environments. As a middle ground we did optimise the performance here. We can revisit this for v17.

Most bundlers will be able to tree-shake this with a little help, in https://github.com/graphql/graphql-js/issues/4075#issuecomment-2094052098 you can find a conclusion with a repo where we discuss a few.

- Next.JS by default replaces [`process.env.NODE_ENV`](https://github.com/vercel/next.js/blob/b0ab0fe85fe8c93792051b058e060724ff373cc2/packages/next/webpack.config.js#L182) you can add `typeof process` linearly
- Vite allows you to specify [`config.define`](https://vitejs.dev/config/shared-options.html#define)
- ESBuild by default will replace `process.env.NODE_ENV` but does not support replacing `typeof process`
- Rollup has a plugin for this https://www.npmjs.com/package/@rollup/plugin-replace

Supersedes https://github.com/graphql/graphql-js/pull/4021
Supersedes https://github.com/graphql/graphql-js/pull/4019
Supersedes https://github.com/graphql/graphql-js/pull/3927

> This now also adds a documentation page on how to remove all of these

### DIFF
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -2,7 +2,6 @@ import { inspect } from './inspect';
 
 /* c8 ignore next 3 */
 const isProduction =
-  // eslint-disable-next-line no-undef
   typeof process !== 'undefined' &&
   // eslint-disable-next-line no-undef
   process.env.NODE_ENV === 'production';
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -2,7 +2,7 @@ import { inspect } from './inspect';
 
 /* c8 ignore next 3 */
 const isProduction =
-  typeof process !== 'undefined' &&
+  globalThis.process &&
   // eslint-disable-next-line no-undef
   process.env.NODE_ENV === 'production';
 
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -1,6 +1,13 @@
 import { inspect } from './inspect';
 
-const isProduction = typeof process !== 'undefined' && process.env && process.env.NODE_ENV === 'production';
+/* c8 ignore next 4 */
+const isProduction =
+  // eslint-disable-next-line no-undef
+  typeof process !== 'undefined' &&
+  // eslint-disable-next-line no-undef
+  process.env &&
+  // eslint-disable-next-line no-undef
+  process.env.NODE_ENV === 'production';
 
 /**
  * A replacement for instanceof which includes an error warning when multi-realm
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -1,11 +1,9 @@
 import { inspect } from './inspect';
 
-/* c8 ignore next 4 */
+/* c8 ignore next 3 */
 const isProduction =
   // eslint-disable-next-line no-undef
   typeof process !== 'undefined' &&
-  // eslint-disable-next-line no-undef
-  process.env &&
   // eslint-disable-next-line no-undef
   process.env.NODE_ENV === 'production';
 
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -1,5 +1,7 @@
 import { inspect } from './inspect';
 
+const isProduction = typeof process !== 'undefined' && process.env && process.env.NODE_ENV === 'production';
+
 /**
  * A replacement for instanceof which includes an error warning when multi-realm
  * constructors are detected.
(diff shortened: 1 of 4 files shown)


benjie (approved):
LGTM

JoviDeCroock:
After talking to @phryneas there seems to be a lot of pushback from his side with regards to this change opting out of tree-shaking

src/jsutils/instanceOf.ts
```diff
@@ -1,5 +1,14 @@
 import { inspect } from './inspect';
 
+/* c8 ignore next 4 */
+const isProduction =
+  // eslint-disable-next-line no-undef
+  typeof process !== 'undefined' &&
+  // eslint-disable-next-line no-undef
+  process.env &&
```
JoviDeCroock, line 8: We can remove this to expedite merging this, this issue should only occur when we have a DOM-node that has an `id` property with the value `process`

```suggestion
  // eslint-disable-next-line no-undef
  process.env &&
```

JoviDeCroock:
CC @graphql/graphql-js-reviewers

src/jsutils/instanceOf.ts
```diff
@@ -1,5 +1,14 @@
...
+  // eslint-disable-next-line no-undef
+  process.env.NODE_ENV === 'production';
+
```
JoviDeCroock, line 11:
```suggestion
/* c8 ignore next 4 */
const isDev =
  // eslint-disable-next-line no-undef
  typeof process !== 'undefined' &&
  // eslint-disable-next-line no-undef
  process.env &&
  // eslint-disable-next-line no-undef
  process.env.NODE_ENV === 'development';

```

src/jsutils/instanceOf.ts
```diff
@@ -9,7 +18,7 @@ import { inspect } from './inspect';
 export const instanceOf: (value: unknown, constructor: Constructor) => boolean =
   /* c8 ignore next 6 */
   // FIXME: https://github.com/graphql/graphql-js/issues/2317
-  globalThis.process && globalThis.process.env.NODE_ENV === 'production'
+  isProduction
```
JoviDeCroock, line 21:
```suggestion
  !isDevelopment
```

n1ru4l (approved):

src/jsutils/instanceOf.ts
```diff
@@ -1,5 +1,14 @@
 import { inspect } from './inspect';
 
+/* c8 ignore next 4 */
+const isProduction =
+  // eslint-disable-next-line no-undef
+  typeof process !== 'undefined' &&
+  // eslint-disable-next-line no-undef
+  process.env &&
+  // eslint-disable-next-line no-undef
+  process.env.NODE_ENV === 'production';
```
JoviDeCroock, line 10: Another alternative which does not fix the `id="process"` bug

```suggestion
/* c8 ignore next 3 */
const isProduction =
  // eslint-disable-next-line no-undef
  typeof process !== 'undefined' &&
  // eslint-disable-next-line no-undef
  process.env.NODE_ENV === 'production';
```

src/jsutils/instanceOf.ts
echocrow: +1 for the dev detection instead of prod detection here. in case of a false-negative, better to hit the prod branch in dev than the dev code in prod.

JoviDeCroock: In the ongoing discussions we concluded this is a breaking change and hence not worth it for v16

echocrow: ah, makes sense - cheers for the info and cross-reference!

saihaj (approved):

JoviDeCroock:
CI is failing with a rate limit issue, I can't force a retry 😅

benjie:
Love love love the documentation page; excellent work! :raised_hands:

Thompson1985 (approved):

PR #4025, opened 2024-03-07
feat: fragmentifyDocument utility

samuelAndalon:
This PR introduces a utility function `fragmentifyDocument` that significantly optimize GraphQL ASTs by programmatically converting `InlineFragments` to `FragmentDefinitions` and replace these `inlineFragments` with `FragmentSpread`s. This transformation aims to enhance the efficiency of operation parsing, validation and execution, especially for large and complex queries.

GraphQL operations can become quite large and complex, leading to increased parsing time and execution latency. This is particularly evident with extensive use of InlineFragments, which, while flexible and powerful, can add significant overhead to query processing. By extracting these inline fragments into separate FragmentDefinitions and referencing them with FragmentSpreads, we can significantly reduce the complexity and size of the initial query sent to the server.

The optimization would consist of:

1. Traverse the AST.
2. Search for InlineFragment nodes in SelectionSet nodes.
3. Create a FragmentDefinition of the SelectionSet of that InlineFragments.
4. Append that FragmentDefinition to the document AST.
5. replace the InlineFragment in the SelectionSet with a FragmentSpread.

This optimization is scoped for operations that heavily rely in polymorphism. It doesn't need a graphQL schema at all, only needs the AST.

I have a test suite pending to add, (already have the tests), but we (ExpediaGroup) tried this in production giving us a 100x smaller operations.

This PR was originally raised [against Apollo Federation](https://github.com/apollographql/federation/pull/2893/files), however, this utility function applies to ASTs and to any GraphQL implementation specification, so graphql-js sound like the best place to have this.

### DIFF
--- a/src/utilities/fragmentifyDocument.ts
+++ b/src/utilities/fragmentifyDocument.ts
@@ -1,0 +1,173 @@
+import type { ObjMap } from '../jsutils/ObjMap.js';
+
+import type {
+  DefinitionNode,
+  DocumentNode,
+  FragmentDefinitionNode,
+  FragmentSpreadNode,
+  InlineFragmentNode,
+  SelectionSetNode
+} from '../language/ast.js';
+import { Kind } from '../language/kinds.js';
+import { visit } from '../language/visitor.js';
+
+interface ReferencedFragmentDefinitionNode {
+  totalReferences: number;
+  fragment: FragmentDefinitionNode;
+}
+
+interface InlineFragmentToFragmentSpreadResult {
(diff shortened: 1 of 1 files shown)


yaacovCR:
@samuelAndalon this looks extremely useful! I think it might fit very well into [graphql-tools](https://github.com/ardatan/graphql-tools) as that is more of a general toolbox, but it could make sense within the reference implementation if the WG ever formalizes something in the spec with regard to canonical documents of some kind.

What do you think?

PR #4002, opened 2023-12-22
allow nested defers at the same level

yaacovCR:
Implements Option A from https://github.com/graphql/defer-stream-wg/discussions/80

PR #4026, opened 2024-03-10
rewrite of incremental delivery

yaacovCR:
following https://github.com/graphql/graphql-spec/pull/1077

requires optimization, diff minimization, etc.

but hopefully will help advance discussions around these spec edits


yaacovCR:
depends on #4002

yaacovCR:

yaacovCR:
@github-actions run-benchmar

yaacovCR:
@github-actions run-benchmark

yaacovCR:

yaacovCR:

yaacovCR:

yaacovCR:

yaacovCR:

yaacovCR:
image

take-two = this PR
remove-filtering-again = #4032

yaacovCR:
we got these performance improvements in the rewrite by:
creating separate functions for completing iterables with and without stream, such that if there is no stream directive, we do not have to check for each item whether we have hit the index at which we should start streaming
inline completePromisedValue => this was an async function which consolidated some work and previously seemed to give us a speedup, now the extra function overhead seemed to be giving us a slowedown => these changes seem to signal the perils of microbenchmarking, but since I was the one who introduced this function during the 17-alpha process for the purpose of the speed-up, I think it's probably better to just take it out.

yaacovCR:
lastest updates:

1. We have removed filtering within this rewrite to better match the spec. #4043 can be looked at if there is interest in what bringing back filtering would look like.
2. In that context, we have reworked this PR to keep completePromisedValue and also introduce a special completePromisedListItemValue. It seems to still help, so we are keeping that approach for now.

yaacovCR:
Latest results:

image

We were able to even improve synchronous list performance, but not asynchronous. Deopt Explorer doesn't show anything being deoptimized. The hit seems to be that we need an extra tick after our Promise.all call to consolidate all of the individual list item `GraphQLResult` into a single `GraphQLResult` for the list.

yaacovCR:
I have extracted #4050 #4051 #4052 #4053 from this PR to aid review.

yaacovCR:
image

src/execution/IncrementalPublisher.ts
```diff
@@ -78,7 +74,10 @@ export interface FormattedInitialIncrementalExecutionResult<
 export interface SubsequentIncrementalExecutionResult<
   TData = unknown,
   TExtensions = ObjMap<unknown>,
-> extends Partial<IncrementalUpdate<TData, TExtensions>> {
+> {
+  pending?: ReadonlyArray<PendingResult>;
+  incremental?: ReadonlyArray<IncrementalResult<TData, TExtensions>>;
+  completed?: ReadonlyArray<CompletedResult>;
```
saihaj, line 80: I think we should keep these required args? better to be explicit imo.


```suggestion
  pending: ReadonlyArray<PendingResult>;
  incremental: ReadonlyArray<IncrementalResult<TData, TExtensions>>;
  completed: ReadonlyArray<CompletedResult>;
```

yaacovCR: This is a bit tricky to follow, but I think it's correct. `SubsequentIncrementalExecutionResult` is part of the new return type of `execute`, and so these keys are optional => if there isn't anything newly pending, we do not send a pending key at all rather than sending a key with an empty array.

Previously, we were using within our `IncrementalPublisher` class an intermediate result type we called `IncrementalUpdate` to collect these entries, and then only sending them if they were populated.

So we had `SubsequentIncrementalExecutionResult` extend `Partial<IncrementalUpdate>`. That itself was confusing, and so we got rid of `IncrementalUpdate` entirely and just put `_incremental` and `_completed` on the `IncrementalPublisher` class instance itself.

saihaj (approved):
excited to land this the bonus perf improvements is always great!

src/execution/execute.ts
```diff
@@ -163,6 +165,8 @@ export interface StreamUsage {
   fieldGroup: FieldGroup;
 }
 
+type GraphQLResult<T> = [T, ReadonlyArray<IncrementalDataRecord>];
```
robrichard, line 168: Maybe call this `GraphQLWrappedResult`, and use `wrappedResult` instead of `acc`?

src/execution/execute.ts
```diff
@@ -561,36 +604,44 @@ function executeFields(
...
-        incrementalDataRecord,
+        errors,
         deferMap,
       );
 
       if (result !== undefined) {
-        results[responseName] = result;
         if (isPromise(result)) {
+          results[responseName] = result.then((resolved) => {
+            appendNewIncrementalDataRecords(acc, resolved[1]);
```
robrichard, line 614: is it faster to push each new incrementalDataRecord onto `acc[1]` here instead of using `appendNewIncrementalDataRecords`, which creates a new array each time?

yaacovCR: Sounds good! 74d092ef1

src/execution/execute.ts
yaacovCR: done! 24136afec

yaacovCR:
reposting benchmarks considering there were a few changes

image

alpha-1 is prior to incremental delivery, current-main is basically alpha-3, and we see that the rewrite improves on it significantly in the synchronous list case. the next set of PRs should hopefully improve further. we can then judge whether we should/want to bring back a form of filtering.

robrichard (approved):

PR #4034, opened 2024-03-27
oneof: add test case

yaacovCR:
this test currently passes. should it?

see:

https://github.com/graphql/graphql-spec/pull/825/#discussion_r588879692
https://github.com/graphql/graphql-spec/pull/825/#discussion_r1540939491

### DIFF
--- a/src/execution/__tests__/oneof-test.ts
+++ b/src/execution/__tests__/oneof-test.ts
@@ -134,6 +134,30 @@ describe('Execute: Handles OneOf Input Objects', () => {
       });
     });
 
+    it('accepts multiple variable key as long as only one has a value', () => {
+      // NOTE: This is an *invalid* query, but it should be an *executable* query.
+      const query = `
+        query ($a: String) {
+          test(input: { a: $a, b: $b }) {
+            a
+            b
+          }
+        }
+      `;
+      const result = executeQuery(query, rootValue, {
+        a: 'abc',
+      });
+
+      expectJSON(result).toDeepEqual({
+        data: {
+          test: {
+            a: 'abc',
+            b: null,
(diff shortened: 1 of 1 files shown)


saihaj (commented):
I think this should throw an error 🤔 based on [All Variable Uses Defined](https://spec.graphql.org/October2021/#sec-All-Variable-Uses-Defined)

benjie:
The question is kind of moot because the query is meaningless having a variable that's not defined; but yes this should fail the "OneOf Input Objects Have Exactly One Field" validation rule:

>     - Let {fields} be the fields provided by {oneofInputObject}.
>     - {fields} must contain exactly one entry.

If you define the variable `$b`, does the issue go away?

yaacovCR:
Have to check back on this

yaacovCR:
Sorry @saihaj  @benjie => I was out of context on this for a while. I updated the operation to include the variable so it doesn't run afoul of "All Variables Used Defined", and the query fails OneOf validation appropriately whether or not the operation variable is included. But it passes execution without a problem, and the question I am asking is if it should throw an error, i.e. whether this invalid query should still execute or not.

I found [this discussion](https://github.com/graphql/graphql-spec/pull/825/#discussion_r588879692), where there seems to have been a request for feedback on this exact issue, without a clear resolution I could find. What the implementation does is let it slide in execution, but from the discussion there it seems like we would have to update the coercion table if that's the choice we make. Not the most live issue, as this is with regards to an invalid query, can certainly wait until after GraphQLConf, hope everyone will enjoy!

benjie:
@yaacovCR I believe this is replaced by your newer PRs that mark this as invalid?

PR #4043, opened 2024-04-05
incremental: use filtering instead of GraphQLResult

yaacovCR:
now that filtering has been removed from #4026, this draft PR could be checked out to see what it would look like to reintroduce it.


yaacovCR:
image

use-undefined-for-empty corresponds to #4046

the above stats show that with filtering, we have significant performance boost for synchronous lists, and we have maintained parity even for asynchronous lists

yaacovCR:
Closing this stale PR.

PR #4046, opened 2024-04-05
perf: use undefined for empty

yaacovCR:
As convention, we will lazily instantiate arrays/sets when adding the first item.

This applies to arrays/sets on execution/incremental context, as well as the second member of the GraphQLWrappedResult tuple holding the array of incremental data records.


yaacovCR:
image

robrichard (approved):

PR #4053, opened 2024-04-07
perf: optimize completion loops

yaacovCR:
by hoisting streamUsage checks out of the loop, requires introducing two versions of the looping functions

depends on #4052

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -1091,10 +1091,97 @@ async function completeAsyncIteratorValue(
     undefined,
   ];
   let index = 0;
-  const streamUsage = getStreamUsage(exeContext, fieldGroup, path);
   // eslint-disable-next-line no-constant-condition
   while (true) {
-    if (streamUsage && index >= streamUsage.initialCount) {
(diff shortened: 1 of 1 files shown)


yaacovCR:
image

saihaj (approved):

yaacovCR:
After offline discussion with @robrichard i am closing this one for now as not worth the extra slight performance bump

by our very synthetic benchmarks, in the very long synchronous list case we’re already faster than the first 17 alpha release and in the asynchronous case we’re still slower either way secondary to incremental delivery, so the small difference from this PR doesn’t really change our performance story overall

in the short term, but not immediate future we can think about #4043 as the real path away from the performance penalty involved with incremental delivery

although the current plan is not to proceed with that immediately as during the pre-and presumably immediate post release. After incremental delivery, there is significant value in having our reference and limitation be more aligned with the specification changes.

#4043 add an optimization that diverges somewhat from the specification, and so we may adopt it within the reference implementation, but the thought is at least not to do so immediately

PR #4078, opened 2024-05-02
Add `@oneOf` support to introspection query

### LINKED ISSUE #4077: Introspection queries don't support `@oneOf`
Related to the issue in https://github.com/graphql/graphql-js/issues/3968, it doesn't appear that the introspection query is selecting the `isOneOf` field from the server, as defined by the [RFC](https://github.com/graphql/graphql-spec/pull/825/files#diff-30a69c5a5eded8e1aea52e53dad1181e6ec8f549ca2c50570b035153e2de1c43).

This has downstream effects like preventing this introspection response from powering codegen that uses that flag to enforce mutual exclusivity amongst the arguments.

I've attempted to add this functionality in https://github.com/graphql/graphql-js/pull/4078

maciesielka:
Fixes #4077

### DIFF
--- a/src/utilities/getIntrospectionQuery.ts
+++ b/src/utilities/getIntrospectionQuery.ts
@@ -32,6 +32,10 @@ export interface IntrospectionOptions {
    * Default: false
    */
   inputValueDeprecation?: boolean;
+
+  // Whether target GraphQL server supports `@oneOf` input objects.
+  // Default: false
+  inputObjectOneOf?: boolean;
 }
 
 /**
@@ -45,6 +49,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
     directiveIsRepeatable: false,
     schemaDescription: false,
     inputValueDeprecation: false,
+    inputObjectOneOf: false,
     ...options,
   };
 
@@ -62,6 +67,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
   function inputDeprecation(str: string) {
     return optionsWithDefault.inputValueDeprecation ? str : '';
   }
+  const inputObjectOneOf = optionsWithDefault.inputObjectOneOf ? 'isOneOf' : '';
 
   return `
     query IntrospectionQuery {
@@ -90,6 +96,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
       name
       ${descriptions}
       ${specifiedByUrl}
+      ${inputObjectOneOf}
       fields(includeDeprecated: true) {
         name
         ${descriptions}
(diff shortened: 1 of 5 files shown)


src/utilities/getIntrospectionQuery.ts
```diff
@@ -32,6 +32,10 @@ export interface IntrospectionOptions {
    * Default: false
    */
   inputValueDeprecation?: boolean;
+
+  // Whether target GraphQL server supports `@oneOf` input objects.
+  // Default: false
+  inputObjectOneOf?: boolean;
```
maciesielka, line 38: Followed the pattern here above for making this configurable, but wasn't sure if that was necessary.

JoviDeCroock (approved):

src/utilities/getIntrospectionQuery.ts
```diff
@@ -32,6 +32,10 @@ export interface IntrospectionOptions {
    * Default: false
    */
   inputValueDeprecation?: boolean;
+
+  // Whether target GraphQL server supports `@oneOf` input objects.
+  // Default: false
```
mmgoodnow, line 37: should this be JSDoc formatted to match the comments above?

src/utilities/getIntrospectionQuery.ts
```diff
@@ -223,6 +230,7 @@ export interface IntrospectionObjectType {
   readonly interfaces: ReadonlyArray<
     IntrospectionNamedTypeRef<IntrospectionInterfaceType>
   >;
+  readonly isOneOf: boolean;
```
mmgoodnow, line 233: looks like this was added to the `IntrospectionObjectType` here in addition to the `IntrospectionInputObjectType` below—is that intentional? The other references in this diff seem to only care about input objects.

src/utilities/getIntrospectionQuery.ts
maciesielka: 🤦🏻 definitely. updated in [3af4434](https://github.com/graphql/graphql-js/pull/4078/commits/3af443467cdb9713f700e2b7b4368fb55c18171c)

src/utilities/getIntrospectionQuery.ts
maciesielka: You're so right. Removed in [633331a](https://github.com/graphql/graphql-js/pull/4078/commits/633331aded549b3db00296646ad4833b5f8b5c7f)

maciesielka:
@JoviDeCroock, @mmgoodnow appreciate your reviews and hopefully I've addressed your feedback. Any idea what it takes to get something like this merged?

benjie (approved):
Thanks for adding this vital feature!

Issue #4075, opened 2024-04-30
`process.env`, `globalThis`, and `typeof process`


phryneas:
This is a writeup to discuss this problem in the GraphQL-WG and the GraphQL-JS-WG. Please give feedback in either of those meetings, or here in the issue, until the next GraphQL-JS-WG meeting, where we want to make a final call on this issue.

Last year, #3887 was released which changed the dev check in `src/jsutils/instanceOf.ts` from
```js
process.env.NODE_ENV === 'production'
```
 to 
```js
globalThis.process?.env.NODE_ENV === 'production'
```
shortly followed by #3923, which changed it to 
```js
globalThis.process && globalThis.process.env.NODE_ENV === 'production'
```
as some bundlers were choking on the optional chaining syntax.

Since then, various issues and PRs have been opened to change into various other forms.

I'll try to give an overview over problems, potential solutions and their shortcomings here so we can decide on a way forward.

### 1. accessing `process.env.NODE_ENV` directly

There is a bunch of environments (e.g. ESM in the browser), where accessing `process.env` directly just crashes, since `process` is not a variable.

### 2. accessing `globalThis.process.env.NODE_ENV`: bundler replacement

Some bundlers do a string replacement of `process.env.NODE_ENV`, which lead to the code being replaced by invalid JavaScript like `globalThis."production"`.
(Afaik, most of these have been reported in the upstream bundlers at this point in time and they have fixed their regular expressions)

### 3. accessing `process.env.NODE_ENV` or `globalThis.process.env.NODE_ENV` while testing for `process`, without checking if the `env` property is set: DOM elements with the id `process`

If a DOM element with the `id` `process` exists (e.g. `<span id="process">...</span>`), this element will be registered as the global variable `process`, so it will be accessible as `process` and `globalThis.process` - but not have a `.env` property, so accessing `process.env.FOO` will crash.

### 4. testing for `process` to be present with `typeof process`: cannot be tree-shaken

Some bundlers (e.g. esbuild) can only replace statements like `foo.bar.baz`, but not statements like `typeof foo` as they rely on AST-level replacement and just don't have support for that kind of replacement.

The do not plan to add this to ESBuild: https://github.com/evanw/esbuild/issues/1954#issuecomment-1023315210

### A) `process.env.NODE_ENV === 'production'`

This would be a rollback to the original code. It works fine with bundlers, but has problem 1.

### B) `globalThis.process && globalThis.process.env.NODE_ENV === 'production'`

[...]

PR #4079, opened 2024-05-03
use try-catch to simplify instanceOf checks

yaacovCR:
for platforms where `process.env` is not globally accessible and defined, just return true

bundlers should have an easy time just replacing `process.env.NODE_ENV` without any qualifiers


yaacovCR:
Closing, this does not seem so tree-shakable, see #4075.

PR #4080, opened 2024-05-06
chore(performance): optimize regressed for..of and array destructuring performance

JoviDeCroock:
One of the issues we uncovered in https://github.com/graphql/graphql-js/issues/3648 is that upgrading to ES6 output maintained the `for..of` and array destructuring in our bundles which leads to a pretty considerable performance regression. This changes our build step to account for these less performant syntaxes and transforms them.

There is more low hanging fruit that can be optimised like the custom accumulatormap/... however starting with a quick win. The babel plugin also still leverages the iterator spec for non-array for loops, [it optimises what it can](https://babeljs.io/docs/babel-plugin-transform-for-of#optimization)

Fixes https://github.com/graphql/graphql-js/issues/3648

The performance implication would be higher if we did something akin to https://github.com/graphql/graphql-js/pull/3687

### DIFF
--- a/.babelrc-deno.json
+++ b/.babelrc-deno.json
@@ -1,6 +1,8 @@
 {
   "plugins": [
     "@babel/plugin-syntax-typescript",
+    "@babel/plugin-transform-for-of",
+    "@babel/plugin-transform-destructuring",
     ["./resources/add-extension-to-import-paths", { "extension": "ts" }],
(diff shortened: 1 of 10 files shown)


JoviDeCroock:
Looks like the performance improvements as seen in https://github.com/graphql/graphql-js/pull/3687 aren't reproducible with the babel plugin

PR #3691, opened 2022-08-11
build/npm: add TS transformation that optimise for-of on arrays

IvanGoncharov:
Context: I wanted to try more automatic alternative to #3687 that
allows us to continue using ES6 features without perfomance lost.

PR #4084, opened 2024-05-08
chore(perf): improve performance of iterator usage

JoviDeCroock:
I realised why I didn't see performance differences in https://github.com/graphql/graphql-js/pull/4080 I was targeting the wrong branch 😅 This takes the transform used in https://github.com/graphql/graphql-js/pull/3691 and updates it to recent TS versions and adds one for Array destructuring.

The transform that takes on `for..of` will only optimize usage that isn't relying on `iterator` i.e. plain arrays vs `Map`.

I have been seeing issues running the benchmarks, looking into that atm.

### DIFF
--- a/resources/build-npm.ts
+++ b/resources/build-npm.ts
@@ -12,6 +12,8 @@ import {
   showDirStats,
   writeGeneratedFile,
 } from './utils.js';
+import { optimizeArrayDestructuring } from './optimize-array-destructuring.js';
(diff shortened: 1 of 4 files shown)


JoviDeCroock:
After getting the benchmarks to run, this seems resolved 😅 sorry for the PR's

yaacovCR:
> sorry for the PR's

Even if this didn’t pan out, it’s super useful that you investigated it. Thank you!

PR #4086, opened 2024-05-08
Reproduction of #4085

ardatan:
Related https://github.com/graphql/graphql-js/issues/4085

PR #4087, opened 2024-05-08
Introduce astFromValueUntyped and handle object values within astFromValue correctly

### LINKED ISSUE #4085: astFromValue fails with a custom scalar serializing to an object value
When you have a scalar with a `serialize` function that returns an object, it is not possible to convert it to AST. 
So it doesn't allow you to print a schema that has an input value with an object as a default value.

Reproduction -> https://github.com/graphql/graphql-js/pull/4086
Proposed fix -> https://github.com/graphql/graphql-js/pull/4087

```ts
const JSONScalar = new GraphQLScalarType({
  name: "JSON",
  serialize(value) {
    return value;
  },
});

  types: [JSONScalar],
    name: "Query",
    fields: {
      test: {
        type: GraphQLString,
        args: {
          i: {
            type: JSONScalar,
            defaultValue: { object: true },
          },
        },
      },
    },
  }),
});
```

ardatan:
Fixes https://github.com/graphql/graphql-js/issues/4085
Closes https://github.com/graphql/graphql-js/pull/4086

This PR introduces a new function called `astFromValueUntyped` that creates an AST without a type.
This function is needed because if a custom scalar returns an object type, it needs to be serialized without a type.

### DIFF
--- a/src/utilities/__tests__/astFromValue-test.ts
+++ b/src/utilities/__tests__/astFromValue-test.ts
@@ -233,6 +233,24 @@ describe('astFromValue', () => {
     expect(() => astFromValue('value', returnCustomClassScalar)).to.throw(
       'Cannot convert value to AST: {}.',
     );
+
+    const returnObjectScalar = new GraphQLScalarType({
+      name: 'ReturnObjectScalar',
+      serialize() {
+        return { some: 'data' };
+      },
+    });
+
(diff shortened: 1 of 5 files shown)


yaacovCR:
I think this should be closed within v17 by #3812 and #4209, so I am going to close this old PR in the name of tidying up, feel free to reopen.

See also: https://github.com/graphql/graphql-js/issues/4085#issuecomment-2385360079

PR #3814, opened 2023-01-02
RFC: Default value validation & coercion

yaacovCR:
[#3049 rebased on main](https://github.com/graphql/graphql-js/pull/3049).

This is the last rebased PR from the original PR stack concluding with #3049.

* Rebased: #3809 [Original: #3092]
* Rebased: #3810 [Original: #3074]
* Rebased: #3811 [Original: #3077]
* Rebased: #3812 [Original: #3065]
* Rebased: #3813 [Original: #3086]
* Rebased: #3814 (this PR) [Original: #3049]

Update: #3044 and #3145 have been separated from this stack.

Changes from original PR:
1. `astFromValue()` is deprecated instead of being removed.

@leebyron comments from #3049, the original PR:
> Implements [graphql/graphql-spec#793](https://github.com/graphql/graphql-spec/pull/793)
> 
> * BREAKING: Changes default values from being represented as an assumed-coerced "internal input value" to a pre-coerced "external input value" (See chart below).
>   This allows programmatically provided default values to be represented in the same domain as values sent to the service via variable values, and allows it to have well defined methods for both transforming into a printed GraphQL literal string for introspection / schema printing (via `valueToLiteral()`) or coercing into an "internal input value" for use at runtime (via `coerceInputValue()`)
>   To support this change in value type, this PR adds two related behavioral changes:
>   
>   * Adds coercion of default values from external to internal at runtime (within `coerceInputValue()`)
>   * Removes `astFromValue()`, replacing it with `valueToLiteral()` for use in introspection and schema printing. `astFromValue()` performed unsafe "uncoercion" to convert an "Internal input value" directly to a "GraphQL Literal AST", where `valueToLiteral()` performs a well defined transform from "External input value" to "GraphQL Literal AST".
> * Adds validation of default values during schema validation.
>   Since assumed-coerced "internal" values may not pass "external" validation (for example, Enum values), an additional validation error has been included which attempts to detect this case and report a strategy for resolution.
> 
> Here's a broad overview of the intended end state:
> 
> GraphQL Value Flow

PR #4088, opened 2024-05-08
astFromValue - JavaScript BigInt support

ardatan:
This PR adds support for [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) values to `astFromValue`.
It coerces whenever possible instead of failing.
This will also allow custom scalars like https://the-guild.dev/graphql/scalars/docs/scalars/big-int to get a `BigInt` as a default value. Similar to #4085

### DIFF
--- a/src/utilities/__tests__/astFromValue-test.ts
+++ b/src/utilities/__tests__/astFromValue-test.ts
@@ -41,9 +41,14 @@ describe('astFromValue', () => {
       value: false,
     });
 
-    expect(astFromValue(1, GraphQLBoolean)).to.deep.equal({
+    expect(astFromValue(1n, GraphQLBoolean)).to.deep.equal({
       kind: 'BooleanValue',
       value: true,
+    });
+
+    expect(astFromValue(0n, GraphQLBoolean)).to.deep.equal({
+      kind: 'BooleanValue',
+      value: false,
     });
 
     const NonNullBoolean = new GraphQLNonNull(GraphQLBoolean);
@@ -67,6 +72,11 @@ describe('astFromValue', () => {
     expect(astFromValue(1e4, GraphQLInt)).to.deep.equal({
       kind: 'IntValue',
       value: '10000',
+    });
+
+    expect(astFromValue(1n, GraphQLInt)).to.deep.equal({
+      kind: 'IntValue',
+      value: '1',
     });
 
     // GraphQL spec does not allow coercing non-integer values to Int to avoid
@@ -94,6 +104,11 @@ describe('astFromValue', () => {
     expect(astFromValue(123.0, GraphQLFloat)).to.deep.equal({
       kind: 'IntValue',
       value: '123',
+    });
+
+    expect(astFromValue(9007199254740993n, GraphQLFloat)).to.deep.equal({
+      kind: 'IntValue',
+      value: '9007199254740993',
     });
 
     expect(astFromValue(123.5, GraphQLFloat)).to.deep.equal({
@@ -131,6 +146,11 @@ describe('astFromValue', () => {
     expect(astFromValue(123, GraphQLString)).to.deep.equal({
       kind: 'StringValue',
       value: '123',
+    });
+
+    expect(astFromValue(9007199254740993n, GraphQLString)).to.deep.equal({
+      kind: 'StringValue',
+      value: '9007199254740993',
     });
 
     expect(astFromValue(false, GraphQLString)).to.deep.equal({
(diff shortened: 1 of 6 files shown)


ardatan:
@yaacovCR Also I am curious what do you think about this?

saihaj (commented):
not sure if we want to merge this 🤔 

[_Numeric integer values larger than 32-bit should either use String or a custom-defined Scalar type, as not all platforms and transports support encoding integer numbers larger than 32-bit._](https://spec.graphql.org/October2021/#note-0770a)

ardatan:
@saihaj Isn't it for `Int` scalar type? However, this PR doesn't change that behavior. `Int` still doesn't support larger integers. This PR just allows to use `bigint` type of JavaScript just like `number` and `string`. So it is not the size of the number but the type of it.

saihaj:
> @saihaj Isn't it for `Int` scalar type? However, this PR doesn't change that behavior. `Int` still doesn't support larger integers. This PR just allows to use `bigint` type of JavaScript just like `number` and `string`. So it is not the size of the number but the type of it. If `Int` supports 32-bit numeric values in string format, it should allow the same values in `bigint` format as well right? Same for `ID` and `Float but without 32-bit limit.

Oh I see what you mean, its like we are coercing `2147483646n` as `2147483646` since it is a valid GraphQL Int after all

src/utilities/__tests__/astFromValue-test.ts
```diff
@@ -69,6 +74,11 @@ describe('astFromValue', () => {
       value: '10000',
     });
 
+    expect(astFromValue(1n, GraphQLInt)).to.deep.equal({
```
saihaj, line 77: we should add a test case for a throw where we give a `bigint` value and it isn't a legal `GraphQLInt`

ardatan: Yes you're right! I added more! What do you think?

saihaj (approved):

yaacovCR:
In terms of default values, that use case should probably be re-examined after #3814 lands, which should hopefully be soon! We are halfway-through the stack.

I suppose we will need to support conversion from BigInt to Int/Float when possible only if the received JSON is parsed with a BigInt supporting parser, so this could be technically necessary otherwise. I think it probably would require some discussion at the WG in terms of whether to include it, but I would be in favor.

yaacovCR:
In terms of output, I am of two minds in terms of whether we should be supporting conversion from bigint to int, because that means that a custom scalar probably should have been used, in case the value was greater than the int threshold. Maybe it's better to fail overall?

For conversion to string or id, I imagine we would want to support it, I can't see why not?

See #3973 and #3109.

JoviDeCroock (approved):

src/utilities/__tests__/astFromValue-test.ts
```diff
@@ -96,6 +121,11 @@ describe('astFromValue', () => {
       value: '123',
     });
 
+    expect(astFromValue(9007199254740993n, GraphQLFloat)).to.deep.equal({
```
yaacovCR, line 124: The change in type of this test is because astFromValue is not safe, set to be replaced by `valueToLiteral()` for all remaining uses in https://github.com/graphql/graphql-js/pull/3814

But a bigint of sufficient size, -- as far as I know -- cannot be converted to a JS float.

We also in this PR do not properly handle GraphQLFloat scalar parsing and serialization of small BigInts, I will push a failing test to demonstrate what I mean.

We also probably should update the logic within the new `valueToLiteral()` utility. It's possible that overall we can just delay landing this PR until after #3814

yaacovCR:
Thank you @ardatan so much for your work on this! `bigint` support belatedly made it into v17 as reworked in https://github.com/graphql/graphql-js/pull/4550

ardatan:
Thanks @yaacovCR ! But like we discussed here earlier -> https://github.com/graphql/graphql-js/issues/4085
v17 seems far from being released any time soon. So I think a patch-fix in v16 would be better to have.

yaacovCR:
> Thanks @yaacovCR ! But like we discussed here earlier -> #4085 v17 seems far from being released any time soon. So I think a patch-fix in v16 would be better to have.

I'm hopeful to (finally!) move v17 to beta over the weekend.

Reopening to help track, however, as I think now that we have this merged into v17 and we know directionally where the code is headed with the new `valueToLiteral()` functionality, etc, perhaps we can backport more of the new changes to v16.

Ideally this would be a non-change for anyone not using the new functionality.

PR #4090, opened 2024-05-13
incremental: add highWaterMark option to apply backpressure with async streams

yaacovCR:
This protects against a potential OOM error if we end up pulling data from a stream faster than it is consumed.

Default is set at 100; after pulling 100 entries, we will pause until some have been flushed.


yaacovCR:
Closing this in favor of #4097 … we can revisit later if we need something more complicated.

PR #4091, opened 2024-05-15
incremental: remove extra ticks

yaacovCR:
consolidates some defer payloads

This is accomplished by replacing a promised result within a `DeferredGroupedFieldSetRecord` with the actual result as soon as possible, so that IncrementalPublisher doesn't wait an extra tick if it doesn't need to.

This same optimization can also be employed with `StreamItemsRecord`s, but it is possible that the stream logic should be overall reworked.


yaacovCR:
This has been reworked and is now part of #4094

PR #4092, opened 2024-05-16
incremental: always only emit single completion error for deferred fragment

yaacovCR:
When multiple deferred grouped field sets within a fragment error causing the fragment to complete in error, only a single completion error should be emitted.


yaacovCR:
Fix for this, and the corresponding test has been integrated into #4094

PR #4094, opened 2024-05-22
incremental: introduce IncrementalGraph class to manage tree of subsequent results

yaacovCR:
A Graph object exposes an iterator for the completed incremental data such that the Incremental Publisher can properly be conceptualized as an operator on a stream, similar to the spec changes.

depends on #4099

### DIFF
--- a/src/execution/IncrementalGraph.ts
+++ b/src/execution/IncrementalGraph.ts
@@ -74,9 +74,8 @@ export class IncrementalGraph {
   }
 
   getNewPending(): ReadonlyArray<SubsequentResultRecord> {
-    const maybeEmptyNewPending = this._newPending;
     const newPending = [];
-    for (const node of maybeEmptyNewPending) {
+    for (const node of this._newPending) {
       if (isDeferredFragmentRecord(node)) {
         if (node.expectedReconcilableResults) {
           this._pending.add(node);
@@ -84,7 +83,7 @@ export class IncrementalGraph {
           continue;
         }
         for (const child of node.children) {
-          this._addNonEmptyNewPending(child, newPending);
+          this._newPending.add(child);
         }
       } else {
         this._pending.add(node);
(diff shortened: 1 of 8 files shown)


src/jsutils/Future.ts
```diff
@@ -0,0 +1,20 @@
+import { invariant } from './invariant.js';
+import { isPromise } from './isPromise.js';
+import type { PromiseOrValue } from './PromiseOrValue.js';
+
+/** @internal */
+export class Future<T> {
```
robrichard, line 6: Can you explain how this class is used?

yaacovCR: Sure =>

I just pushed a version of this PR with many commits to show how when I extracted the incremental graph from the incremental publisher -- and made it mutation-less -- the end of the process (i.e., the second to last commit) introduced some extra ticks.

The Future class get rid of those extra ticks, as well as a whole bunch of others. Basically, `Future<T>` can be used wherever a `PromiseOrValue<T>` can be used => but a `Future<T>` will automatically update its internal value when the promise resolves, so that if you are doing some logic that is `if (isPromise(x)) { /* do A */ } else { /* do B */ }`, using a Future, you don't have to wait an extra tick if the promise has previously been resolved.

You can see now that almost all of the individual commits don't change tests, but now you can see which ones do, and hopefully that makes clearer why I introduced the `Future` class and what it's doing.

yaacovCR: Renamed `Future` to `BoxedPromiseOrValue`

src/execution/IncrementalGraph.ts
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  removeDeferredFragment(
+    deferredFragmentRecord: DeferredFragmentRecord,
+  ): boolean {
+    const deferredFragmentNode = this._deferredFragmentNodes.get(
+      deferredFragmentRecord,
+    );
+    // TODO: add test case?
+    /* c8 ignore next 3 */
```
yaacovCR, line 208: these lines can be reviewed, a test case was added

yaacovCR: these lines can be removed, a test case was added

src/execution/IncrementalGraph.ts
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  completeDeferredFragment(
+    deferredFragmentRecord: DeferredFragmentRecord,
+  ): Array<ReconcilableDeferredGroupedFieldSetResult> | undefined {
+    const deferredFragmentNode = this._deferredFragmentNodes.get(
+      deferredFragmentRecord,
+    );
+    // TODO: add test case?
+    /* c8 ignore next 3 */
```
yaacovCR, line 170: these lines can be removed, a test case was previously added

yaacovCR: 09ae45a2

yaacovCR: wrong spot, deleting comments

src/execution/IncrementalGraph.ts
yaacovCR: 09ae45a2

robrichard (approved):

PR #4100, opened 2024-06-02
require non-empty directive locations

jbellenger:
A [proposed spec edit](https://github.com/graphql/graphql-spec/pull/1089) clarifies a requirement that was always true but was slightly buried: _directive definitions must include 1 or more locations_.

This PR adds explicit validation to graphql-js around non-empty directive locations. [Similar work](https://github.com/graphql-java/graphql-java/pull/3552) was landed in graphql-java


JoviDeCroock (approved):

benjie (approved):
Looks good to me; thanks!

PR #4107, opened 2024-06-14
polish(incremental): remove id mutation

yaacovCR:
depends on #4106


robrichard (approved):

yaacovCR:
As discussed with @robrichard offline, will drop this in favor of assigning ids sequentially during execution, even though if early execution is enabled, this means that ids might be delivered out of order  or even never sent (when results are filtered). Will submit that in a separate PR.

PR #4108, opened 2024-06-16
polish(incremental): refactor getNewPending functionality

yaacovCR:
return newPending alongside incremental results when completing a fragment

= removes need to track `newPending` and `newIncrementalDataRecords` as part of the class
= unifies handling of retrieving new pending for initial result, completed fragments, and stream items incremental entries

### DIFF
--- a/src/execution/IncrementalGraph.ts
+++ b/src/execution/IncrementalGraph.ts
@@ -145,7 +145,7 @@ export class IncrementalGraph {
     const reconcilableResults = Array.from(
       deferredFragmentNode.reconcilableResults,
     );
-    this._removePending(deferredFragmentNode);
+    this._removeRootNode(deferredFragmentNode);
     for (const reconcilableResult of reconcilableResults) {
       for (const otherDeferredFragmentNode of this._fragmentsToNodes(
         reconcilableResult.deferredGroupedFieldSetRecord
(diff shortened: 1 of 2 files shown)


src/execution/IncrementalGraph.ts
robrichard, line 1: Are these function names still accurate now that you renamed `_pending` to `_rootNodes`?
getNewPending
_removePending
_hasPendingFragment

yaacovCR: In 163e96c8 -- I changed:

`_removePending(...)` => `_removeRootNode(...)`

and

`_hasPendingFragment(...)` => `_completesRootNode(...)`

I think `getNewPending()` as a public method still makes sense,  `rootNodes` is an implementation detail on `IncrementalGraph`; the `IncrementalPublisher` class that calls the `getNewPending()` method knows nothing of `rootNodes`, and so it makes sense that it doesn't refer to it by name.

Let me know what you think!

yaacovCR: Actually, I reconsidered and changed `getNewPending()` as well to `getNewRootNodes()` and changed `newPending` to `newRootNodes`. After all, the class is called IncrementalGraph, so the caller knows that we are managing things with a graph, so of course there are root nodes. The caller will assume that the root nodes  are the pending and any new root nodes are able to be convering to new pending result entries for the pending array. This way, we get consistency of nomenclature, and don't overload the use of the term pending as much.

fc849b74

robrichard (approved):

PR #4109, opened 2024-06-17
fix(incremental): properly initiate nested deferred grouped field sets

yaacovCR:
when early execution is disabled, the shouldDefer logic has a subtle bug

see: https://github.com/graphql/defer-stream-wg/discussions/84#discussioncomment-8280294

depends on #4112


yaacovCR:
Closing in favor of https://github.com/graphql/graphql-js/pull/4140

Issue #3938, opened 2023-07-17
RangeError: Maximum call stack size exceeded in OverlappingFieldsCanBeMergedRule


MarkKoester:
I noticed during some testing that sending queries with 10000 fragments causes a
`RangeError: Maximum call stack size exceeded in OverlappingFieldsCanBeMergedRule` error to be thrown from `OverlappingFieldsCanBeMergedRule`. I was wondering if this is the intended behavior and if not, how it should be handled.

This was observed on version 16.6.0 and 16.7.2

Below is the stack trace of the error and the steps to reproduce
### Error
```
RangeError: Maximum call stack size exceeded
      at getReferencedFieldsAndFragmentNames (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:678:45)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:211:5)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:250:5)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:250:5)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:250:5)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:250:5)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:250:5)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:250:5)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:250:5)
      at collectConflictsBetweenFieldsAndFragment (/graphql/validation/rules/OverlappingFieldsCanBeMergedRule.js:250:5)
```

### Reproduction steps
```ts
import {
  parse,
  buildSchema,
  validate,
  OverlappingFieldsCanBeMergedRule,
} from 'graphql';

const n = 10000;
const fragments = Array.from(Array(n).keys()).reduce((fragments, next) => {
  return fragments.concat(`\n
    fragment X${next + 1} on Query {
      ...X${next}
[...]

Issue #4031, opened 2024-03-18
Support for deep input graphs


devkral:
I currently ran in the downstream project graphql-python/graphql-core in an recursion issue (inputs with depths around >200 fail).

It seems to be a construction problem in the base implementation (recursive design). And the token limiting is also useless because such deep graphs will fail roughly at the same level for the js implementation.

There is a simple fix: the "stack free" (not really stack free) generator pattern, you save the current state via generator and pushes nodes and new function calls at two stacks.

here an example (python, sry not ported yet):

````python
from functools import partial

def _foo(level=0, *, get_result):
    if level > 100000:
        yield level
        return
    print("1", level)
    yield partial(_foo, level=level + 1)
    print("2", get_result())
    yield level


def foo():
    result_stack = []

    fn_stack = [_foo(get_result=result_stack.pop)]

    while fn_stack:
        cur_el = fn_stack[-1]
        try:
            next_el = next(cur_el)
            if isinstance(next_el, partial):
                fn_stack.append(next_el(get_result=result_stack.pop))
            else:
                result_stack.append(next_el)
        except StopIteration:
            fn_stack.pop()
    return result_stack.pop()

print("final", foo())
````

Links:
-  https://github.com/graphql-python/graphql-core/issues/216   Issue reported downstream
- https://github.com/devkral/graphene-protector  Project, in which I use the generator pattern

devkral:
I looked through your code in detail:

as far as I can see, you don't use a recursive design like the downstream project. Am I right?
Sry for the noise.

devkral:
I first misread the code. You seems to be affected the same.

One way to prevent the stack issue would be to transform many calls into generators

yaacovCR:
I think a max-depth guard as suggested in the linked issue sounds like a reasonable first step.

This project is open-source and so a PR to introduce “stack-free” execution should be welcome. One issue to consider is that the reference implementation should try to match the specification, which has a recursive design, but that’s not necessarily a complete block — It might depend on how transformative the PR would be. It could also depend on what performance profile the change has. If it comes with a significant improvement, for example, I imagine that it would be easier to advocate for its inclusion.

devkral:
what do you think about the idea to use a rust parser for more performance and to easily synchronize the fix between both projects?

PR #4116, opened 2024-06-19
fix(16.x.x.): avoid stack size limit in overlapping field rules as well as execute

JoviDeCroock:
Fixes https://github.com/graphql/graphql-js/issues/3938

This moves us from our recursion when we are in a fragment to discovering more referenced fragments to a stack-iterator based approach. The memory consumption might be slightly higher during validation now but this will stop us from crashing which is less desired I presume.

EDIT: found some bugs, working on it (fixed in https://github.com/graphql/graphql-js/pull/4116/commits/18f81bd5af0c9a45b11a27380c156db4e16f7b60)

### DIFF
--- a/src/execution/collectFields.ts
+++ b/src/execution/collectFields.ts
@@ -21,10 +21,17 @@ import { typeFromAST } from '../utilities/typeFromAST';
 
 import { getDirectiveValues } from './values';
 
-interface FragmentEntry {
-  fragment: FragmentDefinitionNode;
+interface FieldEntry {
+  selection: FieldNode;
+  name: string;
+}
+
+interface EntryWithSelectionset {
+  selectionSet: SelectionSetNode;
   runtimeType: GraphQLObjectType;
 }
+
+type StackEntry = EntryWithSelectionset | FieldEntry;
 
 /**
  * Given a selectionSet, collects all of the fields and returns them.
@@ -42,32 +49,30 @@ export function collectFields(
   runtimeType: GraphQLObjectType,
   selectionSet: SelectionSetNode,
 ): Map<string, ReadonlyArray<FieldNode>> {
-  const foundFragments: Array<FragmentEntry> = [];
+  const stack: Array<StackEntry> = [{ selectionSet, runtimeType }];
   const fields = new Map();
   const visited = new Set<string>();
-  collectFieldsImpl(
-    schema,
-    fragments,
-    variableValues,
-    runtimeType,
-    selectionSet,
-    fields,
-    visited,
-    foundFragments,
-  );
 
   let entry;
-  while ((entry = foundFragments.pop()) !== undefined) {
-    collectFieldsImpl(
-      schema,
-      fragments,
-      variableValues,
-      entry.runtimeType,
-      entry.fragment.selectionSet,
-      fields,
-      visited,
-      foundFragments,
-    );
+  while ((entry = stack.shift()) !== undefined) {
+    if ('selectionSet' in entry) {
+      collectFieldsImpl(
+        schema,
+        fragments,
+        variableValues,
+        entry.runtimeType,
+        entry.selectionSet,
+        visited,
+        stack,
+      );
+    } else {
+      const fieldList = fields.get(entry.name);
+      if (fieldList !== undefined) {
+        fieldList.push(entry.selection);
+      } else {
+        fields.set(entry.name, [entry.selection]);
+      }
+    }
   }
 
   return fields;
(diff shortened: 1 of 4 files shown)


src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
```diff
@@ -1138,4 +1138,70 @@ describe('Validate: Overlapping fields can be merged', () => {
       },
     ]);
   });
+
+  it('does not hit stack size limits', () => {
+    const n = 10000;
```
JoviDeCroock, line 1143: The n is purposefully high here because some browsers cap at 9.000 and some environments at 1.000

JoviDeCroock:
@github-actions run-benchmark

src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
```diff
@@ -469,6 +483,44 @@ function findConflictsBetweenSubSelectionSets(
...
+const processDiscoveredFragments = (
+  context: ValidationContext,
+  conflicts: Array<Conflict>,
+  cachedFieldsAndFragmentNames: Map<SelectionSetNode, FieldsAndFragmentNames>,
+  comparedFragmentPairs: PairSet,
+  areMutuallyExclusive: boolean,
+  fieldMap: NodeAndDefCollection,
+  discoveredFragments: Array<Array<string>>,
+) => {
+  while (discoveredFragments.length !== 0) {
```
yaacovCR, line 497: tiny suggestion to make TS happy without additional check

```suggestion
  let item;
  while ((item = discoveredFragments.pop()) !== undefined) {
    if (comparedFragmentPairs.has(item[1], item[0], areMutuallyExclusive)) {
```

JoviDeCroock:
@yaacovCR applied your suggestion, thank you!

src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
```diff
@@ -469,6 +483,47 @@ function findConflictsBetweenSubSelectionSets(
   return conflicts;
 }
 
+// (E) Then collect any conflicts between the provided collection of fields
+// and any fragment names found in the given fragment.
+const processDiscoveredFragments = (
```
yaacovCR, line 488: Could this be declared as a named function rather than an arrow function?

JoviDeCroock: Is there a reason for that? 😅 but yes can do, does this stop approvals though?

yaacovCR:
Does this approach get us to validating nested named fragments of any depth? What was the old limit, what is the new? We presumably have a maximum depth secondary to a stack limit in execution as well with respect to field collection…. Is that now comparable?

JoviDeCroock:
This validates nested named fragments as the second test does. I'm not aware of any stack depth limitations in the implementation nor in the spec, this makes it limitless basically so no more runtime crash

src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
yaacovCR: Just style

yaacovCR:
> I'm not aware of any stack depth limitations in the implementation nor in the spec, this makes it limitless basically so no more runtime crash.

There are none in the spec.

But our execution implementation uses synchronous recursion and so we should hit limits both in collect fields and in field execution:

Eg https://github.com/graphql/graphql-js/issues/4031

I am personally not quite sure what they are tbh. When I am no longer on mobile and have enough bandwidth, I can try to take a closer look at this validation algorithm, but at first glance, it seems with your change to actually not necessarily use recursion at all for the nested named fragments, as advertised, dissimilar to collect fields and therefore allow a deeper limit/no limit. I don’t think that’s a blocker to approval => just curious, perhaps it might make sense to do something similar within collect fields.

yaacovCR (approved):

yaacovCR:
I think this can also be merged into main and then back ported

JoviDeCroock:
@yaacovCR you are right though, this still fails in execution https://github.com/graphql/graphql-js/pull/4116/commits/d9c4bd3eefb50dbe6ffa15b0c21b76e5d761bcd5 I can look at this separately or as part of this PR, as you prefer

yaacovCR:
I would be fine with either. I suppose we might want @graphql/tsc to address whether it's OK to change the simple recursive algorithm from the spec into something iterative.

It is certainly valid according to the specification, but perhaps there is some benefit in preserving a closer correspondence for the reference implementation. That small benefit might have to be weighed against the practical costs for this library users.

If I had a vote, I would vote in favor!

JoviDeCroock:
@yaacovCR I've replaced it in https://github.com/graphql/graphql-js/pull/4116/commits/4407b5539d96839ee83caf29fb115ab4946fd057

src/execution/collectFields.ts
```diff
@@ -37,16 +42,34 @@ export function collectFields(
...
     runtimeType,
     selectionSet,
     fields,
-    new Set(),
+    visited,
+    foundFragments,
   );
+
+  let entry;
+  while ((entry = foundFragments.pop()) !== undefined) {
```
JoviDeCroock, line 60: We could switch to leveraging `shift()` here if we are concerned with the order these are processed in, my assumption was that order doesn't matter as we come to a consistent view either way

yaacovCR: Probably better to preserve order, see https://github.com/graphql/graphql-spec/pull/1092

yaacovCR (commented):
Is it possible to unify visited Fragment names set with the new encountered fragments collection?

yaacovCR:
> Is it possible to unify visited Fragment names set with the new encountered fragments collection?

Thinking about this more, I don't think you can.

yaacovCR:
> @yaacovCR I've replaced it in [4407b55](https://github.com/graphql/graphql-js/commit/4407b5539d96839ee83caf29fb115ab4946fd057)

Just for the record, this will fix arbitrarily nested fragments at a single level => which is cool => but it's not like we can have arbitrarily deep queries, the completion algorithm within execution is still recursive, so we will hit a limit. This is I think fine, but just want to make explicit that we are not necessarily solving the issue in #4031 ....

src/execution/collectFields.ts
yaacovCR: Actually if I understand correctly now better, no matter what we have an ordering problem as it seems that all fragments are collected after all fields

saihaj (approved):

yaacovCR:
Preserving the serialized map ordering https://spec.graphql.org/draft/#sec-Serialized-Map-Ordering

I think this whole approach would cause all fragments to appear after fields, so I think that would also have to be fixed.

JoviDeCroock:
@github-actions run-benchmark

JoviDeCroock:
Closing this out due to the lack of real world scenarios where this will occur.

PR #4142, opened 2024-07-09
fix(incremental): do not ignore errors from filtered streams

yaacovCR:
depends on #4141

Early execution may result in non-completed streams after publishing is completed -- these streams must be closed using their return methods. When this occurs, we should pass through any error that occurs in the clean-up function instead of swallowing errors.

Swallowing errors is a bad practice that could lead to memory leaks. The argument in favor of swallowing the error might be that because the stream was "executed early" and the error does not impact any of the returned data, there is no "place" to forward the error.

But there is a way to forward the error, and that's on the next() call that returns { value: undefined, done: true } to end the stream. We will therefore appropriately send all the data and be able to pass through an error. Servers processing our stream should be made aware of this behavior and put in place error handling procedures that allow them to forward the data to clients when they see a payload with { hasNext: false } and then filter any further errors from clients. Servers could also postpone sending { hasNext: false } until the clean-up has been performed, and then optionally send information about the error in the extensions key, if that made sense. It would be up to the server!

### DIFF
--- a/src/execution/IncrementalPublisher.ts
+++ b/src/execution/IncrementalPublisher.ts
@@ -53,6 +53,19 @@ interface SubsequentIncrementalExecutionResultContext {
 }
 
 /**
+ * The IncrementalPublisherState Enum tracks the state of the IncrementalPublisher, which is initialized to
+ * "Started". When there are no more incremental results to publish, the state is set to "Completed". On the
+ * next call to next, clean-up is potentially performed and the state is set to "Finished".
+ *
+ * If the IncrementalPublisher is ended early, it may be advanced directly from "Started" to "Finished".
+ */
+enum IncrementalPublisherState {
+  Started = 1,
+  Completed = 2,
+  Finished = 3,
+}
+
+/**
  * This class is used to publish incremental results to the client, enabling semi-concurrent
  * execution while preserving result order.
  *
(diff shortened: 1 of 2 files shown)


yaacovCR:
There are 2-4 cases in which we might want to call `.return()` on an `asyncIterator` being used to iterate through a list.

1.  If a non-nullable list item errors, and we abort iteration.
    1.  without the use of the `@stream` directive or with erroring indices prior to hitting `initialCount`
    1.  without the use of the `@stream` directive or with erroring indices prior to hitting `initialCount`, even when the entire list is not included in the response anyway because of null bubbling from somewhere else...
    1.  with the use  the `@stream` directive with errors within streamed items, i.e. for nonreconcilable stream items.
    1.  when streams are "filtered" i.e. when early execution results in `asyncIterators` that are encountered but not actually send to the client.
1.  When using incremental delivery, when `.return()` is called on the `subsequentResults` iterator explicitly, we should forward that to the underlying stream `asyncIterator`.
    1.  for streams that would have been sent to the client
    1.  for streams that would have been "filtered"

In each of these scenarios, we would have to think about how to forward errors from `.return()` to the client if we would like to do so. Right now, the current state is that we are not tackling how to return errors from `.return()` at all with respect to (1), and that's probably how it should stay until users of `graphql-js` can weight in on what this API should look like.

PR #4145, opened 2024-07-15
improve(validation): add location to error messages

KonstantinSimeonov:
Having no locations on errors can be frustrating when debugging errors on large schemas. Append a location to each error message in schema validation.

Before:
```
Error: Unknown directive "@id".

Unknown directive "@createdAt".

Unknown directive "@label".

Unknown directive "@updatedAt".

Unknown directive "@searchField".

Unknown directive "@label".
```

After:
```
Error: Unknown directive "@id". (254:11)

Unknown directive "@createdAt". (255:24)

Unknown directive "@label". (255:35)

Unknown directive "@updatedAt". (256:24)

Unknown directive "@searchField". (258:17)

Unknown directive "@label". (258:30)
```

### DIFF
--- a/src/validation/validate.ts
+++ b/src/validation/validate.ts
@@ -108,6 +108,28 @@ export function validateSDL(
 }
 
 /**
+ * Combine multiple errors into a single error, whose message
+ * contains the error messages and their line:column on separate lines.
+ */
+function combineErrorsWithLocation(errors: ReadonlyArray<GraphQLError>): Error {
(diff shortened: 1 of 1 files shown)


JoviDeCroock (commented):
I like this change, we however need to test these changes and as the messages change a lot I would target this to `graphql:main` instead.

KonstantinSimeonov:
@JoviDeCroock I'll look into it

PR #4150, opened 2024-07-22
experiment: lazily create DeferredFragments

yaacovCR:
goal:

avoid creating or passing around the deferMap

methodology:

- each DeferredFragmentRecord will be unique for a given deferUsage and creationPath
- we annotate the path and deferUsage with a "fieldDepth" property representing the number of nested fields in the operation corresponding to the current execution path and the defer directive.
- from a path for a deferredGroupedFieldSet, we can derive the path for the deferredFragment from a given deferUsage by "rewinding" the deferredGroupedFieldSet path to the depth of the given deferUsage
- we also add a "fieldDepth" property to the FieldDetails structure as well so as to not require an additional depth argument to collectFields which would complicate memoization

### DIFF
--- a/src/execution/IncrementalGraph.ts
+++ b/src/execution/IncrementalGraph.ts
@@ -65,10 +65,15 @@ export class IncrementalGraph {
   addCompletedReconcilableDeferredGroupedFieldSet(
     reconcilableResult: ReconcilableDeferredGroupedFieldSetResult,
   ): void {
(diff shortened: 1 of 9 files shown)


yaacovCR:
not sure this is worth it -- but it works!!

yaacovCR:
closing this for now as best to not modify Path, but we can always revisit

PR #4153, opened 2024-07-25
incremental: lazily create DeferredFragments

yaacovCR:
depends on #4152

goal:

avoid creating or passing around the deferMap

methodology:

- each `DeferredFragmentRecord` will be unique for a given `DeferUsage` and creation `Path`.
- we annotate the `DeferUsage` with a `depth` property representing the path `depth` in the response for wherever this `DeferUsage` is used.
- from a given `ExecutionGroup`, we can derive the `path` for its `DeferredFragmentRecord` from the `DeferUsage` and the `ExecutionGroup` `path` by "rewinding" the `ExecutionGroup` `path` to the depth annotated on the corresponding `DeferUsage`.

~~note:~~

~~for performance, we now save the depth of the path on the path elements themselves. this could always be replaced by a call to `pathToArray` as `path.depth === pathToArray(path).length)`~~

### DIFF
--- a/src/execution/DeferredFragments.ts
+++ b/src/execution/DeferredFragments.ts
@@ -1,0 +1,108 @@
+import type { Path } from '../jsutils/Path.js';
+
+import type { DeferUsage } from './collectFields.js';
+import type {
+  PendingExecutionGroup,
+  StreamRecord,
+  SuccessfulExecutionGroup,
+} from './types.js';
+
+export type DeliveryGroup = DeferredFragmentRecord | StreamRecord;
+
+/** @internal */
+export class DeferredFragmentRecord {
+  path: Path | undefined;
+  label: string | undefined;
+  parentDeferUsage: DeferUsage | undefined;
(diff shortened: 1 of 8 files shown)


yaacovCR:
Comparing to #4150, we now annotate the `DeferUsage` with the actual depth in the response rather than the field depth in the operation. This is simpler and results in No Actual Change to the path interface (although technically we still change it by annotating it with the depth, this could always be derived later).

yaacovCR:
once we no longer create the deferMap, collectFields also does not need to return the newDeferUsages.

we do still return a boolean if defers were encountered to update the context so that buildExecutionPlan will be called; this allows us to still skip `buildExecutionPlan` in the wholly non-incremental case, somewhat narrowing the scope of that optimization

yaacovCR:

yaacovCR:

Thompson1985 (approved):

yaacovCR:
Closing this stale PR.

PR #4164, opened 2024-08-14
Proof of concept for release action

JoviDeCroock:
This would make life a bit easier when it comes to releasing as it would automatically pick up the tag and go and release our changes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,82 @@
+name: publish-npm
+on:
+  push:
+    tags: 
+      - v17.*.*
+
+permissions: {}
+jobs:
+  build-npm-dist:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - uses: actions/download-artifact@v4
+        with:
+          name: npmDist
+          path: npmDist
+
+      - name: Publish NPM package
+        run: npm publish --ignore-scripts ./npmDist
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.npm_canary_pr_publish_token }}
```
JoviDeCroock, line 66: We might want to leverage a different token and enable [Provenance](https://docs.npmjs.com/generating-provenance-statements)

PR #4166, opened 2024-08-16
chore(deps): upgrade dependencies

yaacovCR:
Errata:

Left typescript at v5.4.x as docusaurus currently conflicts with typescript v5.5.x. Left eslint v8.x as not all plugins are compatible with the new flat file config format introduced in eslint v9.x. Left at docusaurus 3.4.x as docusaurus-plugin-typedoc-api has a bug with docusaurus v3.5.x.

Code changes, in no particular order:

1. Prettier formatting changes.
2. Prettier moved to an async API, but the `writeGeneratedFile` utility, which previously included prettifying was passed as a callback function to TS and had to stay sync, so prettifying was separated into a separate async function -- the callback function luckily did not seem to actually requiring another round of prettifying, as it just involved renaming. All the other callsites of the new utility had to be made async. In the alternate, I investigated @prettier/sync and the lower-level make-synchronized and make-synchronous packages, but I could not get them working.
3. Plenty of eslint rule changes! I have tried to make sure that the rule list orders now match the linked documentation, so that further updates might be easier.
4. Minor docusaurus config tweaks to get the build to pass.


JoviDeCroock:
@yaacovCR can we hold off on this until we merge #4165 due to the many new conflicts that will arise.

yaacovCR:
@JoviDeCroock -- 

I think we can merge this now that #4171 has landed? If we do decide to go forward with #4165, my understanding is that there are no actual code changes any more from 16.x.x that will need to land, and this would only be for making the git history look pretty, so the conflicts at this point are moot?

JoviDeCroock (approved):

yaacovCR:
Updated again, this time `docusaurus-plugin-typedoc-api` has been upgraded to work with latest `docusaurus`, so I updated both of those, and I noticed that we can actually upgrade to typescript 5.5 as we get some warnings from underlying typedoc used by plugin, but website still compiles fine.

So everything upgraded except eslint, not all of our plugins support v9 even with new flat config file format disabled.

yaacovCR:
Seems like our ci does not like typescript 5.5, may have to roll that back

PR #4168, opened 2024-08-16
fix(validation): catch OverlappingFieldsCanBeMergedRule violations with nested fragments

sachindshinde:
In trying to prevent infinite loops, https://github.com/graphql/graphql-js/pull/3442 introduced a bug that causes certain violations of the [Field Selection Merging](https://spec.graphql.org/draft/#sec-Field-Selection-Merging) validation to not be caught (released in `16.3.0`, and backported to `15.9.0`). This PR fixes this bug, while continuing to prevent infinite loops.

Specifically, the code introduced in that PR misuses `comparedFragmentPairs` in the `collectConflictsBetweenFieldsAndFragment()` function. That function takes in fields/a selection set and checks whether it conflicts with the given fragment, while recursively checking those fields against nested fragments in the given fragment. The `comparedFragmentPairs` data structure is meant to store whether fragments have been checked for conflict against other fragments, but the `collectConflictsBetweenFieldsAndFragment()` does not compare fragments against nested fragments (only the selection set against fragments).

For an example of how this causes violations to be missed, see the test in the [first commit](https://github.com/graphql/graphql-js/pull/4168/commits/7abc1e73c4d4f13a45ed34d83b580fb65c1ac4b7). The first fragment causes an entry to be added to `comparedFragmentPairs` for the pair of the second and third fragments. When the query's selection set introduces a field that conflicts with a field in the third fragment, `collectConflictsBetweenFieldsAndFragment()` will then erroneously skip the third fragment due to the presence in `comparedFragmentPairs`. The [second commit](https://github.com/graphql/graphql-js/pull/4168/commits/94fd7c299f68b659c38efa3a149a6b93215db501) accordingly reverses the non-test changes of https://github.com/graphql/graphql-js/pull/3442.

The [third commit](https://github.com/graphql/graphql-js/pull/4168/commits/e444b16177e80f82f6fcfea14e90ffb0a286b781) shows a naive approach to preventing infinite loops, where we keep track of fragments seen while recursing into nested fragments in `collectConflictsBetweenFieldsAndFragment()`. This approach notably doesn't track the set of seen fragments across field boundaries, which is less memory overhead. However this unfortunately introduces a separate bug, shown by the test in the [fourth commit](https://github.com/graphql/graphql-js/pull/4168/commits/3e7e6f6591a5c0bbc229f180c19a30e39d57221b). This test creates nested fragments across field boundaries in such a way as to cause an infinite loop (`collectConflictsBetweenFieldsAndFragment()` eventually calls itself with the same arguments when it recurses into fields). The [fifth commit](https://github.com/graphql/graphql-js/pull/4168/commits/f24c570e7ee4ac4e858e8b593cfd1e6756eceb3b) accordingly reverses these changes.

The [sixth commit](https://github.com/graphql/graphql-js/pull/4168/commits/4c0a2d1d7f2d5aaf8b53d133766569a3ddd95594) introduces an actual fix. The `comparedFragmentPairs` data structure has been updated to additionally store a pair of fragment and fields/selection set instead of just two fragments. We then memoize comparisons between fields/fragments at the start of the `collectConflictsBetweenFieldsAndFragment()` function (similar to the fragment pair memoization at the start of `collectConflictsBetweenFragments()`). Note that this will increase the size of `comparedFragmentPairs` to scale with the product of the number of selection sets and the number of fragments, whereas previously it scaled with the square of the number of fragments (though I'm unsure if this extra overhead is significant in the context of full GraphQL validation).

### DIFF
--- a/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
+++ b/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
@@ -548,6 +548,33 @@ describe('Validate: Overlapping fields can be merged', () => {
     ]);
   });
 
+  it('reports deep conflict after nested fragments', () => {
+    expectErrors(`
+      fragment F on T {
+        ...G
+      }
+      fragment G on T {
+        ...H
+      }
+      fragment H on T {
+        x: a
+      }
+      {
+        x: b
+        ...F
+      }
+    `).toDeepEqual([
+      {
+        message:
+          'Fields "x" conflict because "b" and "a" are different fields. Use different aliases on the fields to fetch both if this was intentional.',
+        locations: [
+          { line: 12, column: 9 },
+          { line: 9, column: 9 },
+        ],
+      },
+    ]);
+  });
+
   it('ignores unknown fragments', () => {
     expectValid(`
       {
--- a/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
+++ b/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts
@@ -1165,4 +1165,31 @@ describe('Validate: Overlapping fields can be merged', () => {
       },
     ]);
   });
+
+  it('does not infinite loop on recursive fragments separated by fields', () => {
+    expectValid(`
+      {
+        ...fragA
+        ...fragB
+      }
+
+      fragment fragA on T {
+        x {
+          ...fragA
+          x {
+            ...fragA
+          }
+        }
+      }
+
+      fragment fragB on T {
+        x {
+          ...fragB
+          x {
+            ...fragB
+          }
+        }
+      }
+    `);
+  });
 });
--- a/src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
+++ b/src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
@@ -200,7 +200,6 @@ function findConflictsWithinSelectionSet(
         conflicts,
         cachedFieldsAndFragmentNames,
         comparedFragmentPairs,
-        null,
         false,
         fieldMap,
         fragmentNames[i],
@@ -232,15 +231,10 @@ function collectConflictsBetweenFieldsAndFragment(
   conflicts: Array<Conflict>,
   cachedFieldsAndFragmentNames: Map<SelectionSetNode, FieldsAndFragmentNames>,
   comparedFragmentPairs: PairSet,
-  comparedFragmentsForFields: null | Set<string>,
   areMutuallyExclusive: boolean,
   fieldMap: NodeAndDefCollection,
   fragmentName: string,
 ): void {
-  if (comparedFragmentsForFields?.has(fragmentName)) {
-    return;
-  }
-
   const fragment = context.getFragment(fragmentName);
   if (!fragment) {
     return;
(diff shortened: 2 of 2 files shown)


JoviDeCroock (approved):
This makes a lot of sense to me CC @graphql/graphql-js-reviewers for the second review, this seems like a legit bug though

yaacovCR (approved):
Agree that this is a great find and fix!

Only feedback is whether we can perhaps use a separate `compared<*>Map` for when comparing fields and fragments, as otherwise the naming gets confusing.

I took a stab at how that might look: https://github.com/sachindshinde/graphql-js/compare/fix-16.3.0-OverlappingFieldsCanBeMergedRule...yaacovCR:suggestions-for-overlapping-fix

We could always do that in a separate PR.

If we do choose to add a separate map, it might make sense to also add a separate `ruleContext` that bundles all of these maps, which should probably be a separate PR which we can do later:  https://github.com/yaacovCR/graphql-executor/compare/suggestions-for-overlapping-fix...rule-context

yaacovCR:
As an aside, applying this PR (and my first suggestion) onto main, we get the following results for our validation benchmarks as compared to main:

image

(note that the base of this PR is 16.x.x but we will of course need this on main as well -- again, great pickup @sachindshinde !!!!)

sachindshinde:
Thanks for the reviews!

@yaacovCR 
Regarding the first branch you've linked https://github.com/sachindshinde/graphql-js/compare/fix-16.3.0-OverlappingFieldsCanBeMergedRule...yaacovCR:suggestions-for-overlapping-fix , I think there's two things that would be nice to change:
1. The first is that the API for `comparedFieldsAndFragmentPairs` doesn't actually restrict it to compare just fields to fragments. Callers could accidentally call `comparedFieldsAndFragmentPairs.add(fieldMap1, fieldMap2)` or `comparedFieldsAndFragmentPairs.add(fragmentName1, fragmentName2)`.
2. The second is that the branch's logic would mean `comparedFieldsAndFragmentPairs.add(fieldMap, fragmentName)` and `comparedFieldsAndFragmentPairs.add(fragmentName, fieldMap)` would create two separate entries in the underlying `Map<NodeAndDefCollection | string, Map<NodeAndDefCollection | string, boolean>>` instead of one.

I've pushed another commit to this PR (the [seventh commit](https://github.com/graphql/graphql-js/pull/4168/commits/9e5df54b07f87a6c289ffe290e4d13faafc49271)), that addresses the above two issues via Typescript function overloading and by providing a comparator for `comparedFieldsAndFragmentPairs`. (I've also requested re-review from you both.)

yaacovCR (approved):
Great find, great fix, amazing TS wizardry!

yaacovCR:
> that addresses the above two issues

Looking at this again => it seems that our pairs in `comparedFieldsAndFragmentNames` collection are known to be of different types, and we always compare in the same order, so the flexibility of `PairSet` is somewhat unnecessary. Or, in other words, the `comparator` function will always return the same value.

We could introduce an `OrderedPairSet` class for that collection.

https://github.com/sachindshinde/graphql-js/compare/fix-16.3.0-OverlappingFieldsCanBeMergedRule...yaacovCR:graphql-executor:thought

Let me know what you think?

Thanks again for finding this!

yaacovCR:
@sachindshinde

image

here is the benchmark, shows an essential equivalence, but I would still favor that approach, I think it incidentally makes clearer what's going on in `PairSet` => let us know what you think when you get a chance!

yaacovCR:
@sachindshinde -- I took the liberty of pushing my suggestion and merging, if you would like to make additional changes, feel free to open a PR and fix-up as you would like. Thanks again so much for finding this!!

Issue #4169, opened 2024-08-17
`UniqueInputFieldNamesRule` incorrectly identifies field names as duplicates within `and/or` array


vctqs1:
Hi,

I encountered an issue with the`UniqueInputFieldNamesRule` in GraphQL. This rule currently prevents having duplicate field names even within `and/or`, which is problematic for certain queries.

For example, consider the following query: https://developer.salesforce.com/docs/platform/graphql/guide/filter-joins.html#multiple-semi-joins-or-anti-joins
```
query {
      Account (where: { and: [{
        Id: { inq: {
          Contact: { LastName: { like: "G%" } },
          ApiName:"AccountId" },
         },
         Id: { inq: {
           Opportunity : { IsClosed : { eq : false } },
           ApiName:"AccountId" }
         }
       }] }) {
            edges {
              node {
                Id
                Name {
                  value
                }
              }
            }
        }
     }
```

And got error 
```
AggregateError: GraphQL Document Validation failed with 1 errors;
  Error 0: There can be only one input field named "Id".
    at /Users/vctqs1/Documents/developer/manabie/school-portal-admin/src/squads/calendar/service/sf/Account.graphql:32:15
    at /Users/vctqs1/Documents/developer/manabie/school-portal-admin/src/squads/calendar/service/sf/Account.graphql:38:15
```
<img width="893" alt="image" src="https://github.com/user-attachments/assets/fd858520-4aef-47c7-9c00-f2c3f6ddc347">

In this query, the `UniqueInputFieldNamesRule` incorrectly recognizes the field name `Id` as a duplication. 

However, this is not a true duplication issue. In this context, `Id` is used in (Id in ... AND Id in ...), which are logically separate and should not be treated as duplicates.

PR #4170, opened 2024-08-17
[#4169] `UniqueInputFieldNamesRule` incorrectly identifies field names as duplicates within and/or array

vctqs1:
Fixed: #4169

To address the issue, use additional variables to store `ListValue` nodes and check if the duplicate field is present in the `ListValue`

### DIFF
--- a/src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
+++ b/src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
@@ -61,7 +61,7 @@ describe('Validate: Unique input field names', () => {
     `);
     expectValid(`
       {
-        field(arg: { or: [{ f: true }, { f: {f1: true} }] })
+        field(arg: { or: [{ f: { f1: "value1" } }, { f: { f1: "value2" } }] })
       }
     `);
     expectValid(`
--- a/src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
+++ b/src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
@@ -56,7 +56,7 @@ describe('Validate: Unique input field names', () => {
   it('allow and/or with duplicate fields in array', () => {
     expectValid(`
       {
-        field(arg: { and: [{ f: true }, { f: true }] })
+        field(arg: { and: [{ f: "value1" }, { f: "value2" }] })
       }
     `);
     expectValid(`
--- a/src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
+++ b/src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
@@ -69,19 +69,40 @@ describe('Validate: Unique input field names', () => {
         field(arg: { or: [{ f: true }, { f1: {f: true} }] })
       }
     `);
+    expectValid(`
+      {
+        field(arg: { 
+          or: [
+            { field: true },
+            { 
+              deep1: { 
+                deep2: {
+                  and: [{ field: false }, { field: true }]
+                }
+              } 
+            }
+            {
+              deep1: {
+                field: true
+              }
+            }
+          ]
+        })
+      }
+    `);
   });
 
   it('duplicate input object fields in objects of array', () => {
     expectErrors(`
       {
-        field(arg: { or: [{ f: true, f: true }] })
+        field(arg: { or: [{ f: "value1", f: "value2" }] })
       }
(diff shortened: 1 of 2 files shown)


src/validation/rules/UniqueInputFieldNamesRule.ts
```diff
@@ -34,18 +37,54 @@ export function UniqueInputFieldNamesRule(
...
+    ListValue: {
+      enter(node) {
+        node.values.forEach((valueNode) => {
+          if(valueNode.kind === 'ObjectValue') {
+            knownNamesInList.push(valueNode.fields);
+          }
+        });
+      },
+      leave() {
+        knownNamesInList = Object.create([]);
```
JoviDeCroock, line 49: I think this would also need to use a stack else it would make list --> object --> field --> list lose track of the outer object when it bubbles back up

JoviDeCroock (changes requested):

src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
```diff
@@ -53,6 +53,56 @@ describe('Validate: Unique input field names', () => {
...
 
+  it('allow and/or with duplicate fields in array', () => {
+    expectValid(`
+      {
+        field(arg: { and: [{ f: true }, { f: true }] })
+      }
+    `);
+    expectValid(`
+      {
+        field(arg: { or: [{ f: true }, { f: {f1: true} }] })
```
JoviDeCroock, line 64: This doesn't look valid as f here is both Boolean as well as an input object

src/validation/rules/UniqueInputFieldNamesRule.ts
vctqs1: Let me address the case of this `list --> object --> field --> list`.

src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
vctqs1: ah yeah, let me change this

src/validation/rules/UniqueInputFieldNamesRule.ts
vctqs1: I’ve changed the approach to use a stack to hold `listValue[]` and pop the state when it exits. I’ve also added a test for this case. Could you please help me review it again? @JoviDeCroock

src/validation/__tests__/UniqueInputFieldNamesRule-test.ts
JoviDeCroock, line 1: I was checking your tests and all of these also pass in current 16.x.x so I don't think this shows any issue in the current shape

vctqs1: ah yeah, you're right. Sorry, I misread the format; the issue was with the {} inside the single big object (labeled in blue). 
image



something like `[{Id: {}, Id: {}}]`

our tests cases was `[{Id: {}}, {Id: {}}]`

JoviDeCroock: Yes but

```
  it('duplicate input object fields in objects of array', () => {
    expectErrors(`
      {
        field(arg: [{Id: {}, Id: {}}])
      }
    `).toDeepEqual([
      {
        message: 'There can be only one input field named "Id".',
        locations: [
          { line: 3, column: 29 },
          { line: 3, column: 42 },
        ],
      },
    ]);
  });
```

Correctly reports it as an error

vctqs1: Oh, based on that, I think the issue is with Salesforce rather than the current validation rule. Thank you for your response, thank you for spending time to help me validation again.

JoviDeCroock: No worries, happy to help, if anything pops up feel free to reach out/create an issue so we can get to the bottom of this.

vctqs1: I get the error query from Salesforce docuemtn https://developer.salesforce.com/docs/platform/graphql/guide/filter-joins.html#multiple-semi-joins-or-anti-joins

vctqs1: thank you, I'll the close PR right now. @JoviDeCroock

PR #4177, opened 2024-08-23
standardize error messages prior to introducing schema coordinates

yaacovCR:
extracted from #3808

PR #3808 uses schema coordinates to improve GraphQL-JS error messages. To reduce the size of the PR, this commit standardizes error messages according to the general pattern that will be introduced with schema coordinates without introducing the coordinates themselves, in the hopes of aiding review of the later PR.

EDITED 8/26/2024:

I was able to reproduce all of the standardized error messages from #3808 except for the ones in getArgumentValues when it is passed a Field Definition, because the parent type is not passed. Everything else can be calculated for the error messages we are currently printing, although schema coordinates simplifies things.

Extracting these changes out of #3808 and rebasing #3808 on main will therefore will better demonstrate how schema coordinates improves the clarity of some of our error messages (namely, getArgumentValues) and simplifies printing them.

### DIFF
--- a/src/execution/__tests__/variables-test.ts
+++ b/src/execution/__tests__/variables-test.ts
@@ -226,7 +226,7 @@ describe('Execute: Handles inputs', () => {
           errors: [
             {
               message:
-                'Argument "input" has invalid value ["foo", "bar", "baz"].',
+                'Argument "input" of type "TestInputObject" has invalid value ["foo", "bar", "baz"].',
               path: ['fieldWithObjectInput'],
               locations: [{ line: 3, column: 41 }],
             },
@@ -262,7 +262,7 @@ describe('Execute: Handles inputs', () => {
           errors: [
             {
               message:
-                'Argument "input" has invalid value { c: "foo", e: "bar" }.',
+                'Argument "input" of type "TestInputObject" has invalid value { c: "foo", e: "bar" }.',
               path: ['fieldWithObjectInput'],
               locations: [{ line: 3, column: 41 }],
             },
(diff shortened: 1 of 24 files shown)


src/execution/__tests__/variables-test.ts
```diff
@@ -226,7 +226,7 @@ describe('Execute: Handles inputs', () => {
           errors: [
             {
               message:
-                'Argument "input" has invalid value ["foo", "bar", "baz"].',
+                'Argument "input" of type "TestInputObject" has invalid value ["foo", "bar", "baz"].',
```
saihaj, line 229: I think this can be really helpful when debugging!

saihaj (approved):

src/execution/__tests__/variables-test.ts
yaacovCR: I can't take credit for it @leebyron tucked this addition into #3808.

Note that after $3808 lands, this will be:

`Argument "TestType.fieldWithObjectInput(input:)" of type "TestInputObject" has invalid value` and it will be even better.

yaacovCR:
Note that schema coordinates do not make knowing the type of an erroring argument trivial. They make knowing the parent type of an erroring field and of the field on an erroring argument, trivial, so you can easily receive:

`Argument "TestType.fieldWithObjectInput(input:)" of type "TestInputObject" has invalid value ["foo", "bar", "baz"].`

instead of

`Argument "input" of type "TestInputObject" has invalid value ["foo", "bar", "baz"].`

But they are still really great. But when printing errors, it's actually not that hard to have access to the type of the argument. This PR follows @leebyron 's approach in #3808 of adding that information in with select errors that were missing it.

yaacovCR:
Actually, I was able to minimize the diff everywhere except for the errors in `getArgumentValues` when it is passed a Field Definition, because the parent type is not passed. Everything else can be calculated for the error messages we are currently printing, although schema coordinates simplifies things.

PR #4181, opened 2024-09-07
Fix/optimize value coercion check for OneOf type

Cito:
Where a JavaScript value is coerced to a GraphQL OneOf Input Type, a null value should only be reported when the "one of" condition is satisfied. The code block starting at line 158 [here](https://github.com/graphql/graphql-js/blob/1dbdadc6f46d2c97c71c7a4f5c61a2c75d1536ae/src/utilities/coerceInputValue.ts#L158) that accesses `keys[0]` should not be executed if `keys` is empty or contains more than one item.

The PR fixes this by adding an else statement. Alternatively, the "if/else" branches could be reversed, and the `keys.length !== 1` check should become a `keys.length === 1` check.

### DIFF
--- a/src/utilities/coerceInputValue.ts
+++ b/src/utilities/coerceInputValue.ts
@@ -153,16 +153,16 @@ function coerceInputValueImpl(
             `Exactly one key must be specified for OneOf type "${type}".`,
           ),
         );
-      }
-
-      const key = keys[0];
-      const value = coercedValue[key];
-      if (value === null) {
-        onError(
-          pathToArray(path).concat(key),
-          value,
-          new GraphQLError(`Field "${key}" must be non-null.`),
-        );
+      } else {
+        const key = keys[0];
+        const value = coercedValue[key];
+        if (value === null) {
+          onError(
+            pathToArray(path).concat(key),
+            value,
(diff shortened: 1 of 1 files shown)


yaacovCR:
@cito -- do you think there is some value in reporting multiple errors, i.e. if there are multiple keys, and some of them are null, multiple things are wrong? The code at present, heads toward that direction, checks the number of values, and checks the identity of the first, so you would get more detailed errors if there were two keys, first being null, but not more details errors if there were two keys with the second null.

Or possibly I have the flow wrong!

yaacovCR:
@cito from feedback from @benjie on #4195, it seems like the correct style is one error per incorrect value, with the suggestion there to just include a more generic error message to catch all cases, so that if the incorrect number of keys are present or any are null, just emit an error message like:

> "Within OneOf Input Object type "TestOneOfInputObject", exactly one field must be specified, and the value for that field must be non-null."

So I think that's the suggestion I would make here. The idea of #4195 is to add the pre-coercion checks => your bug fix for not emitting more than one error could be landed separately, before that. What do you think?

Thompson1985 (approved):

yaacovCR (approved):

yaacovCR:
We can always combine the error messages in a separate PR.

PR #4185, opened 2024-09-09
fix(IntrospectionType): properly type using TypeKind enum

yaacovCR:


yaacovCR:
type change is breaking

JoviDeCroock (approved):

PR #4096, opened 2024-05-24
implement bundling changes suggested in #4062

phryneas:
This is now ready for review.

Current AreTheTypesWrong output:

```sh
 graphql v16.7.0

 No problems found 🌟


"graphql/language/ast.js"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/language/parser.js"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/jsutils/instanceOf.js"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/execution/execute.js"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/error"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/execution"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/language"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/subscription"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/type"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/utilities"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/validation"

node10: 🟢 
node16 (from CJS): 🟢 (CJS)
node16 (from ESM): 🟢 (ESM)
bundler: 🟢 

***********************************

"graphql/package.json"

[...]

PR #4188, opened 2024-09-09
instanceOf: disable dev instanceOf checks if NODE_ENV not set

yaacovCR:
development behavior should be opt in, not opt out

https://github.com/graphql/graphql-js/issues/4075#issuecomment-2088870697

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "test": "npm run lint && npm run check && npm run testonly:cover && npm run prettier:check && npm run check:spelling && npm run check:integrations",
     "lint": "eslint --cache --max-warnings 0 --rulesdir resources/eslint-internal-rules/ .",
     "check": "tsc --pretty",
-    "testonly": "mocha --full-trace src/**/__tests__/**/*-test.ts",
+    "testonly": "dotenvx run --quiet --env NODE_ENV=development -- mocha --full-trace src/**/__tests__/**/*-test.ts",
     "testonly:cover": "c8 npm run testonly",
     "testonly:watch": "npm run testonly -- --watch",
     "prettier": "prettier --cache --cache-strategy metadata --write --list-different .",
@@ -56,6 +56,7 @@
   "devDependencies": {
     "@docusaurus/core": "3.5.2",
     "@docusaurus/preset-classic": "3.5.2",
+    "@dotenvx/dotenvx": "^1.14.0",
     "@mdx-js/react": "3.0.1",
     "@svgr/webpack": "8.1.0",
     "@types/chai": "4.3.19",
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -1,10 +1,12 @@
 import { inspect } from './inspect.js';
 
 /* c8 ignore next 3 */
-const isDevelopment =
-  globalThis.process != null &&
+const isProduction =
+  globalThis.process == null ||
   // eslint-disable-next-line no-undef
-  process.env.NODE_ENV === 'development';
+  (process.env.NODE_ENV !== 'development' &&
+    // eslint-disable-next-line no-undef
+    process.env.NODE_ENV !== 'test');
 
 /**
  * A replacement for instanceof which includes an error warning when multi-realm
@@ -13,8 +15,13 @@ const isDevelopment =
  * See: https://webpack.js.org/guides/production/
  */
 export const instanceOf: (value: unknown, constructor: Constructor) => boolean =
-  isDevelopment
+  /* c8 ignore next 6 */
+  // FIXME: https://github.com/graphql/graphql-js/issues/2317
+  isProduction
     ? function instanceOf(value: unknown, constructor: Constructor): boolean {
+        return value instanceof constructor;
+      }
+    : function instanceOf(value: unknown, constructor: Constructor): boolean {
         if (value instanceof constructor) {
           return true;
         }
@@ -45,11 +52,6 @@ spurious results.`,
           }
         }
         return false;
-      }
-    : /* c8 ignore next 4 */
-      // FIXME: https://github.com/graphql/graphql-js/issues/2317
-      function instanceOf(value: unknown, constructor: Constructor): boolean {
-        return value instanceof constructor;
       };
 
 interface Constructor extends Function {
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -1,10 +1,10 @@
 import { inspect } from './inspect.js';
 
 /* c8 ignore next 3 */
-const isProduction =
+const isDevelopment =
   globalThis.process != null &&
   // eslint-disable-next-line no-undef
-  process.env.NODE_ENV === 'production';
+  process.env.NODE_ENV === 'development';
 
 /**
  * A replacement for instanceof which includes an error warning when multi-realm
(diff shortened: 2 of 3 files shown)


yaacovCR:
This will still not fix #4017. An alternative to this PR would be to rearchitect the instanceOf checks more completely....

@JoviDeCroock @phryneas @benjie

JoviDeCroock:
@yaacovCR yes as proposed in https://github.com/graphql/graphql-js-wg/issues/125

JoviDeCroock (dismissed):

src/jsutils/instanceOf.ts
```diff
@@ -1,10 +1,10 @@
 import { inspect } from './inspect.js';
 
 /* c8 ignore next 3 */
-const isProduction =
+const isDevelopment =
   globalThis.process != null &&
   // eslint-disable-next-line no-undef
-  process.env.NODE_ENV === 'production';
+  process.env.NODE_ENV === 'development';
```
benjie, line 7: `test` and `development` should both invoke the non-production behavior.

package.json
```diff
@@ -37,7 +37,7 @@
     "test": "npm run lint && npm run check && npm run testonly:cover && npm run prettier:check && npm run check:spelling && npm run check:integrations",
     "lint": "eslint --cache --max-warnings 0 --rulesdir resources/eslint-internal-rules/ .",
     "check": "tsc --pretty",
-    "testonly": "mocha --full-trace src/**/__tests__/**/*-test.ts",
+    "testonly": "dotenvx run --quiet --env NODE_ENV=development -- mocha --full-trace src/**/__tests__/**/*-test.ts",
```
benjie, line 40: Does mocha set NODE_ENV=test by default? If so, this shouldn't be needed.

src/jsutils/instanceOf.ts
benjie: To minimize the size of this PR, I'd personally rewrite this to just:

```ts
const isProduction = process.env.NODE_ENV !== 'development' && process.env.NODE_ENV !== 'test';
```

Then the lower lines in this PR shouldn't be needed.

benjie (changes requested):
I think defaulting to production mode makes sense; however it will probably cause more issues to be filed to be aware of that.

src/jsutils/instanceOf.ts
yaacovCR: 965457044

so this is now:

```ts
const isProduction = globalThis.process == null || (process.env.NODE_ENV !== 'development' && process.env.NODE_ENV !== 'test');
```

retaining the check to see whether process is defined globally

package.json
yaacovCR: it does not... tested locally without dotenvx and fails

also, see:
https://github.com/mochajs/mocha/issues/185

thinking from their end is that you might want to use mocha to test in multiple environments, best if mocha stays out of your env

src/jsutils/instanceOf.ts
```diff
@@ -1,10 +1,12 @@
 import { inspect } from './inspect.js';
 
 /* c8 ignore next 3 */
-const isDevelopment =
-  globalThis.process != null &&
+const isProduction =
+  globalThis.process == null ||
```
JoviDeCroock, line 5: Most bundlers won't eliminate the || you can run tests at https://github.com/JoviDeCroock/graphql-minifier-experiments

src/jsutils/instanceOf.ts
yaacovCR: https://github.com/JoviDeCroock/graphql-minifier-experiments/pull/1

I have not gone through all the bundlers, but I think that flipping the default would require us changing our instructions about what to set globalThis.process to, now we would have to set to false

https://github.com/graphql/graphql-js/blob/16.x.x/website/docs/tutorials/going-to-production.md

and if I remember correctly, the reason that is in at all is to support cloudflare.

So I am not sure if I want to move forward with this PR actually.... as if we have a better long-term solution, we should move to that without this intermediate stage......?

yaacovCR: https://github.com/JoviDeCroock/graphql-minifier-experiments/pull/1

I have not gone through all the bundlers, but I think that flipping the default would require us changing our instructions about what to set globalThis.process to, now we would have to set to false

https://github.com/graphql/graphql-js/blob/16.x.x/website/docs/tutorials/going-to-production.md

and if I remember correctly, the reason that is in at all is to support cloudflare.

So I am not sure if I want to move forward with this PR actually.... as if we have a better long-term solution, we should move to that without this intermediate stage......?

@JoviDeCroock @benjie @phryneas

phryneas: Long-term solution: 3 versions of this file - one with the `development` export condition, without any of this logic, one with the `production` export condition without any of this, and one `default` that falls back to this (or a similar) solution.

phryneas: That does require us to start with an `exports` field though. #4096 might be a step in that direction - reviews welcome.

yaacovCR: Does an exports field help us for this internal function? Do we need an imports map as well?

phryneas: We would have two directions here:
* import map and hope that packagers actually support that nicely (I honestly do not know)
* exports field and we `import from 'graphql/instanceof'`

yaacovCR: It would be best not to export internal functions like instanceof at all. If bundlers respect import maps that would seem to be best.

phryneas: The problem is that e.g. Webpack 4 doesn't support import/export maps at all, but is afaik still used by tools like StoryBook.
Having `exports` in place that just also exist under the same path on the file system as a fallback can work in those cases - but `imports` maps unfortunately cannot :/

yaacovCR: Hmmm.

WebPack 5 was released on 2020.10.10 almost 4 years ago.

In terms of Storybook, which I have zero familiarity with, released v7.0.0 on 2023.04.12 almost 1.5 years ago and with that major [dropped support for Webpack 4.](https://storybook.js.org/docs/7/migration-guide).

phryneas: @markerikson I remember you had a very good reason to support Webpack 4 in the Redux packages. Is that obsolete with this?

benjie (commented):
I don’t like adding a dependency just to set an envvar; but 🤷‍♂️

It seems possible for globalThis.process to exist without globalThis.process.env, which would cause a throw. But that’s an issue with the existing code anyway.

Looks alright; I can’t comment whether it impacts bundling/tree-shaking.

Thompson1985 (approved):

src/jsutils/instanceOf.ts
yaacovCR: Submitted https://github.com/graphql/graphql-js/pull/4221 as an alternative demoing the use of an imports map.

yaacovCR: @phryneas I closed #4221 because once we use an imports map, it becomes tricky to use graphql in browsers in a bundler-free manner. I think unfortunately using the exports map and publishing instanceOf might have the same problem.

phryneas: @yaacovCR What problems are you seeing?
Generally, I'd say that unless we publish only a single file, users will have to adopt import maps, but that's the future of ESM on the web anyways.

yaacovCR: I’m not sure if we want to require the use of an import map but I would definitely defer to @JoviDeCroock  on that.

mcollina (approved):
lgtm

Speedy1991:
I just came across this PR in the v17 roadmap. I tried to follow as much of the discussion as possible (including related issues and PRs), and I wanted to add my two cents.

I’d prefer to stick with:

```const isProduction = globalThis.process && process.env.NODE_ENV === 'production';```

for a few simple reasons: it’s straightforward, readable, and explicit.

Relying on “non-production” checks can be error-prone. A small typo like `testt` or a trailing space such as `test ` would silently be treated as production. In contrast, environment names like test, testing, dev, or development vary widely between projects, while production is fairly universal and well-understood.

I also think the point from @benjie is still valid:
> I don’t like adding a dependency just to set an env var.

yaacovCR:
Closed in favor of #4464

PR #4193, opened 2024-09-14
Upgrade deprecated actions

JoviDeCroock:
In response to some of our actions starting to fail deprecate all of the actions that don't work anymore due to using Node 10/12

- [Looks like upload artefact is safe to upgrade](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) we don't seem to overwrite files
- [Same for download](https://github.com/actions/download-artifact/blob/main/docs/MIGRATION.md)


benjie (approved):
Aligns with my own research 👍

PR #4195, opened 2024-09-18
add additional runtime errors for pre-coercion OneOf errors

yaacovCR:
This changes the implementation to match the specification.

Adds additional tests as well.


src/execution/__tests__/oneof-test.ts
```diff
@@ -178,6 +183,354 @@ describe('Execute: Handles OneOf Input Objects', () => {
             message:
               'Variable "$input" got invalid value { a: "abc", b: null }; Exactly one key must be specified for OneOf type "TestInputObject".',
           },
+          {
+            message:
+              'Variable "$input" got invalid value null at "input.b"; Field "b" of OneOf type "TestInputObject" must be non-null.',
+            locations: [{ line: 2, column: 16 }],
+          },
```
benjie, line 190: Two errors should not be generated for the same value - once it's determined that there was more than one key, no further checks should be performed.

yaacovCR: 70e5c4f91

src/execution/__tests__/oneof-test.ts
```diff
@@ -181,5 +186,338 @@ describe('Execute: Handles OneOf Input Objects', () => {
...
+      expectJSON(result).toDeepEqual({
+        data: {
+          test: null,
+        },
+        errors: [
+          {
+            // A nullable variable in a oneOf field position would be caught at validation-time
+            // hence the vague error message here.
+            message:
+              'Argument "input" of type "TestInputObject!" has invalid value { a: $a, b: $b }.',
```
benjie, line 255: I understand the errors are vague and why; but the `{ a: $a }` error messages below are confusing. Maybe it would be better to replace all the errors with:

``Argument "input" of type "TestInputObject!" has invalid value - exactly one field must be specified, and the value for that field must not be `null`.``

src/execution/__tests__/oneof-test.ts
```diff
@@ -181,5 +186,338 @@ describe('Execute: Handles OneOf Input Objects', () => {
...
+            locations: [{ line: 3, column: 23 }],
+            path: ['test'],
+          },
+        ],
```
benjie, line 371: I had to validate that this is the correct behavior (e.g. do we care about the uncoerced value during coercion) - but yes, we do. Good catch :+1: 

> - If the input object literal or **unordered map** does not contain exactly one entry an error must be thrown.
> - Within the input object literal or unordered map, if the single entry is {null} an error must be thrown.
> - If the **coerced unordered map** does not contain exactly one entry an error must be thrown.
> - If the value of the single entry in the coerced unordered map is {null} an error must be thrown.

benjie (changes requested):
Thanks for your work on this! It seems right; but let's improve the error messages

src/execution/__tests__/oneof-test.ts
yaacovCR: I updated the OneOf error messages to something uniform, but the vague ones are a result of the uniform behavior within `getArgumentValues()`

https://github.com/graphql/graphql-js/blob/78755526c8e89e2182c7f19bec68e36ca2279409/src/execution/values.ts#L232-L237 which does not transmit the original error (but would be caught at validation time).

Within #3813 , part of the rebased Default Value Coercion/Validation stack, @leebyron has fixed this to provide more informative errors.

src/validation/rules/ValuesOfCorrectTypeRule.ts
```diff
@@ -193,7 +193,7 @@ function validateOneOfInputObject(
   if (isNotExactlyOneField) {
     context.reportError(
       new GraphQLError(
-        `OneOf Input Object "${type}" must specify exactly one key.`,
+        `Within OneOf Input Object Type "${type}", exactly one field must be specified, and the value for that field must be non-null.`,
```
benjie, line 196:
```suggestion
        `Within OneOf Input Object type "${type}", exactly one field must be specified, and the value for that field must be non-null.`,
```

src/validation/rules/ValuesOfCorrectTypeRule.ts
```diff
@@ -206,9 +206,12 @@ function validateOneOfInputObject(
 
   if (isNullLiteral) {
     context.reportError(
-      new GraphQLError(`Field "${type}.${keys[0]}" must be non-null.`, {
-        nodes: [node],
-      }),
+      new GraphQLError(
+        `Within OneOf Input Object Type "${type}", exactly one field must be specified, and the value for that field must be non-null.`,
```
benjie, line 210:
```suggestion
        `Within OneOf input object type "${type}", exactly one field must be specified, and the value for that field must be non-null.`,
```

src/validation/rules/ValuesOfCorrectTypeRule.ts
```diff
@@ -221,7 +224,7 @@ function validateOneOfInputObject(
     if (isNullableVariable) {
       context.reportError(
         new GraphQLError(
-          `Variable "$${variableName}" must be non-nullable to be used for OneOf Input Object "${type}".`,
+          `Variable "$${variableName}" must be non-nullable to be used for Within OneOf Input Object Type "${type}".`,
```
benjie, line 227:
```suggestion
          `Variable "$${variableName}" must be non-nullable to be used for within OneOf input object type "${type}".`,
```

src/execution/__tests__/oneof-test.ts
```diff
@@ -152,13 +157,13 @@ describe('Execute: Handles OneOf Input Objects', () => {
           {
             locations: [{ column: 16, line: 2 }],
             message:
-              'Variable "$input" got invalid value { a: "abc", b: 123 }; Exactly one key must be specified for OneOf type "TestInputObject".',
+              'Variable "$input" got invalid value { a: "abc", b: 123 }; Within OneOf Input Object Type "TestInputObject", exactly one field must be specified, and the value for that field must be non-null.',
```
benjie, line 160: This error seems to be a combination of two errors; ideally this `Within` should not have a capital letter since it's in the middle of a sentence. Alternatively; can the preceding semicolon be changed to a fullstop?

benjie (changes requested):
Just being picky over capitalisation now...

Do we generally capitalise `Object Type` / `Interface Type` and so on in error messages? If so, the suggestions to reduce capitalization should be dismissed.

src/execution/__tests__/oneof-test.ts
yaacovCR: I think another way of saying what I am saying that all of the tests within the oneOf test file, including the new ones, actually belong within the tests for `coerceInputValue()` and `coerceInputLiteral`() (formerly known as `valueFromAST()`).

I'll set about to do that. Then these tests won't have opaque error messages in our test files that make it difficult to understand what exactly is being tested, although users who choose not to validate will still get opaque errors, at least until #3813 lands.

src/validation/rules/ValuesOfCorrectTypeRule.ts
yaacovCR: d4d1bcb13

src/validation/rules/ValuesOfCorrectTypeRule.ts
yaacovCR: d4d1bcb13

src/execution/__tests__/oneof-test.ts
yaacovCR: we could definitely do either, but it would have to be in a different PR, a changing this behavior would effect multiple errors across the code base, and should not be mixed with this PR which introduces some new behavior. (We're already stretching things a teense by changing the existing errors while adding new ones, but I am Ok with that as it at least all stays within OneOf)

src/validation/rules/ValuesOfCorrectTypeRule.ts
yaacovCR: we can introduce a separate PR to get consistency across the code base on this, as well as probably consistency of whether we quote the type/schema element "name", but that should be in a separate non OneOf specific PR

src/execution/__tests__/oneof-test.ts
yaacovCR: This has been updated with the last couple of commits, and is now ready for re-review.

src/validation/rules/ValuesOfCorrectTypeRule.ts
yaacovCR: https://github.com/graphql/graphql-spec/issues/1112

yaacovCR: https://github.com/graphql/graphql-js/issues/4196

src/execution/__tests__/oneof-test.ts
yaacovCR: Thinking about this more, moving the tests locations from file to file within this PR which ostensibly is just supposed to add a few new test cases will make the diff confusing. As, honestly, will updating the error messages as you requested. I should probably split this PR into multiple ones to better delineate what's happening.

@cito [has an open PR](https://github.com/graphql/graphql-js/pull/4181) to reduce the number of OneOf errors and I have submitted your suggestion within this PR to him to combine the error messages. So that PR can house the error message changes, and then this PR can be about adding the pre-coercion checks, and then we can do another one if necessary to move the tests around, which may in the end not be necessary.

Thompson1985 (approved):

src/execution/__tests__/oneof-test.ts
benjie: GitHub tip: if you reference PRs in a bulleted list then it will expand them showing their status and title; e.g. instead of:

> at least until #3813 lands.

You could write:

> at least until this PR lands:
>
> - #3813

I mention this because the "open PR" referenced above is now merged:

- #4181

src/utilities/coerceInputValue.ts
```diff
@@ -134,8 +134,28 @@ function coerceInputValueImpl(
       );
     }
 
+    /**
+     * keys with the value "undefined" do not appear within the coerced input
```
benjie, line 138:
```suggestion
     * Keys with the value "undefined" do not appear within the coerced input
```

benjie (changes requested):
This looks correct to me; great work!

benjie (approved):

JoviDeCroock (approved):
Should this be backported to v16

yaacovCR:
This PR ended up turning into a lot of editing of error messages, but contains the following two cases:

1. `{ foo: 123, unknownField: 123 }`

Assuming `unknownField` is an unknown field, besides the error for the unknown field, you should also get an error for OneOf, because there is more than one field specified. The query is invalid anyway, so it's kind of not a huge deal.

2. `{ foo: 123, field: undefined }`

Per non-normative graphql-js specific behavior, when the value for field is the JavaScript value `undefined`, this is equivalent to `{ foo: 123 }` and should be valid.

I should probably extract both of these cases from this PR as two separate small PRs, and then they can be merged to v16.x.x. The diffuse changes to the error messages make this PR difficult to follow.

Thompson1985 (approved):

benjie:
> Assuming unknownField is an unknown field, besides the error for the unknown field, you should also get an error for OneOf, because there is more than one field specified.

Hmmm... Thanks for calling this out! We should only get one error for each mistake. I think when oneOf performs this "one field" check, it's actually checking that exactly one of _its_ fields is set.

From the spec edits:

> a special variant of Input Objects where the type system asserts that exactly one of the fields must be set and non-null

> and thus requires exactly one of its field be provided

There are some places in the spec where it suggests  that it throws an error even if its not one of its fields:

> - If the input object literal or unordered map does not contain exactly one entry an error must be thrown.

This is just saying that an error must be thrown, not that this must throw an _additional_ error. An existing error being thrown for an unknown field would be suitable in this case, as clarified through the coercion table:

| Literal Value            | Variables                        | Coerced Value                                       |
| ------------------------ | -------------------------------- | --------------------------------------------------- |
| `{ b: 123, c: "xyz" }`   | `{}`                             | Error: Unexpected field {c}                         |

This doesn't seem sufficiently clear though, so I'll edit the spec text so that oneof only applies these assertions to its own fields.

benjie:
Upon further reading, this clarification to the spec edits is not required. The spec already states:

> In either case, the input object literal or unordered map must not contain any entries with names not defined by a field of this input object type, otherwise a request error must be raised.

This is the first thing to be stated before visiting any individual rules, thus I think the coercion rules can rely on this assumption.

The coercion table also demonstrates this behavior.

yaacovCR:
Thanks for exploring that one with me, I think you are right.

I should also call out that there is an additional failure mode added by this PR that actually does not yet have a test case, namely, from the table at https://github.com/graphql/graphql-spec/pull/825

> | `{ a: $a, b: $b }`       | `{ a: "abc" }`                   | Error: Exactly one key must be specified            |

If I'm not mistaken, the coerced value will just be `{ a: 'abc' }`, and without checking the pre-coercion values, this would pass, but this PR would fail it because it has that second variable pre-coercion => although I don't think I actually added a test for that case in that PR!

Issue #4201, opened 2024-09-24
`graphql-language-server@5.3.0` causes type errors with `graphql@16.9.0`


phryneas:
This happened over in https://github.com/apollographql/vscode-graphql/pull/175 ([failing CI job](
https://app.circleci.com/pipelines/github/apollographql/vscode-graphql/812/workflows/7ef38837-7497-42de-aa37-f170622a0ff6/jobs/3925)):

```
node_modules/graphql-language-service/dist/parser/types.d.ts:116:37 - error TS2694: Namespace 'Kind' has no exported member 'LIST_NULLABILITY_OPERATOR'.

116     LIST_NULLABILITY_OPERATOR: Kind.LIST_NULLABILITY_OPERATOR;
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/graphql-language-service/dist/parser/types.d.ts:117:30 - error TS2694: Namespace 'Kind' has no exported member 'NON_NULL_ASSERTION'.

117     NON_NULL_ASSERTION: Kind.NON_NULL_ASSERTION;
                                 ~~~~~~~~~~~~~~~~~~

node_modules/graphql-language-service/dist/parser/types.d.ts:118:26 - error TS2694: Namespace 'Kind' has no exported member 'ERROR_BOUNDARY'.

118     ERROR_BOUNDARY: Kind.ERROR_BOUNDARY;
                             ~~~~~~~~~~~~~~
```

I'll open a PR to backport these enum members, as that seems to be the easiest way to prevent this

JoviDeCroock:
I think the main issue here is that `graphql-language-service` is using a newer alpha of `graphql-js` which has breaking changes included 😅 I don't see backporting as a viable option here as this will cause people to defensively code around options that would never be present.

phryneas:
`graphql-language-service` has very broad `peerDependencies`:  `"graphql": "^15.5.0 || ^16.0.0 || ^17.0.0-alpha.2"`

So it doesn't seem to depend with these at all.

phryneas:
Wait... they are bundling the types? Oh boy.

phryneas:
Opened https://github.com/graphql/graphiql/issues/3790, closing here.

PR #4202, opened 2024-09-24
backport `Kind` members from `main` to ensure TS compatibility

phryneas:
This would be one possible solution for https://github.com/graphql/graphql-js/issues/4201


phryneas:
Probably better to fix up the packaging of `graphql-language-server`

PR #4206, opened 2024-09-27
polish(ast): prefer undefined over empty arrays

### LINKED ISSUE #2203: Test all functions (with AST input) that they handle `undefined` instead of array properites

yaacovCR:
see: https://github.com/graphql/graphql-js/pull/2405#issuecomment-579077133

Manually created ASTs always allowed undefined in place of empty arrays; this change simply updates the parser to more closely follow the manual approach. This is therefore not technically a breaking change, but considering that there may be tools not aware of this, we have labelled it a BREAKING_CHANGE to highlight it for consumers.

This closes #2203 as our tests now cover the undefined case by default whenever there is an empty array.


yaacovCR:
Wo! This is much more involved to do consistently => and I am not sure worth the effort. Closing for now, anybody interested can take this as a starting point if they want.

yaacovCR:
Well, it works!

It would seem that it's more accurate to use undefined when an `optionalMany(...)` is skipped entirely and an empty array when it is found, but has an empty list. So, it would seem that this PR is technically more correct.

Peeking around at https://astexplorer.net/ it seems that TS parser has some empty arrays, some undefined, while for Babel I didn't on a very cursory look see any undefined.

But still not 100% sure this is too compelling, with the status quo being somewhat compelling! Comments welcome!

Thompson1985 (approved):

yaacovCR:
Merging this PR allows us to close #2203, improving robustness of our test coverage. Presumably our downstream users were also affected by #2203 and had to account for manually created AST nodes, so they should be robust to this change, but I upgraded it to a breaking change anyway just to call it out.

JoviDeCroock (approved):

JoviDeCroock:
Probably needs a rebase

JoviDeCroock (approved):

PR #4208, opened 2024-09-29
Revive for..of optimization

JoviDeCroock:
This revives the effort for us to use ES6 features without sacrificing performance.

We'll transpile `for..of` statements to 

```js
for (let i = 0; i < x.length; i++) {
  const item = x[i]
}
```

There's a potential optimisation in here to extract `x.length` outside of the loop so we don't have to access `length` every time. Another thing we'll have to transpile is array destructuring.

With the automated approach like here we'll still have a lot of `for..of` where we invoke the iterator of i.e. a map.

Related to https://github.com/graphql/graphql-js/pull/3691


JoviDeCroock:
@github-actions run-benchmark

JoviDeCroock:
To be consistent we'd have to translate Set/Map however we rely on those being mutated during runtime which does not boat well.

PR #4210, opened 2024-10-01
introduce ValidatedExecutionArgs

yaacovCR:
to safely export buildExecutionContext() as validateExecutionArgs()

motivation: this will allow us to:

1. export `executeOperation()` and `executeSubscription()` for those who would like to use directly.
2. add a `perEventExecutor` option to `ExecutionArgs` that allows one to pass a custom context for execution of each execution event, with the opportunity to clean up the context on return, a la #2485 and #3071, addressing #894, which would not require re-coercion of variables.

The signature of the `perEventExecutor` option would be:

```ts
type SubscriptionEventExecutor = ( validatedExecutionArgs: ValidatedExecutionArgs): PromiseOrValue<ExecutionResult>
```

rather than:

```ts
type SubscriptionEventExecutor = ( executionArgs: ExecutionArgs): PromiseOrValue<ExecutionResult>
```

This might be a first step to integrating `subscribe()` completely into `execute()` (see: https://github.com/graphql/graphql-js/pull/3644) but is also a reasonable stopping place.

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -277,7 +277,7 @@ function executeOperation(
   const exeContext: ExecutionContext = {
     validatedExecutionArgs,
     errors: undefined,
-    cancellableStreams: new Set(),
+    cancellableStreams: undefined,
   };
   try {
     const { schema, fragments, rootValue, operation, variableValues } =
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -94,18 +94,21 @@ import {
  */
 const collectSubfields = memoize3(
   (
-    exeContext: ExecutionContext,
+    validatedExecutionArgs: ValidatedExecutionArgs,
     returnType: GraphQLObjectType,
     fieldGroup: FieldGroup,
-  ) =>
-    _collectSubfields(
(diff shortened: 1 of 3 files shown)


yaacovCR:
motivation: this will allow us to:

1. export `executeOperation()` and `executeSubscription()` for those who would like to use directly.
2. add a `perEventExecutor` option to `ExecutionArgs` that allows one to pass a custom context for execution of each execution event, with the opportunity to clean up the context on return, a la #2485 and #3071, addressing #894, which would not require re-coercion of variables.

The signature of the `perEventExecutor` option would be:

```ts
type SubscriptionEventExecutor = ( validatedExecutionArgs: ValidatedExecutionArgs): PromiseOrValue<ExecutionResult>
```

rather than:

```ts
type SubscriptionEventExecutor = ( executionArgs: ExecutionArgs): PromiseOrValue<ExecutionResult>
```

This might be a first step to integrating `subscribe()` completely into `execute()` (see: https://github.com/graphql/graphql-js/pull/3644) but is also a reasonable stopping place.

JoviDeCroock (approved):

PR #4213, opened 2024-10-02
perf: optimizations for buildResolveInfo() and replaceVariables()

yaacovCR:
These can be merged separately with the "Rebase" strategy to get separate commits for use with `git bisect`.

`buildResolveInfo()` was deoptimized within https://github.com/graphql/graphql-js/pull/4015.
`replaceVariables()` was never optimized, uses less memory with simple function rather than visitor.


yaacovCR:
This fixes our benchmarks:

image

JoviDeCroock (approved):

Issue #4190, opened 2024-09-12
Add option to disable Field Suggestions since LexicalDistance Constructor takes Too Long for Unknown Field-Names


walterunt:
If someone sends a simple query with a very long unknown/invalid fieldname (e.g. millions of characters), the validation rule `FieldsOnCorrectTypeRule` takes very long. This may lead to a lot of unecessary processing power wasted, when not used.

See also files:
- https://github.com/graphql/graphql-js/blob/16.x.x/src/validation/rules/FieldsOnCorrectTypeRule.ts
- https://github.com/graphql/graphql-js/blob/16.x.x/src/jsutils/suggestionList.ts

It looks like that the `LexicalDistance` calculation constructor is very slow when generating the `_inputArray` and `_rows` members, if the fieldname is very long.

Example Query (e.g. the content is several MB big (e.g. 40mb)):` {"query":"{ unknownfield_unknownfield_...unknownfield}"}`.

This all comes in place, when calculating `suggestions` for fields.

Please make an option to completely disable the suggestion calculation, i.e. do *not* perform the calculation at all if an option is set.

I have found similar issues:
- https://github.com/graphql/graphql-js/issues/3229
- https://github.com/graphql/graphql-js/issues/2247

but it looks like that the calculation is still in place, when the DisableSuggestion plugin is activated.

JoviDeCroock:
This will be implemented in v17 and has been in #4214

PR #4214, opened 2024-10-04
feat: Allow skipping suggestions

JoviDeCroock:
Resolves https://github.com/graphql/graphql-js/issues/4190
Resolves https://github.com/graphql/graphql-js/issues/2247
Supersedes https://github.com/graphql/graphql-js/pull/3467

This is a best practice already in production however GraphQL JS does not provide an out of the box solution. In the community we have solutions like https://escape.tech/graphql-armor/docs/plugins/block-field-suggestions/ however in v17 it would be worth considering providing a built-in.

The recommendation here would be to do `validate(schema, doc, rules, { hideSuggestions: process.env.NODE_ENV === 'production' })`

### DIFF
--- a/src/execution/collectFields.ts
+++ b/src/execution/collectFields.ts
@@ -55,7 +55,7 @@ interface CollectFieldsContext {
   operation: OperationDefinitionNode;
   runtimeType: GraphQLObjectType;
   visitedFragmentNames: Set<string>;
-  shouldProvideSuggestions: boolean;
+  maskSuggestions: boolean;
 }
 
 /**
@@ -74,7 +74,7 @@ export function collectFields(
   variableValues: VariableValues,
   runtimeType: GraphQLObjectType,
   operation: OperationDefinitionNode,
-  shouldProvideSuggestions: boolean,
+  maskSuggestions: boolean,
 ): {
   groupedFieldSet: GroupedFieldSet;
   newDeferUsages: ReadonlyArray<DeferUsage>;
@@ -88,7 +88,7 @@ export function collectFields(
     runtimeType,
     operation,
     visitedFragmentNames: new Set(),
-    shouldProvideSuggestions,
+    maskSuggestions,
   };
 
   collectFieldsImpl(
@@ -118,7 +118,7 @@ export function collectSubfields(
   operation: OperationDefinitionNode,
   returnType: GraphQLObjectType,
   fieldDetailsList: FieldDetailsList,
-  shouldProvideSuggestions: boolean,
+  maskSuggestions: boolean,
 ): {
   groupedFieldSet: GroupedFieldSet;
   newDeferUsages: ReadonlyArray<DeferUsage>;
@@ -130,7 +130,7 @@ export function collectSubfields(
     runtimeType: returnType,
     operation,
     visitedFragmentNames: new Set(),
-    shouldProvideSuggestions,
+    maskSuggestions,
   };
   const subGroupedFieldSet = new AccumulatorMap<string, FieldDetails>();
   const newDeferUsages: Array<DeferUsage> = [];
@@ -172,18 +172,14 @@ function collectFieldsImpl(
     runtimeType,
     operation,
     visitedFragmentNames,
+    maskSuggestions,
   } = context;
 
   for (const selection of selectionSet.selections) {
     switch (selection.kind) {
       case Kind.FIELD: {
         if (
-          !shouldIncludeNode(
-            selection,
-            variableValues,
-            fragmentVariableValues,
-            context.shouldProvideSuggestions,
-          )
+          !shouldIncludeNode(selection, variableValues, fragmentVariableValues)
         ) {
           continue;
         }
@@ -200,7 +196,6 @@ function collectFieldsImpl(
             selection,
             variableValues,
             fragmentVariableValues,
-            context.shouldProvideSuggestions,
           ) ||
           !doesFragmentConditionMatch(schema, selection, runtimeType)
         ) {
@@ -213,7 +208,6 @@ function collectFieldsImpl(
           fragmentVariableValues,
           selection,
           deferUsage,
-          context.shouldProvideSuggestions,
         );
 
         if (!newDeferUsage) {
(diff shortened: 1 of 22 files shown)


Thompson1985 (approved):

yaacovCR:
It looks like calls to `didYouMean()` also show up within `execute(...)`, namely `coerceInputValue()` via `GraphQLEnumType.parseValue()` and unknown input fields on a `GraphQLInputObjectType`. So we may need a fourth argument to `coerceInputValue()` that is then passed along as a second argument to `GraphQLEnumType.parseValue()`? That would cause the signature of `parseValue()` to be different for scalars than enums => which could be avoided by passing the new argument also to scalars. The specified scalars don't give suggestions, but perhaps some custom scalars do (???) and maybe this argument could be useful. Otherwise, we could just diverge between enums and custom scalars.

Thompson1985 (approved):

JoviDeCroock:
@yaacovCR added it for enums/scalars, just need to add tests now

src/utilities/extendSchema.ts
```diff
@@ -728,7 +728,11 @@ function getDeprecationReason(
 function getSpecifiedByURL(
   node: ScalarTypeDefinitionNode | ScalarTypeExtensionNode,
 ): Maybe<string> {
-  const specifiedBy = getDirectiveValues(GraphQLSpecifiedByDirective, node);
+  const specifiedBy = getDirectiveValues(
+    GraphQLSpecifiedByDirective,
+    node,
+    true,
```
yaacovCR, line 734: it would be nice if we could leave this off

src/utilities/extendSchema.ts
```diff
@@ -737,5 +741,5 @@ function getSpecifiedByURL(
  * Given an input object node, returns if the node should be OneOf.
  */
 function isOneOf(node: InputObjectTypeDefinitionNode): boolean {
-  return Boolean(getDirectiveValues(GraphQLOneOfDirective, node));
+  return Boolean(getDirectiveValues(GraphQLOneOfDirective, node, true));
```
yaacovCR, line 744: same

src/utilities/valueFromAST.ts
```diff
@@ -148,7 +148,7 @@ export function valueFromAST(
     // no value is returned.
     let result;
     try {
-      result = type.parseLiteral(valueNode, variables);
+      result = type.parseLiteral(valueNode, variables, true);
```
yaacovCR, line 151: Considering `valueFromAST()` deprecated, we don't 100% have to support this new feature, but no harm.

yaacovCR (commented):
I think this has been a much-requested feature! Amazing! Left some comments in terms of some suggestions, basically I am hopeful that we can make this non-breaking in terms of function signatures, even though of course it's going in v17, so we could break things if we need to.

src/execution/values.ts
```diff
@@ -159,6 +167,7 @@ export function getFragmentVariableValues(
   fragmentSpreadNode: FragmentSpreadNode,
   fragmentSignatures: ReadOnlyObjMap<GraphQLVariableSignature>,
   variableValues: VariableValues,
+  shouldProvideSuggestions: boolean,
```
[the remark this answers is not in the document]
JoviDeCroock: The thing is, I really like uniform API design which makes this whole naming consistent across the codebase making it a simple search.

src/execution/collectFields.ts
```diff
@@ -66,12 +67,14 @@ interface CollectFieldsContext {
...
  * @internal
  */
+// eslint-disable-next-line @typescript-eslint/max-params
 export function collectFields(
   schema: GraphQLSchema,
   fragments: ObjMap<FragmentDetails>,
   variableValues: VariableValues,
   runtimeType: GraphQLObjectType,
   operation: OperationDefinitionNode,
+  shouldProvideSuggestions: boolean,
```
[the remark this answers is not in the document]
JoviDeCroock: We need it for fragment-arguments as well which can be actual enums, ... so I think we should keep this around, did flick it to `maskSuggestions`

yaacovCR:
> The recommendation here would be to do `validate(schema, doc, rules, { maskSuggestions: process.env.NODE_ENV === 'production' })`

Or `validate(schema, doc, rules, { maskSuggestions: process.env.NODE_ENV !== 'development' })`

yaacovCR:
I added some thoughts in #4219

src/execution/collectFields.ts
```diff
@@ -66,12 +67,14 @@ interface CollectFieldsContext {
...
  * @internal
  */
+// eslint-disable-next-line @typescript-eslint/max-params
 export function collectFields(
   schema: GraphQLSchema,
   fragments: ObjMap<FragmentDetails>,
   variableValues: VariableValues,
   runtimeType: GraphQLObjectType,
   operation: OperationDefinitionNode,
+  maskSuggestions: boolean,
```
JoviDeCroock, line 77: We could replace this with an options object, the coverage aspect will be a bit harder to achieve though

yaacovCR: i think this is fine for now

src/execution/collectFields.ts
```diff
@@ -265,6 +271,7 @@ function collectFieldsImpl(
             fragmentVariableSignatures,
             variableValues,
             fragmentVariableValues,
+            false,
```
yaacovCR, line 274: I think this has to be `maskSuggestions` instead of false because this is where we get the fragment variable values, so it's not just about the built ins.

yaacovCR (commented):
Just a few questions

src/utilities/valueFromAST.ts
JoviDeCroock:
```suggestion
      result = type.parseLiteral(valueNode, variables);
```

src/graphql.ts
```diff
@@ -61,6 +61,7 @@ import type { ExecutionResult } from './execution/types.js';
 export interface GraphQLArgs {
   schema: GraphQLSchema;
   source: string | Source;
+  maskSuggestions?: boolean;
```
yaacovCR, line 64: I guess this could be Maybe <boolean> as well for consistency

yaacovCR (approved):

PR #4217, opened 2024-10-07
chore(workspace): lint and format on commit

JoviDeCroock:
Personally have been really annoyed with the lack of pre-commit in this repository, I have to commit twice every time as I'm not used to manually calling `npm run lint && npm run format`

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+lint-staged
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "build:deno": "node --loader ts-node/esm resources/build-deno.ts",
     "diff:npm": "node --loader ts-node/esm resources/diff-npm-package.ts",
     "gitpublish:npm": "bash ./resources/gitpublish.sh npm npmDist",
-    "gitpublish:deno": "bash ./resources/gitpublish.sh deno denoDist"
+    "gitpublish:deno": "bash ./resources/gitpublish.sh deno denoDist",
+    "prepare": "husky"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -86,6 +89,14 @@
     "typescript": "5.4.5",
     "url-loader": "4.1.1"
   },
+  "lint-staged": {
```
yaacovCR, line 92: could we move the configuration to a `.lintstagedrc.json` file in the project root like out other configs?

.husky/pre-commit
```diff
@@ -0,0 +1 @@
+lint-staged
```
yaacovCR, line 1: instead of running lint-staged directly, could we add a new package script like `npm run format` that combines eslint and prettier and a `npm run format:staged` that just does it for the staged files? not sure if that would work, but I think would then be super transparent what is happening... what do you think?

yaacovCR (approved):
Tested on my local machine, and prevents me from unhealthy committing. :) I added some thoughts to consider, but it could always be done in a later PR.

.husky/pre-commit
JoviDeCroock: I'd personally love to not rewrite the whole staged logic, the thing being that now it only executes these scripts against the staged files, and it effectively differentiates the command based on the extension.

yaacovCR: Then maybe we could drop the `npm run format` script and just have `npm run format:staged` which just calls `lint-staged` with our config. This adds a level of indirection, but it makes sure that someone can easily run our pre-commit hook manually, which I think is a value add.

yaacovCR: Unless there is a different way to do that?

PR #4219, opened 2024-10-07
suggestions

yaacovCR:
includes:

1. changes just to reduce the diff
2. use of Maybe<boolean> for some of our exported helper functions to match rest of our API
3. reverts some of my previous suggestions revolving around collectFields() => i was under the mistaken assumption that we never needed to pass maskSuggestions to collectFields, but because of fragment arguments, we might have complex inputs and we might need to mask suggestions
4. errata

### DIFF
--- a/src/execution/collectFields.ts
+++ b/src/execution/collectFields.ts
@@ -55,7 +55,7 @@ interface CollectFieldsContext {
   operation: OperationDefinitionNode;
   runtimeType: GraphQLObjectType;
   visitedFragmentNames: Set<string>;
-  shouldProvideSuggestions: boolean;
+  maskSuggestions: boolean;
 }
 
 /**
@@ -74,7 +74,7 @@ export function collectFields(
   variableValues: VariableValues,
   runtimeType: GraphQLObjectType,
   operation: OperationDefinitionNode,
-  shouldProvideSuggestions: boolean,
+  maskSuggestions: boolean,
 ): {
   groupedFieldSet: GroupedFieldSet;
   newDeferUsages: ReadonlyArray<DeferUsage>;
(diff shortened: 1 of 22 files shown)


yaacovCR:
I'm not totally sure about the argument ordering, I'm not happy to separate variableValues and fragmentVariableValues, and I'm not happy to put maskSuggestions earlier, because that would be a breaking change, so I settled on putting maskSuggestions later, which is fine, except that if we ever revert the fragment arguments experiments, that would end up being a breaking change... Now, I don't ==think== we are going to revert it.... thoughts on that minor issue?

src/execution/collectFields.ts
```diff
@@ -171,14 +171,20 @@ function collectFieldsImpl(
...
   for (const selection of selectionSet.selections) {
     switch (selection.kind) {
       case Kind.FIELD: {
         if (
-          !shouldIncludeNode(selection, variableValues, fragmentVariableValues)
+          !shouldIncludeNode(
+            selection,
+            variableValues,
+            fragmentVariableValues,
+            maskSuggestions,
```
JoviDeCroock, line 186: He, I knew I was forgetting about arg name suggestions when I removed this but I honestly didn't mind I would just either always suggest for the built-in directives or not at all

JoviDeCroock, line 186: He, I knew I was forgetting about arg name suggestions when I removed this but I honestly didn't mind I would just either always suggest for the built-in directives or not at all

JoviDeCroock (commented):
I'm sorry but in general this feels very nitpicky about argument ordering 😅

src/execution/collectFields.ts
yaacovCR: seems fair to always suggest for built ins!

PR #4220, opened 2024-10-07
move to separate script

yaacovCR:
https://github.com/graphql/graphql-js/pull/4217#discussion_r1790144268

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "testonly:watch": "npm run testonly -- --watch",
     "prettier": "prettier --cache --cache-strategy metadata --write --list-different .",
     "prettier:check": "prettier --cache --cache-strategy metadata --check .",
-    "format:staged": "lint-staged",
+    "precommit": "lint-staged",
     "check:spelling": "cspell --cache --no-progress \"**/*\"",
     "check:integrations": "mocha --full-trace resources/integration-test.ts",
     "serve": "docusaurus serve --dir websiteDist/ --config website/docusaurus.config.cjs",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,6 +42,7 @@
     "testonly:watch": "npm run testonly -- --watch",
     "prettier": "prettier --cache --cache-strategy metadata --write --list-different .",
     "prettier:check": "prettier --cache --cache-strategy metadata --check .",
+    "format:staged": "lint-staged",
```
JoviDeCroock, line 45: To me this just feels like an extra step while it's pretty ecosystem standard for the pre-commit to invoke lint-staged. People can see the config and won't have to go

.husky/pre-commit - package.json - .lintstagedrc.json - package.json

Instead they can just cut out that second step and in most js cases even the first.

yaacovCR: What I'm looking for is a manual way to run lint-staged? Does that exist?

This PR introduces that, and once we introduce that, it seems like the pre-commit hook should point to that method rather than invoking lint-staged directly in that it sets us up to be able to modify args to lint-staged in one place, were we ever to choose to do so.

I guess we are coming at this differently. I want this option because I have worked in repositories/setups where the pre-commit hook kind of fails inexplicably, and it would be nice to be able to directly invoke the error messages right in the terminal exactly the way the pre-commit hook is doing so. I don't want to be able to look at the config, I want to be able to invoke the command and see what's actually happening.

It seems like you are envisioning a scenario in which the commit hook fails, and the contributor doesn't know about npm run format:staged and so has to trace the path from the pre-commit hook to the final scripts to see what's happening. I don't envision that to be a problem, once they see that we're running a script, they can just run it and don't have to check anything else, or if it is a problem, it will only be a problem only once for each contributor once they see our command.

It might be a bit unconventional, but I'm seeing a low potential for harm.

I am changing the name to `npm run precommit` I think that might potentially help here to be even more explicit?

JoviDeCroock (approved):

PR #4221, opened 2024-10-08
use conditional imports to trigger development `instanceOf()`

yaacovCR:
This PR changes this library to use the development version of instanceOf when a condition is set rather than based on a Node JS specific environment variable.

See:
https://github.com/graphql/graphql-js/issues/4075
https://github.com/graphql/graphql-js/pull/4188


yaacovCR:
Closing, this precludes bundler-free web versions

Issue #3409, opened 2021-12-06
The result of `getIntrospectionQuery` does not match `IntrospectionQuery` type


Woodz:
`getIntrospectionQuery` only requests the `name` field from the root `queryType` object
(https://github.com/graphql/graphql-js/blob/6e48d16f92b9a6df8638b1486354c6be2537033b/src/utilities/getIntrospectionQuery.ts)
```
queryType { name }
```

yet the `IntrospectionQuery` datatype defines `queryType` as `IntrospectionNamedTypeRef` (https://github.com/graphql/graphql-js/blob/main/src/utilities/getIntrospectionQuery.ts#L167) which requires `kind` as well
```
export interface IntrospectionNamedTypeRef<
  T extends IntrospectionType = IntrospectionType,
> {
  readonly kind: T['kind'];
  readonly name: string;
}
```

This means that copying the result of the querying a GQL API with the introspection query as JSON and trying to assign to `IntrospectionQuery` will fail with:

> The types of '__schema.queryType' are incompatible between these types.
        Property 'kind' is missing in type '{ name: string; }' but required in type 'IntrospectionNamedTypeRef<IntrospectionObjectType>'.

sibo2000:
Bumping this.

conatus:
Still seeing this issue. At the moment, chucking in a `@ts-ignore` but would be nice to have a firmer fix.

Exact same situation, passing an introspection in JSON into `buildClientSchema`. Using the very latest `graphql` at `16.4.0`.

```javascript
import * as introspectionResult from "../../../graphql.schema.json";

const schema = buildClientSchema(introspectionResult);
```

Produces the following error:
```
Argument of type '{ __schema: { queryType: { name: string; }; mutationType: { name: string; }; subscriptionType: null; types: ({ kind: string; name: string; description: string; fields: null; inputFields: null; interfaces: null; enumValues: null; possibleTypes: null; } | ... 9 more ... | { ...; })[]; directives: ({ ...; } | { ...; })...' is not assignable to parameter of type 'IntrospectionQuery'.
  The types of '__schema.queryType' are incompatible between these types.
    Property 'kind' is missing in type '{ name: string; }' but required in type 'IntrospectionNamedTypeRef<IntrospectionObjectType>'
 ```
 
 Other issues downstream in users of this library:
- https://github.com/apollographql/apollo-tooling/issues/1491
- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34649

[...]

PR #4222, opened 2024-10-10
Include kind in the types

JoviDeCroock:
Supersedes https://github.com/graphql/graphql-js/pull/3910
Fixes https://github.com/graphql/graphql-js/issues/3909
Fixes https://github.com/graphql/graphql-js/issues/3409

This puts the Selection-set in line with the expected type

### DIFF
--- a/src/type/__tests__/introspection-test.ts
+++ b/src/type/__tests__/introspection-test.ts
@@ -32,7 +32,7 @@ describe('Introspection', () => {
     expect(result).to.deep.equal({
       data: {
         __schema: {
-          queryType: { name: 'SomeObject' },
+          queryType: { name: 'SomeObject', kind: 'OBJECT' },
           mutationType: null,
           subscriptionType: null,
           types: [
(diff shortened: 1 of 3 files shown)


yaacovCR:
An issue here is that we are still missing fields/interfaces in that root type position, and doesn't seem to make sense to duplicate those...

Alternative in #4224 is not workable yet.

JoviDeCroock:
@yaacovCR not sure what you mean 😅

yaacovCR:
https://github.com/graphql/graphql-js/blob/45e28a5b74cfa811857e7a4694b095da284d14eb/src/utilities/getIntrospectionQuery.ts#L227-234

The full correct type needs more than the kind

JoviDeCroock:
@yaacovCR that is incorrect, the type for root-types is the following https://github.com/graphql/graphql-js/blob/45e28a5b74cfa811857e7a4694b095da284d14eb/src/utilities/getIntrospectionQuery.ts#L311

Issue #3913, opened 2023-06-18
Support for bigint in ID


vwkd:
It's desirable to represent an ID on the server as a `bigint` instead of a `number`.

Currently, the ID scalar type supports serializing `number` to `string`, but not `bigint`.

I'd like to see support for serialization of `bigint` to `string` similar to `number`.

I think it would require adding this code to https://github.com/graphql/graphql-js/blob/d766c8eb34fb03c5ee63c51380f2b012924e970c/src/type/scalars.ts#L225-L227

```js
if (typeof coercedValue == "bigint") {
  return coercedValue.toString();
}
```

JoviDeCroock:
This is correct, the spec supports 128-bit numbers https://spec.graphql.org/draft/#sec-ID

yaacovCR:
This is completed in v17 by #4550

PR #4223, opened 2024-10-11
Allow for serializing int as BigInt

JoviDeCroock:
Fixes https://github.com/graphql/graphql-js/issues/3913

In the spec we say that we allow for `BigInt` as an `ID` value in [here](https://spec.graphql.org/draft/#sel-GAHXZFDCAACCBshF). However in JS we don't support serializing a response value from `BigInt` to `String`.

### DIFF
--- a/src/type/__tests__/scalars-test.ts
+++ b/src/type/__tests__/scalars-test.ts
@@ -534,6 +534,8 @@ describe('Type System: Specified scalar types', () => {
       expect(parseValue(1)).to.equal('1');
       expect(parseValue(0)).to.equal('0');
       expect(parseValue(-1)).to.equal('-1');
+      expect(parseValue(BigInt(123))).to.equal('123');
+      expect(parseValue(1n)).to.equal('1');
 
       // Maximum and minimum safe numbers in JS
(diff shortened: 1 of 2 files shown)


yaacovCR:
How does this compare to #4088 ?

JoviDeCroock:
@yaacovCR https://github.com/graphql/graphql-js/pull/4088 adds full support to Int/... for BigInt however that isn't really specced out compared to ID which explicitly in the spec states BigInt support.

yaacovCR:
Maybe #3109 then? I had closed the earlier PRs in favor of the more comprehensive, but that could have been the less optimal approach.

JoviDeCroock:
That is indeed similar, that only gets applied to `serialize` not `parseValue`, I can add a `co-authored-by` though as they do deserve credit for this. Thank you for making me aware of this PR, it wasn't linked to the issue so I did not find it

yaacovCR:
> However in JS we don't support serializing a response value from `BigInt` to `String`.

Did you mean ID here?

Did we want to support String as well?

JoviDeCroock:
I meant ID, I don't know what you mean here 😅

Issue #3790, opened 2022-11-29
`validate` does not check that selection sets are non-empty


berkowitze:
The `validate` function does not find any issues if you pass an AST with an empty selection set for an object/interface field.

Running into this when building a query ast based on a form, and toggling form fields sometimes lead to an empty selection set. I'm going to properly handle "empty" selection sets (by removing the node from the ast), but it being caught by validation would be really nice as well as a backup and for correctness.

From [the spec](https://spec.graphql.org/draft/#sel-EALRHHDDBDFAADFBAy5W):
> If selectionType is an interface, union, or object:
The subselection set of that selection must NOT BE empty.

Repro ([sandbox](https://codesandbox.io/s/festive-cray-95jz9v?file=/src/index.ts)):
```ts
import {
  buildSchema,
  DocumentNode,
  Kind,
  OperationTypeNode,
  validate
} from "graphql";

const s = `
type Query {
  dog: Int
}

schema {
  query: Query
}
`;

const schema = buildSchema(s);

const query: DocumentNode = {
  kind: Kind.DOCUMENT,
  definitions: [
    {
      kind: Kind.OPERATION_DEFINITION,
      operation: OperationTypeNode.QUERY,
      selectionSet: {
        kind: Kind.SELECTION_SET,
        selections: []
      }
    }
  ]
};

console.log(validate(schema, query)); // outputs `[]`
```

Happy to fix this but this would be my first OSS contribution so wanted to ask:
- Is this the right place to post this?
- Is this actually an issue?
- What is the general approach to a fix?
- What is the right wording for any error messaging?

yaacovCR:
This seems like a bug!

yaacovCR:
It seems like from the section you are quoting that the ScalarLeafsRule was co-opted in the spec to also say something about non-leaf types and then did not get implemented. We may need another rule, but it might pay to discuss at a wg meeting whether the spec text should be divided as well.

JoviDeCroock:
Technically this case is covered by the parser already so we'd have to manually construct the AST. I do think it's worth covering this case though, in case a consumer is using an alternative parser

JoviDeCroock:
Added in v17 on `main` thank you for reporting! We'll also bring this up at a WG meeting

PR #4228, opened 2024-10-13
Cover empty field-selections

JoviDeCroock:
Closes https://github.com/graphql/graphql-js/issues/3790

This follows the spec closer to the letter as we consider empty selection-sets invalid as well. This however does not cover empty operation-definitions, which should also be considered invalid.

The parser itself considers empty-selections invalid as it requires you to specify a name before a closing curly, hence why testing happens a bit more manual here. Despite our parser not considering this valid, I think this is good to validate when folks use a separate parser or construct documents manually as this is valid in our type-system.

### DIFF
--- a/src/validation/__tests__/ScalarLeafsRule-test.ts
+++ b/src/validation/__tests__/ScalarLeafsRule-test.ts
@@ -1,8 +1,15 @@
 import { describe, it } from 'mocha';
 
-import { ScalarLeafsRule } from '../rules/ScalarLeafsRule.js';
+import { expectJSON } from '../../__testUtils__/expectJSON.js';
 
-import { expectValidationErrors } from './harness.js';
+import type { DocumentNode } from '../../language/ast.js';
+import { OperationTypeNode } from '../../language/ast.js';
+import { Kind } from '../../language/kinds.js';
+
+import { ScalarLeafsRule } from '../rules/ScalarLeafsRule.js';
+import { validate } from '../validate.js';
+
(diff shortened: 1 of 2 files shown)


src/validation/__tests__/ScalarLeafsRule-test.ts
```diff
@@ -35,6 +42,35 @@ describe('Validate: Scalar leafs', () => {
...
+                kind: Kind.FIELD,
+                name: { kind: Kind.NAME, value: 'human' },
+                selectionSet: { kind: Kind.SELECTION_SET, selections: [] },
+              },
+            ],
+          },
+        },
+      ],
+    };
+    const errors = validate(testSchema, doc, [ScalarLeafsRule]);
```
yaacovCR, line 65: maybe just add a comment here that we can't use expectValid like most of the rules do because the parser would throw an error?

does this situation come up anywhere else?

yaacovCR (approved):
Makes sense to match the spec as written, although in theory it would be a good discussion for the WG why the ScalarLeafsRule covers empty selection sets, presumably the empty selection set implies a leaf and that's invalid, but it's a bit of a stretch. I think for now it is the right thing to match the specification with little downside.

src/validation/__tests__/ScalarLeafsRule-test.ts
JoviDeCroock: I haven't noticed it elsewhere, added the comment!

PR #4229, opened 2024-10-13
Use canary tag

JoviDeCroock:
We should leverage the canary tag rather than also publishing it on the alpha/stable channel when publishing PR's to npm.


JoviDeCroock:
Sorry I misread this 😅

Issue #2470, opened 2020-03-02
dev: enable sourceMaps for better debugging experience


pranshuchittora:
### Description
Trying to discuss the need for sourceMaps for better debugging.

### Current behaviour
The build doesn't generate any source maps as of now (correct me if I am wrong).
Hence it's very hard to debug code using breakpoints.

### Expected behaviour
Enabling source maps to enable synchronisation of the breakpoints between the source files & build files.

### Suggested Approach
Enabling `"sourceMaps": "inline"` in the `babelrc` to enable source maps.
But the build size has been increased significantly
(Source maps enabled)
```
     *.js x135 | 1908.47 KB
    *.mjs x135 | 1868.78 KB
*.js.flow x135 |  561.75 KB
   *.d.ts x106 |  152.89 KB
     README.md |    5.36 KB
       LICENSE |    1.05 KB
  package.json |    0.61 KB
---------------------------
         Total |    4.39 MB
```
(Original build size without source maps)
```
     *.js x135 | 600.55 KB
*.js.flow x135 | 561.75 KB
    *.mjs x135 | 517.53 KB
   *.d.ts x106 | 152.89 KB
     README.md |   5.36 KB
       LICENSE |   1.05 KB
  package.json |   0.61 KB
--------------------------
         Total |   1.80 MB
```
This issue can be solved by providing an interface to create two builds i.e. `prod` (the one without source maps) & `dev` (the one with source maps)

### Future prospects
After migrating to TS https://github.com/graphql/graphql-js/issues/2104 the approach can be different.

JoviDeCroock:
In v17 we have sourcemaps enabled!

PR #4231, opened 2024-10-13
Add sourcemaps to our published bundles

JoviDeCroock:
Fixes https://github.com/graphql/graphql-js/issues/2470


yaacovCR:
Do we also need to publish the TS source for this to be useful?

JoviDeCroock:
No, sourcemaps should be able to fully standalone do these translations. There is a sourcesContent key in the sourcemap that basically encodes all that information

yaacovCR (approved):

PR #4234, opened 2024-10-14
Support symbol properties on extension objects

JoviDeCroock:
Supersedes https://github.com/graphql/graphql-js/pull/3511

This adds support for `Symbol` property keys on the extension property of AST nodes. Had to tweak `toObjMap` here to support symbol keys by adding `getOwnPropertySymbols` which is the only way to enumerate over them.

### DIFF
--- a/src/jsutils/ObjMap.ts
+++ b/src/jsutils/ObjMap.ts
@@ -1,13 +1,21 @@
 export interface ObjMap<T> {
-  [key: string | symbol]: T;
+  [key: string]: T;
 }
 
-export type ObjMapLike<T> = ObjMap<T> | { [key: string | symbol]: T };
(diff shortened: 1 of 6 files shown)


src/jsutils/ObjMap.ts
```diff
@@ -1,13 +1,13 @@
 export interface ObjMap<T> {
-  [key: string]: T;
+  [key: string | symbol]: T;
```
yaacovCR, line 2: It would be nice if we didn’t have to change this type as it’s used very often in the code base and I think we only want to support symbols on extensions

JoviDeCroock: We use `toObjMap` for extensions though, let me see what I can do

yaacovCR (approved):

PR #4235, opened 2024-10-14
Alternative approach for findSchemaChanges

JoviDeCroock:
Alternative to https://github.com/graphql/graphql-js/pull/2181

This adds a new `findSchemaChanges` export and deprecates the breaking and dangerous alternative to be removed in v18. While adding safe-schema-changes I noticed that we lack directive arg validation here, we should probably add these to breaking/dangerous changes in the future.

### DIFF
--- a/src/utilities/index.ts
+++ b/src/utilities/index.ts
@@ -104,10 +104,16 @@ export {
 export {
   BreakingChangeType,
   DangerousChangeType,
+  SafeChangeType,
   findBreakingChanges,
   findDangerousChanges,
-} from './findBreakingChanges.js';
-export type { BreakingChange, DangerousChange } from './findBreakingChanges.js';
+  findSchemaChanges,
+} from './findSchemaChanges.js';
+export type {
+  BreakingChange,
+  DangerousChange,
+  SafeChange,
+} from './findSchemaChanges.js';
 
 // Wrapper type that contains DocumentNode and types that can be deduced from it.
 export type { TypedQueryDocumentNode } from './typedQueryDocumentNode.js';
--- a/src/index.ts
+++ b/src/index.ts
@@ -472,8 +472,10 @@ export {
   // Compares two GraphQLSchemas and detects breaking changes.
   BreakingChangeType,
   DangerousChangeType,
+  SafeChangeType,
   findBreakingChanges,
   findDangerousChanges,
+  findSchemaChanges,
 } from './utilities/index.js';
 
 export type {
@@ -501,6 +503,7 @@ export type {
   IntrospectionDirective,
   BuildSchemaOptions,
   BreakingChange,
+  SafeChange,
   DangerousChange,
   TypedQueryDocumentNode,
 } from './utilities/index.js';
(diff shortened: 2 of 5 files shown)


src/utilities/index.ts
```diff
@@ -106,6 +106,7 @@ export {
   DangerousChangeType,
   findBreakingChanges,
   findDangerousChanges,
+  findSchemaChanges,
```
yaacovCR, line 109: I guess we also have to add to main index

src/utilities/index.ts
```diff
@@ -106,6 +106,7 @@ export {
   DangerousChangeType,
   findBreakingChanges,
```
yaacovCR, line 107: I think it might work better to add deprecation notices to the various index files as well — I think in some cases it doesn’t carry through automatically.

JoviDeCroock: That's not how tsdoc works though 😅 it's tied to the func def

JoviDeCroock:
Converting to draft so I can add `SAFE_SCHEMA_CHANGE` as well

src/utilities/index.ts
yaacovCR: ah it was a bug that they fixed https://github.com/microsoft/TypeScript/issues/53754 i read it recently and must have misunderstood, in theory we can remove some of my extra deprecated comments that I have scattered around, but not a priority

yaacovCR (approved):

src/utilities/findBreakingChanges.ts
```diff
@@ -103,10 +129,10 @@ export function findDangerousChanges(
   );
 }
 
-function findSchemaChanges(
+export function findSchemaChanges(
```
yaacovCR, line 132: Should we change the name of the file now or when we remove the old functions? Could be later…

src/utilities/findBreakingChanges.ts
```diff
@@ -131,6 +157,13 @@ function findDirectiveChanges(
     });
   }
 
+  for (const newDirective of directivesDiff.added) {
+    schemaChanges.push({
+      type: SafeChangeType.DIRECTIVE_ADDED,
```
yaacovCR, line 162: Can we really say this safe?

JoviDeCroock: Yes, a directive that gets added isn't neccasrily used so why would that be unsafe. Adding a directive to a field cvould be considered breaking but this never

yaacovCR (approved):

src/utilities/findBreakingChanges.ts
JoviDeCroock: I was hesitant due to it also being a breaking change but generally supportive

JoviDeCroock: Gotcha, in that case I'll add the name change

benjie:
I do not use this method and have no particular opinion on it, so (due to lack of time) I'm going to decline to review it. Perhaps you can find someone who relies on this functionality to review it - maybe The Guild could help here?

Issue #2317, opened 2019-12-30
CI: run tests in browser


IvanGoncharov:
We have some Node specific code in our codebase and it works fine but we can't fallback mechanism.
For example, we use `Error.captureStackTrace` if available but it's always available in Node so we never tested this code:
https://github.com/graphql/graphql-js/blob/576682fe691deaab79970ef103f3013508a4c587/src/error/GraphQLError.js#L203-L210

PR #4238, opened 2024-10-15
remove the enhanced error messages regarding using multiple versions of graphql-js. 

yaacovCR:
This PR removes the enhanced error messages regarding using multiple versions of graphql-js. 

Motivation:
- Gets this library out of the package management business.
- Removes our Node.JS specific code, tracked in #2317
- Removes need for complicated build instructions for package managers (helpfully added in #4022)
- Removes browser incompatibility bug, #4017 (in quite a blunt way).
- Obviates the need for #4188 to switch default to production.

Downside:
- Removes helpful error messages in tricky-to-debug cases where build tools/frameworks/consumers may end up inadvertently using multiple versions of graphql. While the fixes for this error have gotten easier now that additional package management tools have adopted the ability to set resolutions/overrides, and awareness around this issue has perhaps increased, this change will cause more opaque error messages in that hopefully dwindling number of cases.

This PR should be thought of a strawman PR for discussion.

If we do want to re-enter the package management business to some extent, we could combine this with #3915 => although that PR would only protect us going forward with versions that will support `Symbol.for()`.


yaacovCR:
Closing in favor of:

- #4464 

on background of:

- #4448

Issue #4200, opened 2024-09-23
graphql-js.org is only partially deprecated


yaacovCR:
It looks like graphql-js.org has been deprecated in favor of graphql.org/graphql-js. Our README even points to graphql.org/graphql-js.

So we really should:
1. add the latest additional content from graphql-js.org to graphql.org/graphql-js (see https://github.com/graphql/graphql.github.io/pull/1782)
2. redirect from graphql-js.org to graphql.org/graphql-js or point graphql-js.org to graphql.org/graphql-js
3. remove the old website from the graphql-js repo.

I am not in a position to do 2, but [have submitted a PR doing 1](https://github.com/graphql/graphql.github.io/pull/1782), and once 2 is done, I can do 3.

PR #4240, opened 2024-10-15
Convert from docusaurus to nextra

JoviDeCroock:
This converts the existing docusaurus website to nextra just like https://github.com/graphql/graphql.github.io. This is a first step in moving the documentation here and having a redirect from graphql.org to graphql-js.org.

Not sure yet why codecov started failing 😅 when I run `testonly:cover` locally it tells me we are 100% covered.

WDYT about isolating the dependencies for the website in that folder? As seen in https://github.com/graphql/graphql-js/pull/4240/commits/9c7d615bfe9669c01081156e32bc25c7b24f795c this prevents the weird CI leaks that we're seeing

Resolves https://github.com/graphql/graphql-js/issues/4200


yaacovCR:
> WDYT about isolating the dependencies for the website in that folder? As seen in [9c7d615](https://github.com/graphql/graphql-js/commit/9c7d615bfe9669c01081156e32bc25c7b24f795c) this prevents the weird CI leaks that we're seeing

Sounds fine, ideally, we wouldn’t need to do that but we can always try to fix it later

.github/workflows/ci.yml
```diff
@@ -247,16 +246,10 @@ jobs:
         uses: actions/setup-node@v2
         with:
           cache: npm
-          node-version-file: '.node-version'
+          node-version: 18
```
yaacovCR, line 249: I assume we need to pin this because we are still using 17 on v16.x.x and something is broken there? Maybe we could add a comment to eventually change that...

.github/workflows/ci.yml
```diff
@@ -219,7 +219,6 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v2
         with:
```
yaacovCR, line 221: Do we have to drop the cache?

yaacovCR:
1. I get the following error when running `npm run build` locally, not sure if it is an error on my side. `npm run dev` works.

image

2. Just a note that this PR does not (yet) have the extra context at https://graphql-js.org missing at https://graphql.org/graphql-js => https://github.com/graphql/graphql.github.io/pull/1782 => presumably that is for a later PR

3. Existing graphql-js.org version of website has a typedoc view of the API (https://graphql-js.org/api/) => just noting that to open up discussion for whether we will be doing something similar

JoviDeCroock:
@yaacovCR I think I found what could have been causing the issue for you 😅 I purposefully didn't add the new content yet to make the PR easier to review, currently it's basically a copy of the content within graphql.org and moving the website to its own lockfile. Regarding typedoc, I am personally not a big fan as TypeScript provides more value here.

.github/workflows/ci.yml
JoviDeCroock: Next does not work with anything but v18 or higher

package.json
```diff
@@ -79,15 +71,10 @@
     "eslint-plugin-react-hooks": "4.4.0",
     "eslint-plugin-simple-import-sort": "7.0.0",
     "eslint-plugin-tsdoc": "0.2.16",
-    "file-loader": "6.2.0",
     "mocha": "9.2.2",
     "prettier": "2.6.2",
-    "prism-react-renderer": "1.3.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
     "typedoc": "0.22.15",
```
yaacovCR, line 76: can we get rid of this one as well, then?

.github/workflows/ci.yml
yaacovCR:
```suggestion
          # TODO: switch back in v17
          # node-version-file: '.node-version'
          node-version: 18
```

yaacovCR (approved):

website/theme.config.tsx
```diff
@@ -0,0 +1,232 @@
...
+  navbar: {
+    extraContent: <ThemeSwitch lite className="[&_span]:hidden" />,
+  },
+  toc: {
+    backToTop: true,
+  },
+  search: {
+    placeholder: 'Search…',
+  },
+} as DocsThemeConfig;
```
dimaMachina, line 232:
```suggestion
} satisfies DocsThemeConfig;
```

JoviDeCroock: Not supported in TS version

website/package.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "scripts": {
+    "build": "next build",
+    "dev": "next"
+  },
+  "devDependencies": {
+    "@svgr/webpack": "^8.1.0",
+    "@types/node": "^22.7.5",
+    "next": "^14.2.15",
+    "next-with-less": "^3.0.1",
```
dimaMachina, line 17: I think less is only used in graphql.org for mini graphiql
```suggestion
```

website/next.config.mjs
```diff
@@ -0,0 +1,59 @@
+import nextra from 'nextra';
+import path from 'node:path';
+import withLess from 'next-with-less';
```
dimaMachina, line 3:
```suggestion
```
https://github.com/graphql/graphql-js/pull/4240/files#r1806244986

website/next.config.mjs
```diff
@@ -0,0 +1,59 @@
...
+
+const withNextra = nextra({
+  theme: 'nextra-theme-docs',
+  themeConfig: './theme.config.tsx',
+  mdxOptions: {
+    remarkPlugins: [],
+  },
```
dimaMachina, line 10:
```suggestion
```

website/next.config.mjs
```diff
@@ -0,0 +1,59 @@
...
+const sep = path.sep === '/' ? '/' : '\\\\';
+
+const ALLOWED_SVG_REGEX = new RegExp(`${sep}icons${sep}.+\\.svg$`);
+
+/**
+ * @type {import('next').NextConfig}
+ */
+export default withLess(
+  withNextra({
+    // reactStrictMode: true, provoke duplicated codemirror editors
```
dimaMachina, line 22:
```suggestion
```

.gitignore
```diff
@@ -10,7 +10,10 @@
 /.cspellcache
 /.docusaurus
 /node_modules
+/website/node_modules
```
dimaMachina, line 13:
```suggestion
node_modules/
```

website/pages/graphql-clients.mdx
```diff
@@ -0,0 +1,85 @@
...
+  rollDice(numDice: Int!, numSides: Int): [Int]
+}
+```
+
+You could access this from JavaScript with the code:
+
+```js
+var dice = 3;
+var sides = 6;
+var query = /* GraphQL */ `
```
dimaMachina, line 60:
```suggestion
let dice = 3;
let sides = 6;
let query = /* GraphQL */ `
```

website/pages/index.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+// Construct a schema, using GraphQL schema language
+var schema = buildSchema(`
+  type Query {
+    hello: String
+  }
+`);
+
+// The rootValue provides a resolver function for each API endpoint
+var rootValue = {
```
dimaMachina, line 36:
```suggestion
let { graphql, buildSchema } = require('graphql');

// Construct a schema, using GraphQL schema language
let schema = buildSchema(`
  type Query {
    hello: String
  }
`);

// The rootValue provides a resolver function for each API endpoint
let rootValue = {
```

website/theme.config.tsx
```diff
@@ -0,0 +1,232 @@
...
+        >
+          Check out the recorded talks and workshops
+        </NextLink>
+      </>
+    ),
+    key: 'graphqlconf-2024',
+  },
+  logo: graphQLLogo,
+  docsRepositoryBase:
+    'https://github.com/graphql/graphql.github.io/tree/source',
```
dimaMachina, line 213:
```suggestion
    'https://github.com/graphql/graphql-js/tree/16.x.x/website',
```

or

```suggestion
    'https://github.com/graphql/graphql-js/tree/main/website',
```

so Edit On GitHub link will point to right file

website/theme.config.tsx
dimaMachina: maybe in this case refactor `const config: DocsThemeConfig ` and later `export default config` so it will be typesafe config

website/pages/object-types.mdx
```diff
@@ -0,0 +1,142 @@
...
+    roll(numRolls: 3)
+  }
+}
+```
+
+If you run this code with `node server.js` and browse to http://localhost:4000/graphql you can try out these APIs with [GraphiQL](https://github.com/graphql/graphiql).
+
+This way of defining object types often provides advantages over a traditional REST API. Instead of doing one API request to get basic information about an object, and then multiple subsequent API requests to find out more information about that object, you can get all of that information in one API request. That saves bandwidth, makes your app run faster, and simplifies your client-side logic.
+
+So far, every API we've looked at is designed for returning data. In order to modify stored data or handle complex input, it helps to [learn about mutations and input types](/graphql-js/mutations-and-input-types/).
```
dimaMachina, line 142: I think all internal links need to be verified, here will point to incorrect, I guess?

PR #4250, opened 2024-10-20
AbortController support

JoviDeCroock:
This adds support for aborting execution from the outside or resolvers, this adds a few tests and tries to make the support as easy as possible.

Do we want to support having abort support on subscriptions, I guess it makes sense for server-sent events.

I've chosen 2 places to place these interrupts

- `executeFieldsSerially` - every time we start a new mutation we check whether the runtime has interrupted
- `executeFields` - every time we start executing a new field we check whether the runtime has interrupted
  - inside of the catch block as well so we return a singular error, all though this doesn't really matter as the consumer would not receive anything 
  - this here should also take care of deferred fields

When comparing this to `graphql-tools/execute` I am not sure whether we want to match this behavior, this throws a DomException which would be a whole new exception that gets thrown while normally during execution we wrap everything with GraphQLErrors.

Supersedes https://github.com/graphql/graphql-js/pull/3791
Resolves https://github.com/graphql/graphql-js/issues/3764

### DIFF
--- a/integrationTests/ts/tsconfig.json
+++ b/integrationTests/ts/tsconfig.json
@@ -1,7 +1,13 @@
 {
   "compilerOptions": {
     "module": "commonjs",
-    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string"],
+    "lib": [
+      "es2019",
+      "es2020.promise",
+      "es2020.bigint",
+      "es2020.string",
+      "DOM"
+    ],
     "noEmit": true,
     "types": [],
     "strict": true,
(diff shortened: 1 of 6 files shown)


ardatan:
> When comparing this to graphql-tools/execute I am not sure whether we want to match this behavior, this throws a DomException which would be a whole new exception that gets thrown while normally during execution we wrap everything with GraphQLErrors.

In `fetch`, when the request is aborted, it throws `AbortError` as `DOMException`.
I think it uses [`abortSignal.throwIfAborted`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted), so the implementation decides on what to throw.

Issue #4253, opened 2024-10-24
Enums not tree shaking well


jasonkuhrt:
I have redefined the `Kind` and `OperationTypeNode` enums in Graffle because they do not seem to tree shake well. You can see the code and bundle visualization [in this PR](https://github.com/graffle-js/graffle/pull/1215). With my approach already just those two enums are reduced by over 4kb in Graffle.

![CleanShot 2024-10-24 at 16 14 07@2x](https://github.com/user-attachments/assets/e19e056e-5e79-4efa-8c35-fcb527e2a3e4)

Obviously it would be nice if `graphql` was already fully tree shake optimized from the start. Is there an appetite to adjust how the enums are defined in `graphql`? I think my approach would be a breaking change at the type level because TypeScript enum types, even if using string constants at runtime, are not equivalent to them at the type level.

Feel free to close this issue if there's zero chance of change on the `graphql` package side of course. Convesely, happy to help out on this issue.

JoviDeCroock:
Yeah, there are several issues with the TS enum like https://github.com/graphql/graphql-js/issues/3356. However converting to an object won't solve your issue here, that being said, this is a reference implementation and we have made tradeoffs for DX/readability for that matter. The minzipped size of this object when not using TS enums should not exceed 1kb.

We could split this up between schema-kind and executable kind to improve this a bit more for web implementations.

jasonkuhrt:
Right an object wouldn't solve tree shaking. Rather I defined every enum member as its own constant and then use the namespace to group them. That supports tree shaking and I see actually no DX issue here, since TS allows type names that match a namespace name so it ends up feeling identical to the enum before.

JoviDeCroock:
Grouping them under a namespace won't allow them to be used in JS as follows right?

```js
import { Kind } from 'graphql';

if (selection.kind === Kind.FIELD)
```

My main concern would be to still allow this usage so v17 doesn't have too many breaking changes, in #4254 we move away from the `enum` notation which already saves 20% of the bytes in that file. I consider that to be a good start as it also allows for the type to be omitted and to just use the plain string, which in your case would help you get rid of the type-casting.

I would be a huge fan of evolving into a tree-shakable model though, for now I would advise people that are concerned with that to leverage your approach.

jasonkuhrt:
Here's another approach that might seem/be simpler:

[...]

PR #4254, opened 2024-10-25
Remove typescript enums in favor of plain JS/TS solution

JoviDeCroock:
This gets rid of enums for two reasons

- TypeScript does not really compile them size efficiently, see [here](https://www.runpkg.com/?graphql@16.8.1/language/kinds.mjs)
  - Raw size kinds.js before: 2800bytes
  - Raw size kinds.js after: 2200 bytes
- It's not kind to plains strings as seen in https://github.com/graphql/graphql-js/issues/3356

Resolves https://github.com/graphql/graphql-js/issues/3356

This does not intend to fix the tree-shaking issue as seen in https://github.com/graphql/graphql-js/issues/4253, for that we'd need to fully normalize these exports and sacrifice the DX of the namespaces. I do think that it becomes easier as plain strings will be "easier" compared to before so if you're not doing comparisons you can opt out of the `Kind` namespace.

<details>
  <summary>See new output</summary>

```js
/**
 * The set of allowed kind values for AST nodes.
 */
exports.Kind = {
    /** Name */
    NAME: 'Name',
    /** Document */
    DOCUMENT: 'Document',
    OPERATION_DEFINITION: 'OperationDefinition',
    VARIABLE_DEFINITION: 'VariableDefinition',
    SELECTION_SET: 'SelectionSet',
    FIELD: 'Field',
    ARGUMENT: 'Argument',
    FRAGMENT_ARGUMENT: 'FragmentArgument',
    /** Fragments */
    FRAGMENT_SPREAD: 'FragmentSpread',
    INLINE_FRAGMENT: 'InlineFragment',
    FRAGMENT_DEFINITION: 'FragmentDefinition',
    /** Values */
    VARIABLE: 'Variable',
    INT: 'IntValue',
    FLOAT: 'FloatValue',
    STRING: 'StringValue',
    BOOLEAN: 'BooleanValue',
    NULL: 'NullValue',
    ENUM: 'EnumValue',
    LIST: 'ListValue',
    OBJECT: 'ObjectValue',
    OBJECT_FIELD: 'ObjectField',
    /** Directives */
    DIRECTIVE: 'Directive',
    /** Types */
    NAMED_TYPE: 'NamedType',
    LIST_TYPE: 'ListType',
    NON_NULL_TYPE: 'NonNullType',
    /** Type System Definitions */
    SCHEMA_DEFINITION: 'SchemaDefinition',
    OPERATION_TYPE_DEFINITION: 'OperationTypeDefinition',
    /** Type Definitions */
    SCALAR_TYPE_DEFINITION: 'ScalarTypeDefinition',
    OBJECT_TYPE_DEFINITION: 'ObjectTypeDefinition',
    FIELD_DEFINITION: 'FieldDefinition',
    INPUT_VALUE_DEFINITION: 'InputValueDefinition',
    INTERFACE_TYPE_DEFINITION: 'InterfaceTypeDefinition',
    UNION_TYPE_DEFINITION: 'UnionTypeDefinition',
    ENUM_TYPE_DEFINITION: 'EnumTypeDefinition',
    ENUM_VALUE_DEFINITION: 'EnumValueDefinition',
    INPUT_OBJECT_TYPE_DEFINITION: 'InputObjectTypeDefinition',
    /** Directive Definitions */
    DIRECTIVE_DEFINITION: 'DirectiveDefinition',
    /** Type System Extensions */
    SCHEMA_EXTENSION: 'SchemaExtension',
    /** Type Extensions */
    SCALAR_TYPE_EXTENSION: 'ScalarTypeExtension',
    OBJECT_TYPE_EXTENSION: 'ObjectTypeExtension',
    INTERFACE_TYPE_EXTENSION: 'InterfaceTypeExtension',
    UNION_TYPE_EXTENSION: 'UnionTypeExtension',
    ENUM_TYPE_EXTENSION: 'EnumTypeExtension',
    INPUT_OBJECT_TYPE_EXTENSION: 'InputObjectTypeExtension',
};
```

</details>

Currently seeing whether we should disable the redeclare lint-rule and whether there is a better way to type our `nodeKinds`


src/language/tokenKind.ts
```diff
@@ -2,29 +2,29 @@
...
+  BRACE_R: '}' as const,
+  NAME: 'Name' as const,
+  INT: 'Int' as const,
+  FLOAT: 'Float' as const,
+  STRING: 'String' as const,
+  BLOCK_STRING: 'BlockString' as const,
+  COMMENT: 'Comment' as const,
+} as const;
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type TokenKind = (typeof TokenKind)[keyof typeof TokenKind];
```
yaacovCR, line 30: I think it might be a bit confusing to have the type and the raw value be the same,

if somebody in their code wants to import both in the same file, I think the identifiers clash.

same for the others.

@benjie had a suggestion: https://github.com/graphql/graphql-js/pull/3360#issuecomment-963053990

yaacovCR (changes requested):

src/language/tokenKind.ts
yaacovCR: x however many "enums" we have

JoviDeCroock: That's an opinion though 😅 I personally think this is easier to read

yaacovCR: is it? i am getting an error in my editor when i have the following:

```ts
import type { A } from 'somewhere.js' 
import { A } from 'somewhere.js' 
```

although there is no error with the following:

```ts
export type { A } from 'somewhere.js' 
export { A } from 'somewhere.js' 
```

for our users, they will be using the former case.

I could be wrong here, I don't think I've encountered this before.

JoviDeCroock: Because you only need `import { A } from 'somewhere.js' ` as that is both a type as well as a value.

JoviDeCroock: Anyway my preference for this approach was 'least amount of breaking changes', if we can't agree on this approach then I welcome any other one.

yaacovCR: > Because you only need `import { A } from 'somewhere.js' ` as that is both a type as well as a value.

Ah! I see, that should be fine then.

yaacovCR: @jasonkuhrt thanks for joining in, I am following along also with interest at #4253 and #4268

benjie:
I support removing TS enums in v17 :+1: I am not a fan of them.

PR #4258, opened 2024-10-27
fix: handle default and different types passed to abort method

yaacovCR:


src/execution/execute.ts
```diff
@@ -518,7 +518,7 @@ export function validateExecutionArgs(
   } = args;
 
   if (abortSignal?.aborted) {
-    return [locatedError(new Error(abortSignal.reason), undefined)];
+    return [locatedError(abortSignal.reason, undefined)];
```
JoviDeCroock, line 521: I would avoid falling in our unexpected error value bucket, that feels like a footgun

yaacovCR: Way back when, I had a discussion with Ivan about the entire toError() conversion where I suggested we not wrap non-Error thrown values at all, but just change the type of the originalError property, see the discussion at: 

- https://github.com/graphql/graphql-js/pull/3379

So I do agree that the whole wrapping this is a bit odd.

The problem is that as written, the code wraps the default provided `AbortController.abort()` instead of just passing it through.

JoviDeCroock: We could check whether it's an instanceof Error but technically reason should always be a plain value

yaacovCR: I added a commit to make our toError wrapping more generic: 8c3f805d1 => does this address your concern? If this seems beneficial, this could be merged as part of this PR or separately either before or after.

If this doesn't address your concern, maybe you could elaborate what footgun you are seeing? The solution in this PR has the "originalError" property within the GraphQLError we provided passing through any actual `Error` provided to `AbortController.abort(reason)` and wrapping only non-error reasons, just like the rest of our field errors, but the existing code wraps even Error reasons, and in a way that does not preserve the original reason.

yaacovCR: > technically reason should always be a plain value

> If not specified, the reason is set to "AbortError" [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException).

yaacovCR: > technically reason should always be a plain value

[From the docs](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) for the `abort()` method:

> If not specified, the reason is set to "AbortError" [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException).

Where DOMException is indeed an instance of Error, see https://webidl.spec.whatwg.org/#js-DOMException-specialness

I added another commit, showing within tests how with this PR, we will in the default case not wrap the abort reason such that the `originalError` property on the returned GraphQLError within our errors array will be the actual AbortError/DomException or custom Error that is passed to `abort()`.

src/execution/__tests__/abort-signal-test.ts
```diff
@@ -99,7 +196,7 @@ describe('Execute: Cancellation', () => {
       },
       errors: [
         {
-          message: 'Aborted',
+          message: 'Encountered error: "Custom abort error message"',
```
JoviDeCroock, line 199: Not sure I like this Encountered error prefix but 🤷‍♂️ at this point

JoviDeCroock (approved):

PR #4259, opened 2024-10-27
fix: abortion should not cause incremental completion error

yaacovCR:
incremental completion errors are indicated if and only if null bubbling has caused a previously sent response to be null

we must instead just give a partial response with errors within the incremental array

The GraphQL Tools implementation of this feature has different behavior where the next call will throw if the operation is aborted, but we have opted to give partial responses, and so we must adjust.

### DIFF
--- a/src/execution/__tests__/abort-signal-test.ts
+++ b/src/execution/__tests__/abort-signal-test.ts
@@ -100,8 +100,8 @@ describe('Execute: Cancellation', () => {
       errors: [
         {
           message: 'Aborted',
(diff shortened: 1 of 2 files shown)


yaacovCR:
@JoviDeCroock this was my oversight when reviewing, compounded by another mistake I made in that I allowed a discrepancy between the path of the error and the null site in the response, which should only occur in a situation with null-bubbling. This PR also fixes that.

JoviDeCroock (approved):

PR #4260, opened 2024-10-27
adjust prefix for toError

yaacovCR:
should encompass thrown errors and encountered abort errors

### DIFF
--- a/src/error/__tests__/locatedError-test.ts
+++ b/src/error/__tests__/locatedError-test.ts
@@ -17,8 +17,8 @@ describe('locatedError', () => {
 
     expect(error).to.be.instanceOf(GraphQLError);
     expect(error.originalError).to.include({
-      name: 'NonErrorThrown',
-      thrownValue: testObject,
+      name: 'WrappedNonErrorValueError',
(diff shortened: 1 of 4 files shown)


yaacovCR:
This PR emerged from #4258 => but note that comment at https://github.com/graphql/graphql-js/pull/4258#discussion_r1818122057 by @JoviDeCroock notes remaining reservation with respect to the modified prefix.

I have left this PR in draft, because considering the above mentioned reservation, I am not sure if we consider it better than the status quo.

An argument could be made that the status quo is sufficient: although the AbortError is not technically thrown, depending on the implementation, it could have been, as we could have implemented it as a thrown error simply by using `abortSignal.throwIfAborted()` => we chose instead to directly access `abortSignal.aborted` and `abortSignal.reason` directly, for performance reasons on some runtimes, but semantically, the abortSignal is a "throwable" value.

This PR is also another breaking change, technically, although hopefully it doesn't hit too many people, as since I do not like our current state of error wrapping at all, see discussion at
- #3379

yaacovCR:
Closing this for now, we can always reopen if desired.

PR #4263, opened 2024-10-28
stop resolvers after execution ends

yaacovCR:
depends on #4267

addresses: #3792


yaacovCR:
our test show that nested completion of object fields terminates => we also want to also show that the abortSignal is triggered

JoviDeCroock (approved):

PR #4264, opened 2024-10-29
introduce completeMaybePromisedListItemValue

yaacovCR:
depends on #4266

- unifies some logic between completing sync and async iterables
- gets rid of a few `c8 ignore` directives

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -1269,7 +1269,6 @@ async function completeAsyncIteratorValue(
         break;
       }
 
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

yaacovCR:
I have changed the base to #4266

This PR on its own seemed to lead to a drop in performance only in the case of async iterators, ad does #4266, but when combined, there is no drop in performance.

image

yaacovCR:
was able to remove this from the #4267 stack

PR #4265, opened 2024-10-29
use object for GraphQLWrappedResult instead of tuple

yaacovCR:
for readability. benchmarks suggest similar or even better performance

image


JoviDeCroock (approved):
The reason for this performance improvement is that multi-value-shape tuples and especially holey tuples are a deopt.

- First deopt --> the Array is not mono-value i.e. we can have T and nested arrays as well as an explicit undefined
- Second deopt --> the second array can be an explicit undefined (a hole)

PR #4266, opened 2024-10-29
move errors to GraphQLWrappedError rather than context

yaacovCR:
- preserves order of errors with respect to operation
- bubbling errors mask the non-bubbling errors such that extra information does not leak

### DIFF
--- a/src/execution/__tests__/executor-test.ts
+++ b/src/execution/__tests__/executor-test.ts
@@ -680,14 +680,10 @@ describe('Execute: Handles basic execution tasks', () => {
 
     const result = execute({ schema, document });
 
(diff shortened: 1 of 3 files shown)


yaacovCR:
oddly:

image

yaacovCR:
I have rebased #4264 on top of this PR => both PRs seem individually to lead to a drop in performance for async iterator lists, but when combined, the drop disappears. @JoviDeCroock -- my thought is that the function is very sensitive, and so the benchmark in this case may not be reliable, but that overall if we can make the numbers work, we should be able to merge both PRs, and maybe we should open a separate issue to investigate?

yaacovCR:
was able to remove this from the #4267 stack

PR #4267, opened 2024-10-29
cancel execution despite hanging async resolvers

yaacovCR:
Prior to this pull request, cancellation works by checking the abort signal status during execution, and throwing the reason if the abort signal has been triggered. This fails if an asynchronous resolver hangs.
This pull request changes the cancellation method to wrap promises returned by resolvers so that they immediately reject on cancellation.

### DIFF
--- a/src/execution/__tests__/abort-signal-test.ts
+++ b/src/execution/__tests__/abort-signal-test.ts
@@ -45,7 +45,7 @@ async function complete(
 const schema = buildSchema(`
   type Todo {
     id: ID
-    text: String
+    items: [String]
     author: User
   }
 
@@ -91,7 +91,6 @@ describe('Execute: Cancellation', () => {
         todo: async () =>
           Promise.resolve({
             id: '1',
-            text: 'Hello, World!',
             /* c8 ignore next */
             author: () => expect.fail('Should not be called'),
           }),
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

src/execution/Canceller.ts
```diff
@@ -0,0 +1,48 @@
+import { promiseWithResolvers } from '../jsutils/promiseWithResolvers.js';
+
+/**
+ * A Canceller object that can be used to cancel multiple promises
+ * using a single AbortSignal.
+ *
+ * @internal
+ */
+export class Canceller {
```
JoviDeCroock, line 9: NIT: kind of dislike the naming, it's basically a promiseTracker class

yaacovCR: Renamed to PromiseCanceller….. any better?

yaacovCR: Actually, considering that the next step is to hook up early async iterator return to the abort signal, I think PromiseTracker/Canceller won't be accurate ... naming things so hard ...

What this class really does is allow us the ability to set lots of contingent things to be triggered by the AbortSignal without ever hitting a max event listener warning.

I'm going to merge it as we have it for now, but be very open to additional changes. :)

PR #4268, opened 2024-10-29
Allow for tree-shaking the Kind enum

### LINKED ISSUE #4253: Enums not tree shaking well

JoviDeCroock:
Resolves #4253 

This enables tree-shaking `Kind` - I opted to only do this with Kind as the other enums are either tiny or included by default if you are using `parseSchema`.

### DIFF
--- a/src/language/_kinds.ts
+++ b/src/language/_kinds.ts
@@ -1,0 +1,65 @@
+/** Name */
+export const NAME = 'Name';
+
+/** Document */
+export const DOCUMENT = 'Document';
+export const OPERATION_DEFINITION = 'OperationDefinition';
+export const VARIABLE_DEFINITION = 'VariableDefinition';
+export const SELECTION_SET = 'SelectionSet';
+export const FIELD = 'Field';
+export const ARGUMENT = 'Argument';
+export const FRAGMENT_ARGUMENT = 'FragmentArgument';
+
+/** Fragments */
+export const FRAGMENT_SPREAD = 'FragmentSpread';
+export const INLINE_FRAGMENT = 'InlineFragment';
+export const FRAGMENT_DEFINITION = 'FragmentDefinition';
+
+/** Values */
+export const VARIABLE = 'Variable';
+export const INT = 'IntValue';
+export const FLOAT = 'FloatValue';
+export const STRING = 'StringValue';
+export const BOOLEAN = 'BooleanValue';
+export const NULL = 'NullValue';
+export const ENUM = 'EnumValue';
+export const LIST = 'ListValue';
+export const OBJECT = 'ObjectValue';
+export const OBJECT_FIELD = 'ObjectField';
+
+/** Directives */
+export const DIRECTIVE = 'Directive';
+
+/** Types */
+export const NAMED_TYPE = 'NamedType';
+export const LIST_TYPE = 'ListType';
+export const NON_NULL_TYPE = 'NonNullType';
+
+/** Type System Definitions */
+export const SCHEMA_DEFINITION = 'SchemaDefinition';
+export const OPERATION_TYPE_DEFINITION = 'OperationTypeDefinition';
+
+/** Type Definitions */
+export const SCALAR_TYPE_DEFINITION = 'ScalarTypeDefinition';
+export const OBJECT_TYPE_DEFINITION = 'ObjectTypeDefinition';
+export const FIELD_DEFINITION = 'FieldDefinition';
+export const INPUT_VALUE_DEFINITION = 'InputValueDefinition';
+export const INTERFACE_TYPE_DEFINITION = 'InterfaceTypeDefinition';
+export const UNION_TYPE_DEFINITION = 'UnionTypeDefinition';
+export const ENUM_TYPE_DEFINITION = 'EnumTypeDefinition';
+export const ENUM_VALUE_DEFINITION = 'EnumValueDefinition';
+export const INPUT_OBJECT_TYPE_DEFINITION = 'InputObjectTypeDefinition';
+
+/** Directive Definitions */
+export const DIRECTIVE_DEFINITION = 'DirectiveDefinition';
+
+/** Type System Extensions */
+export const SCHEMA_EXTENSION = 'SchemaExtension';
+
+/** Type Extensions */
+export const SCALAR_TYPE_EXTENSION = 'ScalarTypeExtension';
+export const OBJECT_TYPE_EXTENSION = 'ObjectTypeExtension';
+export const INTERFACE_TYPE_EXTENSION = 'InterfaceTypeExtension';
+export const UNION_TYPE_EXTENSION = 'UnionTypeExtension';
+export const ENUM_TYPE_EXTENSION = 'EnumTypeExtension';
+export const INPUT_OBJECT_TYPE_EXTENSION = 'InputObjectTypeExtension';
--- a/src/language/kinds.ts
+++ b/src/language/kinds.ts
@@ -45,7 +45,10 @@ import type {
   VARIABLE_DEFINITION,
 } from './_kinds.js';
 
-export type Kind =
+/**
+ * The set of allowed kind values for AST nodes.
+ */
+export type Kind = 
   | typeof NAME
   | typeof DOCUMENT
   | typeof OPERATION_DEFINITION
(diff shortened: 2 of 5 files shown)


jasonkuhrt:
I closed my duplicate here https://github.com/graphql/graphql-js/pull/4270.

src/language/_kinds.ts
jasonkuhrt, line 1: A name of `kinds_` would keep it co-located with the other module that imports from this fwiw. Not sure what's consistent for this repo.

src/language/kinds.ts
jasonkuhrt, line 1: Another approach that works I think is:

```ts
import * as kinds_ from './kinds_.js'; // eslint-disable-line

 export * as Kind from './kinds_.js';

 /**
  * The set of allowed kind values for AST nodes.
  */
 export type Kind = (typeof kinds_)[keyof typeof kinds_];
```

src/language/kinds.ts
```diff
@@ -1,72 +1,97 @@
...
+  SCHEMA_EXTENSION,
+  SELECTION_SET,
+  STRING,
+  UNION_TYPE_DEFINITION,
+  UNION_TYPE_EXTENSION,
+  VARIABLE,
+  VARIABLE_DEFINITION,
+} from './_kinds.js';
+
+export type Kind =
```
jasonkuhrt, line 48:
```suggestion
/**
 * The set of allowed kind values for AST nodes.
 */
export type Kind = 
```

This will have the desired effect of having JSDoc on both the type and term level.

JoviDeCroock:
@jasonkuhrt looks like not all TS versions support this

jasonkuhrt:
@JoviDeCroock nice catch, yeah https://www.typescriptlang.org/dev/bug-workbench/?ts=4.8.4#code/FAj0oAgAQMwSwDYFMB2BDAtkgXBFB7AEyQH0MiBXZAZ1AHMAnNABwAsBHBUAazhUJ59CAOgAu1YEgAezfA1EQAxvhTUFAMQCSAUQAyAEQgBeCAHJ4SBIVOSZchctUKAQsbMAjUxHAQkoxcK2svJKKmoQAMJupopePn4BQfahThCGJqbW3pAJgUkhogCezEgQWnrpEEUl+DBlOgb5CtWlriYttRDOTVXFpVHtfZ0RPS1pbh11+iBgkLCIqJg4eESk5IRUSLSMLBxcvPygJCRiEnAYwc19EABUEGjUEADSQiQQMAz4GGbCgvzCACtqDZRtcXvwJkM6uDCCQANrcJCFTqTZ6vCAAMggagYfDoAF0QNJLrd7o8Ye9Pt9TL8DiIgSDZtB4Mh0FhcARiGRKDR6Ew2JxQEJpKcencPl8fn8BMdAcCZsAfPNWUtcCxmKLziSAN5oiEAX0pktMOwFCBBjnCMHw+DcAAo6bgYQBKYwAPgg2uAEB9EDgdQdQmMRgyFispldDD8FAYKGyEEItuoXz8rDxvogwH1IBhwnKBnjACIw4RC0yoOIALTEpCKUTVhifBjAa34O2mdxoayunwwNCIECt9sliPx-DcFs29ueHuQceTtsxUc+edDzLLufcIA

Can we drop support for 4.8? Does GraphQL to date support even older versions of TypeScript?

Urigo:
maybe v17 is a good opportunity to drop support of old Typescript and Node versions?

JoviDeCroock:
I think it's fair to cut https://github.com/graphql/graphql-js/blob/main/integrationTests/ts/package.json#L11-L22 down to i.e. 4.9 - recent but that might introduce some issues during migration - @yaacovCR has also expressed their concerns regarding ghosted exports which this also does so we might need to settle that first https://github.com/graphql/graphql-js/pull/4254#discussion_r1821301808

src/language/kinds.ts
```diff
@@ -1,72 +1,97 @@
...
+  | typeof ENUM_VALUE_DEFINITION
+  | typeof INPUT_OBJECT_TYPE_DEFINITION
+  | typeof DIRECTIVE_DEFINITION
+  | typeof SCHEMA_EXTENSION
+  | typeof SCALAR_TYPE_EXTENSION
+  | typeof OBJECT_TYPE_EXTENSION
+  | typeof INTERFACE_TYPE_EXTENSION
+  | typeof UNION_TYPE_EXTENSION
+  | typeof ENUM_TYPE_EXTENSION
+  | typeof INPUT_OBJECT_TYPE_EXTENSION;
```
jasonkuhrt, line 92:
```suggestion
  | NAME
  | DOCUMENT
  | ...
```

src/language/_kinds.ts
```diff
@@ -0,0 +1,65 @@
+/** Name */
+export const NAME = 'Name';
+
+/** Document */
+export const DOCUMENT = 'Document';
+export const OPERATION_DEFINITION = 'OperationDefinition';
+export const VARIABLE_DEFINITION = 'VariableDefinition';
```
jasonkuhrt, line 7:
```suggestion
export const DOCUMENT = 'Document';
export type DOCUMENT = typeof DOCUMENT
export const OPERATION_DEFINITION = 'OperationDefinition';
export type OPERATION_DEFINITION = typeof OPERATION_DEFINITION
export const VARIABLE_DEFINITION = 'VariableDefinition';
// ...
```

jasonkuhrt (changes requested):
While writing this message https://github.com/graphql/graphql-js/pull/4254#discussion_r1822576683 I realized that we have an opportunity at the type level to improve the enum DX further.

_Sometimes_ it is useful to be able to access the enum at the type level just like it would be at the term level e.g. `Kind.FIELD // type level!`.

As posted in that referenced message, here is a link demonstrating how to achieve that: https://www.typescriptlang.org/dev/bug-workbench/?#code/PTAEAEDMEsBsFMB2BDAtvAXKA1tRATAfQDoAXAZwCh4APABwHsAnU0AYwcXNeVAF5QARmr1mrUgE868ULwHCRjFu07dQAI36gATIrGhJ0jVt2UQEGAhTosuAmSq0lrAFSzyoANJ58oSEwZUUAByYmA7ImIAK3JgymhUZwMpGQBvWQAaYwBfPwCg0PCfEhi4vWVDGW8CfkpQAB9ZOsb1SjbzKDgkNEwDAHcGB3jE-XTq31z-QJCwiOjYto4uVnxocjpkUjYACy0ACgiMcYBKfgA+HB9F1R4tceJkSkrZO58H6+XjAXvW581vt7qIA

The difference is that:

1. We define each enum member type next to its term `const` variable.
2. When we define the `Kind` _type_ we do so by importing the _types_ from the module defining the members.

AFAICT this suggestion has no downside. Meanwhile its upside is increased symmetry between the term and type level and easier type level logic when type level access to specific enums are desired.

src/language/_kinds.ts
JoviDeCroock: The reason I did not do these is mainly that it would cause a huge pain with our eslint rule no-redeclare

jasonkuhrt: Would aslant complain about that? If so though, could we just ignore with an ignore directive at the top of the file?

JoviDeCroock:
Going to close this as I won't have time to fine tune this. If someone wants to take this over feel free to, I might get back to it in the future.

jasonkuhrt:
@JoviDeCroock Sure thing I'll take a swing, thanks for clearing the way up to this point!

jasonkuhrt:
@JoviDeCroock @Urigo I've got something ready for review here https://github.com/graphql/graphql-js/pull/4270.

PR #4270, opened 2024-10-29
refactor: treeshakable kind enum

jasonkuhrt:
closes https://github.com/graphql/graphql-js/issues/4253


jasonkuhrt:
Duplicate of https://github.com/graphql/graphql-js/pull/4268.

jasonkuhrt:
@JoviDeCroock needed to focus on other things so #4268 was closed. I am re-opening this PR therefore to finish that work.

integrationTests/ts/package.json
```diff
@@ -8,11 +8,6 @@
   "dependencies": {
     "graphql": "file:../graphql.tgz",
     "graphql-esm": "file:../graphql-esm.tgz",
-    "typescript-4.4": "npm:typescript@4.4.x",
-    "typescript-4.5": "npm:typescript@4.5.x",
-    "typescript-4.6": "npm:typescript@4.6.x",
-    "typescript-4.7": "npm:typescript@4.7.x",
-    "typescript-4.8": "npm:typescript@4.8.x",
```
jasonkuhrt, line 15: Regarding this, was already discussed at https://github.com/graphql/graphql-js/pull/4268#issuecomment-2445185032.

src/language/__tests__/predicates-test.ts
jasonkuhrt, line 1: The `Kind` enum member order is now alphabetical per how ESM namespace keys are serialized (I guess).

I've updated the orders in this tests suite. I assume their order did not matter before.

src/language/__tests__/kind-test.ts
jasonkuhrt, line 1: I've added some sanity checks here. Let me know if there is an established type testing approach in the library. For now I did something quick and simple.

src/language/__tests__/predicates-test.ts
yaacovCR: If it did, we are targeting v17 so this would be ok

yaacovCR (approved):
Do we want to take this approach for all enums?

jasonkuhrt:
Yeah I think there's nothing to lose by doing that and it would be a better DX (e.g. all enum have type level name spaces, consistent).

If you agree then I can update this PR to change them all.

JoviDeCroock (approved):
I think the odds of other enums tree-shaking is a lot lower and the amount of code we add in this approach is quite high. The others like DirectiveLocation, ... are more geared for server-side code either way. I don't merging as is as the bundle impact is reduced by moving off of TS enums

jasonkuhrt:
Right now I don't have a strong preference. I consider the type level symmetry aspect bike shedding territory. Happy to throw my labour at this so no issue there.

PR #4271, opened 2024-10-30
Semantic nullability document directive

twof:


src/execution/__tests__/executor-test.ts
```diff
@@ -1321,3 +1327,199 @@ describe('Execute: Handles basic execution tasks', () => {
...
+    `);
+
+    const result = await execute({
+      schema: new GraphQLSchema({ query: DataType }),
+      document,
+      rootValue: data,
+    });
+
+    let executable = document.definitions?.values().next().value as ExecutableDefinitionNode;
+    let selectionSet = executable.selectionSet.selections.values().next().value;
```
twof, line 1369: These are just tests so it's not the biggest deal, but is there a better way to retrieve the node for use in error construction? This is workable, but unwieldy.

yaacovCR: Because we do not yet enable noUncheckedIndexAccess (see https://github.com/graphql/graphql-js/pull/3867) I think you could just index it via 0 at this point although I think what you are doing is preferred, for example

https://github.com/graphql/graphql-js/blob/423e72dec44feb5656a0ea2a384bac68a462732c/src/execution/subscribe.ts#L217

Has been changed in main to your method.

You could assert that you have an executable definition or selection node instead of using typescript casting….

In fact, I think that needs to be improved in main for the example I gave!

https://github.com/graphql/graphql-js/blob/831c1211d7a88bed1758d5c33416514947247241/src/execution/execute.ts#L2241

twof:
## Notes for reviewers
I'm looking for review on everything but utilities. I'm going to let @benjie handle those. Once I've been given a go ahead on the implementation, I'll write the spec edits.

src/execution/__tests__/executor-test.ts
yaacovCR: https://github.com/graphql/graphql-js/blob/423e72dec44feb5656a0ea2a384bac68a462732c/src/language/predicates.ts#L24

yaacovCR:
Is there a reason for the introduction of two new wrapping types as opposed to a single one? Is it because of an unresolved syntax question or is it expected to remain in the final implementation?

I could see this being useful if we were to actually allow multiple coexisting syntax variants as dictated by arguments on the semantic nullability directive (but I am not in favor of that additional complexity).

twof:
> Is there a reason for the introduction of two new wrapping types as opposed to a single one? Is it because of an unresolved syntax question or is it expected to remain in the final implementation?
> 
> I could see this being useful if we were to actually allow multiple coexisting syntax variants as dictated by arguments on the semantic nullability directive (but I am not in favor of that additional complexity).

In the first commit with the parsing implementation I had a single type.

The printer was the main driver here. In the parser there was a clear place to set the `useSemanticNullability` mode flag, but with the printer it wasn't so obvious.

src/language/parser.ts
```diff
@@ -740,6 +754,7 @@ export class Parser {
    *   - NamedType
    *   - ListType
    *   - NonNullType
+   *   - SemanticNonNullType
```
BoD, line 757:
```suggestion
   *   - SemanticNonNullType
   *   - SemanticNullableType
```

src/execution/__tests__/executor-test.ts
yaacovCR: Paid greater attention when working on https://github.com/twof/graphql-js/pull/5 and I think the answer is that generally the tests use `expectJSON` and some other testing deduplication, added second commit there demonstrating use.

yaacovCR:
> The printer was the main driver here. In the parser there was a clear place to set the `useSemanticNullability` mode flag, but with the printer it wasn't so obvious.

I raised https://github.com/twof/graphql-js/pull/5 as a proof-of-concept for what it might look like with a single new wrapper type.

I stored within the created `GraphQLSchema` whether it was created `usingSemanticNullability` and built printing and execution related changed off of that flag.

You can add in semantically nullable types even in a schema that is not using `usingSemanticNullability` and they will be ignored. In particular, to remain with simply one version of the Introspection types, I modified the built-in Introspection types where I thought to be appropriate to use the new `GraphQLSemanticNullable` wrapping types; these types are unwrapped but otherwise ignored even when `usingSemanticNullability`  is not set.

yaacovCR:
closing in favor of newest proposal: see https://github.com/graphql/graphql-spec/pull/1163

PR #4274, opened 2024-10-31
add cancellation support to async iterable iteration

yaacovCR:
When the abort signal is triggered, any pending `.next()` calls should return immediately, ~~and the underlying async iterables `.return()` method should be called~~

### DIFF
--- a/src/execution/PromiseCanceller.ts
+++ b/src/execution/PromiseCanceller.ts
@@ -35,9 +35,7 @@ export class PromiseCanceller {
     }
 
     const { promise, resolve, reject } = promiseWithResolvers<T>();
-    const abort = () => {
-      reject(this.abortSignal.reason);
-    };
+    const abort = () => reject(this.abortSignal.reason);
     this._aborts.add(abort);
     originalPromise.then(
       (resolved) => {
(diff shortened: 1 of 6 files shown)


yaacovCR:
NOTE: will potentially rename `PromiseCanceller` to something more generic in a later PR.

Something like: `AbortSignalListener`, and potentially even the methods of the class can be switched to functions which take a shared `AbortSignalListener`.

JoviDeCroock (approved):

JoviDeCroock:
ExecutionOrchestrator could be an over-arching term. It essentially listens for a signal and tears down execute

yaacovCR:
This PR has been updated:

1. to be a little less clever => with `cancellableIterable()`, we no longer call the underlying iterable's `.return()` method automatically. This is tricky to get exactly right -- users of the functions would have to know not to call `.return()` themselves -- and is also not necessary in our case, because as soon as the iterable errors with any error, including an abort error, our code calls `.return()` already.

2. to be a little less sloppy => we now call `.disconnect()` hopefully everywhere required to remove all added event listeners.

JoviDeCroock (approved):

Issue #4245, opened 2024-10-18
Add description changes to safe schema changes


JoviDeCroock:
Expand `findSchemaChanges` to find changes to descriptions of schema-elements

PR #4275, opened 2024-10-31
Add safe schema change for descriptions changing

JoviDeCroock:
This adds a new safe schema-change when a description changes of any schema-member. Do note that there are some drive-by fixes in here as I noticed a bug in safe-argument changes where we'd just always report the changes rather than actually see whether something changed 😅 

Resolves https://github.com/graphql/graphql-js/issues/4245

### DIFF
--- a/src/utilities/__tests__/findSchemaChanges-test.ts
+++ b/src/utilities/__tests__/findSchemaChanges-test.ts
@@ -22,6 +22,23 @@ describe('findSchemaChanges', () => {
     ]);
   });
 
+  it('should detect a type changing description', () => {
+    const newSchema = buildSchema(`
+      "New Description"
+      type Type1
+    `);
+
+    const oldSchema = buildSchema(`
+      type Type1
+    `);
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
+      {
+        type: SafeChangeType.DESCRIPTION_CHANGED,
+        description: 'Description of Type1 has changed to "New Description".',
+      },
+    ]);
+  });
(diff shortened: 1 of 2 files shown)


src/utilities/__tests__/findSchemaChanges-test.ts
```diff
@@ -43,6 +60,30 @@ describe('findSchemaChanges', () => {
...
+        foo: String
+        "New Description"
+        bar: String
+      }
+    `);
+    expect(findSchemaChanges(oldSchema, newSchema)).to.deep.equal([
+      {
+        type: SafeChangeType.DESCRIPTION_CHANGED,
+        description:
+          'Description of field Query.bar has changed to "New Description".',
```
yaacovCR, line 82: Generally -- not just here -- we could consider printing the old description as well, giving better context as to what exactly changed.

yaacovCR:
> as I noticed a bug in safe-argument changes where we'd just always report the changes rather than actually see whether something changed

I broke this out into a separate commit on this fix out into a separate commit on this new branch: https://github.com/graphql/graphql-js/tree/break-out-fix

Instead of doing stacked PRs, we can just have multiple commits within a PR and use the "rebase" merge strategy instead of the "squash" strategy. This makes the git history easier to follow in the long run but doesn't lead to (as much) rebasing madness.

yaacovCR (approved):
Left a comment with something to consider, and addressed separately the drive-by changes. I guess we should try our best to minimize those, I know I've been tempted to sneak them in when possible!

PR #4276, opened 2024-11-01
BigInt support in scalars

JoviDeCroock:
Supersedes https://github.com/graphql/graphql-js/pull/4088
Resolves https://github.com/graphql/graphql-js/issues/3913

This adds support for BigInt values when we serialize a response value from JS to send over the wire.

- When the expected value is a String we'll stringify the BigInt
- When the expected value is an ID we'll stringify the BigInt
- When the expected value is an int we'll cast it to a number, when this exceeds or is below min/max int we'll still throw an error
- When the expected value is a float we'll cast it to a number
- When the expected value is a boolean we'll cast it to a boolean value, i.e. 1 true 0 false

For Input Values:

- When the expected value is an ID we'll stringify the BigInt
- When the expected value is an int we'll cast it to a number, when this exceeds or is below min/max int we'll still throw an error
- When the expected value is a float we'll cast it to a number

This also adds support in `valueToAST` even though it's deprecated because BigInt support has always been in the spec when it comes to the ID Scalar.

I've tried adding BigInt to `valueToLiteral` however, for that we'd need to expand the spec to i.e. have an identifier for BigInt as currently we don't really have a way to represent it so we have to treat it like it is a regular number.

### DIFF
--- a/src/type/__tests__/scalars-test.ts
+++ b/src/type/__tests__/scalars-test.ts
@@ -21,6 +21,7 @@ describe('Type System: Specified scalar types', () => {
       expect(coerceInputValue(1)).to.equal(1);
       expect(coerceInputValue(0)).to.equal(0);
       expect(coerceInputValue(-1)).to.equal(-1);
+      expect(coerceInputValue(1n)).to.equal(1);
 
       expect(() => coerceInputValue(9876504321)).to.throw(
         'Int cannot represent non 32-bit signed integer value: 9876504321',
@@ -119,6 +120,7 @@ describe('Type System: Specified scalar types', () => {
       expect(coerceOutputValue(1e5)).to.equal(100000);
       expect(coerceOutputValue(false)).to.equal(0);
       expect(coerceOutputValue(true)).to.equal(1);
+      expect(coerceOutputValue(1n)).to.equal(1);
 
       const customValueOfObj = {
         value: 5,
@@ -190,6 +192,7 @@ describe('Type System: Specified scalar types', () => {
       expect(coerceInputValue(-1)).to.equal(-1);
       expect(coerceInputValue(0.1)).to.equal(0.1);
       expect(coerceInputValue(Math.PI)).to.equal(Math.PI);
+      expect(coerceInputValue(1n)).to.equal(1);
 
       expect(() => coerceInputValue(NaN)).to.throw(
         'Float cannot represent non numeric value: NaN',
@@ -280,6 +283,7 @@ describe('Type System: Specified scalar types', () => {
       expect(coerceOutputValue('-1.1')).to.equal(-1.1);
       expect(coerceOutputValue(false)).to.equal(0.0);
       expect(coerceOutputValue(true)).to.equal(1.0);
+      expect(coerceOutputValue(1n)).to.equal(1n);
 
       const customValueOfObj = {
         value: 5.5,
@@ -380,6 +384,7 @@ describe('Type System: Specified scalar types', () => {
       expect(coerceOutputValue(-1.1)).to.equal('-1.1');
       expect(coerceOutputValue(true)).to.equal('true');
       expect(coerceOutputValue(false)).to.equal('false');
+      expect(coerceOutputValue(9007199254740993n)).to.equal('9007199254740993');
 
       const valueOf = () => 'valueOf string';
       const toJSON = () => 'toJSON string';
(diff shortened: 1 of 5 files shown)


src/type/scalars.ts
```diff
@@ -113,7 +118,7 @@ export const GraphQLFloat = new GraphQLScalarType<number>({
       num = Number(coercedValue);
     }
 
-    if (typeof num !== 'number' || !Number.isFinite(num)) {
+    if (!isNumeric(num)) {
```
yaacovCR, line 121: If I follow this correctly, the result of this coercion could be of JS type `bigint` rather than of type JS `number`.

This would seem to be unexpected.

maybe the line above can be edited to:

>     if (
>      typeof coercedValue === 'bigint' ||
>      (typeof coercedValue === 'string' && coercedValue !== '')
>    ) {

Note: there is a loss of precision when convering from bigint => float. Another option is to error, so that developers can do the conversion on their own and be more aware of the loss of precision.

src/type/scalars.ts
```diff
@@ -122,12 +127,12 @@ export const GraphQLFloat = new GraphQLScalarType<number>({
...
 
   coerceInputValue(inputValue) {
-    if (typeof inputValue !== 'number' || !Number.isFinite(inputValue)) {
+    if (!isNumeric(inputValue)) {
       throw new GraphQLError(
         `Float cannot represent non numeric value: ${inspect(inputValue)}`,
       );
     }
-    return inputValue;
+    return typeof inputValue === 'bigint' ? Number(inputValue) : inputValue;
```
yaacovCR, line 135: As opposed to above, here I see an explicit conversion. :)

src/type/__tests__/scalars-test.ts
```diff
@@ -280,6 +283,7 @@ describe('Type System: Specified scalar types', () => {
       expect(coerceOutputValue('-1.1')).to.equal(-1.1);
       expect(coerceOutputValue(false)).to.equal(0.0);
       expect(coerceOutputValue(true)).to.equal(1.0);
+      expect(coerceOutputValue(1n)).to.equal(1n);
```
yaacovCR, line 286:
```suggestion
      expect(coerceOutputValue(1n)).to.equal(1);
```

Assuming we take the change below.

src/utilities/__tests__/astFromValue-test.ts
```diff
@@ -143,6 +168,11 @@ describe('astFromValue', () => {
       kind: 'NullValue',
     });
 
+    expect(astFromValue(9007199254740993n, GraphQLString)).to.deep.equal({
+      kind: 'StringValue',
+      value: '9007199254740993',
```
yaacovCR, line 173:
```suggestion
      value: '9007199254740992',
```

There is a loss of precision when converting a bigint to a float

yaacovCR (changes requested):
Would be great to get this in. Some comments inline are with respect to how to handle bigint => float

In terms of `valueToLiteral()`, I think it can be upgrade to handle bigint in the same way as `astFromValue`. `valueToLiteral` is a typed function that ends up calling the individual scalar `.valueToLiteral()` methods just like `astFromValue()` and so when supplied a bigint `someBigInt` like so: `valueToLiteral(someBigInt, GraphQLInt)`, it would call `GraphQLInt.valueToLiteral()` and the idea would be that this should work as long as the bigint can be coerced.

This would not be about adding a bigint literal type.

yaacovCR:
> This also adds support in `valueToAST`

I think typo => `astFromValue`

yaacovCR:
Taking a step back:

1. upgrading `coerceOutputValue`/`serialize` to handle `bigint` seems like a sometimes win, as resolvers can now return `bigint`s and we convert automatically to `GraphQLString` or `GraphQLID` => but also a sometimes loss, as converting `bigint` to `GraphQLInt` or `GraphQLFloat` might be an error/lose precision. Even if we can coerce in an individual scenario, it suggests an underlying developer error, because we can't *always coerce*. It seems like forcing the developer to coerce might be the safest option.

2. upgrading `coerceInputValue`/`parse` to handle `bigint` is not as big of a win from the get-go, as the natural path of value flow into GraphQL is via JSON, and JSON does not support `bigint`. We also in general are stricter on coercion for input => For example, in output we would convert a boolean to GraphQLInt and a number to GraphQLBoolean, but we do not do that for input types. In short, I am not sure if we should upgrade `coerceInputValue`/`parse` to handle `bigint` .

3. upgrading `astFromValue` or `valueToLiteral` to support bigint means allows us to print programmatically set default values that were set using `bigint`. In v16 and earlier, the default values when supplied programatically are the internal values, and  so we call `serialize` to print them, even though that may not be defined on input values. In v17, we call `valueToLiteral` to print them, which is well defined. In v16, that means that once we upgrade serialize, essentially `astFromValue` has already been upgraded. In v17, there is less to little need to upgrade `valueToLiteral` because the schema author should be able to supply the correct type instead of `bigint` when specifying the default value.

Just some longer thoughts, meant to rehearse my understanding of [Lee's diagram from the default value PR](https://github.com/graphql/graphql-js/pull/3049#issue-866953281).

Issue #4237, opened 2024-10-14
Add directive-arguments changes to `findSchemaChanges`


JoviDeCroock:
# Feature requests

Currently we are missing changes to directive-arguments in our `findSchemaChanges` function, we'll have to cover similar cases to field-arguments.

PR #4277, opened 2024-11-01
Detect directive argument changes

JoviDeCroock:
Resolves https://github.com/graphql/graphql-js/issues/4237

This mirrors the fields/input-object argument logic but specifically for directives.

### DIFF
--- a/src/utilities/__tests__/findSchemaChanges-test.ts
+++ b/src/utilities/__tests__/findSchemaChanges-test.ts
@@ -2,7 +2,12 @@ import { expect } from 'chai';
 import { describe, it } from 'mocha';
 
 import { buildSchema } from '../buildASTSchema.js';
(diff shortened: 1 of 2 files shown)


yaacovCR (commented):
Questions:


1. Directive arg change logic matches field arg change logic with respect to safe/dangerous/breaking if the directives are used in operations. Is that always true with respect to directives used to annotate schemas written in SDL?

2. Not directly related to this PR, but => default argument removal seems to marked as dangerous, but it could be breaking if the argument is non-null (because it makes the argument Required from the operation perspective). This fact is not specific to directives, it's true for fields also, but I couldn't find the logic where this was handled for either directives or fields. Is this a separate bug?

JoviDeCroock:
1. we also don't make the distinction between schema and document directives in the directive removed, we outright flag it as breaking. If we want to change that logic, we should also change it for removal of directives.

2. yes, separate issue

Issue #3421, opened 2021-12-09
changing the type-signature of parse and validate functions to return a MaybePromise


n1ru4l:
Both `parse` and `validate` do 100% sync logic, so this request might seem obscure. Please let me try to break this down and propose why we might want to consider changing the `parse` and `validate`  return types.

Today all major GraphQL server and frameworks built on `graphql-js` wrap the functions `parse`, `validate`, `execute` and `subscribe`.

`execute` and `subscribe` both already return a `MaybePromise` as the underlying logic can either be resolved sync or async, depending on the resolvers called or the errors occurring.

Many GraphQL servers allow overwriting the `parse`, `validate`, `execute`, and `subscribe` functions. In user-land functionality of those functions can be customized by wrapping the original `graphql-js` functions and then forwarding them to the servers.

For example, caching for parse could be simply achieved in the following way:

**parse in memory cache**
```ts
import { parse as originalParse } from "graphql"
import { buildCacheKey } from "./buildCacheKey"

// very cheap implementation with memory leaks :)
const cheapCache = {}

function parse(args) {
  const cacheKey = buildCacheKey(args)
  let cachedValue = cheapCache[cacheKey]
  if (cachedValue != undefined) {
    return cachedValue 
  }
  const result = originalParse(args)
  cheapCache[cacheKey] = result
  return result
}
```

For very big operations and GraphQL server replicas or serverless/edge worker environments, we might want to use a shared cache between our GraphQL handlers. E.g. we want to store our parsed operations or validation results within Redis instead of simply in memory.

**async validate cache using redis**
```ts
import { validate as originalValidate } from "graphql"
import { buildCacheKey } from "./buildCacheKey"
import { redisClient } from "./redisClient"

async function validate(args) {
  const cacheKey = buildCacheKey(args)
  let cachedValue = await redisClient.get(cacheKey)
  if (cachedValue != undefined) {
    return cachedValue 
  }
  const result = originalValidate (args)
  await redisClient.set(cacheKey, result)
  return result
}
```

We can easily build this function, however, if passed to a server those libraries will not expect `validate` to return a Promise and usually raise an error.

This leads me to the following conclusion:

If the `parse` and `validate` functions return a `MaybePromise` instead of "only" a sync value, graphql-js enforces that server framework implementors ensure a `Promise` returned from `parse` and `validate` is correctly handled.

Today adopters of different server frameworks have to convince the framework maintainers individually to allow a `validate` or `parse` functions to return a `Promise`.

[...]

PR #4278, opened 2024-11-01
Allow for parse to return a Promise

JoviDeCroock:
This PR enables users to return a Promise from parsing, this is mainly intended so GraphQL servers are prepared for middleware approaches that can inject async caches.

I've opted to not do validate as this could create the impression that rules can run in the background, while all rules logic is synchronous.

I've also opted to not do the `parseValue` and other parse equivalents as these seem less relevant for the caching bit.

😅 still have to fix a lot of types but first gauging the appeal of this.

Resolves https://github.com/graphql/graphql-js/issues/3421

### DIFF
--- a/src/language/parser.ts
+++ b/src/language/parser.ts
@@ -1,3 +1,4 @@
+import { isPromise } from '../jsutils/isPromise.js';
 import type { Maybe } from '../jsutils/Maybe.js';
 
 import type { GraphQLError } from '../error/GraphQLError.js';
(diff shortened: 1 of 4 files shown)


src/language/parser.ts
```diff
@@ -127,6 +128,18 @@ export function parse(
   return document;
 }
 
+export function parseSync(
+  source: string | Source,
+  options?: ParseOptions | undefined,
+): DocumentNode {
+  const result = parse(source, options);
+  /* c8 ignore next 3 */
+  if (isPromise(result)) {
```
JoviDeCroock, line 137: As we can't really test this I've opted to ignore the lines, it's more so that consumers will be prepared to handle a Promise

yaacovCR:
I do not feel comfortable changing the type signature of `parse()` to async if it does not do async work.

I've raised #4285 as a separate feature that I think has independent value (allowing for optional simple user-provided caches) that change the type signature because the return value may be async.

PR #4284, opened 2024-11-04
Provide people with tabs so they can use classes as well

JoviDeCroock:
This provides people with the option to choose between the template approach or the classes approach.

Screenshot 2024-11-04 at 20 52 24


yaacovCR (approved):
I haven’t found the time yet to get through every line, but I love the idea of tabs!

One question I have is whether we want to call these SDL-first vs code-first …. Our SDL-first currently requires using the root value/parent method of providing resolvers but the eventual plan would be to allow resolvers as additional arguments to buildschema just like graphql tools makeExecutableSchema…

PR #4286, opened 2024-11-06
fix: properly type `extensions` in GraphQLFormattedError

tpoisseau:
Following typescript documentation, it's not possible override interface property, we can only add new props.

Since it's declared as unknown dict, even if we merge `GraphQLErrorExtensions`, we can't access to our extensions without workaround and verbose typescript in source-code.

It's annoying since apollo expose `GraphQLFormattedError` instead `GraphQLError`

Refs: https://www.typescriptlang.org/docs/handbook/declaration-merging.html
Refs: https://github.com/apollographql/apollo-client/pull/11789

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -20,6 +20,19 @@ export interface GraphQLErrorExtensions {
   [attributeName: string]: unknown;
 }
 
+/**
+ * Custom formatted extensions
+ *
+ * @remarks
+ * Use a unique identifier name for your extension, for example the name of
+ * your library or project. Do not use a shortened identifier as this increases
+ * the risk of conflicts. We recommend you add at most one extension field,
+ * an object which can contain all the values you need.
+ */
+export interface GraphQLFormattedErrorExtensions {
+  [attributeName: string]: unknown;
+}
+
 export interface GraphQLErrorOptions {
   nodes?: ReadonlyArray<ASTNode> | ASTNode | null;
   source?: Maybe<Source>;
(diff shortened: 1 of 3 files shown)


src/error/GraphQLError.ts
```diff
@@ -275,7 +275,7 @@ export interface GraphQLFormattedError {
    * Reserved for implementors to extend the protocol however they see fit,
    * and hence there are no additional restrictions on its contents.
    */
-  readonly extensions?: { [key: string]: unknown };
+  readonly extensions?: GraphQLErrorExtensions;
```
yaacovCR, line 278: What do you think about...

```suggestion
  readonly extensions?: GraphQLFormattedErrorExtensions;
```

with the new interface:

```ts
/**
 * Custom extensions
 *
 * @remarks
 * Use a unique identifier name for your extension, for example the name of
 * your library or project. Do not use a shortened identifier as this increases
 * the risk of conflicts. We recommend you add at most one extension field,
 * an object which can contain all the values you need.
 */
export interface GraphQLFormattedErrorExtensions {
  [attributeName: string]: unknown;
}
```

tpoisseau: I'm not convinced of the point of having two different types, but it won't cost any more.

I'll do it

yaacovCR (approved):

JoviDeCroock (approved):
The general idea behind this looks good to me, I did have one thought which is mainly testing related. Currently it looks like the strict typing won't be applied to manually creating a `GraphQLError` however due to the loosely typed nature of it it won't block you. Just wanted to highlight that currently this static typing will only apply to the result type.

I've marked this as a feature, as this allows you to overwrite this to make your type more strict.

PR #4290, opened 2024-11-07
Write about `@oneOf`

JoviDeCroock:
Currently input-unions and by extension the `@oneOf` directive aren't present in the documentation. I have opted to put this into the advanced section. The copy might be up for improvement, honestly fire away if there's more cases to cover, just wanted to get the ball rolling here.

CC @benjie

### DIFF
--- a/website/pages/input-unions.mdx
+++ b/website/pages/input-unions.mdx
@@ -1,0 +1,72 @@
+---
+title: Input Unions
+---
+
+import { Tabs } from 'nextra/components';
+
+Some inputs will behave differently depending on what input we choose. Let's look at the case for
+a field named `product`, we can fetch a `Product` by either its `id` or its `name`. Currently we'd
+make a tradeoff for this by introducing two arguments that are both nullable, now if both are passed
+as null we'd have to handle that in code. To fix this the `@oneOf` directive was introduced so we
+can create these input-unions without sacrificing the strictly typed nature of our GraphQL Schema.
+
+<Tabs items={['SDL', 'Code']}>
+    <Tabs.Tab>
+```js
+const schema = buildSchema(`
+  type Product {
+    id: ID!
+    name: String!
+  }
+
+  input ProductInput @oneOf {
+    id: ID
+    name: String
+  }
+
+  type Query {
+    product(input: ProductInput!): Product
+  }
+`);
+```
+    </Tabs.Tab>
+    <Tabs.Tab>
+```js
+const Product = new GraphQLObjectType({
+  name: 'Product',
+  fields: {
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+    },
+    name: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+  },
+});
+
+const ProductInput = new GraphQLInputObjectType({
+  name: 'ProductInput',
+  isOneOf: true,
+  fields: {
+    id: { type: GraphQLID },
+    name: { type: GraphQLString },
+  },
+});
+
+const schema = new GraphQLSchema({
+  query: new GraphQLObjectType({
+    name: 'Query',
+    fields: {
+      product: {
+        type: Product,
+        args: { input: { type: ProductInput } },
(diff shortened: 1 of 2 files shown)


website/pages/input-unions.mdx
```diff
@@ -0,0 +1,72 @@
+---
+title: Input Unions
```
benjie, line 2: `@oneOf` is not equivalent to "input union". You can _use_ it to form an input union, but its use is broader than that. I would title it "OneOf input objects" or similar.

(Example: `input UserSpecifier @oneOf { databaseId: Int, username: String, hospitalNumber: Int }` - note that both `databaseId` and `hospitalNumber` are `Int` and can uniquely identify a user in different ways - so this isn't simply a union.)

website/pages/input-unions.mdx
```diff
@@ -0,0 +1,72 @@
...
+---
+
+import { Tabs } from 'nextra/components';
+
+Some inputs will behave differently depending on what input we choose. Let's look at the case for
+a field named `product`, we can fetch a `Product` by either its `id` or its `name`. Currently we'd
+make a tradeoff for this by introducing two arguments that are both nullable, now if both are passed
+as null we'd have to handle that in code. To fix this the `@oneOf` directive was introduced so we
```
benjie, line 10:
```suggestion
as null (or both non-null) we'd have to handle that in code - the type system wouldn't indicate that exactly one was required. To fix this, the `@oneOf` directive was introduced so we
```

website/pages/input-unions.mdx
```diff
@@ -0,0 +1,72 @@
...
+can create these input-unions without sacrificing the strictly typed nature of our GraphQL Schema.
```
benjie, line 11:
```suggestion
can create this "exactly one option" constraint without sacrificing the strictly typed nature of our GraphQL Schema.
```

website/pages/input-unions.mdx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  input ProductInput @oneOf {
+    id: ID
+    name: String
+  }
+
+  type Query {
+    product(input: ProductInput!): Product
+  }
```
benjie, line 29: `ProductInput` is generally used for CRUD-style mutations. Let's use a different input object name here:

```suggestion
  input ProductSpecifier @oneOf {
    id: ID
    name: String
  }

  type Query {
    product(by: ProductSpecifier!): Product
  }
```

website/pages/input-unions.mdx
```diff
@@ -0,0 +1,72 @@
...
+      },
+    },
+  }),
+});
+```
+    </Tabs.Tab>
+</Tabs>
+
+It doesn't matter whether you have 2 or more inputs here, all that matters is
+that your user will have to specify one, and only one, for this input to be valid.
```
benjie, line 72:
```suggestion
that your user will have to specify one, and only one, for this input to be valid.
The values are not limited to scalars, lists and other input object types are also allowed.
```

website/pages/input-unions.mdx
benjie: Maybe we should expand this so it demonstrates usage with non-scalars too?

```suggestion
  input ProductLocation {
    aisleNumber: Int!
    shelfNumber: Int!
    positionOnShelf: Int!
  }
  
  input ProductSpecifier @oneOf {
    id: ID
    name: String
    location: ProductLocation
  }

  type Query {
    product(by: ProductSpecifier!): Product
  }
```

website/pages/input-unions.mdx
JoviDeCroock: My hesitancy was mainly that it's less of a clickthrough as the name is opaque to people while a general known word helps. I do agree that they are not the same

benjie: I agree; perhaps the URL would be oneof and the title would be "OneOf and input unions" or similar?

benjie:
The SDL changes need reflection in the code version I think?

PR #4292, opened 2024-11-08
Expose tokenCount on the DocumentNode

JoviDeCroock:
Backports https://github.com/graphql/graphql-js/pull/4251 from v17

Backporting this so we can document the `maxTokens` for production use cases. The performance of this change looks to be neutral


JoviDeCroock:
@github-actions run-benchmark

yaacovCR (approved):

PR #4297, opened 2024-11-19
introduces mapSchemaConfig utility function

yaacovCR:
motivation:

1. rebuilds `lexicographicSortSchema()` on this generic utility, better demonstrating only the sorting
2. rebuilds `extendSchemaImpl()` on this utility
3. can be used as base to more easily enhance `extendSchema()`/`buildASTSchema()`/`buildSchema()` to take resolvers, etc
4. can be used to expose a generic safe `mapSchemaConfig()` utility
5. improved performance -- not the initial motivation, but turns out these changes bring an 80% speed improvement to `buildSchema()`.

### DIFF
--- a/src/utilities/mapSchemaConfig.ts
+++ b/src/utilities/mapSchemaConfig.ts
@@ -1,0 +1,464 @@
+import { inspect } from '../jsutils/inspect.js';
+import { invariant } from '../jsutils/invariant.js';
+import type { Maybe } from '../jsutils/Maybe.js';
+
+import type {
+  GraphQLArgumentNormalizedConfig,
+  GraphQLEnumTypeNormalizedConfig,
+  GraphQLEnumValueConfig,
+  GraphQLFieldNormalizedConfig,
+  GraphQLFieldNormalizedConfigArgumentMap,
+  GraphQLFieldNormalizedConfigMap,
+  GraphQLInputFieldConfig,
+  GraphQLInputObjectTypeNormalizedConfig,
+  GraphQLInterfaceTypeNormalizedConfig,
+  GraphQLNamedType,
+  GraphQLObjectTypeNormalizedConfig,
+  GraphQLScalarTypeNormalizedConfig,
+  GraphQLType,
+  GraphQLUnionTypeNormalizedConfig,
+} from '../type/definition.js';
+import {
+  GraphQLEnumType,
+  GraphQLInputObjectType,
+  GraphQLInterfaceType,
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLScalarType,
+  GraphQLUnionType,
+  isEnumType,
+  isInputObjectType,
+  isInterfaceType,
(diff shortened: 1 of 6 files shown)


yaacovCR:
note: diff is easier to scan when hiding whitespace

yaacovCR:
This re-implementation brings a 25% performance boost to `buildSchema()` and a small drop in memory use.

image

I'm not sure yet what exactly brings the performance boost, I was just hoping for parity.

yaacovCR:
Just poking around a bit with [deopt-explorer](https://github.com/microsoft/deoptexplorer-vscode), which, as usual, is beyond me, but the summary for main is:

image

and the summary for map is:

image

For clarity, within the map version, we rename `getWrappedType()` to `typeFromAST()` and `getNamedType()` to `namedTypeFromAST()` because it allows us to use `getNamedType()` in the `SchemaContext` closure without a clash -- and these functions essentially are a pre-GraphQLSchema-creation version of our `typeFromAST()` utility with the exact same algorithm.

For clarity, I also renamed `buildType()` to `buildNamedType()`.

So the thing that leaps out at me is that the number of deoptimizations in `buildType()`/`buildNamedType()` and `extendSchemaImpl()` drop, the former significantly. I am not 100% on why. I can only guess that splitting the functions into smaller bits helped.

yaacovCR:
UPDATE:

- I have separated out the commits introducing `mapSchemaConfig()` and those used to rebuild `lexicographicSortSchema()` and `extendSchemaImpl()`, the latter underlying `buildSchema()` and `extendSchema()`.
- The initial commit has full test coverage on its own.
- I have introduced another "helper" method available to the mapper functions, `getNamedTypes()` which means that we don't have to manually track the added types and then concatenate them to the pre-existing types within `extendSchemaImpl()`.

Benchmark results from current version are even better than before, an ~80% improvement on main:

image

yaacovCR:

yaacovCR:

yaacovCR:

yaacovCR:

JoviDeCroock (approved):

src/utilities/mapSchemaConfig.ts
```diff
@@ -0,0 +1,464 @@
...
+    return mapper == null ? mappedConfig : mapper(mappedConfig);
+  }
+}
+
+const stdTypeMap = new Map(
+  [...specifiedScalarTypes, ...introspectionTypes].map((type) => [
+    type.name,
+    type,
+  ]),
+);
```
ab-pm, line 464: It seems this is now duplicated between https://github.com/graphql/graphql-js/blob/ef042cf131b27979269ce5b0e5ef1caac22b3d1d/src/utilities/extendSchema.ts#L574-L579 and https://github.com/graphql/graphql-js/blob/ef042cf131b27979269ce5b0e5ef1caac22b3d1d/src/utilities/mapSchemaConfig.ts#L459-L464
One of them should suffice :-)

PR #4298, opened 2024-11-22
use the node --import option as suggested

yaacovCR:
replacing --loader ts-node/esm/transpile-only, which gives the following warning:

> (node:33248) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
> --import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("ts-node/esm/transpile-only", pathToFileURL("./"));'


yaacovCR:
woops, we can't quite do that as that option doesn't work on our older supported versions!

PR #4299, opened 2024-11-22
Make deprecated.reason non-nullable

martinbonnin:
See https://github.com/graphql/graphql-spec/pull/1040


leebyron:
Approved! ensure this goes in a major version change since it could cause type recompiles which might have CI implications for some users

yaacovCR (approved):

PR #4300, opened 2024-11-25
fix for mixture of default and custom root type names

yaacovCR:
Is it legal to use “extend schema” if the schema definition is omitted and the default types could have been used?

### DIFF
--- a/src/utilities/__tests__/buildASTSchema-test.ts
+++ b/src/utilities/__tests__/buildASTSchema-test.ts
@@ -1033,6 +1033,22 @@ describe('Schema Builder', () => {
     expect(schema.getSubscriptionType()).to.include({ name: 'Subscription' });
   });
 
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
I am not sure I understand the question about the legality of this operation? If your question is whether it's fine to leverage `extend Schema` without there being an explicit `Schema` definition I think that's legal yes, as the `Schema` definition in both SDL as well as Code first is often implicit so I'd see this as a safe way of overriding the Query default type.

All though this does beg the question where the implicit schema type is `Schema { query: Query }` and we change it to `Schema { query: MyQuery }` over a type-extension. That's potentially a footgun in fully removing all the related fields hat are defined in `Query`.

yaacovCR:
That's a great catch on the spec requiring the schema to be defined! :) I think we can close this.

PR #4301, opened 2024-11-26
frontport v16 work to main

yaacovCR:
EDITED:

this branch:

- tracks all of the work done on v16 and frontports it to main (essentially the website changes and #4286)
- upgrades eslint to v9 as nextra and v8 do not play well together
- upgrades our devDependencies while we are at it

the only thing that is failing is our netlify checks, which should be removed as part of the transition to nextra

NOTE: this branch should be merged to main with the "rebase" strategy so as to preserve the original commits in history


yaacovCR:
this is ready for review, converted this PR to an integration branch for frontporting all of the v16 work to v17, basically just the website work and #4286

JoviDeCroock (approved):

PR #4303, opened 2024-11-26
adjust GraphQLResolveInfo to remove breaking change from v16

yaacovCR:
#3811 internally preserved the sources of variable values so that we could properly replace variables within complex scalars, but it also exposed this within the GraphQLResolveInfo interface available to resolvers in a breaking manner, by changing the type of the `variableValues` property.

This PR reverts the change to the `variableValues` property, but exposed the extended variable information under a new `variableValuesWithSources` property.

### DIFF
--- a/src/execution/__tests__/executor-test.ts
+++ b/src/execution/__tests__/executor-test.ts
@@ -222,6 +222,7 @@ describe('Execute: Handles basic execution tasks', () => {
       'rootValue',
       'operation',
       'variableValues',
+      'variableValuesWithSources',
     );
 
     const operation = document.definitions[0];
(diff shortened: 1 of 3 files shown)


yaacovCR:
Not 100% sure about this:

from the what's new in v17 WIP at #4198:

> The return type of `getVariableValues()` has been updated, as have the provided arguments to `getArgumentValues()`, `coerceInputLiteral()`, and `collectFields()`, and `getDirectiveValues()`.

I forgot to write there explicitly that we changed the type of `variableValues`, although it's partly what I meant when I wrote "throughout the code base." I guess I'm wondering if we need to change it consistently throughout the code base. Those doing lookahead using `GraphQLResolveInfo` will probably need the variable sources for the same reason we do (to replace variable values safely within complex scalar literals), but they can access that in a new properly.

As of now, `collectFields()` doesn't seem to require the variable sources, we could probably just pass the coerced variables, but because the functions it depends on are more generic, they now take the full `VariableValues` with the sources.

Is there a way to make this even less breaking? Should we skip this PR and just be 100% consistent. Questions!

yaacovCR:
I'm favoring today just leaving things as is for consistency, feedback welcome!

yaacovCR:
Closing for now!

PR #4305, opened 2024-11-28
Make gql js faster

JoviDeCroock:
While researching the numbers on v15 vs v16 I realised that our printer became _much_ slower, the reason for this is two-fold.

- We started leveraging a really expensive `join` function that `filters` even when it's not needed
- During `visit` we always recreate the _full_ tree

Now that second point we should discuss, as changing this might be considered a breaking change.

The way `visit` currently behaves is that it is a reducer, whether or not the user returns a value it will always _fully_ recreate the AST without any attempt at reusing parts of it, or as is the issue with our printer, it will recreate the full AST even if we aren't even remotely returning anything AST-like, we are returning strings 😅 

This PR can serve as an exploration of how we can improve things in v17 - I've seen issues flying around about refactoring the visitor, which this seems like a good trigger judging from how the performance is.

### DIFF
--- a/src/language/visitor.ts
+++ b/src/language/visitor.ts
@@ -214,8 +214,7 @@ export function visit(
           node = node.slice();
 
           let editOffset = 0;
-          for (let i = 0; i < edits.length; i++) {
-            const { 0: editKey, 1: editValue } = edits[i];
+          for (const [editKey, editValue] of edits) {
             const arrayKey = editKey - editOffset;
             if (editValue === null) {
               node.splice(arrayKey, 1);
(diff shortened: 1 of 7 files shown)


JoviDeCroock:
@github-actions run-benchmark

JoviDeCroock:
@github-actions run-benchmark

src/language/visitor.ts
```diff
@@ -222,10 +224,7 @@ export function visit(
             }
           }
         } else {
-          node = Object.defineProperties(
```
JoviDeCroock, line 225: This is a massive deopt

yaacovCR (approved):
looks amazing

yaacovCR:
> looks amazing

Might want to make the PR title a bit more specific tho :)

JoviDeCroock:
After evaluation it doesn't look safe to remove property-descriptors 😅 this would be a breaking change

PR #4311, opened 2024-12-15
Remove a few deopts already

JoviDeCroock:
I've started digging through the codebase on wins with regards to performance, I've seen a lot of them.... However, some of them also come down to our ... weird usage of TypeScript.

Let's take the `Token` as an example, we did a great thing in creating it as a class as it enforced consistent monomorphic backing shapes, however we do a non-null assertion, for the reason that `value` can actually be undefined. When we use `Token` we always assume that `value` is defined which makes all of this deopt as we will call functions with `string | undefined` the whole thing while our code is geared for `string`.

I've attempted to make `value` the correct type which resulted in a lot more issues.

In addition to that we have a lot of call-sites for the `.slice` method, we however do polymorphic access on it as adjacent call-sites will invoke `String.slice` right next to `Array.slice` which makes `.slice` not optimizeable, instead we can resort to converting `String.slice` into `String.substring` to avoid this issue all together.

The last noLocation parse bailout remaining is the `.slice` case of `blockstring.ts`

A tooltip showing us that a slice on a newly created array through .map deopts over polymorphic access

The improvements on the parser benchmark are currently minimal, I am trying to discover the deopts and how we can have a way forward.

### DIFF
--- a/src/language/lexer.ts
+++ b/src/language/lexer.ts
@@ -38,7 +38,7 @@ export class Lexer {
   lineStart: number;
 
   constructor(source: Source) {
-    const startOfFileToken = new Token(TokenKind.SOF, 0, 0, 0, 0);
+    const startOfFileToken = new Token(TokenKind.SOF, 0, 0, 0, 0, undefined);
 
     this.source = source;
     this.lastToken = startOfFileToken;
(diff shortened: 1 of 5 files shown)


src/language/lexer.ts
```diff
@@ -248,39 +247,123 @@ function readNextToken(lexer: Lexer, start: number): Token {
...
+          undefined,
+        );
       case 0x0029: // )
-        return createToken(lexer, TokenKind.PAREN_R, position, position + 1);
+        return createToken(
+          lexer,
+          TokenKind.PAREN_R,
+          position,
+          position + 1,
+          undefined,
```
JoviDeCroock, line 287: Consistently calling `createToken` with 5 arguments helps V8 inline the cache

PR #4314, opened 2024-12-19
reintroduce incremental delivery support for subscriptions

yaacovCR:
depends on #4313

Motivation: based on discussion at most recent WG, a concern for v17 migration is that there are breaking changes to incremental delivery support from prior alphas/experimental branches.

This PR reduces one of those changes, i.e. restores an entry point for allowing incremental delivery support with subscriptions.

### DIFF
--- a/src/execution/__tests__/flattenAsyncIterable-test.ts
+++ b/src/execution/__tests__/flattenAsyncIterable-test.ts
@@ -1,0 +1,219 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { expectPromise } from '../../__testUtils__/expectPromise.js';
+
+import { flattenAsyncIterable } from '../flattenAsyncIterable.js';
+
+describe('flattenAsyncIterable', () => {
+  it('flatten nested async generators', async () => {
+    async function* source() {
+      yield await Promise.resolve({
+        value: 1,
+        nestedIterable: (async function* nested(): AsyncGenerator<
+          number,
+          void,
+          void
+        > {
+          yield await Promise.resolve(1.1);
+          yield await Promise.resolve(1.2);
+        })(),
+      });
+      yield await Promise.resolve({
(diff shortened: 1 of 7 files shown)


JoviDeCroock (commented):
I personally don't feel like we should support incremental delivery in subscriptions. They are already push based so whether all fields take 10ms to calculate or 500ms shouldn't make that much of a difference imho.

This is one of those that I wouldn't feel strongly about leaving out in v17, stronger I would encourage us to leave it out. We should assess whether this is widely used before rolling back to this state. Mind linking to the incremental delivery WG or GraphQL WG that touched on subscriptions incremental delivery specifically?

If this is from the last GraphQL WG, I would reduce the scope of this backtracking work to prior payloads but not subscriptions.

JoviDeCroock:
Yes, I would def not add this back, it looks like a footgun and it just broadens the API surface even more.

yaacovCR:
I'm not super devoted to this change, but I assumed that structurally it raises the same issues as the response format changes... I would be happy to not need to merge it. It's handy to have around to know that everything still works (with the tweaks I included). I'll change it to draft for now!

benjie:
One of the guiding principles of stream/defer has been that the server may choose to ignore them if it feels that using the traditional format is more optimal. As such, so long as you don't outright reject requests that have stream/defer on subscriptions (or, if  you do, you do so via a validation rule that can be turned off) then we should be able to punt this without breaking backwards compatibility?

yaacovCR:
> One of the guiding principles of stream/defer has been that the server may choose to ignore them if it feels that using the traditional format is more optimal. As such, so long as you don't outright reject requests that have stream/defer on subscriptions (or, if you do, you do so via a validation rule that can be turned off) then we should be able to punt this without breaking backwards compatibility?

Closing this for now in light of this helpful feedback from @benjie as well as @JoviDeCroock comments further up.

PR #4315, opened 2024-12-19
convert executor to class

yaacovCR:
1. introduces Executor class so we can more easily add generic type parameters to the necessary functions
2. extracts generic Payload Publisher from the IncrementalPublisher class and makes Executor generic
3. adds legacy entrypoints

TODO: tests


yaacovCR:
Closing for now, use case was for #4316 => which may have to go in a different direction.

PR #4316, opened 2024-12-20
introduce legacy incremental delivery entrypoint

yaacovCR:
depends on #4315

add legacy incremental delivery format entrypoint:  `legacyExecuteIncrementally`

Motivation: based on discussion at most recent WG, a concern for v17 migration is that there are breaking changes to incremental delivery support from prior alphas/experimental branches.


yaacovCR:
Closing for now, this naive approach does not include the legacy verion of "filtering".

PR #4317, opened 2024-12-31
Allow configuration of the `ofType` introspection depth

### LINKED ISSUE #4321: Reduce introspection complexity programmatically
# Use case 

I'm trying to introspect[ Gitlab API schema ](https://gitlab.com/api/graphql) with [default introspection query ](https://graphql-js.org/api/function/getintrospectionquery/) .

Unfortunately Giltab API fails with the error: `Query has complexity of 220, which exceeds max complexity of 217`

I noticed that in a [default query ](https://github.com/graphql/graphql-js/blob/16.x.x/src/utilities/getIntrospectionQuery.ts#L140)we use a static construction like which takes up to 10 nested levels:
```
fragment TypeRef on __Type {
      kind
      name
      ofType {
        kind
        name
        ofType {
          kind
          name
          ofType {
            kind
            name
            ofType {
              kind
              name
              ofType {
                kind
                name
                ofType {
                  kind
                  name
                  ofType {
                    kind
                    name
                    ofType {
                      kind
                      name
                      ofType {
                        kind
                        name
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
```

However in case of Gitlab API I'm ok to intentionally reduce complexity.

So if I use postman and just remove one nested level out then introspection works and I can test it in postman.

Of course - I can store this schema in a local file and then work it just by loading. But problem is - there are tons of different tools ( especially from The Guild ) which relies on this module and introspect using this default query.

I don't see here conflicts with a spec if we some explicitly tell what is our limit for the max depth.

# Possible solution

If we introduce one more option depthLimit in IntrospectionOptions, Then we can just build mentioned above segment programmtically in a loop.

It may default to 10  as it is working now.

What do you think about this idea?

Thank you!

Nols1000:
This allows for a better configuration in case the server restricts the maximum query depth.

---

Added by @benjie:

- Fixes #4321

### DIFF
--- a/src/utilities/getIntrospectionQuery.ts
+++ b/src/utilities/getIntrospectionQuery.ts
@@ -90,16 +90,18 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
     return optionsWithDefault.experimentalDirectiveDeprecation ? str : '';
   }
   const oneOf = optionsWithDefault.oneOf ? 'isOneOf' : '';
-  function ofType(level = 9): string {
+  function ofType(level: number, indent: string): string {
     if (level <= 0) {
       return '';
     }
+    if (level > 100) {
+      throw new Error("Please set typeDepth to a reasonable value; the default is 9.");
+    }
     return `
-            ofType {
-              name
-              kind
-              ${ofType(level - 1)}
-            }`;
+${indent}ofType {
+${indent}  name
+${indent}  kind${ofType(level - 1, indent + "  ")}
+${indent}}`;
   }
 
   return `
@@ -174,8 +176,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
 
     fragment TypeRef on __Type {
       kind
-      name
-      ${ofType(optionsWithDefault.typeDepth)}
+      name${ofType(optionsWithDefault.typeDepth ?? 9, "      ")}
     }
   `;
 }
--- a/src/utilities/getIntrospectionQuery.ts
+++ b/src/utilities/getIntrospectionQuery.ts
@@ -179,7 +179,7 @@ ${indent}}`;
 
     fragment TypeRef on __Type {
       kind
-      name${ofType(optionsWithDefault.typeDepth ?? 9, '      ')}
+      name${ofType(optionsWithDefault.typeDepth, '      ')}
     }
   `;
 }
--- a/src/utilities/getIntrospectionQuery.ts
+++ b/src/utilities/getIntrospectionQuery.ts
@@ -96,7 +96,9 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
       return '';
     }
     if (level > 100) {
-      throw new Error('Please set typeDepth to a reasonable value; the default is 9.');
+      throw new Error(
+        'Please set typeDepth to a reasonable value between 0 and 100; the default is 9.',
+      );
     }
     return `
 ${indent}ofType {
--- a/src/utilities/getIntrospectionQuery.ts
+++ b/src/utilities/getIntrospectionQuery.ts
@@ -46,9 +46,10 @@ export interface IntrospectionOptions {
   oneOf?: boolean;
 
   /**
-   * How deep to recurse into nested types. Larger values will result in more
-   * accurate results, but have a higher load. Some servers might restrict the
-   * maximum query depth. If thats the case, try decreasing this value.
+   * How deep to recurse into nested types, larger values will result in more
+   * accurate results, but have a higher load on the server.
+   * Some servers might restrict the maximum query depth or complexity.
+   * If that's the case, try decreasing this value.
    *
    * Default: 9
    */
@@ -95,12 +96,12 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
       return '';
     }
     if (level > 100) {
-      throw new Error("Please set typeDepth to a reasonable value; the default is 9.");
+      throw new Error('Please set typeDepth to a reasonable value; the default is 9.');
     }
     return `
 ${indent}ofType {
 ${indent}  name
-${indent}  kind${ofType(level - 1, indent + "  ")}
+${indent}  kind${ofType(level - 1, indent + '  ')}
 ${indent}}`;
   }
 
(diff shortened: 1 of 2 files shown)


src/utilities/getIntrospectionQuery.ts
```diff
@@ -52,6 +61,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
     schemaDescription: false,
     inputValueDeprecation: false,
     oneOf: false,
+    typeDepth: 9,
```
saihaj, line 64: so this default is a breaking change. If we want this to land in v16 then we should do `Infinity` and in `v17` we can restrict the limits.

benjie: @saihaj I think you may have misunderstood which depth this relates to - it's the `ofType { ofType { ofType { ... } } }` depth of the generated query. We couldn't set that to infinity.

src/utilities/getIntrospectionQuery.ts
```diff
@@ -70,6 +80,17 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
...
+    if (level <= 0) {
+      return '';
+    }
+    return `
+            ofType {
+              name
+              kind
+              ${ofType(level - 1)}
+            }`;
+  }
```
benjie, line 93:
```suggestion
  function ofType(level: number, indent: string): string {
    if (level <= 0) {
      return '';
    }
    if (level > 100) {
      throw new Error("Please set typeDepth to a reasonable value; the default is 9.");
    }
    return `
${indent}ofType {
${indent}  name
${indent}  kind${ofType(level - 1, indent + "  ")}
${indent}}`;
  }
```

src/utilities/getIntrospectionQuery.ts
```diff
@@ -140,42 +161,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
     fragment TypeRef on __Type {
       kind
       name
```
benjie, line 163:
```suggestion
      name${ofType(optionsWithDefault.typeDepth, "      ")}
```

src/utilities/getIntrospectionQuery.ts
```diff
@@ -140,42 +161,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
...
-                      }
-                    }
-                  }
-                }
-              }
-            }
-          }
-        }
-      }
+      ${ofType(optionsWithDefault.typeDepth)}
```
benjie, line 164:
```suggestion
```

benjie (dismissed):
Let's keep this nicely formatted :)

Nols1000:
@benjie Thanks for the suggestions. I've commited them my branch. I think for the scope of this PR that is the right move. 

I'd also like to take this as an opportunity to discuss formatting the introspection query. I think it would be simpler to just keep the query in one line, without indentions. This could clean up the code and save a few byte over the wire. It would of course reduce the readability. Maybe we can document the query better, so that would become less of an issue.

benjie:
You can always minify it by parsing it and using a GraphQL minifier. I don't think our source code should concern itself with that too much - we should make it easy to read and edit. What sort of comments do you have in mind?

ThePlenkov:
@Nols1000  can you run please `npm run lint -- --fix` in your branch to get rid of a lint error?

ThePlenkov:
@Nols1000  also please cover this line with a test case:
image

You can see it here https://app.codecov.io/gh/graphql/graphql-js/pull/4317

Thanks!

I decided not to submit identical MR, so now also interested in introducing this change

src/utilities/getIntrospectionQuery.ts
```diff
@@ -38,6 +38,15 @@ export interface IntrospectionOptions {
    * Default: false
    */
   oneOf?: boolean;
+
+  /**
+   * How deep to recurse into nested types. Larger values will result in more
+   * accurate results, but have a higher load. Some servers might restrict the
+   * maximum query depth. If thats the case, try decreasing this value.
```
JoviDeCroock, line 45:
```suggestion
   * How deep to recurse into nested types, larger values will result in more
   * accurate results, but have a higher load on the server.
   * Some servers might restrict the maximum query depth or complexity.
   * If that's the case, try decreasing this value.
```

src/utilities/getIntrospectionQuery.ts
```diff
@@ -70,6 +80,19 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
     return optionsWithDefault.inputValueDeprecation ? str : '';
   }
   const oneOf = optionsWithDefault.oneOf ? 'isOneOf' : '';
+  function ofType(level: number, indent: string): string {
+    if (level <= 0) {
+      return '';
+    }
+    if (level > 100) {
+      throw new Error("Please set typeDepth to a reasonable value; the default is 9.");
```
JoviDeCroock, line 88: This line also needs a test

```suggestion
      throw new Error('Please set typeDepth to a reasonable value; the default is 9.');
```

src/utilities/getIntrospectionQuery.ts
```diff
@@ -70,6 +80,19 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
...
+    }
+    return `
+${indent}ofType {
+${indent}  name
+${indent}  kind${ofType(level - 1, indent + "  ")}
```
JoviDeCroock, line 93:
```suggestion
${indent}  kind${ofType(level - 1, indent + '  ')}
```

src/utilities/getIntrospectionQuery.ts
```diff
@@ -139,43 +162,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
...
-                      }
-                    }
-                  }
-                }
-              }
-            }
-          }
-        }
-      }
+      name${ofType(optionsWithDefault.typeDepth ?? 9, "      ")}
```
JoviDeCroock, line 165:
```suggestion
      name${ofType(optionsWithDefault.typeDepth ?? 9, '      ')}
```

JoviDeCroock (approved):
This looks like a good feature to have, great work, left some comments but pending those changes this should be good to go

Nols1000:
Hi @ThePlenkov, I'll add tests tomorrow.

Nols1000:
@benjie At least for me it's not 100% clear how the query works. I would have expected more documentation e.g. here [https://graphql.org/learn/introspection/](https://graphql.org/learn/introspection/). For me a breakdown of this introspection query would have been a great learning resource.

benjie (approved):
Looks good to me :+1:

PR #4319, opened 2025-01-07
introduce transformer from latest incremental format to legacy format

yaacovCR:
The last attempt to support the both the current and legacy incremental delivery format in #4316 was a flop.[^1] This PR introduces a new one:

```js
import { legacyExecuteIncrementally } from `graphql`;

const result = legacyExecuteIncrementally( ... );
```

The `legacyExecuteIncrementally()` function works just like `experimentalExecuteIncrementally()` from the latest graphql v17 alpha but returns the legacy format.[^2]

The plan if this approach meets goals would probably be to introduce this as a separate library, perhaps capable of even more complex transformations. For now, you can use it instead of the latest graphql v17 alpha via canary release `graphql@canary-pr-4319`

[^1]: In that version we built a non-standard execution plan with one execution group per delivery group where the different execution groups had duplicated fields, instead of the spiffy non-duplicated version of the current format.

    Why did this not work? 
    
    Consider the following operation:
    
    ```graphql
    query {
      hero { nonNullName }
      ... @defer { hero { name } }
    }
    ```
        
    Assuming `Hero.nonNullName` is a non-nullable field and happens to return `null`, the `null` will "bubble up" to `hero`, i.e. the initial response for our new algorithm will simply be:
    
    ```json
    {
      "data": {
        "hero": null,
      },
      "errors": [
        {
          "message": "Cannot return null for non-nullable field Hero.nonNullName.",
          "locations": [{ "line": 4, "column": 11 }],
          "path": ["hero", "nonNullName"],
        },
      ],
    }
    ```
    
    No deferred fragment will be delivered, because there is no where for a client to store this data => without undoing a received null. Our new algorithm manages this by simply discarding any requests to send deferred data whenever a `null` is returned.

    This could be kind of tricky; after all, the deferred fragment is declared at the root, before `hero` is encountered. The way this works is the all overlapping fields simultaneously at each level, and uses the information gathered to decide when to initiate deferral in a separate "execution plan step". For the root fields in our example, no special action is taken even though there is a deferred fragment at root, because the deferred `hero` field overlaps with the non-deferred `hero`. When `hero` returns `null`, we discard the request for nested deferred data.

    If deferral is not initiated, though, our algorithm still saves details about any requested deferral on all of the child selections.  So for the non-error case, when planning the subfields of `hero`,  the algorithm can still "remember" that `Hero.name` was requested as deferred and deliver just that necessary bit.
    
    In #4316, we hacked that "execution plan step" to create that separate deferred fragment as if that overlap didn't occur, but we forgot that this will break the previous case! If we kick off a deferred fragment separately at root, we will return that fragment even if the initial result returns `null`.

[^2]: `legacyExecuteIncrementally()` uses `experimentalExecuteIncrementally()` under the hood, transforming the resulting latest format into the legacy format. If [the new type of null-bubbling introduced within the latest format](https://github.com/graphql/defer-stream-wg/discussions/69) is encountered, the error will bubble up to the root of the deferred fragment.


yaacovCR:
All defer/stream tests currently passing except:

```graphql
query {
  nestedObject {
    nestedFriendList @stream(initialCount: 0) {
      id
    }
  }
  nestedObject {
    ... @defer {
      nestedFriendList @stream(initialCount: 0) {
        id
        name
      }
    }
  }
}
```

Supporting this case is unfortunately tricky with the current approach.

yaacovCR:

yaacovCR:

yaacovCR:

yaacovCR:

yaacovCR:
published this as an external package at:

npm: https://www.npmjs.com/package/@yaacovcr/transform
github: https://github.com/yaacovCR/transform

yaacovCR:
> The last attempt to support the both the current and legacy incremental delivery format in #4316 was a flop.

I have read through the explanation, and I'm not 100% sure that it was wrong at the time (although I suspect it was) but it seems wrong now. We keep track of when to drop deferred fragments  by taking a look at which paths were nulled (if using filtering), as implemented at the moment on the `next` branch or which deferred fragments were returned (if using a wrapped GraphQL result containing the field level result with any new deferred fragments, as implemented in the most recent prior version).

So this should work. Proof of concept is in #4546

PR #4323, opened 2025-01-13
further refine incremental result types

yaacovCR:
continue work with #4313

Fixes the types for Subsequent Execution Results. Currently, the generic argument for subsequent results is passed to deferred and streamed results. By providing two different generic arguments, we can get properly types individual members of the incremental array.

### DIFF
--- a/src/execution/types.ts
+++ b/src/execution/types.ts
@@ -36,12 +36,13 @@ export interface FormattedExecutionResult<
 
 export interface ExperimentalIncrementalExecutionResults<
   TInitial = ObjMap<unknown>,
-  TSubsequent = unknown,
+  TData = ObjMap<unknown>,
+  TItems = ReadonlyArray<unknown>,
   TExtensions = ObjMap<unknown>,
 > {
   initialResult: InitialIncrementalExecutionResult<TInitial, TExtensions>;
   subsequentResults: AsyncGenerator<
-    SubsequentIncrementalExecutionResult<TSubsequent, TExtensions>,
+    SubsequentIncrementalExecutionResult<TData, TItems, TExtensions>,
     void,
     void
   >;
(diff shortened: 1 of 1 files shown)


src/execution/types.ts
```diff
@@ -36,12 +36,13 @@ export interface FormattedExecutionResult<
 
 export interface ExperimentalIncrementalExecutionResults<
   TInitial = ObjMap<unknown>,
-  TSubsequent = unknown,
+  TData = ObjMap<unknown>,
+  TItems = ReadonlyArray<unknown>,
```
JoviDeCroock, line 40: Not sure what we are trying to achieve here but should we be explicit about what these types can be?

i.e. if Data is always going to be an object should we use `extends Record<string, unknown>`

Furthermore `TItems` sounds like it's always going to be an array so why don't we make this a `TIncrementalItem` and wrap it in the `ReadonlyArray` ourselves? Is the purpose of this wrapping so that folks can avoid passing in a readonly one? If so then we still need to push people into the happy path by constraining the inputs.

yaacovCR: Hmmm, basically, I ran into type conflicts when working on the transformer in #4319 because TData was mapped to TItems, and this was a simple fix for that.

Taking a step back, I think we should honestly at this point remove all of the generics from the incremental results @robrichard . As far as I understand, these are used by tools like GraphQL CodeGen to provide a typed experience based on the query, but it is possible/likely that typed access to the incremental data will be from the merged object anyway, after all the incremental bits have been merged in, and we don't need to provide these generics at all.

I think we may have simply started off supplying them because the IncrementalDeferResult is modelled after an ExecutionResult, and the ExecutionResult provides this. @JoviDeCroock @robrichard @benjie  any thoughts?

JoviDeCroock: I think it's an anti-pattern to leave untyped bits, I can see a world where the incremental result is typed as

```ts
type PossibleIncrementalResults = 
  | UserFragment
  | TodoFragment
```

yaacovCR: Just a factoid to add to this, because of inlining we are not 100% on the shape of the initial result either.

JoviDeCroock: If the shape of the response, be that initial isn't predictable. Isn't that a big deviation from the typed nature of graphql?

benjie: The shape of the response is as predictable as a response using `@skip` and `@include` where the value of the variables isn't factored into codegen.

JoviDeCroock: Gotcha, yes, I thought I was missing something so

> because of inlining we are not 100% on the shape of the initial result either.

is just the addition of i.e. optional properties

yaacovCR: c72159a45

yaacovCR:
This PR fixes the types for Subsequent Execution Results by splitting up the generics between the deferred result data and the streamed item type, so that the constituent types can now be properly typed.

yaacovCR:
closing in favor of fresher #4560

PR #4324, opened 2025-01-13
Backport GraphQLError toJSON method to v15

enisdenjo:
Eases up the transition to future v17 that has completely removed the `formatError` utility.


src/error/GraphQLError.d.ts
```diff
@@ -4,6 +4,9 @@ import { ASTNode } from '../language/ast';
 import { Source } from '../language/source';
 import { SourceLocation } from '../language/location';
 
+// eslint-disable-next-line import/no-cycle
+import { GraphQLFormattedError } from './formatError';
```
benjie, line 8: Might be better to move the definition of GraphQLFormattedError here to avoid the cycle; then you can re-export it from `formatError` for backwards compatibility.

benjie (approved):

src/error/GraphQLError.d.ts
enisdenjo: I'll have to move both formatError and the type, sure - give me a sec.

PR #4329, opened 2025-01-16
fix sidebar instability

yaacovCR:
when navigating to https://www.graphql-js.org/api-v16/graphql/ using the top menu bar, the sidebar content expands with the individual mdx files from the api-v16 folder, as if the api/v16 option had been in the sidebar in the first place.

This PR adds the API/v16 to the sidebar (in addition to the placement within the menu bar) to avoid this instability.

image

### DIFF
--- a/website/pages/_meta.ts
+++ b/website/pages/_meta.ts
@@ -21,10 +21,15 @@ const meta = {
   'defer-stream': '',
   '-- 3': {
     type: 'separator',
+    title: 'API',
+  },
+  'api-v16': 'v16',
+  '-- 4': {
+    type: 'separator',
     title: 'FAQ',
   },
   'going-to-production': '',
-  'api-v16': {
+  'api-menu': {
     type: 'menu',
     title: 'API',
     items: {


website/pages/_meta.ts
```diff
@@ -20,11 +20,16 @@ const meta = {
...
+    title: 'API',
+  },
+  'api-v16': 'v16',
+  '-- 4': {
     type: 'separator',
     title: 'FAQ',
   },
   'going-to-production': '',
-  'api-v16': {
+  'api-menu': {
```
JoviDeCroock, line 32: The goal was to not so this, need to reach out to Dima on how to reset this context.

yaacovCR: Is the final goal to have the subheadings within the API show up somewhere?

JoviDeCroock: Only when you surf to the api docs else that would not scale to i.e. also having v17

JoviDeCroock:
Fixed in https://github.com/graphql/graphql-js/pull/4331

PR #4336, opened 2025-01-30
add deprecated note to assertValidExecutionArguments

yaacovCR:
to notify theoretical users via deep import of imminent removal


yaacovCR:
Removed polish label as it is observable!

PR #4192, opened 2024-09-14
[WIP] Experimental support for semantic-non-null

benjie:
**UPDATE 2025-03-27**: updated to latest version of spec proposal:

- https://github.com/graphql/graphql-spec/pull/1065

and rebased on top of:

- #4364

<details>
<summary>Previous description</summary>

**DISCLAIMER**: I'm _very_ jetlagged right now, and it's late. This is almost certainly full of issues.

The main aim of this PR is so we can start experimenting with both semantic non-null and disabling error bubbling.

Semantic non-null has been added using the asterisk (`Int*`) syntax, because that was easiest (and I'm doing this from an airport lounge whilst I wait for boarding...). I know this isn't what we've most recently discussed, but it's a smaller lift and means we can start experimenting.

To disable error bubbling, pass `errorPropagation: false` as part of the execution args: `graphql({ schema, source, errorPropagation: false })`.

To see the full types, pass `nullability: 'FULL'` to `getIntrospectionQuery`, e.g. `getIntrospectionQuery({ nullability: 'FULL' })`. (I don't like this current enum, but it'll do for now.)

Note: this implementation does NOT align with my previous thoughts exactly, and it is definitely not final.

</details>

TODO:

- [x] Address TODO comments, most notably the memoization ones
- [x] Reflect in spec
- [ ] Finalize naming
- [ ] Add reasonable descriptions to introspection
- [ ] Add lots of tests

PR #4337, opened 2025-02-02
Semantic nullability rfc implementation

JoviDeCroock:
## Introduction

This is the working implementation for semantic nullability, this originates from work done by @twof and @benjie in https://github.com/graphql/graphql-js/pull/4271 and https://github.com/graphql/graphql-js/pull/4192. This PR intends to supersede this work and drive it home.

When Semantic nullability is on we change the semantics of the schema-type system, currently there are two types of nullability, it's either nullable or not. When semantic nullability is introduced we expand this, any field can be null, non-null or semantically null.
Semantically null fields can only return a `null` value when there's an error associated with the field, when we return null and there's no error associated we'll bubble up the error to the nearest concretely nullable field.

Semantic nullability will be introduced either by means of a directive in the SDL `@SemanticNullability` which converts the SDL document to use semantic nullability or by switching on the `parse` configuration.

## Documentation

One of the most important parts for adoption would be to adequately document this feature and rope in folks maintaining schema builder libraries.

## Open questions

Currently we employ quite a bit of new semantics to achieve the scenario, we introduce both a  semantic null and non-null type. This might confuse some people in terms of explicitness but helps a lot with the `print` implementation as it clearly delineates the new behavior types.

My proposal would be to only introduce 1 new AST type which would be SEMANTIC_NON_NULLABLE, the complication would arise for the printer but it would be the least changes for folks creating schema builders and other tools in our ecosystem.

Should we be more concrete about input-types not being able to express semantic nullability?

There is a lot of work still to be done on test coverage but I wanted to raise the PR already to let yall know that I'm working on it. This is a place for discussion as well, any alternatives can be raised, I am still familiarizing myself more on the matter and will probably be doing so for a few days.

## Update (5th of February 2025)

We merged https://github.com/graphql/graphql-js/pull/4337/commits/869ca46f24446ef152b54165dc866f428aa2dfea which reduces the increase in AST-nodes to 1, this helps us a bit in terms of introducing new concepts and backwards compatibility of AST traversal. In an ideal world we would only introduce a new syntax so this becomes easier for printing as well and is backwards compatible when it comes to SDL.

## Update (12th of February 2025)

I've made all CI checks pass and the more I run into `print` issues I am convinced that we should not break the SDL syntax. Having one common symbol for nullable, Semantic non-nullable and non-nullable would go a long way in making both the spec more comprehensible as well as staying backwards compatible regardless of the presence of the document directive.

With specialized syntax we would need any print/parse flags and instead we could allow directive-less usage of semantic non-null in future majors.

I also took the opportunity to split our `SchemaOutputType` from the `InputType` as these become increasingly confusing with semantic nullability present. We were implicitly allowing input-types and variable types to be semantically nullable which now correctly only allows the return type of fields to be semantically nullable.


yaacovCR:
Amazing you are taking this up. I posted https://github.com/graphql/graphql-js/pull/4338 which is just the latest hacking I did on the work done by @benjie and @twof, tests seem to pass there with just the one wrapping type.

yaacovCR:
Don't have the bandwidth to progress there further, just posting in case that helps.

JoviDeCroock:
@yaacovCR That does seem to be the inverse of what I'm proposing, I would only have a wrapping type for semantic-nullability which in your PR seems to make the newly introduced type the nullable one, I think that for clarity of the community we would do good in making the SEMANTIC_NON_NULL the only newly introduced AST node. I have yet to explore the feasibility though, I got caught up in a few things

yaacovCR:
@JoviDeCroock I find this feature very difficult in terms of the upgrade story => definitely part of the challenge of making this work. I see the advantage in having the new wrapper type be the new `SemanticNonNull` wrapping type, but I do see a parallel here in the debate within the SDL in terms of whether to introduce a new notation such as `Type*` to indicate SemanticNonNull versus `Type?` to indicate `SemanticNull`.

My pre-existing PR, now rebased in #4338, like this PR and like #4271 (as opposed to #4192) is adopting `Type?` for the SDL, so I assumed that for consistency, we would do the same in terms of AST nodes. As you demonstrate in #4339, that is not strictly necessary, we could come to different answers in terms of the SDL and the AST nodes in terms of what the unadorned type means, but that -- to me -- adds an additional layer of complexity to this already complex story.

Either way, I think the upgrade story is a bit confusing, however. For example, we have a function called `getNullableType()` which takes a possible type possible wrapped by a `GraphQLNonNull` and returns the unadorned type. With the approach in this PR, it can still return an unadorned type. With my approach in #4338, this returns a type that is nullable in the old sense, but in the new sense, it is unadorned, so it is not semantically nullable, => confusing!

My feedback to the spec folks if they are looking for it is that I find it extremely difficult to reason constantly in terms of the two different types of nullability and that we should separate these concepts more broadly into different terms, for example, perhaps `Catching` fields and `Optional` fields, where `Catching` fields can be null for an error, and `Optional` fields are semantically nullable. And then if we wanted to have the base unadorned type be catching and non-optional, our two wrapping types would be something like `GraphQLNonCatching` and `GraphQLOptional`.

benjie:
I'd advise against using "optional" on output types, it implies that the field itself can be omitted, rather than that it can be null (i.e. the difference between `type T = {foo?: number}` and `type T = {foo: number | null}` in TypeScript if you pretend that `undefined` and `null` are the same value :wink: )

src/execution/__tests__/semantic-nullability-test.ts
```diff
@@ -0,0 +1,219 @@
...
+      document,
+      rootValue: data,
+    });
+
+    expect(result).to.deep.equal({
+      data: {
+        a: 'Apple',
+      },
+    });
+  });
```
twof, line 218: I don't think these test cases are necessary anymore now that `SemanticNullable` has been removed.

twof:
> when we return null and there's no error associated we'll bubble up the error to the nearest concretely nullable field.
I think we corrected this in Discord, and it looks like SemanticNonNullable has the correct behavior in code.

src/execution/execute.ts
```diff
@@ -152,6 +154,13 @@ export interface ExecutionArgs {
   fieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
   typeResolver?: Maybe<GraphQLTypeResolver<any, any>>;
   subscribeFieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
+  /**
+   * Set to `false` to disable error propagation. Experimental.
+   * TODO: describe what this does
```
twof, line 159: Outstanding TODO. We might want to say something here about how this is being used and include a link to the RFC or spec edits.

src/language/__tests__/schema-printer-test.ts
```diff
@@ -180,4 +180,41 @@ describe('Printer: SDL document', () => {
...
+    ).to.equal(dedent`MyType`);
+  });
+
+  it('prints SemanticNullableType', () => {
+    expect(
+      print(parseType('MyType?', { allowSemanticNullability: true }), {
+        useSemanticNullability: true,
+      }),
+    ).to.equal(dedent`MyType?`);
+  });
```
twof, line 198: I think anything printed with `allowSemanticNullability`, will need to be prefixed with the document directive (`@SemanticNullability`) because otherwise the contents are liable to misinterpretation.

src/type/introspection.ts
```diff
@@ -204,6 +206,40 @@ export const __DirectiveLocation: GraphQLEnumType = new GraphQLEnumType({
...
+  AUTO = 'AUTO',
+  TRADITIONAL = 'TRADITIONAL',
+  SEMANTIC = 'SEMANTIC',
+  FULL = 'FULL',
+}
+
+// TODO: rename
+export const __TypeNullability: GraphQLEnumType = new GraphQLEnumType({
+  name: '__TypeNullability',
+  description: 'TODO',
```
twof, line 220: Outstanding TODO

src/language/__tests__/schema-printer-test.ts
benjie: It's less _allow_ and more just on or off, right? What it's set to changes how a type is interpreted:

|      | Traditional  | @semanticNullability |
| ---- | ------------ | -------------------- |
| Int? | ERROR!       | Int                  |
| Int  | Int          | SemanticNonNull<Int> |
| Int! | NonNull<Int> | NonNull<Int>         |

Document directives must come at the top of the document, and in my mental model they effectively change the "mode" of the rest of the parsing. You might think of other document directives that might change the mode, e.g. `@commentsAreDescriptions` to re-enable the ancient parsing feature where comments (rather than strings) before fields would become descriptions; or `@prefixedDirectives` which might put directives _before_ things rather than _after_. (I'm not proposing either of these be made real, I'm just trying to show that the space for document directives is not just a set of size 1.)

Once you hit your first non-ignored non-document-directive token the mode is then frozen, and you can pass that mode into `parseType` and other parsing related functions.

Thus I think it would be more like:

```suggestion
  it('prints SemanticNullableType', () => {
    expect(
      print(parseType('MyType?', { mode: { semanticNullability: true }}), {
        useSemanticNullability: true,
      }),
    ).to.equal(dedent`MyType?`);
  });
```

I'm not sure "mode" is really the right name to use for the result of applying document level directives, suggestions welcome. It could be that we just take the document-level directives and store their literal values into the object; for example:

```graphql
@semanticNullability
@behavior(onError: PROPAGATE, onDeprecated: ERROR)
@compositeScalars(serialize: true)

{ __typename }
```

might have a "mode" of:

```js
const parseOptions = {
  mode: {
    semanticNullability: {},
    behavior: { onError: "PROPAGATE", onDeprecated: "ERROR" },
    compositeScalars: { serialize: true },
  }
}
```

benjie: (Note that this mode should also be an output of parsing, e.g. we might use a document directive in an executable document to enable `@noErrorPropagation` - and we'll need to see that. Originally I thought that `@noErrorPropagation` would go on the operation itself rather than the document, but the issue with that is that fragments will have two modes if they're used in two operations with different `@noErrorPropagation` settings, so codegen becomes complicated. Instead, making it a document directive means it applies to all operations and fragments in a single document in unison.)

src/type/__tests__/introspection-test.ts
```diff
@@ -640,6 +659,37 @@ describe('Introspection', () => {
...
+              specifiedByURL: null,
+              fields: null,
+              inputFields: null,
+              interfaces: null,
+              enumValues: [
+                {
+                  name: 'AUTO',
+                  isDeprecated: false,
+                  deprecationReason: null,
+                },
```
twof, line 674: We're going to cut 'AUTO' and 'SEMANTIC' introspection modes

src/type/introspection.ts
```diff
@@ -204,6 +206,40 @@ export const __DirectiveLocation: GraphQLEnumType = new GraphQLEnumType({
   },
 });
 
+// TODO: rename enum and options
+enum TypeNullability {
+  AUTO = 'AUTO',
+  TRADITIONAL = 'TRADITIONAL',
+  SEMANTIC = 'SEMANTIC',
+  FULL = 'FULL',
+}
```
twof, line 215: Already noted this in the tests, but we can get rid of Semantic and Auto here.

src/utilities/__tests__/printSchema-test.ts
```diff
@@ -779,7 +782,7 @@ describe('Type System Printer', () => {
         name: String!
         description: String
         args(includeDeprecated: Boolean = false): [__InputValue!]!
-        type: __Type!
+        type(nullability: __TypeNullability! = AUTO): __Type!
```
twof, line 785: The default mode should be Traditional since we don't want to expose unsuspecting clients to `SemanticallyNonNull` fields

src/utilities/getIntrospectionQuery.ts
```diff
@@ -52,6 +63,7 @@ export function getIntrospectionQuery(options?: IntrospectionOptions): string {
     schemaDescription: false,
     inputValueDeprecation: false,
     oneOf: false,
+    nullability: null,
```
twof, line 66: If I'm understanding this right, I think this should be `'TRADITIONAL' `

twof (changes requested):
Great work! Thanks for putting on the finishing touches. I've left some comments, and it looks like we're missing some code coverage, but after that I think we'll be good to go.

twof:
Oh one thing I missed. I think we need to define the document directive in `directives.ts`

src/utilities/printSchema.ts
```diff
@@ -36,6 +36,8 @@ import type { GraphQLSchema } from '../type/schema';
 
 import { astFromValue } from './astFromValue';
 
+// TODO: we might need to add a flag to print the schema with semantic-non-null types
```
JoviDeCroock, line 39: CC @twof @benjie how do we expect to tackle this, document directives aren't allowed as it stands and we'd need to store this on the schema itself. We do have the possibility of a schema directive but that might span different documents.

twof: Yeah we'd need to add a new directive scope

src/language/ast.ts
```diff
@@ -519,9 +521,20 @@ export interface ConstDirectiveNode {
...
+export interface SemanticNonNullTypeNode {
+  readonly kind: Kind.SEMANTIC_NON_NULL_TYPE;
+  readonly loc?: Location;
+  readonly type: NamedTypeNode | ListTypeNode;
+}
+
 /** Type Reference */
 
 export type TypeNode = NamedTypeNode | ListTypeNode | NonNullTypeNode;
+export type SchemaOutputTypeNode =
```
twof, line 533: This is good. I'm glad input and output are being separated. They already have different behavior.

src/language/parser.ts
```diff
@@ -258,6 +271,16 @@ export class Parser {
...
   parseDefinition(): DefinitionNode {
+    const directives = this.parseDirectives(false);
+    // If a document-level SemanticNullability directive exists as
+    // the first element in a document, then all parsing will
+    // happen in SemanticNullability mode.
+    for (const directive of directives) {
+      if (directive.name.value === 'SemanticNullability') {
+        this._options.allowSemanticNullability = true;
+      }
+    }
```
twof, line 282: I think this would also be the first directive that applies to the parse phase right?

twof: Do you foresee that causing issues?

JoviDeCroock: I don't think it would no

PR #4339, opened 2025-02-02
Propose single wrapping type

JoviDeCroock:
This reduces the new AST-nodes to only be for the newly introduced type, this does make it so that when we invoke `print` we have to rely on the user to either specify that we're in semantic nullability mode _or_ we could do a pre-traverse and when we enter a node with semantic-non-null we toggle it on ourselves.

The main reasoning behind removing the new name for our existing null type is that I would prefer to be backwards compatible in terms of schema structure. This because it might become complex for people to reason about composed schemas, i.e. a lot of individually parsed schemas that later on compose into a larger one.

I know that _technically_ this is covered because in the classic ones we'll have the non wrapped null type and in the modern ones we'll have the semantic nullable wrapped type. For schema-builders like pothos and others I think this is rather complex to reason about _and_ to supply us with. I would instead choose to absorb this complexity in the feature and stay backwards compatible.

This also sets us up for the SDL not being a breaking change, we only add one AST-type, what's left now is to settle on a semantic non-null syntax and making everything backwards compatible.

### DIFF
--- a/src/execution/__tests__/semantic-nullability-test.ts
+++ b/src/execution/__tests__/semantic-nullability-test.ts
@@ -10,7 +10,6 @@ import {
   GraphQLNonNull,
   GraphQLObjectType,
   GraphQLSemanticNonNull,
-  GraphQLSemanticNullable,
 } from '../../type/definition';
 import { GraphQLString } from '../../type/scalars';
 import { GraphQLSchema } from '../../type/schema';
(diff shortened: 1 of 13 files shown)


twof (approved):
I think this works, and I definitely prefer this over having two types. Go ahead an merge it into your other branch, and I'll give you a more thorough review of that one tomorrow.

JoviDeCroock:
@twof my main "desire" would be to only introduce new syntax for the new type as well but I feel that is a more sensitive matter at this moment in time so I'll leave it at that and merge the descriptions as well

PR #4340, opened 2025-02-04
Improve flow of documentation around GraphiQL

benjie:
See graphql/graphql.github.io#1951

For someone following the tutorial they may well get as far as running the server (`node server.js`) and then attempt to visit their new API and get confused because they're met with an error such as `{"errors":[{"message":"Missing query"}]}`.

This PR adds some joining text to make it clear this is the expected outcome, and they must read on to get the GraphiQL IDE set up so that they can write queries.


benjie:
(Copied from https://github.com/graphql/graphql.github.io/pull/1956)

yaacovCR (approved):

PR #4342, opened 2025-02-07
Validate against `@stream` on different instances of the same field

robrichard:
See discussion https://github.com/graphql/defer-stream-wg/discussions/100

### DIFF
--- a/src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
+++ b/src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
@@ -707,12 +707,14 @@ function findConflict(
 
   const directives1 = node1.directives ?? [];
   const directives2 = node2.directives ?? [];
-  if (!sameStreams(directives1, varMap1, directives2, varMap2)) {
-    return [
-      [responseName, 'they have differing stream directives'],
-      [node1],
-      [node2],
-    ];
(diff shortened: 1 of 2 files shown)


src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
```diff
@@ -830,28 +832,27 @@ function getStreamDirective(
...
     // check if both fields have equivalent streams
-    return sameArguments(
-      stream1.arguments,
-      varMap1,
-      stream2.arguments,
-      varMap2,
-    );
+    if (sameArguments(stream1.arguments, varMap1, stream2.arguments, varMap2)) {
+      // This was allowed in previous alpha versions of `graphql-js`.
+      return 'they have differing stream directives. See https://github.com/graphql/defer-stream-wg/discussions/100';
```
yaacovCR, line 850: nit: this seemingly should be 'overlapping stream directives', we don't know if they differ in terms of selection sets.

yaacovCR (commented):
left one small error message suggestion, otherwise looks good.

src/validation/rules/OverlappingFieldsCanBeMergedRule.ts
robrichard: @yaacovCR make sense, updated.

JoviDeCroock (approved):

PR #4345, opened 2025-02-16
Remove process.env check

JoviDeCroock:
This is an alternative approach to removing our `instanceof` check, when we detect a mismatched version of GraphQL we throw an error regardless of us being in development or production. This is justified because multiple GraphQL version will lead to weird behaviour either way, rather than embedding the whole string we link to our docs where the user can find steps for `pnpm`, ... we can keep expanding this.


src/jsutils/instanceOf.ts
```diff
@@ -1,56 +1,28 @@
...
+        ? value[Symbol.toStringTag]
+        : value.constructor?.name;
+    if (className === valueClassName) {
+      throw new Error(
+        `Multiple GraphQL instances detected, Cannot use ${className} from another module or realm. Read more at https://graphql-js.org/errors#1`,
+      );
+    }
+  }
+  return false;
+}
```
benjie, line 25: I think this is a really hot path, are we sure we want to add more logic to it in production?

src/jsutils/instanceOf.ts
```diff
@@ -1,56 +1,28 @@
...
+  if (typeof value === 'object' && value !== null) {
+    const className = constructor.prototype[Symbol.toStringTag];
+    const valueClassName =
+      // We still need to support constructor's name to detect conflicts with older versions of this library.
+      Symbol.toStringTag in value
+        ? value[Symbol.toStringTag]
+        : value.constructor?.name;
+    if (className === valueClassName) {
+      throw new Error(
+        `Multiple GraphQL instances detected, Cannot use ${className} from another module or realm. Read more at https://graphql-js.org/errors#1`,
```
benjie, line 20: I recommend that you don't use monotonically increasing error numbers as it will cause conflicts when you have multiple branches being worked on concurrently and introducing new errors. I also recommend that you use a page per error otherwise the singular error page could get overwhelming (and fixing it later may break existing links in deployed applications).

At Graphile we use our "errored" domain for errors, e.g. https://err.red/gev2 - the first character indicates which website it goes to (`g` for `https://grafast.org/errors/`) and the `ev2` is a code for the error which links to the relevant page under that. This further allows adding sub-anchors such as `https://err.red/gev2#stream`.

src/jsutils/instanceOf.ts
JoviDeCroock: The alternative is that we completely scrap `instanceof` and actually make multiple versions of `graphql` work alongside each other. The `process.env` check has caused so much grievances

benjie: > actually make multiple versions of graphql work alongside each other

I, personally, think this would be a mistake. What if one supports oneof and the other doesn't? Or stream/defer? Fragment arguments?

> The benchmarks showed no signs of serious regressions

Great; glad we found my concerns unwarranted :) Carry on :)

yaacovCR: On my machine, I get a performance regression with this change:

image

yaacovCR:
@github-actions run-benchmark

src/jsutils/instanceOf.ts
yaacovCR: Running the benchmark github action...

PR #4348, opened 2025-02-21
Add support for @experimental_disableErrorPropagation

martinbonnin:
~This pull request adds support for `@onError(action: NULL)` to disable error propagation for error aware clients:~

This pull request adds support for `@experimental_disableErrorPropagation` to disable error propagation for error aware clients:

```graphql
"""
Disables error propagation.
"""
directive @experimental_disableErrorPropagation on QUERY | MUTATION | SUBSCRIPTION
```

I'm not super used to write TypeScript, feel free to amend the PR as needed but I figured it'd be good to have.

The logic is unconditional. The matching [graphql-java](https://github.com/graphql-java/graphql-java/pull/3772) PR has a specific opt-in flag so that it's not enabled by accident in the very unlikely event that a schema already contains a matching directive. Let me know if this is an issue.

Many thanks @JoviDeCroock for pointing me in the right direction 🙏

See https://github.com/graphql/nullability-wg/discussions/85
See https://github.com/graphql/graphql-spec/pull/1050


src/type/directives.ts
```diff
@@ -276,6 +281,54 @@ export const GraphQLOneOfDirective: GraphQLDirective = new GraphQLDirective({
...
+export const ErrorAction = {
+  PROPAGATE: 'PROPAGATE' as const,
+  NULL: 'NULL' as const,
+} as const;
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type ErrorAction = (typeof ErrorAction)[keyof typeof ErrorAction];
+
+export const _ErrorAction = new GraphQLEnumType({
+  name: '_ErrorAction',
```
martinbonnin, line 296: This is `_ErrorAction` and not `__ErrorAction` because I get a validation error else. But ultimately, I'd vote for this to be `__ErrorAction`

src/execution/execute.ts
```diff
@@ -170,6 +174,7 @@ export interface ExecutionContext {
   abortSignalListener: AbortSignalListener | undefined;
   completed: boolean;
   cancellableStreams: Set<CancellableStreamRecord> | undefined;
+  propagateErrors: boolean;
```
JoviDeCroock, line 177: Should we allow people to toggle this on, i.e. allow for a GraphQL server to say we don't allow for this at the moment as this is an experimental feature? My inclination is that if we PR this to v16 we should default to off and v17 we could just do it this way, kind of want to spark the conversation here.

src/type/directives.ts
JoviDeCroock: What's the validation error? Sorry can't run the code atm

JoviDeCroock (approved):
Looks good to me, a few formatting things but those would be nitpicks. Is this good to go from a WG standpoint CC @twof @benjie

src/execution/execute.ts
martinbonnin: We added such a flag in the [graphql-java PR](https://github.com/graphql-java/graphql-java/pull/3772). I'm generally in favor. Not too much for potential clashes, I think those are very very low probability but more for signaling the experimental status of the directive.

src/type/directives.ts
martinbonnin: Complains that types starting with `__` are reserved for builtin types. Which makes sense but since it is not standard yet... But also, it might become standard...

benjie:
We've not really debated what this should be yet, I'd propose going with an argument-less and obviously experimental directive such as `@experimental_disableErrorPropagation`. We may also want the directive to have an impact on introspection (for example returning semantic non-null types rather than strictly non-null types for output fields), so the proposals may or may not be tied together - using an explicitly experimental directive will allow people to play with the execution behavior without committing us to the result shape of introspection/etc.

src/execution/execute.ts
```diff
@@ -976,7 +988,7 @@ function handleFieldError(
 
   // If the field type is non-nullable, then it is resolved without any
   // protection from errors, however it still properly locates the error.
-  if (isNonNullType(returnType)) {
+  if (exeContext.propagateErrors && isNonNullType(returnType)) {
```
benjie, line 991: This is almost exactly the same as I had in #4192 except you called it propagateErrors whereas I called it `errorPropagation` :laughing: 

https://github.com/graphql/graphql-js/blob/eb9b6c8a77193fc5ed3685d1e95d32b99c9df9d9/src/execution/execute.ts#L609

(I don't have a particular preference between the terms.)

src/type/directives.ts
benjie: You need to add it to the introspection types I think:

https://github.com/graphql/graphql-js/blob/eb9b6c8a77193fc5ed3685d1e95d32b99c9df9d9/src/type/introspection.ts#L644-L655

That's where it should be defined too.

benjie: It's an interesting one because it's _reserved_ but it's not actually part of _introspection_. So it should either become a well known (like `String`, `ID`, `Int`, `Boolean`, `Float`) or it should be reserved (`__`) which is currently "reserved for introspection". Since it can't be populated by a variable we don't need the type name in a document so it's not super important, but also feels weird to have an `__` type in a directive that we're expecting users to use heavily.

martinbonnin: >  it should be reserved (__) which is currently "reserved for introspection".

[TIL](https://spec.graphql.org/draft/#sel-IAFNZVDDABBBD0hT). I always thought of `__` as reserved irrespective of introspection. Is there a specific reason why this mentions `__` explicitely here?

src/execution/execute.ts
martinbonnin: I renamed it to `errorPropagation`

src/type/directives.ts
martinbonnin: Now that the directive doesn't take arguments, the problem has disappeared 😄 . Should we explicitely reserve `@experimental_disableErrorPropagation` somewhere? Guessing it's the same issue with `@defer` and `@oneOf`. Is there a process to "reserve" directive names?

src/execution/execute.ts
martinbonnin: Now that the directive is[ named `@experimental_disableErrorPropagation`](https://github.com/graphql/graphql-js/pull/4348/commits/f96492c096c373f2287cffc3328258245a020b82), I'm tempted to say no need for an opt-in flag. The name of the directive should add enough friction that users realize they're doing something experimental.

martinbonnin:
@benjie I renamed to `@experimental_disableErrorPropagation`. I like that it's very explicit 👍 . It's also very long but in my case, the directive is going to be added automatically by the Apollo Kotlin compiler so it's perfectly fine.

src/type/directives.ts
benjie: No need to reserve it, it’s sufficiently complex IMO, extremely unlikely someone would implement this to do something else or to require arguments.

martinbonnin: Perfect 👍

src/execution/execute.ts
benjie: Users should still have to add it to their schema, we don’t want it printed in every printSchema call with no opt-out.

benjie:
I've written up a counter proposal because I think that error propagation should be disabled on a per-client basis rather than a per-operation basis; but I have no issues with this being merged so that people can start experimenting with disabling null propagation - that's the entire point of having experimental directives people can use!

Counter-proposal is to make this a request parameter rather than an operation directive: https://github.com/graphql/nullability-wg/discussions/86

martinbonnin:
@benjie agree a request parameter would be better. Modifying the document is always taking the risk that what the user types in GraphiQL doesn't match what the client is ultimately sending over the wire. I think we have the same issue with fragment arguments. 

One nice thing with the directive though is that feature discovery works out of the box. Whereas if we use a request parameter, we'd have to introduce a change in the response or something like this to indicate support. Which feels like a bigger lift.

All in all, it's probably not the end of the road but I would still like this to be merged so that we can experiment in parallel of discussing other alternatives.

benjie:
> One nice thing with the directive though is that feature discovery works out of the box. Whereas if we use a request parameter, we'd have to introduce a change in the response or something like this to indicate support. Which feels like a bigger lift.

One nice thing about the request approach is you don't _need_ feature discovery - you simply turn it on if you're an "error handling client". If the server doesn't support it your request should still succeed and you'll just have some selection sets be blown up as it would have been anyway, so you're no worse off than you were.

martinbonnin:
Alright, if we agree that we don't need anything in the response, it might be worth going this road directly. 

My only objection then is that it potentially requires pull requests in all the servers out there, right (apollo, yoga, etc... )? Because those are the ones calling `execute()` in the first place? So it'll take yet another round of review/discussions before end users can actually try it?

Whereas the directive solution is "just" a matter of "update your graphql-js" dependency and you can start using the new directive. 

If that's the case, then I'd stick with the long, explicitely experimental directive for now. We can always ask everyone to update all their libs later.

benjie:
I'm not arguing against using the directive for now, I think we should do that. But long term, I think the request parameter is better.

martinbonnin:
Sounds good 👍 . I'll wait for @JoviDeCroock's input and if we all agree, I'll update the [graphql-java PR](https://github.com/graphql-java/graphql-java/pull/3772) to match

JoviDeCroock:
The request parameter would take the shape of a `RequestExtension` I presume or do we have other opinions on that?

I am personally a bigger fan of letting language features be described by the `Document` rather than by aligning some request parameter to have a say over the execution mode of operations contained within the `Document`.

martinbonnin:
> The request parameter would take the shape of a RequestExtension I presume or do we have other opinions on that?

I think @benjie had a top-level key in mind?

https://github.com/graphql/graphql-js/pull/4192/files#diff-1d705c6a4c73cd3ce46190029e75abd3015b49de3ec250357dcf9b44a35cb7d0R162

> I am personally a bigger fan of letting language features be described by the Document rather than by aligning some request parameter to have a say over the execution mode of operations contained within the Document.

No strong opinion here besides the fact that we don't have directives on document (yet)?

> With removing the arguments from the directive we are setting ourselves up for poor scalability of the directive, are we open to doing this?

I'm fine with this. As the long experimental name implies, it's clearly not the final form of this so we can always change later.

Also, we could decide to go with `@abortOnError` and a validation rule if we wanted to keep the current `@disableErrorPropagation`. Not as nice IMO but another option.

benjie:
> The request parameter would take the shape of a `RequestExtension` I presume or do we have other opinions on that?

No; the GraphQL Foundation should never put anything in extensions; the point of extensions is to be a reserved namespace for third parties to put what they want in it - if we start putting stuff there, it no longer serves that purpose. It would be a top-level key like `query`, `variables`, `operationName`, `extensions`.

> I am personally a bigger fan of letting language features be described by the `Document` rather than by aligning some request parameter to have a say over the execution mode of operations contained within the `Document`.

You should probably give that feedback against my proposal directly; however do think about what this means for users - users who shouldn't need to know how their client works will now need to remember to include this directive in every document (or have tooling add it for them), and in 10 years time, the vast majority of GraphQL documents should have this directive, making it effectively a `"use strict"` style of annotation. I'd argue, as I have before, that this is actually a client concern. The client is explicitly stating "I'd like to do the error handling on the client side, no need for you to do it for me on the server side" - the _consumer_ (the person writing the GraphQL document) doesn't actually care which way it's done in most cases.

> With removing the arguments from the directive we are setting ourselves up for poor scalability of the directive, are we open to doing this?

We could add arguments in future with a non-breaking change so long as either they're nullable or they have defaults? But also, it's explicitly `@experimental_*` so this shouldn't be a big deal anyway.

JoviDeCroock:
> You should probably give that feedback against my proposal directly; however do think about what this means for user

I'm going to digest my thoughts more before I do but generally I am really opposed to little option flags on a top-level payload

src/execution/__tests__/errorPropagation-test.ts
```diff
@@ -0,0 +1,74 @@
...
+import { expectJSON } from '../../__testUtils__/expectJSON.js';
+
+import type { PromiseOrValue } from '../../jsutils/PromiseOrValue.js';
+
+import { parse } from '../../language/parser.js';
+
+import { buildSchema } from '../../utilities/buildASTSchema.js';
+
+import { execute } from '../execute.js';
+import type { ExecutionResult } from '../types.js';
```
JoviDeCroock, line 12:
```suggestion
import { expectJSON } from '../../__testUtils__/expectJSON.js';
import type { PromiseOrValue } from '../../jsutils/PromiseOrValue.js';
import { parse } from '../../language/parser.js';
import { buildSchema } from '../../utilities/buildASTSchema.js';
import { execute } from '../execute.js';
import type { ExecutionResult } from '../types.js';
```

martinbonnin: [Looks like lint didn't like that](https://github.com/graphql/graphql-js/actions/runs/13496436014/job/37704680944?pr=4348)...

JoviDeCroock: Ugh 😅 we have really annoying lint rules and we should really get rid of some of them

JoviDeCroock: `pnpm lint --fix` should have you covered though

JoviDeCroock: Yes, if adds it back that's okay. Was looking at other test files and assumed the grouping was okay, I can take care of that later as well when I tweak the rules

martinbonnin: [Done](https://github.com/graphql/graphql-js/pull/4348/commits/aec37aa057e950871821be917a854eb1cacab639)

PR #4357, opened 2025-03-20
fix(incremental): failed fragments should never be re-added to the graph 

yaacovCR:
After an execution group "fails," i.e. a null has bubbled up too far, it fails the entire associated deferred fragment. But if the query includes a shared execution group between the failing deferred fragment and a non-failing deferred fragment with some unique subfields, those subfields will still be added to the graph for the failed deferred fragment, which is not present.

This throws an error, because the code assumes that in the situation above, i.e. a shared execution group between two fragments and some unique subfields, both of the deferred fragments must already be in the graph. So a runtime-error is triggered because some of the invariants we rely on when adding deferred fragments are violated.

But even if we didn't have an error, this would be an erroneous situation, as we should never add back a node to the graph that has been failed! The fix should be to retain the failure state on the deferred fragment so as not to attempt to readd the node to the graph.

### DIFF
--- a/src/execution/IncrementalGraph.ts
+++ b/src/execution/IncrementalGraph.ts
@@ -132,14 +132,8 @@ export class IncrementalGraph {
     return { newRootNodes, successfulExecutionGroups };
   }
 
-  removeDeferredFragment(
-    deferredFragmentRecord: DeferredFragmentRecord,
-  ): boolean {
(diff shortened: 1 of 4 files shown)


JoviDeCroock (approved):

yaacovCR:
Closing in favor of #4358

PR #4358, opened 2025-03-21
incremental: add new deferred fragments to the wrapped graphql result

yaacovCR:
alternative solution to #4357 in which we add the new deferred fragments to the wrapped graphql result

After an execution group "fails," i.e. a null has bubbled up too far, it fails the entire associated deferred fragment. But if the query includes a shared execution group between the failing deferred fragment and a non-failing deferred fragment with some unique subfields, those subfields will still be added to the graph for the failed deferred fragment, which is not present.

This throws an error, because the code assumes that in the situation above, i.e. a shared execution group between two fragments and some unique subfields, both of the deferred fragments must already be in the graph. So a runtime-error is triggered because some of the invariants we rely on when adding deferred fragments are violated.

But even if we didn't have an error, this would be an erroneous situation, as we should never add back a node to the graph that has been failed!

After this change, new deferred fragments are explicitly returned as part of the GraphQLWrappedResult and are only added at a single point when building the graph such that this error can be avoided entirely.

### DIFF
--- a/src/execution/IncrementalGraph.ts
+++ b/src/execution/IncrementalGraph.ts
@@ -132,14 +132,8 @@ export class IncrementalGraph {
     return { newRootNodes, successfulExecutionGroups };
   }
 
-  removeDeferredFragment(
-    deferredFragmentRecord: DeferredFragmentRecord,
-  ): boolean {
-    if (!this._rootNodes.has(deferredFragmentRecord)) {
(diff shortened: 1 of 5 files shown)


robrichard (approved):

robrichard:
I like this approach, i think this will translate well to the spec algorithms

Issue #1428, opened 2018-07-21
merge ast trees


terion-name:
# Feature request

Another feature that is extremely useful and utils are lacking it: merging ast trees.

Primary use case: merging two queries (selections) to dynamically construct queries. There are plenty of cases where it is needed, mainly in libraries and in cases of  graphql to graphql communication.

PS
It would be great if somebody can say how to do this properly now (simple deep merge doesn't wotk properly)

IvanGoncharov:
@terion-name Merging is pretty easy by itself if you don't need to handle conflicts:
```graphql
{
  foo(arg: 1)
}
```
```graphql
{
  foo(arg: 2)
}
```
Would be merged into: 
```graphql
{
  foo(arg: 1)
  foo(arg: 2)
}
```
Which is invalid query constructed from two valid ones.

So do you need conflict resolution or simply merging ASTs would be enough?

terion-name:
@IvanGoncharov common scenario: I have an application that utilizes a backend that has graphql interface (it could be a remote api, a service alike prisma or postgraphile). 
I have a type that has calculated fields. Let's assume `Home { avgRating }`. `avgRating` is calculated from `ratings { stars }`. So to calculate `avgRating` I need to have `ratings { stars }` in request that is passed to remote backend.  So I need to ensure this selection. When a client queries `Home { avgRating }` I need to transform it to `Home { avgRating ratings { stars } }`.
So in a resolver I have `info.fieldNodes[0].selectionSet` (an AST tree), I can construct a desired selection set with `{ ratings { stars } }` (also AST tree) and I need to merge them recursively to properly combine selections.

It is a simple example, there are many more deep and complex situations where one needs to properly deeply merge (combine) several AST trees

IvanGoncharov:
@terion-name It's pretty complex to handle all corner cases even in such small example, e.g. user sens following query:
```graphql
{
  Home {
    avgRating
    ratings: someOtherField
  }
}
```
Or he provide different parameters to the same field:
```graphql
{
   Home {
     avgRating
     ratings(first: 5) {
       stars
     }
   }
}
```
In both cases, you can simply merge and you need to do some tricks with field aliases.
AFAIK, `graphql-tools` and `graphql-bindings` have some functionality to do that.

Would be great to have something like that `graphql-js` but it should cover all edge cases and have an extensive set of test cases to be merged.

ivome:
I have thought about this too quite a bit. 

Here is an approach that could work IMO: 

[...]

PR #4359, opened 2025-03-22
feat(mergeAST): add mergeAST utility

metrue:
This PR is adding utility of merge AST by simply migrating it from [graphiql](https://github.com/graphql/graphiql/blob/8fbac78ae43a2b245aad0203cb8269f38e0c105f/packages/graphiql-toolkit/src/graphql-helpers/merge-ast.ts#L109)

As first step to resolve this issue https://github.com/graphql/graphql-js/issues/1428#issuecomment-2676341366

### DIFF
--- a/src/utilities/__tests__/mergeAST-test.ts
+++ b/src/utilities/__tests__/mergeAST-test.ts
@@ -1,0 +1,182 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { parse } from '../../language/parser';
+import { print } from '../../language/printer';
+
+import { GraphQLInt, GraphQLObjectType, GraphQLString } from '../../type/definition';
+import { GraphQLSchema } from '../../type/schema';
+
+import { mergeAST } from '../mergeAST';
+
+const schema = new GraphQLSchema({
+  query: new GraphQLObjectType({
+    name: 'Test',
+    fields: {
+      id: {
+        type: GraphQLInt,
+      },
+      name: {
+        type: GraphQLString,
+      },
+    },
+  }),
+});
+
+describe.only('mergeAST', () => {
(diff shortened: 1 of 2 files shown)


benjie:
One thing we should very much keep in mind with `mergeAST` is that it's a potential DOS vector. Having it in the frontend (i.e. GraphiQL) is one thing, but putting it in GraphQL.js means that we will need to have a lot stronger protection against things that could cause explosion of complexity or even infinite loops. If we could require that the document was already validated (and thus had passed all our checks) before calling into mergeAST then that would give us some confidence, but I imagine that people will want to mergeAST separately and we need to protect them from bad inputs. I'm not sure the solution to this, just wanted to surface the concern.

metrue:
Thanks @benjie . 

You're absolutely right: mergeAST introduces complexity risks, especially if used on unvalidated or hostile input. My intention with duplicated the utility from GraphiQL is to support use cases like: dynamically composing selection sets, merging query documents for internal middleware, or some other client side use cases, or Building tooling that operates on statically constructed or pre-validated ASTs.  But yeah it might introduce reliability and security risks in server-side GraphQL environments.   

Given that context, I'd be happy to explore ways to make the function safer or more clearly scoped, I am thinking, 
* We can rename it to `mergeValidatedAST`, explicitly require/assume that inputs to mergeAST are the result of parse() followed by validate()
* Add shallow complexity safeguards (depth, cost)

It might not be able to prevent all the malicious input I think.  Would love your thoughts on whether any of these sound like a reasonable direction. I'm happy to iterate and contribute safeguards in this PR or a follow-up depending on what fits best with the project’s expectations.

benjie:
It might be best that you break it out into its own library for now. I'd want to carefully test (and probably rewrite) it before incorporating it into GraphQL.js, even with the rename.

PR #3974, opened 2023-10-09
Forbid @skip and @include directives in subscription root selection

benjie:
This is an implementation of https://github.com/graphql/graphql-spec/pull/860

The spec calls for a separate `CollectSubscriptionFields` algorithm due to executing without `variableValues`; but I figured that maintenance would be easier with the algorithms synchronized.

Please feel free to make any changes you need to this PR; it's just to get the ball rolling.

Related GraphQL WG action item (from 2021!): https://github.com/graphql/graphql-wg/issues/695

PR #4360, opened 2025-03-23
polish: remove check for collecting forbidden directives from execution

yaacovCR:
follow on from #3974

### DIFF
--- a/benchmark/validateSubscription-benchmark.js
+++ b/benchmark/validateSubscription-benchmark.js
@@ -1,0 +1,67 @@
+import { parse } from 'graphql/language/parser.js';
+import { buildSchema } from 'graphql/utilities/buildASTSchema.js';
+import { validate } from 'graphql/validation/validate.js';
+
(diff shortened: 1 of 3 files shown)


JoviDeCroock (commented):
I personally find the boolean approach a lot cleaner, js engines will also have an easier time optimising the top level function with a flag compared to the adhoc created func.

It's also a lot harder to reason about as we add a layer of indirection and the forbidden directives become a hook. I.e. we create an exact replica of shouldIncludeNode that just has an extra step to push to an outer scope array.

yaacovCR:
@JoviDeCroock

I suppose this is a style issue, I prefer the changes in this PR as I find it simpler to reason about, but not strongly enough that I would push this.

In terms of performance, I added a benchmark to demonstrate that there is no significant/reproducible performance penalty for the dynamic function, which also confirms my suspicion that there is no significant/reproducible performance hit for the current approach, so it's all in terms of readability, and I am fine to stick with the status quo, although see #4361.

PR #4361, opened 2025-03-23
polish: use closure instead of CollectFieldsContext

yaacovCR:
While working on #4360, I felt myself trying to reach for opportunity to remove the CollectFieldsContext type/object that I previously introduced; this is accomplished in this separate PR.

Benchmarking on my machine, this seems essentially no better and no worse in the performance realm. I am evaluating this in terms of readability, both in the local sense and in the overall algorithmic sense of better understanding the method and limit of how we are recursing.

Overall, I think it's a minor improvement.

### DIFF
--- a/src/execution/collectFields.ts
+++ b/src/execution/collectFields.ts
@@ -62,17 +62,6 @@ export interface FragmentDetails {
   variableSignatures?: ObjMap<GraphQLVariableSignature> | undefined;
 }
 
-interface CollectFieldsContext {
-  schema: GraphQLSchema;
-  fragments: ObjMap<FragmentDetails>;
-  variableValues: VariableValues;
-  runtimeType: GraphQLObjectType;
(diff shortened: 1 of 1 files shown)


yaacovCR:
Note: this would probably be most helpfully reviewed by ignoring whitespace differences.

yaacovCR:
Also subtracts about 50 lines of code, which is always good.

From my machine:

image

yaacovCR:
> Also subtracts about 50 lines of code, which is always good.

Mostly whitespace/formatting though, so your mileage may vary, but it helps make the code more readable for me.

yaacovCR:
closing for now, I suppose it's somewhat helpful to have the arguments for `shouldIncludeNode` be more explicitly declared (rather than referencing the variables from the enclosure). we can always revisit this if it seems more compelling.

PR #4364, opened 2025-03-27
Implement onError proposal

benjie:
Spec PR:
- ~~https://github.com/graphql/graphql-spec/pull/1153~~
- https://github.com/graphql/graphql-spec/pull/1163


benjie:

JoviDeCroock (approved):
I am not opposed to this but what are we doing with the directive that is present on `main`/v17?

src/execution/execute.ts
```diff
@@ -152,6 +156,15 @@ export interface ExecutionArgs {
...
   subscribeFieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
+  /**
+   * Experimental. Set to NO_PROPAGATE to prevent error propagation. Set to ABORT to
+   * abort a request when any error occurs.
+   *
+   * Default: PROPAGATE
+   *
+   * @experimental
+   */
+  onError?: GraphQLErrorBehavior;
```
JoviDeCroock, line 167: Should we move this to the new `options?:` object that is in here, could maybe facilitate these experimental things better. Alternatively we stick to a flat config

benjie: It's specified to be part of the "GraphQL request", so I would expect it top level. I would consider moving some of the non-request properties to be inside an options object though: fieldResolver, typeResolver, subscribeFieldResolver.

JoviDeCroock:
Regarding the introspection, would we have a non-repeatable schema-directive then that indicates what kind of error-propagation we are dealing with? I think that would make sense, the server can internally dictate what the default is and it can be externally requested to be different by means of the request-parameter.

I think we can merge these separately though, the default for all would be PROPAGATE at the moment so there's no need yet for the schema-directive. The name `errorPolicy` sounds quite good to me for the introspection field.

benjie:
Exactly that, yes. `schema @behavior(onError: NO_PROPAGATE) { query: Query }` or some such.

We _can_ do them separate, but I'd like to make it so that the presence of the field in introspection also implies the support for the parameter and vice versa.

benjie:
This is now up to date with the latest version of the spec RFC, including the removal of `onError` on responses and the addition of `__Schema.defaultErrorBehavior` to introspection and `@behavior(onError: __ErrorBehavior! = PROPAGATE)` to SDL.

benjie:

martinbonnin:
>  It's going to be confusing to have both. I'll open a PR on this PR to do so (and also open a PR to keep graphql-java in sync if we confirm this is the way we're going).

Never mind, I didn't notice this PR is against `16.x.x`. So I'll wait until this lands and I'll open a PR to remove the directive just after (or anyone feel free to do it!).

PS: strong preference for this PR rather than the [other one](https://github.com/graphql/graphql-js/pull/4384)

src/utilities/extendSchema.ts
```diff
@@ -165,6 +168,14 @@ export function extendSchemaImpl(
     }
   }
 
+  let defaultErrorBehavior: Maybe<GraphQLErrorBehavior> = schemaDef
+    ? getDefaultErrorBehavior(schemaDef)
+    : null;
+  for (const extensionNode of schemaExtensions) {
```
yaacovCR, line 174: Can an extension override a previous extension or the original schema definition with respect to an aspect of behavior?

martinbonnin: looks ok to me

benjie: Hmmm. No; it shouldn't. Good catch. So essentially we need "default" to be its own value (`null` is an option), then store a value when set, and don't allow setting it a second time. That or don't allow it on the `extend schema` 

But also this is kind of irrelevant because this implementation is out of date vs the latest spec proposal; we'll be doing it via `service { capabilities { ... } }` now instead (actually... what would an `extend service` look like?) and that should cause a conflict if the same key is set more than once.

martinbonnin: @benjie what do you think of removing the introspection bits + default value from this PR? Just have the request `onError`? We can add the default value later on. Baby steps.

benjie: The client needs to know if the server supports `onError` or not; the server should ignore `onError` if it doesn't understand it, so they kind of go hand in hand: the signalling that the feature is supported and the feature itself. Further, we should tie the indication of the server's default error behavior to the concept of error behaviors, so that clients and servers support both together, without that the migration is a lot more complex. This has been discussed before at the WGs and conclusion is we shouldn't ship onError without solving how to tell a client that it's supported, and out-of-band communications is not ideal for this.

martinbonnin: 👍 [back to capabilities then](https://github.com/graphql/graphql-spec/pull/1163#issuecomment-3237058626).

martinbonnin:
Should this target the v17 branch?

egoodwinx:
hey benjie - wondering the status on this and when it will be available in 17? I want to look at implementing this in some js libraries.

src/error/locatedError.ts
```diff
@@ -43,15 +73,24 @@ export function locatedError(
...
     positions: (originalError as GraphQLError).positions,
     path,
+    pathNonNull,
     originalError,
   });
 }
 
 function isLocatedGraphQLError(error: any): error is GraphQLError {
-  return Array.isArray(error.path);
+  return Array.isArray(error.path) && Array.isArray(error.pathNonNull);
```
benjie, line 92: This is the main potentially breaking change in this PR. Options:

1. (This option) If an error does not contain `pathNonNull` then it's (no longer) a located error, and should be wrapped with a GraphQL error that is.
2. Allow _poorly_ located errors - those without `pathNoNull`, through when they originate in user code.

IMO user code should not be raising errors with `path` anyway (and doing so potentially enables them to break GraphQL spec compliance), so I personally feel this is a moot point and am happy adding this change.

src/execution/execute.ts
benjie: It wasn't clear to me (and using `unknown` in the types didn't help resolve) if `undefined` is possible any more. I've moved the `fieldDef` check into this parent function from `executeField`, so it won't return `undefined` directly any more, but I'm not sure if deeper function calls might?

My instincts tell me we can delete it, but I've left it here for safety.

src/execution/execute.ts
```diff
@@ -694,13 +735,27 @@ function executeFields(
...
       const result = executeField(
         exeContext,
         parentType,
         sourceValue,
         fieldNodes,
         fieldPath,
+        fieldDef,
       );
 
       if (result !== undefined) {
```
benjie: Similarly, this check may always be truthy now.

src/jsutils/Path.ts
```diff
@@ -54,11 +101,5 @@ export function addPath(
...
   path: Maybe<Readonly<Path>>,
 ): Array<string | number> {
-  const flattened = [];
-  let curr = path;
-  while (curr) {
-    flattened.push(curr.key);
-    curr = curr.prev;
-  }
-  return flattened.reverse();
+  return [...pathToDigest(path).path];
```
benjie, line 102: Annoyingly we said this was mutable before, so we need to clone the array. (This allows us to make `pathToDigest` memoized in future.)

benjie (commented):
This implements the latest version of https://github.com/graphql/graphql-spec/pull/1163:

- Removed all the introspection code
- Removed "default error behavior" as discussed at last Spec WG
- Introduced `pathNonNull` as discussed at last Spec WG
- Merged the latest 16.x.x branch
- Added comments and extended tests to pass linting

I'm working on a new project that'll show how this new error/nullability system can shine across all clients with minimal code; but I don't think it'll be ready until late September. In the mean time, it would be helpful to have an experimental release of this out for people to experiment with.

Note: I've used the `@experimental` TSDoc tag to mark methods/interfaces/etc as experimental, this should make it easier for people to adopt whilst still being aware they're using something that may change in a patch release. https://tsdoc.org/pages/tags/experimental/

src/execution/execute.ts
benjie: Scratch that; I proved it with the type system by changing `unknown` to `{} | null`.

benjie:
Good timing @egoodwinx! I've just spent the day bringing this up to date. If we get this merged into v16 we can forward-port it to v17 or vice-versa. I've not attempted to reconcile it with incremental delivery yet, but I've tried to be careful in the wording to allow for it (e.g. for the `ABORT` mode it is essentially described as bubbling to the highest position it can, typically the operation root; but in incremental delivery it will be the individual incremental units).

benjie:
GraphQL v17 port, mostly performed by codex: https://github.com/graphql/graphql-js/pull/4846

PR #4366, opened 2025-04-04
feat(execution): add max coercion errors option to execution context

cristunaranjo:
This PR add `executionOptions` property to the `ExecutionArgs`. Currently only one option can be configured, as is the one I need, but I have built an object so it can easily be extended in the future.

The property allows the configuration of the maximum number of errors (`maxErrors`) for coercion. This allows the current hardcoded limit (`50`) to be reduced to a small number to avoid possible DoS attacks.

> Also, it updates the execution docs to reflect this new change. I think the documentation was outdated since functions were using positional arguments and now they only accept an object.

No longer updates the docs.

### DIFF
--- a/website/pages/api-v16/execution.mdx
+++ b/website/pages/api-v16/execution.mdx
@@ -29,20 +29,16 @@ const { execute } = require('graphql'); // CommonJS
 ### execute
 
 ```ts
-export function execute({
+export function execute(
   schema: GraphQLSchema,
-  document: DocumentNode,
-  rootValue?: unknown,
-  contextValue?: unknown,
+  documentAST: Document,
+  rootValue?: mixed,
+  contextValue?: mixed,
   variableValues?: { [key: string]: mixed },
   operationName?: string,
-  fieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
-  typeResolver?: Maybe<GraphQLTypeResolver<any, any>>;
-  subscribeFieldResolver?: Maybe<GraphQLFieldResolver<any, any>>;
-  options?: { maxCoercionErrors?: number };
-}): PromiseOrValue<ExecutionResult>;
+): MaybePromise<ExecutionResult>;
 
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):

website/pages/api-v16/execution.mdx
```diff
@@ -61,21 +65,32 @@ a GraphQLError will be thrown immediately explaining the invalid input.
 executing the query, `errors` is null if no errors occurred, and is a
 non-empty array if an error occurred.
 
+#### options
+
+##### maxCoercionErrors
+
+Set the maximum number of errors allowed for coercing (defaults to 50).
```
JoviDeCroock, line 72: Can we leave the rest of the docs untouched, the changes seem invalid 😅 like from where I sit the only thing that we need to add is `options?: { maxCoercionErrors?: number };` as an argument, not an object-field to the docs

cristunaranjo: I think you are wrong. I cannot add `options?` as a positional argument without adding the other missing arguments (`fieldResolver`, `typeResolver`, `subscribeFieldResolver`). I don't know why the function accepts positional arguments or an object, I guess to avoid a breaking change, but the code looks like prefer the object solution and the documentation should reflect that. 
I might be wrong but I think my changes make sense

cristunaranjo: From the code:

```
'graphql@16 dropped long-deprecated support for positional arguments, please pass an object instead.',
```

JoviDeCroock: That's fair but the docs should still show both methods rather than this kind of driveby change

cristunaranjo: I have reverted the changes to the docs. My only concern now is that this new options are only documented in the code, so not sure how many people will benefit from them...

It would be nice to update the docs for v17.

PR #4367, opened 2025-04-04
fix(coerce-input-value): input object coercion rejects arrays

### LINKED ISSUE #4365: Input object coercion should reject arrays directly
If an array is fed into input object coercion it should be rejected directly, rather than passing the object check and then failing on the keys checks.

See this repro:

https://gist.github.com/benjie/58ba75340a85b0da4b403973bea2c5fb

Seems like adding a `!Array.isArray(...)` check should cover it?

cristunaranjo:
This PR closes #4365 

Rejects arrays when input type is an object. Works for top level arrays and deep ones.

### DIFF
--- a/src/execution/__tests__/executor-test.ts
+++ b/src/execution/__tests__/executor-test.ts
@@ -11,7 +11,6 @@ import { Kind } from '../../language/kinds';
 import { parse } from '../../language/parser';
 
 import {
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):
This is a great find! Thank you so much

PR #4372, opened 2025-04-12
Refactor every code-first example to leverage resolve

JoviDeCroock:
This refactors the code-first examples to use inline-resolvers rather than the root-value to show a difference between SDL and code-first.


saihaj (approved):
unrelated to this PR but maybe we should call it "Schema first" ?

PR #4384, opened 2025-05-01
Implement onError proposal (but with `ErrorBehavior` rather than `__ErrorBehavior`)

benjie:
Alternative to:
- #4364 

Using `__ErrorBehavior` causes older GraphiQL to baulk. I'm hoping that this approach will not.

### DIFF
--- a/src/__tests__/starWarsIntrospection-test.ts
+++ b/src/__tests__/starWarsIntrospection-test.ts
@@ -37,7 +37,7 @@ describe('Star Wars Introspection Tests', () => {
             { name: 'Droid' },
             { name: 'Query' },
             { name: 'Boolean' },
-            { name: '__ErrorBehavior' },
+            { name: 'ErrorBehavior' },
             { name: '__Schema' },
             { name: '__Type' },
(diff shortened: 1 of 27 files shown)


benjie:

benjie:

src/__tests__/starWarsIntrospection-test.ts
```diff
@@ -37,6 +37,7 @@ describe('Star Wars Introspection Tests', () => {
             { name: 'Droid' },
             { name: 'Query' },
             { name: 'Boolean' },
+            { name: 'ErrorBehavior' },
```
JoviDeCroock, line 40: I am unsure whether we should expose this as a `schema` property, this is in fact an internal property that tells us just how a GraphQL API will behave. Most GraphQL code I've seen treats it as such where when walking the introspection there is an explicit check for properties starting with `__` to be skipped.

benjie:
Closing in favour of https://github.com/graphql/graphql-js/pull/4364 to keep things tidier. That will also need updating.

PR #4389, opened 2025-05-05
fix(replaceVariables): preserve sources for fragment variable values

yaacovCR:
The new flow for coercing input literals -- see https://github.com/graphql/graphql-js/pull/3814 -- introduces a `replaceVariables()` prior to calling `coerceInputLiteral()` such that we go:

from `ValueNode` => `ConstValueNode` => coerced value

The main advantages being that:
(1) we can trivially support embedding variables within complex scalars -- despite this being non-spec defined behavior!
(2) userland implementations of custom scalar `coerceInputLiteral()` methods do not need to worry about handling variables or fragment variables.

Prior to this PR, we did not properly handle this in the case of fragment definition variables/fragment spread arguments. `replaceVariableValues()` assumes that the uncoerced value is preserved as source, but instead of grabbing this value from the argument on the spread, we were incorrectly attempting to retrieve the already stored value from existing variables.

This was not caught because we previously did not have any actual tests for this custom unspecified behavior and were using incorrect types and bespoke builders in our tests for `replaceVariables()`.

This PR:

(a) fixes the bug by storing the argument from the spread
(b) fixes our bespoke builders in `replaceVariables()`
(c) add explicit tests for embedding variables within custom scalars to protect against future changes.

As a post-script, because within `getFragmentVariableValues()` we now end up within a `ValueNode` stored on source, to  coerce this value, we can extract a new helper `coerceArgumentValue()` from `experimentalGetArgumentValues()`, rather than calling it after we are done for all the values, which has some additional overhead.

This has the side benefit of removing the need for a separate `experimentalGetArgumentValues()` function altogether. We initially introduced it to have a more flexible signature for `getArgumentValues()` that encompasses argument for fragment spreads, but now this is taken care of using our more directed helper.

### DIFF
--- a/src/execution/__tests__/variables-test.ts
+++ b/src/execution/__tests__/variables-test.ts
@@ -938,6 +938,26 @@ describe('Execute: Handles inputs', () => {
         },
       });
     });
+
+    it('allows custom scalars with embedded nested fragment variables', () => {
+      const result = executeQueryWithFragmentArguments(`
+        {
+          ...JSONFragment(input1: "foo")
+        }
+        fragment JSONFragment($input1: String) on TestType {
+          ...JSONNestedFragment(input2: $input1)
+        }
+        fragment JSONNestedFragment($input2: String) on TestType {
+          fieldWithJSONScalarInput(input: { a: $input2, b: ["bar"], c: "baz" })
+        }
+      `);
+
+      expectJSON(result).toDeepEqual({
+        data: {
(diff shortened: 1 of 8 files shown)


yaacovCR:
I broke this up into separate commits to aid review.

First commit: introduces the failing test.
Second commit: introduces a new type `FragmentVariableValues` to the code base, for now identical to `VariableValues`
Third commit: fix bug by updating `FragmentVariableValues` so that the source value for a variable is of type `ConstValueNode`

JoviDeCroock (approved):

PR #4392, opened 2025-05-13
Catch unhandled exception in abstract resolution

### LINKED ISSUE #4375: `defaultTypeResolver` can result in unhandled promise rejection
I was just looking over the source code around polymorphism and I think I found an issue here:

https://github.com/graphql/graphql-js/blob/7dd7812e2b8a9b0f2dc9272c62b462678fd8d1b5/src/execution/execute.ts#L1000-L1018

Specifically: if we're looking over types A, B and C, and A yields a promise, and B returns a match, we will return `B.name` and the promise we added to `promisedIsTypeOfResults` for A will never be handled. If that promise were to reject then Node would raise an unhandled promise rejection error.

I think we can solve this by adding a line above `return type.name`:

```diff
       } else if (isTypeOfResult) {
+        // Ignore errors from promises already created
+        if (promisedIsTypeOfResults.length) {
+          void Promise.allSettled(promisedIsTypeOfResults);
+        }
         return type.name;
       }

```

JoviDeCroock:
Fixes #4375

The assertion was correct that we have an unhandled Promise rejection going on when we have resolved a type and the rejection comes in at a later time.

We really have to revamp our tooling, the things that the linter is surfacing in tests and especially uncovered branches in tests... it really makes for a bad DX on this repo

### DIFF
--- a/src/execution/__tests__/union-interface-test.ts
+++ b/src/execution/__tests__/union-interface-test.ts
@@ -12,7 +12,7 @@ import {
 import { GraphQLBoolean, GraphQLString } from '../../type/scalars';
 import { GraphQLSchema } from '../../type/schema';
 
-import { executeSync } from '../execute';
+import { execute, executeSync } from '../execute';
 
 class Dog {
   name: string;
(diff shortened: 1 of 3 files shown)


src/execution/execute.ts
```diff
@@ -1002,6 +1002,9 @@ export const defaultTypeResolver: GraphQLTypeResolver<unknown, unknown> =
         if (isPromise(isTypeOfResult)) {
           promisedIsTypeOfResults[i] = isTypeOfResult;
         } else if (isTypeOfResult) {
+          if (promisedIsTypeOfResults.length) {
+            Promise.allSettled(promisedIsTypeOfResults);
```
benjie:
```suggestion
            // Explicitly ignore any promise rejections
            void Promise.allSettled(promisedIsTypeOfResults);
```

src/execution/__tests__/union-interface-test.ts
```diff
@@ -154,6 +154,77 @@ odie.mother.progeny = [odie];
...
+  fields: () => ({
+    id: { type: GraphQLString },
+    nameA: { type: GraphQLString },
+  }),
+  isTypeOf: (_value, _context, _info) => {
+    return new Promise((_resolve, reject) => {
+      setTimeout(() => {
+        reject(new Error('TypeA_isTypeOf_rejected'));
+      }, 10);
+    });
```
benjie: I think the lint settings would rather you wrote this without braces:

```suggestion
  isTypeOf: (_value, _context, _info) =>
    new Promise(
      (_resolve, reject) =>
        void setTimeout(
          () => void reject(new Error("TypeA_isTypeOf_rejected")),
          10,
        ),
    ),
```

But this is just stylistic guidance so that all code looks as self-similar as possible, to make it easier on readers/reviewers. As the maintainer you're completely free to change these rules. I personally think it's fine to have arrow functions with block bodies; but I also think the `void` keyword should be used more :)

saihaj (approved):

PR #4393, opened 2025-05-14
docs: add page on abstract types

sarahxsanders:
New page on abstract types, part of the effort to expand GraphQL.js docs

PR #4394, opened 2025-05-14
docs: editorial on abstract types page

benjie:
Editorial on #4393

Excellent guide! The main correction was that in GraphQL v16 support for returning the object type from `resolveType` was turned into an error, and it is removed in v17 - strings are the only supported approach going forward. I also made some more of the text more accurate and added a little detail in a couple places. Overall, an excellent resource!

I would like to see all the code in GraphQL.js using ESM though. Something to chat with @JoviDeCroock about, but now Node 22 is the LTS version and it supports `require(ESM)`, writing our code samples in ESM will be much more natural and also allows for copy/paste into TypeScript without requiring editing.

cc @sarahxsanders


JoviDeCroock (approved):

JoviDeCroock:
@benjie This has conflicts since my ESM transition

PR #4391, opened 2025-05-10
docs: cursor-based pagination guide

sarahxsanders:
Adds a guide on cursor-based pagination

PR #4383, opened 2025-04-29
docs: N+1 problem and DataLoader

sarahxsanders:
Adds guide "Solving the N+1 Problem with `DataLoader`"

PR #4382, opened 2025-04-29
docs: understanding graphql errors

sarahxsanders:
Adds a guide "Understanding GraphQL.js Errors" 

I added some additional resources, feel free to suggest others/more if needed

PR #4381, opened 2025-04-26
docs: anatomy of a resolver

sarahxsanders:
adds anatomy of a resolver guide

This is a part of the effort to expand GraphQL.js documentation

PR #4395, opened 2025-05-14
docs: editorial for recent documentation updates

benjie:
Relevant to these PRs:

- #4380
- #4381
- #4382
- #4383
- #4391

In general this content was excellent - thanks @sarahxsanders! - but I wanted to add a few minor updates. I've included comments in a few places with reasoning.

If this PR is too much to review all at once, I'm happy to break it up as I did review of each PR separately.

### DIFF
--- a/website/pages/docs/advanced-custom-scalars.mdx
+++ b/website/pages/docs/advanced-custom-scalars.mdx
@@ -100,25 +100,36 @@ describe('DateTime scalar', () => {
 Integrate the scalar into a schema and run real GraphQL queries to validate end-to-end behavior.
 
 ```js
-const { graphql, buildSchema } = require('graphql');
+const { graphql, GraphQLSchema, GraphQLObjectType } = require('graphql');
+const { DateTimeResolver as DateTime } = require('graphql-scalars');
 
-const schema = buildSchema(`
+const Query = new GraphQLObjectType({
+  name: 'Query',
+  fields: {
+    now: {
+      type: DateTime,
+      resolve() {
+        return new Date();
+      },
+    },
+  },
+});
+
+/*
   scalar DateTime
 
   type Query {
     now: DateTime
   }
-`);
-
-const rootValue = {
-  now: () => new Date('2024-01-01T00:00:00Z'),
-};
+*/
+const schema = new GraphQLSchema({
+  query: Query,
+});
 
 async function testQuery() {
   const response = await graphql({
     schema,
     source: '{ now }',
-    rootValue,
   });
   console.log(response);
 }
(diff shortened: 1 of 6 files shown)


website/pages/docs/advanced-custom-scalars.mdx
```diff
@@ -100,25 +100,36 @@ describe('DateTime scalar', () => {
...
+    now: {
+      type: DateTime,
+      resolve() {
+        return new Date();
+      },
+    },
+  },
+});
 
-const schema = buildSchema(`
```
benjie, line 105: The `buildSchema` approach didn't allow the custom `DateTime` scalar to be implemented, so I replaced it with the class-based approach.

website/pages/docs/advanced-custom-scalars.mdx
```diff
@@ -181,13 +192,22 @@ If you need domain-specific behavior, you can wrap an existing scalar with custo
...
+    return email;
+  },
+  parseLiteral(literal, variables) {
+    const email = EmailAddressResolver.parseLiteral(literal, variables);
+    if (!email.endsWith('@example.com')) {
       throw new TypeError('Only example.com emails are allowed.');
     }
-    return EmailAddressResolver.parseValue(value);
+    return email;
   },
```
benjie, line 211: We commented that it was essential to implement `parseLiteral` alongside `parseValue` so I figured we should do that! I also restructured so the underlying scalar does the hard work (especially for literal!) and then we just check the end result. Further, I showed renaming the scalar.

website/pages/docs/cursor-based-pagination.mdx
```diff
@@ -18,15 +20,15 @@ that works well with clients.
 
 Cursor-based pagination typically uses a structured format that separates
 pagination metadata from the actual data. The most widely adopted pattern follows the
-[Relay Cursor Connections Specification](https://relay.dev/graphql/connections.htm). While
+[GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm). While
```
benjie, line 23: We renamed this spec to the GraphQL Cursor Connections Specification by agreement at the GraphQL WG a couple years back. Relay hosts it, but it's not really a "relay" spec any more - it's very widely used.

website/pages/docs/cursor-based-pagination.mdx
```diff
@@ -192,7 +194,7 @@ const usersField = {
     let start = 0;
     if (args.after) {
       const index = decodeCursor(args.after);
-      if (index != null) {
+      if (Number.isFinite(index)) {
```
benjie, line 197: This is me being a bit pedantic on the security front, but it's possible that a user could pass something that when you `parseInt` it would become `NaN` (e.g. `cursor:frog`), so the isFinite check ensures that the number is reasonable. Really the hardening should take place in `decodeCursor` but this felt a cleaner edit.

website/pages/docs/n1-dataloader.mdx
```diff
@@ -2,7 +2,7 @@
 title: Solving the N+1 Problem with `DataLoader`
 ---
 
-When building a server with GraphQL.js, it's common to encounter
+When building your first server with GraphQL.js, it's common to encounter
```
benjie, line 5: I wanted to soften the wording here, don't want people saying "See! Even the GraphQL documentation says that it has major performance issues!" without bothering to read on. Wanted to make this clear this is a mistake you might make early on, but once you've built a few GraphQL schemas using batching will be your bread and butter.

website/pages/docs/resolver-anatomy.mdx
```diff
@@ -33,15 +33,16 @@ When GraphQL.js executes a resolver, it calls the resolver function
...
 ```
 
 Each argument provides information that can help the resolver return the
 correct value:
 
 - `source`: The result from the parent field's resolver. In nested fields,
-`source` contains the value returned by the parent object. For root fields,
-it is often `undefined`.
+`source` contains the value returned by the parent object (after resolving any
+lists). For root fields, it is the `rootValue` passed to GraphQL, which is often
```
benjie, line 44: This parenthesis might not be the best way of saying this, but `source` is essentially the list item inside a list if there is a list, and that's quite important for users to realise.

JoviDeCroock (approved):

PR #4398, opened 2025-05-15
docs: update "going to production" guide

sarahxsanders:
Updates "Going to Production" guide to cover a larger scope. The goal is to cover most bases on what devs should consider before deploying to production. I also added a checklist.

I will preface that this is a bit of a long page, but useful to have

Not ready for review yet, still proofreading/cleaning up


website/pages/docs/going-to-production.mdx
```diff
@@ -1,53 +1,66 @@
 ---
-title: Going to Production
+title: Best Practices for Production Readiness
```
JoviDeCroock: NIT: when I open the preview, the title feels a bit long

website/pages/docs/going-to-production.mdx
```diff
@@ -124,3 +134,296 @@ export default [
...
+    return resolver(parent, args, context, info);
+  };
+}
+```
+
+For more details, see the [Authentication and Middleware](./authentication-and-express-middleware/) guide.
+
+### Apply rate limiting
+
+To prevent abuse, you can limit how often clients send queries. Basic rate limiting can be added
```
JoviDeCroock: Should we link to https://github.com/teamplanes/graphql-rate-limit/blob/master/README.md#field-config instead

website/pages/docs/going-to-production.mdx
```diff
@@ -2,52 +2,65 @@
...
+This guide covers key practices to prepare a server built with GraphQL.js for production use.
 
-### Vite
+## Optimize your build for production
+
+In development, GraphQL.js includes validation checks to catch common mistakes like invalid schemas
+or resolver returns. These checks are not needed in production and can increase runtime overhead.
+
+You can disable them by setting `process.env.NODE_ENV` to `'production'` during your build process.
+GraphQL.js will automatically strip out development-only code paths.
```
benjie:
```suggestion
GraphQL.js will automatically skip over development-only code paths.
```

website/pages/docs/going-to-production.mdx
```diff
@@ -2,52 +2,65 @@
...
+
+Bundlers are tools that compile and optimize JavaScript for deployment. Most can be configured to
+replace environment variables `process.env.NODE_ENV` at build time.
```
benjie:
```suggestion
replace environment variables such as `process.env.NODE_ENV` at build time,
allowing for unused code (such as development only code paths) to be elided by
minification tools.
```

website/pages/docs/going-to-production.mdx
```diff
@@ -124,3 +134,297 @@ export default [
...
+- Enforcing authentication and authorization
+- Applying rate limits
+
+These techniques can help protect your server from accidental misuse or intentional abuse.
+
+### Control schema introspection
+
+Introspection lets clients query the structure of your schema, including types and fields. While
+helpful during development, it may expose internal details you don't want to reveal in
+production.
```
benjie: Your schema shouldn't contain internal details that you don't want to reveal, and turning off introspection is no guarantee that such details won't leak[^1].

```suggestion
Introspection lets clients query the structure of your schema, including types
and fields. While helpful during development, it may be an unnecessary in
production and disabling it may reduce your API's attack surface.
```

[^1]: Even things that you can't query without introspection, such as field descriptions, may leak in a production environment with introspection disabled under certain circumstances. For example, if code generation reflected the field descriptions as TSDoc comments, it might be possible to view these comments via sourcemaps from the compiled code. Disabling introspection gives a false sense of security in many cases.

website/pages/docs/going-to-production.mdx
```diff
@@ -124,3 +134,297 @@ export default [
...
+```js
+function requireRole(role, resolver) {
+  return (parent, args, context, info) => {
+    if (context.user?.role !== role) {
+      throw new Error('Not authorized');
+    }
+    return resolver(parent, args, context, info);
+  };
+}
+```
```
benjie:
````suggestion
GraphQL doesn't include built-in authentication. Instead, you can attach user data to the request
using middleware, and pass this through to the business logic where
authorization should take place:

```js
// From your business logic
const postRepository = {
  getBody({ user, post }) {
    if (user?.id && (user.id === post.authorId)) {
      return post.body
    }
    return null
  }
}

// Resolver for the `Post.body` field:
function Post_body(source, args, context, info) {
  // return the post body only if the user is the post's author
  return postRepository.getBody({ user: context.user, post: obj })
}
```
````

benjie (changes requested):
This is really good! I'm going to follow up with a commit to add trusted documents to the advice, users who use trusted documents can mostly skip over disabling introspection, query cost analysis, and similar concerns.

benjie (approved):

benjie:
Follow up with trusted documents can be found here: https://github.com/graphql/graphql-js/pull/4418

PR #4401, opened 2025-05-19
docs: add guide on directives

sarahxsanders:
Adds guide "Using Directives in GraphQL.js"

### DIFF
--- a/website/pages/docs/using-directives.mdx
+++ b/website/pages/docs/using-directives.mdx
@@ -3,6 +3,8 @@ title: Using Directives in GraphQL.js
 sidebarTitle: Using Directives
 ---
 
+# Using Directives in GraphQL.js
+
 Directives let you customize query execution at a fine-grained level. They act like
 annotations in a GraphQL document, giving the server instructions about whether to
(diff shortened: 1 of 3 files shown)


website/pages/docs/using-directives.mdx
```diff
@@ -0,0 +1,276 @@
...
+
+This guide covers how GraphQL.js handles built-in directives, how to define and apply
+custom directives, and how to implement directive behavior during execution.
+
+## How GraphQL.js handles built-in directives
+
+GraphQL defines built-in directives that control query execution or describe metadata
+about the schema. These include: 
+
+- `@include` and `@skip`: Conditionally include or skip fields
```
JoviDeCroock: This is for execution language and the other for SDL  should we clarify that

JoviDeCroock (approved):

sarahxsanders:
Resolved that feedback, thank you @JoviDeCroock!!

benjie (commented):
Excellent docs; the only thing I felt was missing was that we should advise users not to use directives where field arguments could be used instead. I've raised this advice as https://github.com/graphql/graphql-js/pull/4417

PR #4402, opened 2025-05-20
docs: add guide for operation complexity controls

sarahxsanders:
Adds guide "Operation Complexity Controls"

### DIFF
--- a/website/pages/docs/operation-complexity-controls.mdx
+++ b/website/pages/docs/operation-complexity-controls.mdx
@@ -1,17 +1,23 @@
 ---
-title: Query Complexity Controls
+title: Operation Complexity Controls
 ---
 
-# Query Complexity Controls
+# Operation Complexity Controls
 
 GraphQL gives clients a lot of flexibility to shape responses, but that
 flexibility can also introduce risk. Clients can request deeply nested fields or 
-large volumes of data in a single query. Without controls, these operations can slow
+large volumes of data in a single operation. Without controls, these operations can slow
 down your server or expose security vulnerabilities.
 
-This guide explains how to measure and limit query complexity in GraphQL.js
+This guide explains how to measure and limit operation complexity in GraphQL.js
 using static analysis. You'll learn how to estimate the cost
-of a query before execution and reject it if it exceeds a safe limit.
+of an operation before execution and reject it if it exceeds a safe limit.
+
+<Callout type="info" emoji="ℹ️">
+  In production, we recommend using [trusted documents](https://graphql.org/learn/persistence/)
+  rather than analyzing arbitrary documents at runtime. Complexity analysis can still be
+  useful at build time to catch expensive operations before they're deployed.
+</Callout>
 
 ## Why complexity control matters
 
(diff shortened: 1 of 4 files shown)


benjie (changes requested):
This is really good, and I'm excited to see us addressing this topic! However, ideally we should not promote a particular module for complexity analysis (unless it's one we control), otherwise we open ourselves to adding that to GraphQL.js' security surface area, and we artificially limit competition/innovation within the ecosystem. Instead we should talk about it in more general terms (and perhaps give a few examples of modules that can be used for this). In addition:

1. We should strongly encourage the use of trusted documents over complexity limits at runtime
2. We should note that complexity limits can be subject to attack themselves (for example, via fragment cycles), so it's generally best to run them after validation but before execution unless we know they're safe to run alongside validation
3. The validation phase doesn't have access to variables (validation tends to happen once per document independent of variables) so complexity limits depending on variables will need special plumbing
4. Certain field types make operations more complex, in particular lists (and, to a lesser extent, abstract types); we should comment how these "multiply up" the complexity.
5. We should note that this should form part of a layered security approach - referencing https://graphql.org/learn/security/
6. We should note that with trusted documents, complexity analysis can still be valuable to warn engineers about expensive operations, and can mostly be done at build-time rather than run-time.
7. We should be careful with the use of the word "query", because we're talking about mutations and subscriptions too really. Consider "operation" or "document" as appropriate ([glossary](https://github.com/graphql/graphql-wg/blob/main/rfcs/glossary/Appendix%20C%20--%20Glossary.md)), or simply elide "query" entirely: "complexity controls".

I don't think you need to change the document very much to accommodate this feedback - e.g. you can keep all the examples, just make it very clear that they're just examples from one of the modules and other approaches are available.

sarahxsanders:
@benjie thanks for much for the thorough feedback! My next commit has the following updates:
- Renamed the guide to "Operation Complexity Controls" and replaced mentions of "query complexity"
- Clarified tool neutrality, making it clear that `graphql-query-complexity` is one example among several (happy to expand on this more if needed)
- Added some callouts to address some of your other points (trusted documents, cycle safety and validation timing, etc.)
- Reworked the best practices to align with updates from your feedback
- Tried to position complexity analysis as part of a layered security approach instead of a standalone solution

Please let me know if there's anything else we should update! :)

website/pages/docs/operation-complexity-controls.mdx
```diff
@@ -0,0 +1,262 @@
...
+## Why complexity control matters
+
+GraphQL lets clients choose exactly what data they want. That flexibility is powerful, 
+but it also makes it hard to predict the runtime cost of a query just by looking
+at the schema.
+
+Without safeguards, clients could:
+
+- Request deeply nested object relationships
+- Use recursive fragments to multiply field resolution
```
benjie: GraphQL validation has always explicitly forbid recursive fragments ([fragment spreads must not form cycles](https://spec.graphql.org/July2015/#sec-Fragment-spreads-must-not-form-cycles)) so servers that allow this are not GraphQL compliant. Let's be more precise what we mean here:

```suggestion
- Use nested fragments to multiply field resolution
```

website/pages/docs/operation-complexity-controls.mdx
```diff
@@ -0,0 +1,262 @@
...
+useful, just in a different way. You can run it at build time to:
+
+- Warn engineers about expensive operations during development
+- Track changes to operation cost across schema changes
+- Define internal usage budgets by team, client, or role
+
+## Best practices
+
+- Use trusted documents in production when possible.
+- Use complexity analysis as a development-time safeguards.
```
benjie:
```suggestion
- Use complexity analysis as a development-time safeguard.
```

website/pages/docs/operation-complexity-controls.mdx
```diff
@@ -0,0 +1,262 @@
...
+In environments that use persisted or precompiled operations, complexity analysis is still
+useful, just in a different way. You can run it at build time to:
+
+- Warn engineers about expensive operations during development
+- Track changes to operation cost across schema changes
+- Define internal usage budgets by team, client, or role
+
+## Best practices
+
+- Use trusted documents in production when possible.
```
benjie:
```suggestion
- Only accept trusted documents in production when possible.
```

website/pages/docs/operation-complexity-controls.mdx
```diff
@@ -0,0 +1,262 @@
...
+flexibility can also introduce risk. Clients can request deeply nested fields or 
+large volumes of data in a single operation. Without controls, these operations can slow
+down your server or expose security vulnerabilities.
+
+This guide explains how to measure and limit operation complexity in GraphQL.js
+using static analysis. You'll learn how to estimate the cost
+of an operation before execution and reject it if it exceeds a safe limit.
+
+<Callout type="info" emoji="ℹ️">
+  In production, we recommend using [trusted documents](https://graphql.org/learn/persistence/)
```
benjie: I think this link was hallucinated, but we can link to #4418 now instead?

This URL is a guess; we should check it:

```suggestion
  In production, we recommend using [trusted documents](/docs/going-to-production#only-allow-trusted-documents)
```

benjie (changes requested):
I think we just need to update the URL and then we're good to go (with the following small tweaks):

benjie (approved):

PR #4406, opened 2025-05-22
docs: add guide on subscriptions

sarahxsanders:
Guide for subscriptions

### DIFF
--- a/website/pages/docs/subscriptions.mdx
+++ b/website/pages/docs/subscriptions.mdx
@@ -1,0 +1,151 @@
+---
+title: Subscriptions
+---
+
+# Subscriptions
+
+Subscriptions allow a GraphQL server to push updates to clients in real time. Unlike queries and mutations, which use a request/response model, subscriptions maintain a long-lived connection between the client and server to stream data as events occur. This is useful for building features like chat apps, live dashboards, or collaborative tools that require real-time updates.
+
+This guide covers how to implement subscriptions in GraphQL.js, when to use them, and what to consider in production environments.
+
+## What is a subscription?
+
+A subscription is a GraphQL operation that delivers ongoing results to the client when a specific event happens. Unlike queries or mutations, which return a single response, a subscription delivers data over time through a persistent connection.
+
+GraphQL.js implements the subscription execution algorithm, but it's up to you to connect it to your event system and transport layer. Most implementations use WebSockets for transport, though any full-duplex protocol can work.
+
+## How execution works
+
+The core of subscription execution in GraphQL.js is the `subscribe` function. It works similarly to `graphql()`, but returns an `AsyncIterable` of execution results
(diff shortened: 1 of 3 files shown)


sarahxsanders:
@JoviDeCroock the prettier ci check is failing for `_meta.ts` but it looks correct 

**DISREGARD :)

website/pages/docs/subscriptions.mdx
```diff
@@ -0,0 +1,151 @@
...
+- A transport layer to maintain the connection
+
+The following examples use the [`graphql-subscriptions`](https://github.com/apollographql/graphql-subscriptions) package to set up an in-memory event system.
+
+### Install dependencies
+
+Start by installing the necessary packages:
+
+```bash
+npm install graphql graphql-subscriptions
```
JoviDeCroock, line 68: This is a package from Apollo, I don't think we should use thtat as the goto for the documentation

JoviDeCroock (approved):

benjie (commented):
Absolutely excellent :raised_hands:

website/pages/docs/subscriptions.mdx
```diff
@@ -0,0 +1,151 @@
...
+```
+
+To serve subscriptions over a network, you’ll also need a transport implementation. One option is [`graphql-ws`](https://github.com/enisdenjo/graphql-ws), a community-maintained WebSocket library. This guide focuses on schema-level implementation.
+
+### Set up a pub/sub instance
+
+Create a `PubSub` instance to manage your in-memory event system:
+
+```js
+import { PubSub } from 'graphql-subscriptions';
```
ardatan, line 78: Besides this unmaintained library, we have an alternative package `@graphql-yoga/subscriptions` that uses the standard [`EventTarget`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget), and it also allows you to use other event sources like Redis.
```ts
import { createPubSub } from '@graphql-yoga/subscriptions'
import { Redis } from 'ioredis'
import { createRedisEventTarget } from '@graphql-yoga/redis-event-target'
 
const publishClient = new Redis()
const subscribeClient = new Redis()
 
const eventTarget = createRedisEventTarget({
  publishClient,
  subscribeClient
})
 
const pubSub = createPubSub({ eventTarget })
```
And it is fully type safe compared to `graphql-subscriptions`;
```ts
const pubSub = createPubSub<{
  randomNumber: [randomNumber: number]
}>()
 
// Usage outside a GraphQL subscribe function
async function subscribe() {
  const eventSource = pubSub.subscribe('randomNumber')
 
  for await (const value of eventSource) {
    console.log(value)
    // dispose subscription after the first event has been published.
    eventSource.return()
  }
}
 
subscribe()
 
pubSub.publish('randomNumber', 3)
```

PR #4407, opened 2025-05-22
Minor validation tweaks from review

benjie:
Whilst reviewing that https://github.com/graphql/graphql-js/pull/3814 aligned with https://github.com/graphql/graphql-spec/pull/793, I spotted these two minor improvements.

### DIFF
--- a/src/type/validate.ts
+++ b/src/type/validate.ts
@@ -812,7 +812,7 @@ function createInputObjectDefaultValueCircularRefsValidator(
   ): void {
     // Start with an empty object as a way to visit every field in this input
     // object type and apply every default value.
-    return detectValueDefaultValueCycle(inputObj, {});
+    return detectValueDefaultValueCycle(inputObj, Object.create(null));
   };
 
   function detectValueDefaultValueCycle(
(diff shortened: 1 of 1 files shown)


src/type/validate.ts
```diff
@@ -910,7 +910,7 @@ function createInputObjectDefaultValueCircularRefsValidator(
 
     // Check to see if there is cycle.
     const cycleIndex = fieldPathIndex[fieldStr];
-    if (cycleIndex !== undefined && cycleIndex > 0) {
+    if (cycleIndex !== undefined) {
```
benjie, line 913: `cycleIndex` will never be <= 0 since it is only populated via `fieldPath.push()` which can never return a number less than `1`. I considered adding an invariant inside the if, but it didn't seem worthwhile.

JoviDeCroock (approved):

benjie:
@yaacovCR  / @JoviDeCroock If one of you wanted to add a :heavy_check_mark: on https://github.com/graphql/graphql-spec/pull/793 to confirm it aligns, that would be appreciated! Pretty sure it does though.

yaacovCR (approved):

PR #4411, opened 2025-05-24
docs: add guide on caching strategies

sarahxsanders:
Adds guide on caching strategies


website/pages/docs/caching-strategies.mdx
```diff
@@ -0,0 +1,281 @@
+---
+title: Caching Strategies
+---
+
+# Caching Strategies
+
+Caching is a core strategy for improving the performance and scalability of GraphQL
+servers. Because GraphQL allows clients to specify exactly what they need, the server often
+does more work per request compared to REST APIs.
```
benjie:
```suggestion
does more work per request (but fewer requests) compared to many other APIs.
```

website/pages/docs/caching-strategies.mdx
```diff
@@ -0,0 +1,281 @@
...
+export const resolvers = {
+  Query: {
+    async product(_, { id }, context) {
+      const cached = productCache.get(id);
+      if (cached) return cached;
+
+      const product = await context.db.products.findById(id);
+      productCache.set(id, product);
+      return product;
+    },
```
benjie: This code is vulnerable to race conditions where many requests for the same data can be kicked off in parallel during the initial fetch. By caching the promise rather than the result, concurrent requests to the same data can be cached without introducing a race condition:

```suggestion
    product(_, { id }, context) {
      const cached = productCache.get(id);
      if (cached) return cached;

      const productPromise = context.db.products.findById(id);
      productCache.set(id, productPromise);
      return productPromise;
    },
```

website/pages/docs/caching-strategies.mdx
```diff
@@ -0,0 +1,281 @@
...
+```
+
+The next example shows how to wrap your execution logic to check the cache first and store results
+afterward:
+
+```js
+// graphql/executeWithCache.js
+import { getCachedResponse, cacheResponse } from '../cache/queryCache.js';
+
+export async function executeWithCache({ cacheKey, executeFn }) {
```
benjie: This also suffers from race conditions, so I've wrapped it with code to track inflight requests:

```suggestion
/**
 * Stores in-flight requests to executeWithCache such that concurrent
 * requests with the same cacheKey will only result in one call to
 * `getCachedResponse` / `cacheResponse`. Once a request completes
 * (with or without error) it is removed from the map.
 */
const inflight = new Map();

export function executeWithCache({ cacheKey, executeFn }) {
  const existing = inflight.get(cacheKey);
  if (existing) return existing;
  
  const promise = _executeWithCacheUnbatched({ cacheKey, executeFn });
  inflight.set(cacheKey, promise);
  return promise.finally(() => inflight.delete(cacheKey));
}

async function _executeWithCacheUnbatched({ cacheKey, executeFn }) {
```

website/pages/docs/caching-strategies.mdx
```diff
@@ -0,0 +1,281 @@
...
+
+### Guidelines
+
+- Don't cache personalized or auth-sensitive data unless you scope the cache per user or token.
+- Invalidation is nontrivial. Consider TTLs, cache versioning, or event-driven purging.
+
+## Schema caching
+
+Schema caching is useful when your schema construction is expensive, for example, when
+you have dynamically generating types, you are stitching multiple schemas, or fetching
```
benjie:
```suggestion
you are dynamically generating types, you are stitching multiple schemas, or fetching
```

benjie (changes requested):
Really good! Since I know people have a tendency to copy/paste example code I've attempted to tighten it up so it doesn't suffer from race conditions.

benjie (approved):
Excellent :raised_hands:

PR #4412, opened 2025-05-27
internal: use empty merge commit to clean up git diff from 16.x.x

yaacovCR:
This no-op merge commit "pretends" to take in all of the changes from the 16.x.x branch since this branch was created from the 16.4 point release. These changes were actually cherry-picked from 16.x.x separately into this branch, but this merge commit allows us to pretend that the branch was created from 16.11.


yaacovCR:
This gets us to green when comparing (currently):

image

Or with
```bash
$ git merge-base --is-ancestor 16.x.x with-empty-merge-commit \
  && echo "✅ 16.x.x is ancestor of with-empty-merge-commit → fast-forward OK" \
  || echo "❌ no fast-forward possible (16.x.x is not ancestor of with-empty-merge-commit)"
✅ 16.x.x is ancestor of with-empty-merge-commit → fast-forward OK
```

Because all of the commits merged in by this PR have no code-changes, i.e. this is a no-op merge, git bisect will still work. However, we are still technically non-linear (`git log --graph --oneline --decorate with-empty-merge-commit`):

image

....

image

....

image

yaacovCR:
What else does this get us?

Well, after this, doing a compare in the reverse direction (https://github.com/yaacovCR/graphql-js/compare/with-empty-merge-commit...graphql:graphql-js:16.x.x) will show us which commits are new on 16.x.x since we last did this, and could remind us what to cherry-pick. Now, for example, it will show us that we are completely up to date:

image

But the utility of that will fade immediately after we cherry-pick -- unless we do this no-op again.

JoviDeCroock (approved):

yaacovCR:
Sigh. This didn't work because of the merge strategy used. We have to do it directly on the branch.

PR #4405, opened 2025-05-21
docs: add guide on nullability

sarahxsanders:
Adds "Nullability in GraphQL.js" guide

PR #4416, opened 2025-05-29
Editorial for #4405 (nullability)

benjie:


website/pages/docs/nullability.mdx
```diff
@@ -55,8 +55,7 @@ You can use `GraphQLNonNull` with:
 
 - Field types
 - Argument types
-- Input object fields
-- Return types for resolvers
```
benjie, line 59: "Return types for resolvers" is the same as field types.

JoviDeCroock (approved):

PR #4422, opened 2025-05-31
Updates to schema coordinates (#3044)

magicmark:
Updates to https://github.com/graphql/graphql-js/pull/3044

Implements the new spec changes:

- Split apart the AST nodes and parsing for the new coordinate types (`FieldCoordinate`, `ArgumentCoordinate` etc)
- Added `MyEnum::value` separator syntax
- Explicit error handling for "containing elements"
- [ ] (todo?) throw on returning meta-fields

### DIFF
--- a/src/language/ast.ts
+++ b/src/language/ast.ts
@@ -182,7 +182,11 @@ export type ASTNode =
   | UnionTypeExtensionNode
   | EnumTypeExtensionNode
   | InputObjectTypeExtensionNode
-  | SchemaCoordinateNode;
+  | TypeCoordinateNode
+  | MemberCoordinateNode
+  | ArgumentCoordinateNode
+  | DirectiveCoordinateNode
+  | DirectiveArgumentCoordinateNode;
 
 /**
  * Utility type listing all nodes indexed by their kind.
@@ -288,7 +292,13 @@ export const QueryDocumentKeys: {
   UnionTypeExtension: ['name', 'directives', 'types'],
   EnumTypeExtension: ['name', 'directives', 'values'],
   InputObjectTypeExtension: ['name', 'directives', 'fields'],
-  SchemaCoordinate: ['name', 'memberName', 'argumentName'],
+
+  // Schema Coordinates
+  TypeCoordinate: ['name'],
+  MemberCoordinate: ['name', 'memberName'],
+  ArgumentCoordinate: ['name', 'memberName', 'argumentName'],
+  DirectiveCoordinate: ['name'],
+  DirectiveArgumentCoordinate: ['name', 'argumentName'],
 };
 
 const kindValues = new Set<string>(Object.keys(QueryDocumentKeys));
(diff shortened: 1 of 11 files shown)


src/language/parser.ts
```diff
@@ -1481,12 +1486,42 @@ export class Parser {
...
+      return this.node<MemberCoordinateNode>(start, {
+        kind: Kind.MEMBER_COORDINATE,
+        name,
+        memberName,
+      });
+    }
+
+    return this.node<TypeCoordinateNode>(start, {
+      kind: Kind.TYPE_COORDINATE,
       name,
```
magicmark: this is a bit stinky, it seems like it would be more idiomatic to delegate parsing each intermediate node into its own function. [Y/n]?

src/language/ast.ts
```diff
@@ -288,7 +292,13 @@ export const QueryDocumentKeys: {
   UnionTypeExtension: ['name', 'directives', 'types'],
   EnumTypeExtension: ['name', 'directives', 'values'],
   InputObjectTypeExtension: ['name', 'directives', 'fields'],
-  SchemaCoordinate: ['name', 'memberName', 'argumentName'],
+
+  // Schema Coordinates
+  TypeCoordinate: ['name'],
+  MemberCoordinate: ['name', 'memberName'],
+  ArgumentCoordinate: ['name', 'memberName', 'argumentName'],
```
martinbonnin: Since this is always for fields at the moment, we could just use `fieldName`:

```suggestion
  ArgumentCoordinate: ['name', 'fieldName', 'argumentName'],
```

src/language/parser.ts
martinbonnin: This LGTM but I'm really not the best TS person do don't take me into account too much ^^

martinbonnin (approved):
Thanks for looking into this ❤️

src/language/ast.ts
```diff
@@ -288,7 +292,13 @@ export const QueryDocumentKeys: {
   UnionTypeExtension: ['name', 'directives', 'types'],
   EnumTypeExtension: ['name', 'directives', 'values'],
   InputObjectTypeExtension: ['name', 'directives', 'fields'],
-  SchemaCoordinate: ['name', 'memberName', 'argumentName'],
+
+  // Schema Coordinates
+  TypeCoordinate: ['name'],
+  MemberCoordinate: ['name', 'memberName'],
```
JoviDeCroock: Can we call this `EnumMemberCoordinate`, `MemberCoordinate` feels like it could be a field or an enum member, linearly should we split between enum-member and field then? I am personally just conflicted about `EnumMember` vs `TypeField`

JoviDeCroock:
Code looks good, mainly feeling conflicted about conflating enum-members and type-fields. Input-fields and type-fields feel similar compared to enum-members

src/language/ast.ts
martinbonnin: I was wondering the same. Maybe we need `FieldCoordinate`, `InputFieldCoordinate` and `EnumMemberCoordinate`?

benjie: Syntactically these are all the same: `Foo.bar`; there's no way to tell which is which without access to a schema. Member coordinate seems right to me.

src/language/parser.ts
benjie: I don't like the continual conditions on `ofDirective`; I'd rather see:

```ts
if (ofDirective) {
  if (argumentName) return ...
  else return ...
} else if (memberName) {
  if (argumentName) return ...
  else return ...
} else return ...
```

But that's just a mild preference.

src/language/ast.ts
magicmark: Leaving as is for now until there's consensus on this.

I'm leaning towards keeping it as is for the following 2 reasons:

1. This distinction is only able to be understood in the context of a schema, which we don't yet have at the time of parsing the schema coordinate. (Even though the spec says coordinates are only valid in the context of a schema, which I think is a fine as a semantic, we may still want to be able to construct an "invalid" coordinate in code as has been pointed out previously.)
2. It would lead to a messier implementation, since parsing the schema coordinate now needs to take in the schema object too. Doable but yucky.

martinbonnin: Good points 👍

martinbonnin (approved):

yaacovCR (approved):

benjie:
@JoviDeCroock has a good point, and perhaps it would actually be better to represent this in the schema coordinates syntax itself:

```
Type # Any named type
Type.field # A field (e.g. on an object, interface, or inputObject type)
Type.field(argName:) # An argument (on an object or interface type)
Type::value # an Enum value
```

`::` is the syntax used to reference an enum value in [C++](https://en.cppreference.com/w/cpp/language/enum.html), [Rust](https://doc.rust-lang.org/reference/items/enumerations.html), [PHP](https://www.php.net/manual/en/language.enumerations.basics.php), [Perl](https://manpages.ubuntu.com/manpages/questing/man3/enum.3pm.html) and others.

benjie:
I have raised https://github.com/magicmark/graphql-spec/pull/1; if we agree then I'll also make these changes to this branch and get it merged into https://github.com/graphql/graphql-js/pull/3044

magicmark:
ack. @benjie done! added the enum changes.

src/language/ast.ts
magicmark: addressed, this is split out into :: for enums.

benjie (approved):

benjie:
Merging this even though it doesn't pass CI under the instruction of the champion :+1:

Issue #4362, opened 2025-03-24
GraphQLSchema does not completely validate names


ab-pm:
The `validateName` function that is used to validate lots of schema parts is currently only checking that names are not reserved:
https://github.com/graphql/graphql-js/blob/6b253e7159a6a6c64c05bcc6bb863fef5a42eeb7/src/type/validate.ts#L206-L217
It should also check that the `name` complies with https://spec.graphql.org/October2021/#Name, e.g. by testing against the regex `/^(?!__)[A-Za-z_][A-Za-z0-9_]*$/`.
Otherwise it's possible to construct schemas (via the constructor, not by parsing) that upon printing would lead to invalid syntax, or fields which could never be queried.

benjie:
Hi @ab-pm! Are you using the `validateSchema()` function to validate your schema has no naming issues or other problems?

ab-pm:
I personally don't, I just thought it would be useful.
(What I'd need is `print` calling `validateName` on AST tokens, since I deal with potentially bogus/malicious AST inputs for query documents - but I understand that in `print` this might have a noticeable performance impact and it's not really the responsibility of a printer to do validation.)

I just looked through the source of graphql-js (as the GraphQL reference implementation) to see whether there is something useful to validate strings as `Name`s, and found this function which (in the comment) claims to "*Ensure names are valid*" but doesn't really live up to that.
In a schema-first style of development this will never happen, but I could see some code-first (or even generated) schemas being constructed with invalid names, and an early error might be helpful for their authors. I admit this hasn't happened to me, so feel free to close this if you consider this a non-issue.

yaacovCR:
Just looking this over, the strange naming of this function seems to be an artifact of refactoring, see https://github.com/graphql/graphql-js/pull/3288

We moved most of the validation of naming into the type constructors, so that except for the underscoring, conformity to the spec is guaranteed.

There is no context in that PR as to why it is preferable to fail at schema construction for name errors (rather than leaving that to the validation step with well formatted errors).

Totally guessing => perhaps @IvanGoncharov was motivated by a potential security concern with prototype pollution and so wanted to make sure that even non-validated schemas wouldn't have this issue?

Things to do:
1. clarify motivation of above PR
2. rename validateName function or add comment explaining hx

ab-pm:
> *@ab-pm*: I just looked through the source of graphql-js (as the GraphQL reference implementation) to see whether there is something useful to validate strings as Names, and found this function

[...]

PR #4423, opened 2025-06-03
move validation of names into validateSchema

yaacovCR:
See discussion in #4362.

As noted by @ab-pm, this is a reversion of #3288 which itself reverted some of #1132. 

We might need some more context on the motivation behind #3288.


yaacovCR:

removing these checks is only a modest performance win, up to ~3% in my testing, and so it does not seem worth it to remove.

So closing this PR in favor of other solutions like #4556

PR #4424, opened 2025-06-03
buildClientSchema: build enum type value maps lazily

yaacovCR:
The option of passing a thunk was made available by: https://github.com/graphql/graphql-js/pull/4018 .

Always passing the thunk helps contribute to a "lazy" build process, which gives the following performance boost:

image

### DIFF
--- a/src/type/definition.ts
+++ b/src/type/definition.ts
@@ -1563,7 +1563,7 @@ export class GraphQLEnumType /* <T> */ implements GraphQLSchemaElement {
 
   private _values:
     | ReadonlyArray<GraphQLEnumValue /* <T> */>
-    | (() => GraphQLEnumValueConfigMap);
+    | (() => ReadonlyArray<GraphQLEnumValue>) /* <T> */;
 
   private _valueLookup: ReadonlyMap<any /* T */, GraphQLEnumValue> | null;
   private _nameLookup: ObjMap<GraphQLEnumValue> | null;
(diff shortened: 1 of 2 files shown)


src/type/definition.ts
```diff
@@ -1534,13 +1534,7 @@ export class GraphQLEnumType /* <T> */ implements GraphQLSchemaElement {
...
     this.extensionASTNodes = config.extensionASTNodes ?? [];
 
-    this._values =
-      typeof config.values === 'function'
-        ? config.values
-        : Object.entries(config.values).map(
-            ([valueName, valueConfig]) =>
-              new GraphQLEnumValue(this, valueName, valueConfig),
-          );
+    this._values = defineEnumValues.bind(undefined, this, config.values);
```
benjie, line 1579: I'd be surprised if the `.bind()` method ends up faster than the simpler/shorter/clearer version:

```suggestion
    this._values = () => defineEnumValues(this, config.values);
```

Both require memory allocation, and traditionally `.bind()` has ended up slower than just using a new function, though the V8 team may have optimised that difference away by now!

yaacovCR: I could benchmark when I get a chance I think we are just using the convention from elsewhere in the file.

benjie: I'd go with the simpler code if there's no perf difference. My guess is that `.bind()` was assumed to be more efficient, but I don't think it is.

yaacovCR: Hmmm. It seems in some scenarios to carry a small performance cost. I am wary of microbenchmarks but they are all we have at this point.

https://github.com/graphql/graphql-js/pull/4470

benjie: Happy that the v8 team have optimised this! Always struck me as weird that a new function was cheaper in the past. Looks like the relevant V8 fix went out in Node 8.3… so a little while ago now! https://github.com/davidmarkclements/v8-perf?tab=readme-ov-file#partial-application-currying-and-binding

yaacovCR:
this change as implemented will result in bad enum value names not being checked unless `type.getValues()` is called in parallel to how bad field names are not checked unless `type.getFields()` is called.

relevant to: https://github.com/graphql/graphql-js/issues/4362

benjie:
Are they checked by the time `validateSchema(schema)` completes? If not, that does seem a concern.

yaacovCR:
Yes, they are. The linked issue requests earlier errors on bad names, so relevant there, but validation cannot pass with bad names.

PR #4427, opened 2025-06-05
pass assumeValid to constructors

yaacovCR:
context: https://github.com/graphql/graphql-js/issues/4362

image


yaacovCR:

removing these checks currently seems to be only a modest performance win, up to ~3% in my current testing, and so it does not seem worth it to remove (although the numbers above suggest a performance win of maybe 5%, perhaps that could be further explored).

Anyway, in the meantime, closing this PR in favor of other solutions like #4556.

PR #4428, opened 2025-06-05
Keep order of specified definitions consistent with the spec

martinbonnin:
See https://github.com/graphql/graphql-spec/pull/1037#discussion_r2026533747

### DIFF
--- a/src/type/__tests__/introspection-test.ts
+++ b/src/type/__tests__/introspection-test.ts
@@ -226,17 +226,6 @@ describe('Introspection', () => {
                   deprecationReason: null,
                 },
                 {
-                  name: 'specifiedByURL',
-                  args: [],
-                  type: {
-                    kind: 'SCALAR',
-                    name: 'String',
-                    ofType: null,
-                  },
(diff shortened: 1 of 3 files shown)


benjie:
For context: we want to add a new appendix to the GraphQL spec that covers the full introspection schema. We want that to be produced by printing the GraphQL.js introspection schema, but we also want it to line up with the fields and arguments defined in the introspection chapter, so field order matters. This PR better aligns the field order in GraphQL.js with the specification.

benjie (approved):
I'm approving this, but with the caveat this might not be the final edit to field order in introspection!

martinbonnin:
Made a few polls and people are predominantly sorting fields semantically. 

Screenshot 2025-06-20 at 15 36 14
Screenshot 2025-06-20 at 15 36 36
Screenshot 2025-06-20 at 15 36 48


I'll close this PR and update the spec text instead. @benjie if you have any preferred order or if you'd prefer to change the order in both graphql-js and the spec, now is the time!

PR #3402, opened 2021-12-03
Add descriptions to executable definitions

IvanGoncharov:
Implements https://github.com/graphql/graphql-spec/pull/892

PR #4430, opened 2025-06-06
Descriptions on executable documents

fotoetienne:
Enhance GraphQL parser and printer to support descriptions for operations, variables, and fragments

This commit is based on https://github.com/graphql/graphql-js/pull/3402 , rebased onto graphql-js 16 and with added tests

Implements https://github.com/graphql/graphql-spec/pull/1170


src/language/parser.ts
```diff
@@ -286,24 +292,14 @@ export class Parser {
...
-        case 'fragment':
-          return this.parseFragmentDefinition();
         case 'extend':
+          if (hasDescription) {
+            throw syntaxError(
+              this._lexer.source,
+              this._lexer.token.start,
+              'Unexpected description, descriptions are not supported on type extensions.',
+            );
+          }
```
benjie: Currently this changes the behavior of the `this.unexpected()` fallback lower down. I think we should restore the previous format - two switch statements with a "no descriptions" assertion between them.

benjie (changes requested):
Looks good! We should have a test to ensure that a description on an anonymous query results in a sensible parse error if we don't already have one, plus the comment I made before. Other than that, I think this is good to go!

```graphql
"""Invalid"""
{ __typename }
```

src/language/parser.ts
yaacovCR: @fotoetienne @benjie 

I took a stab at addressing this feedback slightly differently to better preserve existing behavior, please see

https://github.com/fotoetienne/graphql-js/pull/1

benjie:
Merged via https://github.com/graphql/graphql-js/pull/4482

PR #4431, opened 2025-06-06
move name validation in production to separate validation step

yaacovCR:
Motivation:

- Improve known-valid schema construction time as much as possible.

Background:
- All schema validation except name validation is currently performed separately in the `validateSchema(...)`, which optimizes schema construction for known-valid schemas.
- Name validation is an exception, which provides useful localized dev feedback if a naming mistake has been made, see #4362 and especially [this comment](https://github.com/graphql/graphql-js/issues/4362#issuecomment-2910123920) which details some history of the code changes.
- This PR adds name validation to the `validateSchema(...)` step and makes it so that in a production environment only, name validation is skipped during schema construction.

Depends on:

- #4426

Prior proposals:
- #4423
- #4427

### DIFF
--- a/src/__tests__/env-test.ts
+++ b/src/__tests__/env-test.ts
@@ -1,0 +1,20 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { getEnv, setEnv } from '../setEnv.js';
+
+describe('Env', () => {
+  it('should return undefined if environment is not set', () => {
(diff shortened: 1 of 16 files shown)


yaacovCR:

removing these checks is only a modest performance win, up to ~3% in my testing, and so it does not seem worth it to remove.

So closing this PR in favor of other solutions like #4556

PR #4439, opened 2025-06-22
fix(benchmark): fix benchmark shell command built from environment values

odaysec:
https://github.com/graphql/graphql-js/blob/ba4b411385507929b6c4c7905eb04b3e6bd1e93c/benchmark/benchmark.js#L32-L33
https://github.com/graphql/graphql-js/blob/ba4b411385507929b6c4c7905eb04b3e6bd1e93c/benchmark/benchmark.js#L65-L79
https://github.com/graphql/graphql-js/blob/ba4b411385507929b6c4c7905eb04b3e6bd1e93c/benchmark/benchmark.js#L89-L90

Fix the issue should avoid passing dynamically constructed shell commands to `cp.execSync`. Instead, we can use `cp.execFileSync`, which allows us to pass the command and its arguments as separate parameters. This approach ensures that the shell does not interpret special characters in the arguments, mitigating the risk of command injection or misinterpretation.
1. Replace the `exec` function to use `cp.execFileSync` instead of `cp.execSync`.
2. Update all calls to `exec` to pass the command and arguments as separate parameters.

## Recommendation
If possible, use hard-coded string literals to specify the shell command to run, and provide the dynamic arguments to the shell command separately to avoid interpretation by the shell. Alternatively, if the shell command must be constructed dynamically, then add code to ensure that special characters in environment values do not alter the shell command unexpectedly.

### DIFF
--- a/benchmark/benchmark.js
+++ b/benchmark/benchmark.js
@@ -29,8 +29,8 @@ function localDir(...paths) {
   return path.join(__dirname, '..', ...paths);
 }
 
-function exec(command, options = {}) {
-  const result = cp.execSync(command, {
+function exec(command, args = [], options = {}) {
+  const result = cp.execFileSync(command, args, {
     encoding: 'utf-8',
     stdio: ['inherit', 'pipe', 'inherit'],
     ...options,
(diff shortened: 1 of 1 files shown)


benjie:
Can you explain how `hash` and `repoDir` variables could be supplied with malicious values that would cause command injection without the person who is running the code having done it deliberately to themselves?

yaacovCR:
Closing this for now, feel free to reopen as needed.

PR #4445, opened 2025-07-01
Avoid crash when oneOf input variable is not defined

### LINKED ISSUE #4443: `validate` throws a TypeError: Cannot read properties of undefined (reading 'type')
https://github.com/graphql/graphql-js/blob/c18e9f6aada9ae086ddf836e4d822cf1426f3868/src/validation/rules/ValuesOfCorrectTypeRule.ts#L219

I've hit a `TypeError` being thrown from this line when testing against some invalid queries.

I'm not using this library directly, but instead via https://github.com/apollographql/apollo-server (4.12.2), so maybe I should report it there, let me know if so. It's not really easy to provide a reproducer at this level either, so maybe that's another signal...

Image

JoviDeCroock:
CC @benjie should this be added to the spec?

Fixes https://github.com/graphql/graphql-js/issues/4443

### DIFF
--- a/src/validation/rules/ValuesOfCorrectTypeRule.ts
+++ b/src/validation/rules/ValuesOfCorrectTypeRule.ts
@@ -216,6 +216,13 @@ function validateOneOfInputObject(
   if (isVariable) {
     const variableName = value.name.value;
     const definition = variableDefinitions[variableName];
+
+    // If the variable definition is missing, skip validation here.
+    // This should be caught by other validation rules.
+    if (!definition) {
+      return;
+    }
+
     const isNullableVariable = definition.type.kind !== Kind.NON_NULL_TYPE;
 
     if (isNullableVariable) {
--- a/src/validation/__tests__/ValuesOfCorrectTypeRule-test.ts
+++ b/src/validation/__tests__/ValuesOfCorrectTypeRule-test.ts
@@ -1125,6 +1125,19 @@ describe('Validate: Values of correct type', () => {
         },
       ]);
     });
+
+    it('Undefined variable in oneOf input object', () => {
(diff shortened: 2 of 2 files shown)


src/validation/rules/ValuesOfCorrectTypeRule.ts
```diff
@@ -216,6 +216,13 @@ function validateOneOfInputObject(
   if (isVariable) {
     const variableName = value.name.value;
     const definition = variableDefinitions[variableName];
+
+    // If the variable definition is missing, skip validation here.
+    // This should be caught by other validation rules.
+    if (!definition) {
+      return;
+    }
```
benjie, line 224: Technically "Values of Correct Type" in the spec only applies to literals ("For each literal Input Value {value} in the document:"), what you're actually dealing with here is maybe "All Variable Usages Are Allowed"?

But yes, I think in general the spec probably needs all of its validation rules to say "if <X> doesn't exist, return" where that situation has already been covered by a different validation rule explicitly checking that X exists.

yaacovCR: In the spec, it looks like we touched up where the validation happened around the time of https://github.com/benjie/graphql-spec/pull/1

In v17, it looks like we accomplished this in https://github.com/graphql/graphql-js/pull/4194

In v16, it looks like we touched up VariablesInAllowedPositions by https://github.com/graphql/graphql-js/pull/4363

So what remains to be done in v16 is to remove the variable bits entirely from ValuesOfCorrectType (created a branch to test around with this: https://github.com/yaacovCR/graphql-js/tree/alternative)

Pain points of multiple long-running development branches.... grumble, grumble

PR #4447, opened 2025-07-02
chore: drop unsupported node versions

yaacovCR:
we will test only versions that are not EOL: i.e. 20, 22, 24, and update the engines field to those same versions ~~give warnings on anything below 20~~.

we may further choose to require minimum subversions of 20, 22, and 24 that enable require(esm), but this PR holds off on that just yet.


package.json
```diff
@@ -26,7 +26,7 @@
     "graphql-js"
   ],
   "engines": {
-    "node": "^16.19.0 || ^18.14.0 || >=19.7.0"
+    "node": ">=20.0.0"
```
benjie: Let's set the required version to those that support `require(esm)` even if we don't end up needing it ourselves, we might need it to update our dependencies.

```suggestion
    "node": "^20.19.0 || ^22.12.0 || >=24.0.0"
```

benjie (approved):

package.json
benjie: We should at least exclude the EOL releases 21.x and 23.x - we don't want to support these.

```suggestion
    "node": "^20.0.0 || ^22.0.0 || >=24.0.0"
```

PR #4450, opened 2025-07-04
No `{Ignored}` tokens when parsing schema coordinates

magicmark:
I took a stab at @benjie's suggestion (which I 100% agree with) to disallow ignored tokens in schema coordinates.

(This caught an instance of whitespace in on of our tests!)

cc @yaacovCR @martinbonnin


src/language/lexer.ts
```diff
@@ -6,6 +6,18 @@ import { isDigit, isNameContinue, isNameStart } from './characterClasses.js';
...
+/**
+ * Configuration options to control lexer behavior
+ */
+export interface LexerOptions {
+  /**
+   * By default, ignored tokens are valid syntax and ignored when lexing.
+   * This may be disabled for certain grammars that specifically disallow
+   * ignored tokens (e.g. schema coordinates).
+   */
+  noIgnoredTokens?: boolean | undefined;
```
benjie: Just a "naming things" thing, but should we call it "forbid" or similar instead, to make the meaning clearer?

```suggestion
  forbidIgnoredTokens?: boolean | undefined;
```

src/language/parser.ts
```diff
@@ -199,9 +208,12 @@ export function parseType(
  */
 export function parseSchemaCoordinate(
   source: string | Source,
-  options?: ParseOptions,
+  options: ParseOptions & { noIgnoredTokens?: true } = {},
```
benjie: I didn't spot this before, but `parseSchemaCoordinate` should have its own options as they won't overlap with the main options hardly at all:

- `noIgnoredToken` - will always be true
- `noLocation` - the only relevant option?
- `maxTokens` - the grammar ends so rapidly there's literally no value in this
- `experimentalFragmentArguments` - irrelevant

```suggestion
  options?: { noLocation?: boolean | undefined },
```

benjie (changes requested):
I agree with your concerns on this, but do not have a better suggestion currently. Perhaps the documentation for `noIgnoredTokens`/`forbidIgnoredTokens` should explicitly note that setting the value to `true` will break parsing of the vast majority of GraphQL documents, and is _only_ intended for schema coordinates?

There's no way to form the query `{a b}` without ignored tokens.

I suggested an inline options object here, but really you ought to use a named export called `ParseSchemaCoordinateOptions` or similar.

benjie:
@magicmark I was starting to write an alternative here, but decided to just implement it, so check out https://github.com/graphql/graphql-js/pull/4451 - it adds a single commit on top of this PR that refactors to using an alternative lexer for schema coordinates. (If you like it just cherry-pick that commit into this PR and we can close the other one.)

src/language/parser.ts
magicmark: yeah makes sense

magicmark:
@benjie done! cherry picked.

I also added a test for `SchemaCoordinateLexer` (was getting a coverage report issue on `get [Symbol.toStringTag]()`)

src/language/__tests__/lexer-test.ts
```diff
@@ -1189,6 +1193,33 @@ describe('Lexer', () => {
   });
 });
 
+describe('SchemaCoordinateLexer', () => {
+  it('can be stringified', () => {
+    const lexer = new SchemaCoordinateLexer(new Source('Name.field'));
+    expect(Object.prototype.toString.call(lexer)).to.equal(
```
benjie, line 1199: This should be roughly equivalent?
```suggestion
    expect(String(lexer)).to.equal(
```

benjie (approved):
Looks good to me, but @JoviDeCroock / @yaacovCR should sign off in the changes to the public API

src/language/parser.ts
```diff
@@ -199,9 +221,11 @@ export function parseType(
  */
 export function parseSchemaCoordinate(
   source: string | Source,
-  options?: ParseOptions,
+  options?: ParseSchemaCoordinateOptions,
 ): SchemaCoordinateNode {
-  const parser = new Parser(source, options);
+  // Ignored tokens are excluded syntax for the schema coordinates.
+  const _options = { ...options, Lexer: SchemaCoordinateLexer };
+  const parser = new Parser(source, _options);
```
JoviDeCroock: NIT: we can just inline this new options creation

JoviDeCroock (approved):

src/language/parser.ts
magicmark: felt like a bit too much logic for one line but nbd

src/language/__tests__/lexer-test.ts
magicmark: deferring to prior art / established convention https://github.com/search?q=repo%3Agraphql%2Fgraphql-js%20Object.prototype.toString.call&type=code

(but I agree with you)

PR #4451, opened 2025-07-04
No {Ignored} tokens when parsing schema coordinates (take 2)

benjie:
This builds on #4450 with an alternative approach that allows the user to replace the lexer. I think this is a valuable feature independent of schema coordinates - e.g. if someone wanted to implement a different security mechanism (more powerful than "maxTokens") this would give them the ability to do so.


magicmark:
merged into https://github.com/graphql/graphql-js/pull/4450

PR #4458, opened 2025-07-15
Sibling errors should not be added after propagation

yaacovCR:
Implementation of:

https://github.com/graphql/graphql-spec/pull/1184

@benjie

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -114,8 +114,41 @@ export interface ExecutionContext {
   fieldResolver: GraphQLFieldResolver<any, any>;
   typeResolver: GraphQLTypeResolver<any, any>;
   subscribeFieldResolver: GraphQLFieldResolver<any, any>;
-  errorPositions: Set<Path | undefined>;
-  errors: Array<GraphQLError>;
+  collectedErrors: CollectedErrors;
+}
+
+/**
+ * @internal
+ */
+class CollectedErrors {
+  private _errorPositions: Set<Path | undefined>;
+  private _errors: Array<GraphQLError>;
+  constructor() {
+    this._errorPositions = new Set<Path | undefined>();
+    this._errors = [];
+  }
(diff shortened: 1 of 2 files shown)


src/execution/execute.ts
```diff
@@ -605,12 +612,32 @@ function handleFieldError(
...
+  path: Path | undefined,
+): boolean {
+  if (errorPositions.has(path)) {
+    return true;
+  }
+  if (path === undefined) {
+    return false;
+  }
+  return hasNulledPosition(errorPositions, path.prev);
+}
```
benjie: Recursion can be expensive in JS, let's use a loop instead.

```suggestion
function hasNulledPosition(
  errorPositions: Set<Path | undefined>,
  startPath: Path | undefined,
): boolean {
  let path = startPath;
  while (path !== undefined) {
    if (errorPositions.has(path)) {
      return true;
    }
    path = path.prev;
  }
  return errorPositions.has(undefined);
}
```

benjie (dismissed):
This looks pretty good, minimal additional memory cost, reasonably efficient checks for existing errors, and zero cost if errors don't occur. The main memory cost is retaining `path` objects beyond their previous lifecycle, but that's incredibly marginal unless you have _millions_ of errors - and if that's the case you have bigger issues.

Let's make `hasNulledPosition` more efficient, but that's my only requested change. WDYT, @JoviDeCroock?

benjie (dismissed):
My one concern with this is that in future iteration people may not know to update `errors` and `errorPositions` concurrently.

Can we add a method to execution context to register errors instead (e.g. `exeContext.addError(error, path)`) and mark errors and errorPositions as readonly in ExecutionContext to discourage alternative manipulation?

yaacovCR:
@benjie i added a CollectedErrors utility class to more tightly couple `errors` and `errorPositions`

PR #4464, opened 2025-07-29
use `enableDevMode()` or development condition instead of NODE_ENV

yaacovCR:
extracted from / based on:

- #4437

Motivation:

This makes our build product cross-platform by dropping all Node.js specific code.

Differences:
- the old PR drops esm support, this PR does not
- the old PR uses `import 'graphql/dev'` to enable development mode on platforms that do not support setting custom conditions, while this PR uses: `import { enableDevMode } from 'graphql'; enableDevMode()`

From the docs included in this PR:

> In development mode, GraphQL.JS can provides additional runtime check appropriate for development-time errors, including primarily the erroneous inclusion of multiple GraphQL.JS modules.
>
> Unlike earlier versions of GraphQL.JS, by default, development mode is disabled. This is to best ensure that production builds do not incur the performance and bundle size penalties associated with the additional checks.
>
> Also, unlike earlier versions, development mode is not configured by use of environment variables, which are accessed in disparate ways on the various platforms. In particular, the `NODE_ENV` environment variable now has no effect on triggering development mode. Rather, development mode is either enabled:
> 1. explicitly, by calling `enableDevMode()` or
> 2. implicitly, by setting the 'development' condition, which is possible only on platforms that support `package.json` conditional exports and custom conditions.
>
> Conditional exports are supported by: Node.js, Deno (canary), Bun, Webpack 5, Rspack, Rollup (via the `node-resolve` plugin), esbuild, Vite, and Rsbuild. create-react-app and Next.js support conditional exports when using Webpack 5 as their bundler.
>
> Conditional exports are not supported by Deno (current), Webpack 4, Rollup (without the `node-resolve` plugin), or swc. create-react-app and Next.js do not support conditional exports when using Webpack 4 as their bundler, nor does Next.js yet
support conditional exports when using Turbopack (see https://github.com/vercel/next.js/discussions/78912).
>
> We encourage enabling development mode in a development environment, either explicitly or implicitly. This allows for an additional check to ensure that only a single GraphQL.JS module is used. Additional development-time checks may also be added in the future.


yaacovCR:
updated integration tests

thanks to @JoviDeCroock offline help, jest test working (demonstrated also at: https://github.com/apollographql/apollo-server/pull/8137)

in terms of allowing switching back to prod mode, I still cannot reverse performance degradation in offline benchmarking even after 100,000 invocations of prod function, so will leave this simply as `enableDevMode()` which in general will not need to be called on any platform that supports custom conditions.

PR #4465, opened 2025-07-30
increase memory allowance for benchmarks

yaacovCR:
esm appears to require slightly more memory than common-js


yaacovCR:
as well as a performance hit

https://github.com/nodejs/node/issues/44186

note that in my testing, breaking the live bindings does not improve the performance

PR #4468, opened 2025-08-06
predicates: replace instanceof with symbol based brand check

yaacovCR:
This PR updates our predicates to rely on a brand check rather than `instanceof`.

We still pass brand check to our `instanceOf` utility and in development mode use the same logic as previously to throw our error with multiple instances of `graphql`.

### DIFF
--- a/src/jsutils/instanceOf.ts
+++ b/src/jsutils/instanceOf.ts
@@ -7,19 +7,30 @@ const isProduction =
   process.env.NODE_ENV === 'production';
 
 /**
- * A replacement for instanceof which includes an error warning when multi-realm
- * constructors are detected.
+ * A utility which throws an error warning when multi-realm constructors are detected,
+ * triggered by a positive brand check and a negative instanceof (for an object).
+ *
+ * When not in development mode, only the brand check is used.
+ *
  * See: https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production
  * See: https://webpack.js.org/guides/production/
  */
-export const instanceOf: (value: unknown, constructor: Constructor) => boolean =
-  /* c8 ignore next 6 */
+export const instanceOf: (
+  brandCheck: true | undefined,
+  value: unknown,
+  constructor: Constructor,
+) => boolean =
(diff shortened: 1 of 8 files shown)


yaacovCR:
image

yaacovCR:
actually, we can't change the dev behavior at all, we must still throw the error for clashes with previous versions of this lib

saihaj (approved):

src/jsutils/instanceOf.ts
```diff
@@ -9,17 +9,28 @@ const isProduction =
...
+  value: unknown,
+  constructor: Constructor,
+) => boolean =
+  /* c8 ignore next 9 */
   // FIXME: https://github.com/graphql/graphql-js/issues/2317
   isProduction
-    ? function instanceOf(value: unknown, constructor: Constructor): boolean {
-        return value instanceof constructor;
+    ? function instanceOf(typeBrand: true | undefined): boolean {
+        return typeBrand === true;
```
JoviDeCroock: If a `symbol` is used then you can remove `instanceof` even in dev, i.e. `scalar.brand === symbol('scalar')` is much more effective and won't risk people having a correct dev-setup with deduped graphql and a bad eventual bundle.

People could run `node index.js` but bundle in production and run into the issue of dual realm.

The approach with symbol ensures that

- prod and dev behave the same way entirely
- you can still check the className for the dev-warning

yaacovCR: That makes sense. Let me benchmark that.

JoviDeCroock: It's an equality check, it should be 100% the same thing. That being said correctness > performance, the perf issues of graphql-js are a lot more architectural compared to micro opts

yaacovCR: @JoviDeCroock  I took a stab at a symbol-based approach, not 100% sure if this is along the lines of what you had in mind.

Micro-benchmarked, for the little it is worth, the only difference is that it's faster on one benchmark, so overall 👍 lgtm.

image

JoviDeCroock (approved):

PR #4469, opened 2025-08-08
perf: do not use undefined for empty error array

yaacovCR:
Revert previously attempted micro-optimization as it no longer produces (never produced?) a true performance benefit.

### DIFF
--- a/src/execution/IncrementalPublisher.ts
+++ b/src/execution/IncrementalPublisher.ts
@@ -32,7 +32,7 @@ import {
 export function buildIncrementalResponse(
   context: IncrementalPublisherContext,
   result: ObjMap<unknown>,
-  errors: ReadonlyArray<GraphQLError> | undefined,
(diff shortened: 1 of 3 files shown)


yaacovCR:
odd, reverting both changes singly seems fine, but when benchmarking together, reverting does have a small perf hit (on our synthetic microbenchmarks)

yaacovCR:
> odd, reverting both changes singly seems fine, but when benchmarking together, reverting does have a small perf hit (on our synthetic microbenchmarks)

going to retain just the changes to the errors array rather than to the incrementalDataRecords array. the fix in terms of the latter is really to remove the entire idea of the `GraphQLWrappedResult` and instead filter `incrementalDataRecords` based on the bubbled null positions after execution, but this is something we will probably attempt only after incremental data delivery lands so that the implementation at this point can be closer to the spec.

yaacovCR:
image

PR #4481, opened 2025-09-08
Fix incorrect `FormattedCompletedResult` fields and add a `FormattedPendingResult`

### LINKED ISSUE #4333: Inconsistencies in formatted types
There are a couple of `Formatted...Result` types that don't seem to be used anywhere. Therefore, they are also not checked for consistency.

For instance, there is a `FormattedCompletedResult`, but not a `FormattedPendingResult`, and the definition of `FormattedCompletedResult` seems to be wrong, it should have an "id" instead of "path" and "label".

I can contribute a fix, but it's unclear to me what purpose these types have if they are not used or referenced anywhere.

Particularly, should `FormattedCompletedResult` be changed or simply removed and replaced with `CompletedResult` like `PendingResult`? Or should the `Formatted...` be aliases if they are the same as the unformatted types?

jerelmiller:
Fixes #4333 

This change fixes the fields in `FormattedCompletedResult` to match `CompletedResult` (i.e. `id` and `errors`) using the `Formatted*` type variants. This change also adds a `FormattedPendingResult` as an counterpart to `PendingResult` to match the type patterns already established with other `Formatted*` types. 

I ran into this issue when working on adding support for the new incremental format in Apollo Client (https://github.com/apollographql/apollo-client/pull/12906).

### DIFF
--- a/src/execution/types.ts
+++ b/src/execution/types.ts
@@ -62,7 +62,7 @@ export interface FormattedInitialIncrementalExecutionResult<
   TExtensions = ObjMap<unknown>,
 > extends FormattedExecutionResult<TData, TExtensions> {
   data: TData;
-  pending: ReadonlyArray<PendingResult>;
+  pending: ReadonlyArray<FormattedPendingResult>;
   hasNext: boolean;
   extensions?: TExtensions;
 }
@@ -83,7 +83,7 @@ export interface FormattedSubsequentIncrementalExecutionResult<
   TExtensions = ObjMap<unknown>,
 > {
   hasNext: boolean;
-  pending?: ReadonlyArray<PendingResult>;
+  pending?: ReadonlyArray<FormattedPendingResult>;
   incremental?: ReadonlyArray<FormattedIncrementalResult<TData, TExtensions>>;
   completed?: ReadonlyArray<FormattedCompletedResult>;
   extensions?: TExtensions;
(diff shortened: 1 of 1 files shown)


src/execution/types.ts
```diff
@@ -156,15 +156,20 @@ export interface PendingResult {
   label?: string;
 }
 
+export interface FormattedPendingResult {
```
jerelmiller: I know there is a bit of back and forth on whether this should exist, but I went ahead and added it to continue the pattern of using `Formatted*` types. This is also useful as a library author so I don't have to guess why `PendingResult` exists, but not `FormattedPendingResult`.

I'm happy to pull this out if ultimately you decide it should use `PendingResult`. The real change I'm interested in for this fix anyways is the change to `FormattedCompletedResult`.

JoviDeCroock: Let's pull this out, it feels redundant as they are identical

JoviDeCroock (approved):

jerelmiller:
@JoviDeCroock done with [2091d65](https://github.com/graphql/graphql-js/pull/4481/commits/2091d653d1244969f7ec579d08f928a4f744a33d). Appreciate the review!

JoviDeCroock:
It's a bit odd that Completed and FormattedCompletedResult are identical, feels like these Formatted result types are redundant in the first place. Will take a deeper look on Monday and merge :)

PR #4482, opened 2025-09-17
Implement changes for executable descriptions

JoviDeCroock:
Revives and addresses comments from #4430

This is already in the spec and the impl is lagging behind so we might need to get this in 😅


src/language/__tests__/parser-test.ts
```diff
@@ -657,4 +728,101 @@ describe('Parser', () => {
...
+        }
+      `);
+      // Find the operation definition
+      const opDef = result.definitions.find(
+        (d) => d.kind === Kind.OPERATION_DEFINITION,
+      );
+      if (!opDef || opDef.kind !== Kind.OPERATION_DEFINITION) {
+        throw new Error('No operation definition found');
+      }
+      expect(opDef.description?.value).to.equal('Operation description');
```
phryneas: This one is purely stylistic, but it would probably be more in line with the rest of the document (and easier on the eye) to use `expectJSON(result).toDeepEqual({` instead of making a lot of individual `expect` calls.

phryneas (approved):
This looks good to me, only one stylistic remark.

martinbonnin (approved):

PR #4483, opened 2025-09-20
Adjust `items` type for `IncrementalStreamResult` and `FormattedIncrementalStreamResult`

jerelmiller:
Take the following:

```ts
declare const chunk: FormattedIncrementalResult<Record<string, unknown>>;

if ("items" in chunk) {
  const { items } = chunk;

  // TypeScript error
  items.slice(0);
}
```

When providing a generic to `*IncrementalResult`, the `items` property on `*IncrementalStreamResult` is typed incorrect due to the `Array<*>` being part of the default type instead of the `items` property. This means calling array methods on `items` results in a TypeScript error since TypeScript doesn't know its an array. This is incorrect since `items` can only be an array type and nothing else. Here is the reproduction in [TS Playground](https://www.typescriptlang.org/play/?ssl=10&ssc=1&pln=11&pc=1#code/JYWwDg9gTgLgBAbzgMWiAhjGBTAJgSQDsBjKbEbQmdAGwCVsBnAVxvgF84AzKCEOAEQBzKOjAALAI40BAbjgB6BXBgBPMEwBccAIwB2AHQAGYwFpaE9AYCcAWABQD3NmI10ZOMQiFG8YuOZCAGttVCgMLDwiUnJKanomVhgAHgYvKFxk3yhgQiEAGjhAoMIIAHdCAD5K2QcHYC44AAoBYBwQRgE4XM8A4IBKRAc4T29fRG72xjhOAF5e4tr7EaURtbgAPQB+B05sGkZsIeXRn3gkXEx0Gbh5-0XhxWV1zZ37diA).

This PR fixes this issue by moving the `Array<*>` wrapper to the `items` type so that this works as expected.

### DIFF
--- a/src/execution/types.ts
+++ b/src/execution/types.ts
@@ -114,13 +114,13 @@ export interface FormattedIncrementalDeferResult<
   extensions?: TExtensions;
 }
 
-interface StreamItemsRecordResult<TData = ReadonlyArray<unknown>> {
+interface StreamItemsRecordResult<TData = unknown> {
   errors?: ReadonlyArray<GraphQLError>;
-  items: TData;
+  items: ReadonlyArray<TData>;
 }
 
 export interface IncrementalStreamResult<
-  TData = ReadonlyArray<unknown>,
+  TData = unknown,
   TExtensions = ObjMap<unknown>,
 > extends StreamItemsRecordResult<TData> {
   id: string;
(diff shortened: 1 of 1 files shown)


src/execution/types.ts
```diff
@@ -129,11 +129,11 @@ export interface IncrementalStreamResult<
 }
 
 export interface FormattedIncrementalStreamResult<
-  TData = Array<unknown>,
+  TData = unknown,
   TExtensions = ObjMap<unknown>,
 > {
   errors?: ReadonlyArray<GraphQLFormattedError>;
-  items: TData;
+  items: ReadonlyArray<TData>;
```
jerelmiller, line 136: Everything else seems to use `ReadonlyArray`, so I figured this should probably be the same instead of `Array` like it was before.

yaacovCR:
Thanks for pointing this out. Turns out there were numerous typing errors (found so far!), and the types currently in v17 have been reworked to address, and it seems that among those changes, this issue was also addressed.

Apologies for the tremendous gap in feedback. We probably should have just merged this simple fix at the time and then done whatever rework on top of it. Your time and contribution is very much appreciated.

jerelmiller:
No problem! Glad to hear this change made it in in some form! Thanks for the update.

PR #4485, opened 2025-09-27
Adopt Error.cause in favor of originalError

### LINKED ISSUE #4484: Can we align `GraphQLError` to utilize `Error.cause`?
ECMAScript has standardized around `new Error(message, { cause })` pattern. We have the `originalError` option. I'd hope we could alias and deprecate `originalError` in favor of `cause` to bring uniformity. 

Maybe do a `super(message, { cause })` call?
Maybe do a `Object.defineProperty`?

And I also guess the `cause` is not necessarily is an `Error`, but is `unknown` per spec.

Malien:
The `@depreacted` marks might be heavy-handed. I am completely fine with reversing those.

For environments that don't have `Error.cause` (ES2021) `@deprecated` might be annoying to deal with, since they don't have the alternative.

Closes #4484

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -46,8 +46,13 @@ export interface GraphQLErrorOptions {
   positions?: Maybe<ReadonlyArray<number>>;
   /** Response path where this error occurred during execution. */
   path?: Maybe<ReadonlyArray<string | number>>;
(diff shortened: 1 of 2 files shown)


yaacovCR:
Thanks for this PR!!!!

Moved to `17.x.x` where we can deprecate at will. Perhaps parts can be backported?

yaacovCR (approved):

PR #4486, opened 2025-09-27
docs: production build optimization

sarahxsanders:
First guide in the new GraphQL production guide series, focusing specifically on build optimization for production deployments. Content covers:
- Environment variable configuration for GraphQL.js optimization
- Dead code elimination and tree shaking setup
- Browser compatibility handling
- Code splitting strategies
- Build tool configurations for Webpack, Rollup, and esbuild
- Bundle analysis and verification steps

I'd love feedback and a review, but please do not merge until the entire production guide series is ready to ship. The focused guides will replace the production monolith that exists today

### DIFF
--- a/website/pages/docs/_meta.ts
+++ b/website/pages/docs/_meta.ts
@@ -74,6 +74,7 @@ const meta = {
   },
   'development-mode': '',
   'going-to-production': '',
+  'production-build-optimization': '',
   'scaling-graphql': '',
 };
 
--- a/website/pages/docs/_meta.ts
+++ b/website/pages/docs/_meta.ts
@@ -50,6 +50,7 @@ const meta = {
     type: 'separator',
     title: 'Production & Scaling',
   },
+  'production-build-optimization': '',
   'going-to-production': '',
   'scaling-graphql': '',
 };
--- a/website/pages/docs/production-build-optimization.mdx
+++ b/website/pages/docs/production-build-optimization.mdx
@@ -1,0 +1,344 @@
+---
+title: Optimize your GraphQL build for production
+description:
+---
+
+When you deploy your GraphQL application to production, you 
+need to remove development-only code and minimize your file 
+sizes. This guide shows you principles and techniques for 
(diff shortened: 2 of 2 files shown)


sarahxsanders:
this guide and the following others are ready for review whenever:
https://github.com/graphql/graphql-js/pull/4508
https://github.com/graphql/graphql-js/pull/4509
https://github.com/graphql/graphql-js/pull/4490
https://github.com/graphql/graphql-js/pull/4516

yaacovCR (commented):
- A general question/comment: this PR and the related PRs should hopefully take over for the current "going to production" section, right?

- in terms of this one specifically, I think we need to distinguish between what `graphql-js` itself does in dev vs production mode and the capabilities that `graphql-js` provides to libraries/services/tooling/clients using `graphql-js`, some of which might be useful only in `development`.

In particular, the only thing that `graphql-js` currently does internally in `development` mode both in v16/v17 (and I believe v15 and perhaps earlier?) is enable some extra checks within every internal `instanceOf()` call that powers all `graphql-js` predicates and causes a minor increase in bundle size and a performance hit.

A good example of the second category of capabilities that `graphql-js` provides is the 'assumeValid` option that when used will cause the automatic call to `validateSchema` within `validate()` and `execute()` to short-circuit, i.e. returning without errors. And something similar within `assumeValidSDL` within `buildSchema`/`buildASTSchema`/`extendSchema`.

That's in terms of options, but in terms of functions, any client code that directly calls `validateSchema()` on a trusted schema or even `validate()` on a trusted document can be skipped in production.

Note that in v17, we are changing the internal `instanceOf` to no longer be directed by the `NODE_ENV`, but rather by the `production` vs `development` condition (building on our new support of conditional exports) and with `production` behavior now enabled b default. See: https://github.com/graphql/graphql-js/pull/4551

sarahxsanders:
thanks for the helpful feedback!! I went back through the source and reworked the guide :) take a look and let me know what you think

PR #4492, opened 2025-10-07
Fix int/float coercion according to the execution algo

### LINKED ISSUE #4488: Handling of `NaN` on `Int` scalar violates GraphQL Specification
When resolving an output value through `Int` scalar, `NaN` causes an error to be thrown:

> Int cannot represent non-integer value: NaN

```js
  const schema = makeExecutableSchema({
    typeDefs: /* GraphQL */ `
    type Query {
      testInt: Int
    }
  `,
    resolvers: {
      Query: {
        testInt: () => NaN,
      },
    },
  });
  const query = parse(/* GraphQL */ `query {
    testInt
  }`);
  const result = executeSync({
    schema,
    document: query,
  });
  expect(result).toEqual({ data: { testInt: null }, errors: [expect.objectContaining({ message: 'Int cannot represent non-integer value: NaN' })] });
```

However, the [GraphQL Specification](https://spec.graphql.org/June2018/#sec-Value-Completion) states that `NaN` should be coerced to `null` and returned: 

> 2. If result is null (**or another internal value similar to null such as undefined or NaN**), return null.

JoviDeCroock:
Resolves #4488 

This is kind of ... hard to assess where to best put this. I've put it in the scalar serialize as it felt the most appropriate, we could also put it in `completeValue` and return `null` regardless of whether this is an int/float scalar. We should however not throw errors so instead we'd return NaN from the scalars then instead of null and catch that in `completeValue`.

I found it a bit awkward to disregard the value completely for other types so I opted to put it in the scalars for that reason, i.e. there could be legitimate use cases for test-results or others where N/A is different from null.

We are sacrificing some form of accuracy here as a non-nullable field will return "cannot return null for x" when NaN is supplied while we do become more accurate by returning null for a nullable field rather than an error.

### DIFF
--- a/src/type/__tests__/scalars-test.ts
+++ b/src/type/__tests__/scalars-test.ts
@@ -161,17 +161,14 @@ describe('Type System: Specified scalar types', () => {
       expect(() => serialize(-1e100)).to.throw(
         'Int cannot represent non 32-bit signed integer value: -1e+100',
       );
-      expect(() => serialize('one')).to.throw(
-        'Int cannot represent non-integer value: "one"',
-      );
 
       // Doesn't represent number
       expect(() => serialize('')).to.throw(
         'Int cannot represent non-integer value: ""',
       );
-      expect(() => serialize(NaN)).to.throw(
-        'Int cannot represent non-integer value: NaN',
-      );
+
+      expect(serialize(NaN)).to.equal(null);
+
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
Closing due to inactivity, feel free to re-open if there's interest

PR #4504, opened 2025-10-28
polish: unify our formatted and non-formatted execution result types

yaacovCR:
via using a new generic type parameter with this pattern:

```ts
export interface ExecutionResult<
  TData = ObjMap<unknown>,
  TError extends GraphQLError | GraphQLFormattedError = GraphQLError,
> {
  errors?: ReadonlyArray<TError>;
  data?: TData | null;
  extensions?: TExtensions;
}

export interface FormattedExecutionResult<
  TData = ObjMap<unknown>,
> extends ExecutionResult<TData, TExtensions, GraphQLFormattedError> {}
```

This should remove maintenance burden, as we explicitly say that the formatted types only differ from the unformatted in that they have a formatted error type.

See:
https://github.com/graphql/graphql-js/pull/4481/
https://github.com/graphql/graphql-js/issues/4333

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -51,7 +51,7 @@ export interface GraphQLErrorOptions {
  * and stack trace, it also includes information about the locations in a
  * GraphQL document and/or execution result that correspond to the Error.
  */
-export class GraphQLError extends Error {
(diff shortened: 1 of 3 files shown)


yaacovCR:
closing for now, see #4557 and #4558

PR #4505, opened 2025-10-29
fix(buildSyncStreamItemQueue): fix early termination check timing

yaacovCR:
just tightening the loop!


yaacovCR:
closed in favor of #4515

PR #4511, opened 2025-11-22
fix: check the actual `GraphQLDirective` instance in `isSpecifiedDirective`

ardatan:
Related https://github.com/ardatan/graphql-tools/pull/7716

Fixes the bug when a `GraphQLDirective` instance is passed with the same name as specified directives, `isSpecifiedDirective` returns true unexpectedly.

### DIFF
--- a/src/type/__tests__/predicate-test.ts
+++ b/src/type/__tests__/predicate-test.ts
@@ -700,5 +700,13 @@ describe('Directive predicates', () => {
     it('returns false for custom directive', () => {
       expect(isSpecifiedDirective(Directive)).to.equal(false);
     });
+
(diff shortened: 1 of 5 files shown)


yaacovCR:
The spec assumes that all directives with the names of specified directives are the actual specified directives, but it seems to me we should be more permissive as possible within the implementation, especially to enable custom/experimental forms of the different directives.

In terms of future-proofing this, is the contention that we only need one function describing the actual specified directives, or might we need two different functions, one version that retains the current behavior (returning `true` for custom directives that override the specified ones) and one with the "fixed" behavior?

And basically we have the parallel question with `isSpecifiedScalar`...

PR #4515, opened 2025-12-03
introduce queues for streams and the overall graph allowing backpressure management

yaacovCR:
This PR replaces our IncrementalGraph implementation with a formalized WorkQueue construct with its own spec-like behavior.

The PR also enhances Queues to be more Repeater-like, handling pushing async values in order, as well as backpressure, even while maintaining the ability to be eager and batching results.

In terms of test changes, the re-architecture:
 - most significantly, demonstrates how we can transparently manage backpressure with streams even while allowing early execution. The amount that a producer is allowed to run ahead of the consumer is configurable, currently on a per-overall execution basis.
 - incidentally coalesces some stream items together where possible
 - uncovers at least one stream bug, as demonstrated in the additional test
 - avoids some tricky logic regarding errors that bubble to the root of a deferred fragment or stream if there are pre-existing errors

In terms of execution/types changes, the re-architecture:
- allows for removal of many of the confusing intermediate result types we had needed previously.
- unifies handling of sync and async stream queue creation
- overall simplifies the spec changes needed within execution

This version comes with a specification describing behavior of a generic incremental WorkQueue that should motivate the execution changes necessary to get incremental delivery merged within the spec.

@robrichard @brainkim


yaacovCR:
closing in favor of #4537

PR #4518, opened 2025-12-23
Fix TypeInfo.getInputType() for custom scalar list literals.

### LINKED ISSUE #4512: Custom scalars should allow variables in list literals
Given this file `repro.mjs`:

```javascript
import { buildSchema, validateSchema, parse, validate } from "graphql";

const schema = buildSchema(`
scalar GeoPoint
scalar Float

type Query {
  restaurantsNear(loc: GeoPoint!): [String]
}
`)

const schemaErrors = validateSchema(schema);
if (schemaErrors.length > 0) {
  console.dir(schemaErrors);
  throw new Error("Invalid schema");
}

const doc = parse(`
query ($lat: Float!) {
  works: restaurantsNear(loc: [1, 2])
  breaks: restaurantsNear(loc: [$lat, 2])    # <---- Validation error here
  alsoWorks: restaurantsNear(loc: {_wrapper: [$lat, 2]})
}
`)

const errors = validate(schema, doc);
if (errors.length > 0) {
  console.dir(errors);
  throw new Error("validate failed");
}
```

Run with `npm i graphql && node repro.mjs`. (Credits to @benjie for the idea of using a .mjs file for a self-contained repro)

**Expected**: no errors in all three fields
**Actual**: The second field named `breaks` causes a validation error:

```
[
  GraphQLError: Variable "$lat" of type "Float!" used in position expecting type "GeoPoint".
      at Object.leave (/path/to/project/node_modules/graphql/validation/rules/VariablesInAllowedPositionRule.js:64:17)
      at Object.leave (/path/to/project/node_modules/graphql/language/visitor.js:321:32)
      at Object.leave (/path/to/project/node_modules/graphql/utilities/TypeInfo.js:411:21)
      at visit (/path/to/project/node_modules/graphql/language/visitor.js:194:21)
      at validate (/path/to/project/node_modules/graphql/validation/validate.js:91:24)
      at file:///path/to/project/load-standalone.mjs:26:16
      at ModuleJob.run (node:internal/modules/esm/module_job:377:25)
      at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:691:26)
      at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
    path: undefined,
    locations: [ [Object], [Object] ],
    extensions: [Object: null prototype] {}
  }
]
```

### Use case

I want to define a query where part of the scalar is provided by user input and the other half is fixed (using [trusted documents](https://benjie.dev/graphql/trusted-documents)). The lat/lng use case is just an example. If it's not convincing enough, consider other scalars such as `Range` (`[low, high]`, where the user should only be able modify, say, `low`) or `Vector`. Or even `Raw` / `JSON`.

I looked at the current draft spec (re: [VariablesInAllowedPositionRule](https://spec.graphql.org/draft/#sec-All-Variable-Usages-Are-Allowed)) and it does not seem to forbid variables within custom scalars. I also looked at the proposed clarification https://github.com/graphql/graphql-spec/pull/1156 and what I'm trying to do isn't against the spirit. Note that while  `(loc: [$lat, 2])` does not work, `(loc: {_wrapper: [$lat, 2]})` works. I don't see adding a dummy object wrapper changes the interpretation of the spec here. So I believe it's just a bug / limitation in graphql-js that 

BTW, this also happens in the GraphQL LSP, which I believe share the same validation rules.

(I can work around it by adding an alternative serialization format to my scalar: `{_wrapper: [the actual array here]}` but it is a hack. Especially, according to GraphQL custom scalar implementation guide, I also need to support it as a serialization format too for compliance. I don't want to commit to supporting this wrapper format as part of my API.)

yuchenshi:
Fix #4512.

Also improves test coverage of TypeInfo and the affected validation rule. See linked issue.

I've noticed an interesting existing behavior: When traversing a list literal, `TypeInfo.getInputType()` returns the list *element* type during `enter(ListValue)`. As opposed to, the list type (such as `[String]`), as some may expect. I've add a comment to document this behavior. I'm assuming we don't want to introduce a breaking change here but a maintainer can let me know if we should treat it as a bug.

The two `list literal` tests would have been broken without the fix. Other tests are added purely for extra coverage as I don't want to break things as I fix others.

### DIFF
--- a/src/utilities/TypeInfo.ts
+++ b/src/utilities/TypeInfo.ts
@@ -116,6 +116,8 @@ export class TypeInfo {
     return this._inputTypeStack.at(-1);
   }
 
+  // Note: continues to expose the closest enclosing valid input type if
+  // traversal descends into syntax with no corresponding GraphQL input type.
   getParentInputType(): Maybe<GraphQLInputType> {
     return this._inputTypeStack.at(-2);
   }
--- a/src/utilities/TypeInfo.ts
+++ b/src/utilities/TypeInfo.ts
@@ -254,7 +254,7 @@ export class TypeInfo {
         const listType: unknown = getNullableType(this.getInputType());
         const itemType: unknown = isListType(listType)
           ? listType.ofType
-          : listType;
+          : undefined;
         // List positions never have a default value.
         this._defaultValueStack.push(undefined);
         this._inputTypeStack.push(isInputType(itemType) ? itemType : undefined);
--- a/src/utilities/__tests__/TypeInfo-test.ts
+++ b/src/utilities/__tests__/TypeInfo-test.ts
@@ -451,10 +451,10 @@ describe('visitWithTypeInfo', () => {
       ['enter', 'ObjectField', null, '[String]'],
       ['enter', 'Name', 'stringListField', '[String]'],
       ['leave', 'Name', 'stringListField', '[String]'],
-      ['enter', 'ListValue', null, 'String'],
+      ['enter', 'ListValue', null, 'String' /* the item type, not list type */],
       ['enter', 'StringValue', null, 'String'],
       ['leave', 'StringValue', null, 'String'],
-      ['leave', 'ListValue', null, 'String'],
+      ['leave', 'ListValue', null, 'String' /* the item type, not list type */],
       ['leave', 'ObjectField', null, '[String]'],
       ['leave', 'ObjectValue', null, 'ComplexInput'],
     ]);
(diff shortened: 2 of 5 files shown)


yaacovCR:
It does seem like the `getInputType` for the `itemType` should be undefined rather than the wrapping ListType if it's a custom scalar. So I think this does qualify as a bug.

Fixing this does seem to unlock allowing embedding variables inside list literals in positions expecting custom scalars (although only validation demonstrated here).

We do allow embedding variables inside object literals within a position expecting a custom scalars, although that is currently non-specified, but supported behavior within graphql-js, and has been for a long time.

I am not sure why historically we do not currently support allowing variables within list literals passed to custom scalars, as there would seem to be a reasonable parallel there, but I am a bit wary about introducing further non-specified behavior into the implementation without broader discussion from the WG.

@graphql/tsc any thoughts on this?

yaacovCR:
Ah, I see there is a linked issue basically giving the go ahead from @benjie => i would consider this a bug fix/feature rather than a breaking change, no existing queries will become invalid.

benjie:
That was the go ahead to essentially investigate/suggest a fix, not the go ahead to merge… just for clarity. Will review in more detail in the new year. Thanks for your work on this, both!

benjie:
One concern here is lists have fun behaviours in GraphQL; we do auto coercion of value to list (i.e. you can feed `7` into `[Int]`) and i would want to ensure this doesn’t break under this fix. I think it might, purely because we would process the list first?

benjie:
Either way, the literal behaviour ought to match the value behaviour. If we accept a list value for a custom scalar but not a list literal then that’s inconsistent and we should figure out what the correct behaviour should be. V17 is a good opportunity to introduce that fix.

yaacovCR:
@yuchenshi based on the above comment from @benjie, this would be deferred until v17, which currently corresponds to the `next` branch as the PR base.

Also, a non-emergent reminder re: the CLA.

yuchenshi:
> One concern here is lists have fun behaviours in GraphQL; we do auto coercion of value to list (i.e. you can feed `7` into `[Int]`) and i would want to ensure this doesn’t break under this fix. I think it might, purely because we would process the list first?

Good point, let me see how I can add a test for this.

> Either way, the literal behaviour ought to match the value behaviour. If we accept a list value for a custom scalar but not a list literal then that’s inconsistent and we should figure out what the correct behaviour should be. V17 is a good opportunity to introduce that fix.

Re: what the correct behavior should be, will this be addressed under graphql/graphql-spec#1156 and/or graphql/graphql-spec#1179 ? Does either of these block this PR or the v17 release?

> @yuchenshi based on the above comment from @benjie, this would be deferred until v17, which currently corresponds to the `next` branch as the PR base.

Let me rebase the PR then.

Is there a target date for the v17 release? The linked bug is affecting us in production and even a ballpark estimation helps. I understand this repo doesn't have a fixed release schedule -- I just want to understand if we're quite close to v17 or it's more like months away.

> Also, a non-emergent reminder re: the CLA.

I'm working with my employer on the CLA and I'll try to get it done before v17.

yuchenshi:
> > One concern here is lists have fun behaviours in GraphQL; we do auto coercion of value to list (i.e. you can feed `7` into `[Int]`) and i would want to ensure this doesn’t break under this fix. I think it might, purely because we would process the list first?
> 
> Good point, let me see how I can add a test for this.
> 

Added coverage in `src/execution/__tests__/variables-test.ts`. These tests pass without the TypeInfo change and continue to pass after. I'd need another Vercel approval please.

> > @yuchenshi based on the above comment from @benjie, this would be deferred until v17, which currently corresponds to the `next` branch as the PR base.
> 
> Let me rebase the PR then.

Rebased to `next`. Feel free to take another look or wait until...

> > Also, a non-emergent reminder re: the CLA.
> 
> I'm working with my employer on the CLA and I'll try to get it done before v17.

Unfortunately, I still don't have an update here. I'm not aware of any blockers right now but legal review tend to get slow these days. I'm committed to push this through since we'd very much like to start contributing to upstream.

Is the v17 release coming up soon? Will we miss the chance and have to wait until v18?

yaacovCR:
We'd like to get v17 out soon, but we do not have date yet. I am hopeful to move from alpha=>beta next week.

@benjie I suppose it's the new year already and so this issue is fair game => it seems to me that this is a bone fide bug, probably only lately discovered as custom scalars provided with list values/literals are presumably uncommon. I think it's probably safe to release as a bug fix in v16 or in v17.

Having said that, I do not think it is advisable to implement GeoLocation as a custom scalar rather than an object type, I will comment on the linked issue.

yuchenshi:
@yaacovCR Thanks for taking another look.

I've finally got the CLA done. I think I do need another authorization for Vercel though. And feel free to let me know if you'd like anything else changed on the PR.

yaacovCR:
@yuchenshi looks good to me, made some minor edits

yaacovCR (approved):

yuchenshi:
@yaacovCR Thanks! Would you mind approving the Vercel check on GitHub and let it run?

Also, I don't think I have permissions to merge. Should a maintainer do that?

yaacovCR:
Thanks @yuchenshi !!!!!

PR #4521, opened 2026-01-06
Add experimental support for directives on directive definitions

BoD:
Allow directives on directive definitions, based on [this spec PR](https://github.com/graphql/graphql-spec/pull/1206) which introduces this syntax:

```graphql
directive @onDirective on DIRECTIVE_DEFINITION

directive @foo @onDirective on OBJECT

directive @baz @deprecated(reason: "...") on OBJECT

extend directive @quux @deprecated(reason: "...")
```

### Disclaimer

First time on this codebase and am also not a JS/TS person, so obvious mistakes and/or missing pieces are very likely 😅 Any help and feedback to improve this PR are very welcome 🙏.  The tests seem to pass but additional tests may be needed.


src/language/parser.ts
```diff
@@ -1204,6 +1206,8 @@ export class Parser {
           return this.parseEnumTypeExtension();
         case 'input':
           return this.parseInputObjectTypeExtension();
+        case 'directive':
+          return this.parseDirectiveDefinitionExtension();
```
martinbonnin: Can we have this under a feature flag in `ParseOption`?

src/language/parser.ts
```diff
@@ -1399,6 +1420,7 @@ export class Parser {
     this.expectToken(TokenKind.AT);
     const name = this.parseName();
     const args = this.parseArgumentDefs();
+    const directives = this.parseConstDirectives();
```
martinbonnin: Same here, we need a feature flag I believe

src/language/parser.ts
martinbonnin: @BoD I added the feature flag in github.com/graphql/graphql-js/pull/4521/commits/c86e332d6e753a198ae2344d4d8b2c52bdbf152f. 

I also removed some of the tests (`kitchenSinkSDL` mostly). Until this graduates, I didn't want to pollute the test base with too many `{ experimentalDirectivesOnDirectiveDefinitions: true }`.

BoD: Thanks! 🙏 

> I also removed some of the tests (kitchenSinkSDL mostly). Until this graduates, I didn't want to pollute the test base with too many { experimentalDirectivesOnDirectiveDefinitions: true }.

Works for me 👍

leebyron (approved):

leebyron:
Update includeDeprecated arg per @benjie 's comments, then we'll merge

BoD:
Just made the change for `includeDeprecated: Boolean!` + added one test.

src/language/parser.ts
```diff
@@ -1204,6 +1218,13 @@ export class Parser {
           return this.parseEnumTypeExtension();
         case 'input':
           return this.parseInputObjectTypeExtension();
+        case 'directive':
+          if (
+            this._options.experimentalDirectivesOnDirectiveDefinitions === true
```
jerelmiller: Since `experimentalDirectivesOnDirectiveDefinitions` is typed as a `boolean`, I think its reasonable to check for "truthy" instead of a strict comparison with `true` so this can be simplified to just:


```suggestion
            this._options.experimentalDirectivesOnDirectiveDefinitions
```

That said, I don't know this code base super well and whether strict boolean comparisons are the norm. If so, leaving this is fine 👍

src/language/parser.ts
```diff
@@ -1399,6 +1437,10 @@ export class Parser {
     this.expectToken(TokenKind.AT);
     const name = this.parseName();
     const args = this.parseArgumentDefs();
+    const directives =
+      this._options.experimentalDirectivesOnDirectiveDefinitions === true
```
jerelmiller: Same here, you might be able to check truthy instead of a strict comparison with boolean:


```suggestion
      this._options.experimentalDirectivesOnDirectiveDefinitions
```

src/utilities/__tests__/extendSchema-test.ts
```diff
@@ -1318,5 +1318,31 @@ describe('extendSchema', () => {
...
+        extend directive @isDeprecated @deprecated(reason: "use another directive")
+      `,
+        { experimentalDirectivesOnDirectiveDefinitions: true },
+      );
+      const extendedSchema = extendSchema(schema, extendAST);
+
+      const someDirective = assertDirective(
+        extendedSchema.getDirective('isDeprecated'),
+      );
+      expect(someDirective).to.include({
```
jerelmiller: Small nit, I'd call this `iseprecatedDirective` since that is what is points to, rather than `someDirective`:


```suggestion
      const isDeprecatedDirective = assertDirective(
        extendedSchema.getDirective('isDeprecated'),
      );
      expect(isDeprecatedDirective).to.include({
```

That or just call it `directive` since its in a test and you don't necessarily need to be super specific. Mostly I just don't like the "some" here since its value isn't that ambigous:


```suggestion
      const directive = assertDirective(
        extendedSchema.getDirective('isDeprecated'),
      );
      expect(directive).to.include({
```

src/type/directives.ts
jerelmiller, line 59: Would it make sense to add an `isDeprecated` getter on this class? Not sure if other types do the same, but since we use `isDeprecated` in other places (like introspection), this might be a nice shortcut for those defining directives instances.

```ts
get isDeprecated() {
  return this.deprecationReason == null;  
}
```

Which means you can use this in places like [introspection.ts](https://github.com/graphql/graphql-js/pull/4521/changes#diff-0692ebfea743a0a525e35affba2414bb9ab2f1217c22c7b472acb99d705f8375R86)

```diff
- .filter((directive) => directive.deprecationReason == null),
+ .filter((directive) => directive.isDeprecated),
// ...
- resolve: (directive) => directive.deprecationReason != null,
+ resolve: (directive) => directive.isDeprecated,
```

Would love for someone with more knowledge of the codebase to chime in on whether this is a good idea or against the design principles of this library.

src/utilities/extendSchema.ts
```diff
@@ -138,6 +139,7 @@ export function extendSchemaImpl(
   // Collect the type definitions and extensions found in the document.
   const typeDefs: Array<TypeDefinitionNode> = [];
   const typeExtensionsMap = Object.create(null);
+  const directiveExtensionsMap = Object.create(null);
```
jerelmiller: It would be great to add a specific type for this to avoid the need for type casting in other parts of the changes.


```suggestion
  const directiveExtensionsMap: Record<string, DirectiveExtensionNode[]> = Object.create(null);
```

src/utilities/extendSchema.ts
```diff
@@ -415,6 +432,20 @@ export function extendSchemaImpl(
     return opTypes;
   }
 
+  function extendDirective(directive: GraphQLDirective): GraphQLDirective {
+    const config = directive.toConfig();
+    const extensions = directiveExtensionsMap[config.name] ?? [];
+    const deprecatedReason = extensions
+      .map((ext: DirectiveExtensionNode) => getDeprecationReason(ext))
+      .find((reason: Maybe<string>) => reason != null);
```
jerelmiller: Assuming the change to add an explicit type for `directiveExtensionsMap`, you can remove the type casts here since these can be properly inferred:


```suggestion
    const deprecatedReason = extensions
      .map((ext) => getDeprecationReason(ext))
      .find((reason) => reason != null);
```

src/utilities/extendSchema.ts
```diff
@@ -415,6 +432,20 @@ export function extendSchemaImpl(
     return opTypes;
   }
 
+  function extendDirective(directive: GraphQLDirective): GraphQLDirective {
+    const config = directive.toConfig();
+    const extensions = directiveExtensionsMap[config.name] ?? [];
+    const deprecatedReason = extensions
```
jerelmiller: [nit] If you call this `deprecationReason`:
```suggestion
    const deprecationReason = extensions
```


you can use the object shorthand below:
```diff
return new GraphQLDirective({
  ...config,
+ deprecationReason,
  extensionASTNodes: config.extensionASTNodes.concat(extensions),
});
```

src/utilities/extendSchema.ts
```diff
@@ -187,6 +198,11 @@ export function extendSchemaImpl(
     typeMap[name] = stdTypeMap[name] ?? buildType(typeNode);
   }
 
+  const directiveMap = Object.create(null);
```
jerelmiller:
```suggestion
  const directiveMap: Record<string, GraphQLDirective> = Object.create(null);
```

Always better to type this explicitly where possible, otherwise the type is `any`.

src/utilities/extendSchema.ts
```diff
@@ -199,12 +215,13 @@ export function extendSchemaImpl(
   };
 
   // Then produce and return a Schema config with these types.
+  const directives: Array<GraphQLDirective> = Object.values(directiveMap);
```
jerelmiller: Totally optional, but assuming the type change to `directiveMap`, you can use type inference here to get the same value since `directeMap` would no longer be an `any` type:


```suggestion
  const directives = Object.values(directiveMap);
  // will be `Array<GraphQLDirective>`
```

jerelmiller (commented):
Had a few suggestions, mostly around types, but otherwise looks good from my end!

You probably want to get a proper review from someone who knows the conventions in this codebase better than I do, but wanted to try and help move this along in some way!

src/language/parser.ts
BoD: Agreed!

src/type/directives.ts
BoD: I think that would be a good change, but we'd want to do the same to other types like fields, arguments, enum values etc. for consistency. Although doing that would be a bit out of scope in this PR.

src/utilities/extendSchema.ts
BoD: ESLint wanted to change it to `{ [key: string]: GraphQLDirective }`, I think it's the same thing but correct me if I'm wrong 😅

benjie:
I've not fully reviewed this PR, but the changes since Lee's review all look fine to me - thanks @BoD! :+1:

src/utilities/extendSchema.ts
jerelmiller: Yep its the same thing 🙂

yaacovCR:
@BoD  dropped in a few commits with fixes, possibly spec for some of them has to be correlated => have to check out ci failures a bit later….

yaacovCR (approved):

PR #4541, opened 2026-01-28
convert ExecutionContext interface and functions to Executor

yaacovCR:
This is most of the remaining `Executor.ts` file.


yaacovCR:
No significant performance change.

image

PR #4546, opened 2026-02-01
introduce legacy incremental executor

yaacovCR:
#4319 claimed this was not possible, but I cannot understand why from my previous comment.

Note: the legacy incremental executor is a branching executor, it uses the legacy format that duplicates fields within the response as well as the underlying legacy behavior in which those fields are re-executed, with potentially conflicting field values. In particular, nulls may bubble up in one branch and not another. The transformer approach from #4319 keeps the legacy format without re-executing because it transforms the non-branching incremental executor response.


yaacovCR:
diff is only the second commit as based on #4545 :

https://github.com/graphql/graphql-js/pull/4546/changes/b73204ec88a17a02a484bd11ef9994f74606a9c4

PR #4556, opened 2026-02-12
add assumeValidNames to allow skipping name validation when building a trusted schema

yaacovCR:
i.e. a schema from a parser that would throw with invalid names

see #4362


yaacovCR:
testing on my local PC:

image

about a 2-3% performance improvement

so not a game-changer, but a decent, minimal improvement (note: with a large schema)

this is implemented as backwards compataible, so we need not delay v17 for it to land.

yaacovCR:
(tests are failing -- right now -- only for coverage)

PR #4557, opened 2026-02-15
relax GraphQLFormattedError to allow undefined for optional properties

yaacovCR:
at @IvanGoncharov's suggestions
=> allows GraphQLError class to implement GraphQLFormattedError interface

alternative to parts of #4504

PR #4558, opened 2026-02-15
redefine FormattedExecutionResult as special case of ExecutionResult 

yaacovCR:
where for `FormattedExecutionResult` we "undo" the default to revert back to the broader "GraphQLFormattedError" rather than the specific GraphQLError class, which is the immediate (unformatted/unfiltered/unmasked) error result.

depends on #4557

### DIFF
--- a/src/error/GraphQLError.ts
+++ b/src/error/GraphQLError.ts
@@ -51,7 +51,7 @@ export interface GraphQLErrorOptions {
  * and stack trace, it also includes information about the locations in a
  * GraphQL document and/or execution result that correspond to the Error.
  */
-export class GraphQLError extends Error {
+export class GraphQLError extends Error implements GraphQLFormattedError {
   /**
    * An array of `{ line, column }` locations within the source GraphQL document
    * which correspond to this error.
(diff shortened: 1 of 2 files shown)


yaacovCR:
Motivation: to reduce maintenance burden on our types to prevent divergence.

`ExecutionResult` is now a special case of FormattedExecutionResult essentially, but with this version ExecutionResult is the generic which is odd. If we change this to:

```ts
export interface ExecutionResult<
  TData = ObjMap<unknown>,
> extends FormattedExecutionResult<TData, TExtensions, GraphQLError> {}

export interface FormattedExecutionResult<
  TData = ObjMap<unknown>,
  TError = GraphQLFormattedError,
> {
  errors?: ReadonlyArray<TError>;
  data?: TData | null;
  extensions?: TExtensions;
}
```

then it's maybe a bit better. But the name of `FormattedExecutionResult` is a bit off, it represents the broader type, so it should be something like `GraphQLExecutionResult`, i.e. formatted to GraphQL spec, but the name is actually the old name meant to signify that it's been formatted correctly for the client.

I think I'll retire this for now, and just to eat the extra maintenance burden.

PR #4561, opened 2026-02-16
let GraphQLArgs inherit all parse/validate/execute options

yaacovCR:
motivation:

- reduces maintenance burden
- increases utility/flexibility of the `graphql` pipeline function

potential downside:

- means that we have to police any ambiguity or clashes regarding option property names between `ParseOptions`, `ValidationOptions`, and `ExecutionArgs`.

### DIFF
--- a/src/__tests__/graphql-test.ts
+++ b/src/__tests__/graphql-test.ts
@@ -1,0 +1,164 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { GraphQLError } from '../error/GraphQLError.js';
+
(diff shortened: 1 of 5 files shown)


yaacovCR:
> * we have to police any ambiguity or clashes regarding option property names between `ParseOptions`, `ValidationOptions`, and `ExecutionArgs`.

I think that's ok for now, in a future (major?) change, we need to reorganize all our options and that should help, see comment @  https://github.com/graphql/graphql-js/pull/4364/changes#r2037627873 (you may have to expand resolved/minimized comments to get to see the relevant bit)

PR #4564, opened 2026-02-17
OneOf Inhabitability

jbellenger:
This PR implements the validation described at https://github.com/graphql/graphql-spec/pull/1211

More discussion is at the link above, but the core issue is around uninhabited OneOf types -- the property that a type can be defined in a way that a value cannot be generated for it.

The simplest example of an uninhabited type is this recursive OneOf:
```
input A @oneOf { a:A }
```

While this type definition is currently valid by the rules of OneOf types and circular input types, it violates the spirit of the [Circular References](https://spec.graphql.org/draft/#sec-Input-Objects.Circular-References) because it does not allow construction of a finite value.

This PR updates the existing cycle validation to consider OneOf types. I've tried to follow the existing conventions, though I'm also new here and might have missed something. Any feedback on style or substance is welcome.

### DIFF
--- a/src/type/validate.ts
+++ b/src/type/validate.ts
@@ -786,71 +786,185 @@ function validateOneOfInputObjectField(
 }
 
 // Implements the spec's InputObjectHasUnbreakableCycle algorithm.
-// Tracks already checked types to maintain O(N) and to ensure that types
-// are not redundantly checked.
+// Tracks already checked types to ensure that types are not redundantly checked.
 function createInputObjectUnbreakableCycleCheck(): (
   inputObj: GraphQLInputObjectType,
 ) => boolean {
-  const knownNoCycle = new Set<GraphQLInputObjectType>();
-  const visited = new Set<GraphQLInputObjectType>();
+  const cycleMemo = new Map<GraphQLInputObjectType, boolean>();
 
   return inputObjectHasUnbreakableCycle;
 
   function inputObjectHasUnbreakableCycle(
     inputObj: GraphQLInputObjectType,
   ): boolean {
-    if (knownNoCycle.has(inputObj)) {
-      return false;
-    }
-    if (visited.has(inputObj)) {
-      return true;
+    const cachedResult = cycleMemo.get(inputObj);
+    if (cachedResult !== undefined) {
+      return cachedResult;
     }
 
-    visited.add(inputObj);
+    // Unknown reachable types mapped to types that depend on them.
+    const candidates = new AccumulatorMap<
+      GraphQLInputObjectType,
+      GraphQLInputObjectType
+    >();
+    for (const candidate of collectReachableInputObjects(inputObj)) {
+      candidates.set(candidate, []);
+    }
 
-    let result: boolean;
+    // Tracks unresolved fields for normal Input Objects.
+    const normalTypeStates = new Map<
+      GraphQLInputObjectType,
+      { hasKnownCycleField: boolean; unresolvedFieldCount: number }
+    >();
 
-    if (inputObj.isOneOf) {
-      // OneOf Input Objects have an unbreakable cycle if every field has one.
-      result = true;
-      for (const field of Object.values(inputObj.getFields())) {
-        if (!inputFieldTypeHasUnbreakableCycle(field.type)) {
-          result = false;
-          break;
+    // Types proven not to have cycles and ready to remove.
+    const typesToRemove: Array<GraphQLInputObjectType> = [];
+
+    for (const candidate of candidates.keys()) {
+      const fields = Object.values(candidate.getFields());
+
+      if (candidate.isOneOf) {
+        // OneOf Input Objects have an unbreakable cycle if every field leads to an unbreakable cycle.
+        if (fields.length === 0) {
+          typesToRemove.push(candidate);
+          continue;
+        }
+
+        for (const field of fields) {
+          const target = getUnbreakableCycleTarget(candidate, field.type);
+
+          if (target == null) {
+            typesToRemove.push(candidate);
+            break;
+          }
+
+          const targetResult = cycleMemo.get(target);
+          if (targetResult === false) {
+            typesToRemove.push(candidate);
+            break;
+          }
+
+          if (targetResult === undefined) {
+            candidates.add(target, candidate);
+          }
+        }
+      } else {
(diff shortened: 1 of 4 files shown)


benjie (commented):
Great tests! (It's 10:30pm so I've not tried to review the code... 😅)

src/type/validate.ts
```diff
@@ -403,14 +406,25 @@ function validateTypes(context: SchemaValidationContext): void {
...
     }
   }
+
+  // Report errors for Input Object types that have unbreakable cycles.
+  for (const type of typesWithUnbreakableCycles) {
+    const cyclePath = traceUnbreakableCycle(type, typesWithUnbreakableCycles);
+    const pathStr = cyclePath.map((p) => p.fieldStr).join(', ');
+    context.reportError(
+      `Input Object ${type} references itself via the required fields: ${pathStr}.`,
+      cyclePath.map((p) => p.astNode),
```
martinbonnin: Out of curiosity, why move the generation of the error message to a second traversal vs building it while traversing the type before?

martinbonnin: Okay. Guessing this is problematic for stuff like this:

```graphql
input A @oneOf {
  b: B
  c: C
}

input B {
  a: A!
}

input C {
  a: A!
}
```

Current errors are:

```
Input Object A references itself via the required fields: A.b, B.a.
Input Object B references itself via the required fields: B.a, A.b.
Input Object C references itself via the required fields: C.a, A.b, B.a.
```

Which is a bit misleading as those fields aren't really required. Not 100% sure how to improve the diagnostics though.

jbellenger: This is a good callout, in particular the validation error for `C` is not very helpful here.

My gut feeling is that this kind of validation is a property of a _graph_ of types, rather than individual types, and that this will probably be easier to understand if reported as a single error that describes the participating types and at least one illegal path.

Something like (just riffing):
"Input Objects A, B, and C cannot be provided finite values because they form an unbreakable cycle through fields: A.b, A.c, B.a, and C.a."

wdyt?

martinbonnin: Agreed, feels like something like [SCC](https://en.wikipedia.org/wiki/Strongly_connected_component).

This is a diagnostics issue and this PR already provides a fix. We could merge this one as is and make a second one down the road.

jbellenger: I took a closer look at the existing behavior. Contrary to what we discussed in graphql-wg, todays behavior is not exactly 1 error per type in a cycle -- it's closer to 1 error per distinct cycle.

The most representative test case for this behavior is [this one](https://github.com/graphql/graphql-js/blob/3b1c8774d311aba160599ccfa158328f311af11b/src/type/__tests__/validation-test.ts#L933-L963). I've updated the validator in this PR to follow this same pattern. 

There is also a case of overlapping cycles, which today get reported as separate errors ([test](https://github.com/graphql/graphql-js/blob/3b1c8774d311aba160599ccfa158328f311af11b/src/type/__tests__/validation-test.ts#L965-L1009)). 

For example, the schema you posted above is a good illustration of 2 distinct cycles that overlap through type A:
```graphql
input A @oneOf { b:B, c:C }
input B { a:A! }
input C { a:A! }
``` 
While the validator is already halfway to SCC, going all the way would, I think, probably result in reporting errors for this example as a single error (rather than 2), since all 3 types are strongly connected.

So in the latest revision, this graph produces these 2 errors, which match the existing behavior:
```
Input Object A cannot be provided a finite value because it references itself through fields: A.b, B.a.
Input Object A cannot be provided a finite value because it references itself through fields: A.c, C.a.
```

martinbonnin: That works for me 👍 . I was mainly curious but this shouldn't prevent this to be merged as it fixes a real validation issue (and error reporting looks good!).

yaacovCR:
@jbellenger 

I finally think I understand everything here and this is awesome.

I pushed some changes/reorg added another memoization layer.

yaacovCR:
@jbellenger 

to add some detail

image

- meaning adding another layer of memoization in b9acf03f8 grabbed back some of the perf drop
- using that same map when reporting in 9ed3ad6c5 / `HEAD` doesn't add much perf wise, although it perhaps cleans the code a bit (?)

it's all somewhat slower than baseline `17.x.x` but of course it's doing more (and correct) validation.

yaacovCR:
ok, we now have parity:

image

we avoid a full reachability calculation, but just iterate through all input objects once saving metadata, and then backwards propagate

yaacovCR:
@jbellenger

Do we want (A) an error per input object cycles, or do we want (B) one error per input object listing all the cycles it has? Let's assume that both options filter so that we never report the same cycle more than once, but see the change in the last commit, we do (B), which would apply to non-OneOf cycles as well? In v16 (and prior to that commit ) we did (A), and have for a while with regard to non-OneOf cycles, but I think whatever we do should be consistent for both.

I feel like @benjie weighed in on this at the graphql-wg discussion about it?

jbellenger:
@yaacovCR Thanks for the polish you've put on this!

Re reporting, I think the wg discussion was here: https://youtu.be/0JP4x-Uf2qw?si=qhrpHaOaKFxYKsoO&t=997

My takeaway was that we'd want one error per _cause_, which I think means option B. This would also have the nice property of providing complete diagnostic information to an agent.

yaacovCR:
Shoot, I edited my comment. I think A and B were technically the same but I think it's right now.

yaacovCR:
Hmmm I guess the preference I read was toward A ie keep doing what we were doing, we should get a different error per cycle as the cycle was the cause not the input object per se…. I’m going to roll back that last commit…..

yaacovCR:
Basically, we can table that debate for this PR as it also changes how errors have been reported with regard to circular Input Objects for quite some time. If we want to change how multiple cycles starting at the same Input Object works, we should do so in lockstep across non-OneOf and OneOf cycles, and do so in a separate PR. I am pretty much agnostic, leaning toward grouping together; at the bottom line, we have all of the information in both variations both with regards to the messages and erroring location, the question is only the grouping.

jbellenger:
@yaacovCR 
> If we want to change how multiple cycles starting at the same Input Object works, we should do so in lockstep across non-OneOf and OneOf cycles, and do so in a separate PR.

Big plus one.

yaacovCR (approved):

PR #4582, opened 2026-02-20
Add mergeAST utility for merging multiple AST documents

### LINKED ISSUE #1428: merge ast trees

veeceey:
## Summary

- Adds a new `mergeAST` utility function that merges multiple `DocumentNode`s into a single document by combining their selection sets
- Fields with the same response name (alias or field name) and identical arguments are recursively merged; fields with different arguments are kept separate
- Operations are matched by name and type (query/mutation/subscription); fragment definitions are deduplicated by name
- Variable definitions and directives are also merged with deduplication
- Includes 20 test cases covering: simple merging, deduplication, nested selections, arguments, aliases, named operations, variables, fragments, inline fragments, and the motivating use case from the issue

Closes #1428

### Motivation

As discussed in #1428, there are many use cases where dynamically merging GraphQL AST trees is needed, particularly:
- Resolvers that need to ensure additional fields are present in requests forwarded to backend services
- Libraries that compose queries from multiple sources
- React Server Components combining multiple GraphQL queries into a single request

The implementation follows the suggestion from @benjie to migrate/adapt the approach from GraphiQL's `merge-ast` utility, while providing a simpler, focused API that fits naturally alongside the existing `concatAST` utility.

### API

```typescript
import { mergeAST } from 'graphql';

const merged = mergeAST([documentA, documentB]);
```

## Test plan

- [x] 20 unit tests covering all merging scenarios
- [x] Existing `concatAST` tests still pass
- [ ] Verify CI passes

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -452,6 +452,8 @@ export {
   coerceInputValue,
   // Concatenates multiple AST together.
   concatAST,
+  // Merges multiple AST documents, combining selection sets and deduplicating fields.
+  mergeAST,
   // Separates an AST into an AST per Operation.
   separateOperations,
   // Strips characters that are not significant to the validity or execution of a GraphQL document.
--- a/src/index.ts
+++ b/src/index.ts
@@ -505,4 +505,5 @@ export type {
   TypedQueryDocumentNode,
   // Schema Coordinates
   ResolvedSchemaElement,
+  MergeASTOptions,
 } from './utilities/index';
--- a/src/utilities/__tests__/mergeAST-test.ts
+++ b/src/utilities/__tests__/mergeAST-test.ts
@@ -320,3 +320,96 @@ describe('mergeAST', () => {
     `);
   });
 });
+
+describe('mergeAST depth guard', () => {
+  it('allows merging within the default depth limit', () => {
+    const docA = parse('{ a { b { c { d { e } } } } }');
+    const docB = parse('{ a { b { c { d { f } } } } }');
+
+    expect(print(mergeAST([docA, docB]))).to.equal(dedent`
+      {
+        a {
+          b {
+            c {
+              d {
+                e
+                f
+              }
+            }
+          }
+        }
+      }
+    `);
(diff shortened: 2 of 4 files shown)


yaacovCR:
There is some feedback by @benjie on this idea from PR #4359

veeceey:
Thanks for the pointer @yaacovCR. I went through benjie's feedback on #4359 -- the DOS concern with mergeAST on unvalidated input is a fair point. I'll look into adding depth/complexity guards so it's safer to use on arbitrary documents. If the preference is still to keep this as a separate package first, I'm open to that too. Let me know how you'd like to proceed.

veeceey:
Went through benjie's feedback on #4359 about the DOS concern -- totally valid point. I've added a depth guard to the recursive merge path:

- New `maxDepth` option on `mergeAST` (defaults to 20)
- Depth is tracked through `mergeSelectionSets` -> `mergeFieldNodes` / `mergeInlineFragments` recursion
- Throws a clear error if the limit is hit, with a message suggesting the option if the depth is intentional
- Added tests covering the depth limit for both field merging and inline fragment merging

Default of 20 should be more than enough for any real-world query while still catching adversarial nesting. The option is there if someone has a legitimate deep schema and wants to raise it.

Also exported the `MergeASTOptions` type from the public API so consumers can use it.

veeceey:
hey, just checking in - anything else needed on this one?

yaacovCR:
> hey, just checking in - anything else needed on this one?

Haven't as of yet had a chance to review, but the good news is that this is a backwards compatible change and so we can get into both v16 and v17 whenever bandwidth opens up for review.

PR #4583, opened 2026-02-20
validation: align MaxIntrospectionDepthRule with v16 backport

yaacovCR:
See prior PRs:
- #4118
- #4119


src/validation/rules/MaxIntrospectionDepthRule.ts
```diff
@@ -50,7 +50,6 @@ export function MaxIntrospectionDepthRule(
     if (
       node.kind === Kind.FIELD &&
       // check all introspection lists
-      // TODO: instead of relying on field names, check whether the type is a list
```
benjie, line 53: Should restore this TODO (or implement it), otherwise newly added list fields in introspection will bypass the list depth limits.

PR #4634, opened 2026-03-24
Use Object.create(null) over {} to avoid prototype issues

benjie:
Forward-port of #4631. Description from #4631 on v16

> Object.create(null) is generally safer since it is not vulnerable to prototype pollution in user code. To avoid breaking changes I've returned { ...obj } thus ensuring that the returned object still has the default Object prototype. 

But this PR skips that latter portion, return `obj` rather than `{ ...obj }`, a BREAKING CHANGE for v17 that improves performance, compare `obj-create-null-17` vs `restore-prototype`

image

### DIFF
--- a/src/execution/values.ts
+++ b/src/execution/values.ts
@@ -206,7 +206,7 @@ export function getArgumentValues(
   fragmentVariableValues?: Maybe<FragmentVariableValues>,
   hideSuggestions?: Maybe<boolean>,
 ): { [argument: string]: unknown } {
-  const coercedValues: { [argument: string]: unknown } = {};
+  const coercedValues: { [argument: string]: unknown } = Object.create(null);
 
   const argumentNodes = node.arguments ?? [];
   const argNodeMap = new Map(argumentNodes.map((arg) => [arg.name.value, arg]));
@@ -224,7 +224,7 @@ export function getArgumentValues(
       hideSuggestions,
     );
   }
-  return coercedValues;
+  return { ...coercedValues };
 }
 
 // eslint-disable-next-line max-params
--- a/src/execution/values.ts
+++ b/src/execution/values.ts
@@ -205,8 +205,8 @@ export function getArgumentValues(
   variableValues?: Maybe<VariableValues>,
   fragmentVariableValues?: Maybe<FragmentVariableValues>,
   hideSuggestions?: Maybe<boolean>,
-): { [argument: string]: unknown } {
-  const coercedValues: { [argument: string]: unknown } = Object.create(null);
+): ObjMap<unknown> {
+  const coercedValues: ObjMap<unknown> = Object.create(null);
 
   const argumentNodes = node.arguments ?? [];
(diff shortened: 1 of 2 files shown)


src/execution/values.ts
```diff
@@ -224,7 +224,7 @@ export function getArgumentValues(
       hideSuggestions,
     );
   }
-  return coercedValues;
+  return { ...coercedValues };
```
yaacovCR: is this strictly necessary?

src/utilities/coerceInputValue.ts
```diff
@@ -109,7 +109,7 @@ export function coerceInputValue(
       }
     }
 
-    return coercedValue;
+    return { ...coercedValue };
```
yaacovCR: same here?

src/execution/values.ts
benjie: Without it, it's a breaking change (assuming this is reachable from user code) since you'd be returning a null prototype object rather than a prototypeful object. Users who did `` console.log(`Blah blah ${obj} blah blah`) `` would have their code change from logging `Blah blah [object Object] blah blah` to **throwing an error**.

```
> const a = {}
undefined
> const b = Object.create(null)
undefined
> String(a)
'[object Object]'
> String(b)
Uncaught TypeError: Cannot convert object to primitive value
    at String (<anonymous>)
```

The reason I didn't merge this myself is because _ideally_ we'd just switch to null prototype objects and get rid of the problem entirely. But I didn't want to introduce a breaking change, hence including this careful back-compat.

Personally I vote to remove this.

yaacovCR: Did a little bit more research on this as it was raised directly in #4653, with some relevant history found to include:

- https://github.com/graphql/graphql-js/pull/1056
- https://github.com/graphql/express-graphql/issues/177

Note that https://github.com/graphql/graphql-js/pull/1056 does not fully address the problem raised by https://github.com/graphql/express-graphql/issues/177. In particular, while @leebyron gave a prototype to the args map itself, an input object arg still itself has no prototype -- just as ExecutionResult.data and its descendants have no prototype.

Basically, we are thinking of reverting part of #1056 => this would definitely be a breaking change as you say. I plan on benchmarking it when I have a moment to better understand the tradeoffs.

benjie: Yes, it’s much less relevant than it once was

yaacovCR: Results of benchmarking:

image

As it appears there is a measurable regression when constructing without a prototype, it seems prudent for v17 to at least not re-add the prototype post-construction, as this brings a bigger hit.

yaacovCR:
```suggestion
  return coercedValues;
```

src/utilities/coerceInputValue.ts
yaacovCR:
```suggestion
    return coercedValue;
```

yaacovCR (dismissed):

PR #4652, opened 2026-04-02
Fix valueFromAST variable own-property checks

abishekgiri:
## Summary
Fixes an issue in valueFromAST where own-property checks were not properly enforced, which could lead to incorrect handling of inherited properties.

## Changes
- Updated own-property checks to ensure only direct properties are considered
- Prevent unintended prototype chain access

## Testing
- Verified behavior manually with edge cases
- Existing tests pass locally

### DIFF
--- a/src/utilities/valueFromAST.ts
+++ b/src/utilities/valueFromAST.ts
@@ -48,7 +48,7 @@ export function valueFromAST(
 
   if (valueNode.kind === Kind.VARIABLE) {
     const variableName = valueNode.name.value;
-    if (variables == null || !hasOwnProperty(variables, variableName)) {
+    if (variables == null || variables[variableName] === undefined || !hasOwnProperty(variables, variableName)) {
       // No valid return value.
       return;
     }
--- a/src/utilities/valueFromAST.ts
+++ b/src/utilities/valueFromAST.ts
@@ -48,7 +48,11 @@ export function valueFromAST(
 
   if (valueNode.kind === Kind.VARIABLE) {
     const variableName = valueNode.name.value;
-    if (variables == null || variables[variableName] === undefined || !hasOwnProperty(variables, variableName)) {
+    if (
+      variables == null ||
+      variables[variableName] === undefined ||
+      !hasOwnProperty(variables, variableName)
+    ) {
       // No valid return value.
       return;
     }
@@ -168,7 +172,9 @@ function isMissingVariable(
 ): boolean {
   return (
     valueNode.kind === Kind.VARIABLE &&
-    (variables == null || !hasOwnProperty(variables, valueNode.name.value))
+    (variables == null ||
+      variables[valueNode.name.value] === undefined ||
+      !hasOwnProperty(variables, valueNode.name.value))
   );
 }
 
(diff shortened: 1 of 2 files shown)


abishekgiri:
Status update

- Fixed `valueFromAST()` to use own-property checks when deciding whether a variable was provided.
- Added regression coverage for omitted prototype-colliding names like `toString` across scalar, list, and input-object cases.
- Results: targeted `mocha`, `eslint`, and `prettier` checks passed for this change.
- `tsc` in this environment still reports existing external errors from `../node_modules/@types/react-dom`, which are outside this PR.

src/utilities/valueFromAST.ts
```diff
@@ -168,6 +168,10 @@ function isMissingVariable(
 ): boolean {
   return (
     valueNode.kind === Kind.VARIABLE &&
-    (variables == null || variables[valueNode.name.value] === undefined)
+    (variables == null || !hasOwnProperty(variables, valueNode.name.value))
```
yaacovCR:
```suggestion
    (variables == null || !Object.hasOwn(variables, valueNode.name.value) || variables[valueNode.name.value] === undefined)
```

yaacovCR:
```suggestion
    (variables == null || variables[valueNode.name.value] === undefined || !hasOwnProperty(variables, valueNode.name.value))
```

yaacovCR:
@abishekgiri this looks to me like a legitimate bug fix, although seemingly for a longstanding bug. Thanks for this contribution! Made a small suggestion above, and there is a lint/prettier issue.

src/utilities/valueFromAST.ts
abishekgiri: Updated this to preserve the previous undefinedbehavior as well. I changedisMissingVariable()to treat explicitlyundefinedvariables as missing, and added regression coverage for both scalar and list cases so it stays aligned with the mainvalueFromAST() check.

abishekgiri:
@yaacovCR Pushed a follow-up update addressing the review feedback. This keeps the own-property fix for prototype-colliding names, restores the previous behavior for variables explicitly set to undefined, and adds regression tests for those cases. Targeted mocha, eslint, and prettier checks are passing on the updated branch.

yaacovCR:
@abishekgiri

This seems right to me. `valueFromAST` is deprecated in v17, but it has not been removed, and this seems like an opportunity for an additional PR to forward-port. Let me know if you have time, otherwise I will try to take care of it. In v17 we can use `Object.hasOwn`, but otherwise it should be pretty similar.

abishekgiri:
@yaacovCR Thanks, that sounds good to me. I’m glad the 16.x.xfix looks right as-is. I also checked the current17.x.xbranch and the samevalueFromAST own-property issue still appears to be present there, so a small forward-port PR makes sense. I can put that together if helpful; otherwise I’m also happy for you to handle that follow-up.

yaacovCR:
Check was apparently dropped back in #553 !

yaacovCR:
forward-ported in #4671 => thanks @abishekgiri !!!

PR #4653, opened 2026-04-02
fix: avoid prototype-colliding names in execution values

abishekgiri:
## Summary
Fixes an issue where prototype-colliding property names (e.g., `__proto__`, `constructor`) could interfere with execution value handling.

## Changes
- Ensured own-property checks are used when accessing execution values
- Prevent potential prototype chain collisions

## Motivation
JavaScript objects can inherit properties from the prototype chain. Without proper checks, this can lead to unexpected behavior when handling execution values.

## Testing
- Verified manually with edge-case inputs involving prototype properties
- Existing tests pass locally

### DIFF
--- a/src/execution/values.ts
+++ b/src/execution/values.ts
@@ -58,9 +58,8 @@ type VariableValuesOrErrors =
  * provided variable definitions and arbitrary input. If the input cannot be
  * parsed to match the variable definitions, a GraphQLError will be thrown.
  *
- * Note: The returned value is a plain Object with a prototype, since it is
- * exposed to user code. Care should be taken to not pull values from the
- * Object prototype.
+ * Note: The `coerced` and `sources` properties of VariableValues use null
+ * prototype to avoid collisions with JavaScript's own property names.
  */
 export function getVariableValues(
   schema: GraphQLSchema,
@@ -195,9 +194,8 @@ export function getFragmentVariableValues(
  * Prepares an object map of argument values given a list of argument
  * definitions and list of argument AST nodes.
  *
- * Note: The returned value is a plain Object with a prototype, since it is
- * exposed to user code. Care should be taken to not pull values from the
- * Object prototype.
+ * Note: The returned value uses a null prototype to avoid collisions with
+ * JavaScript's own property names.
  */
 export function getArgumentValues(
   def: GraphQLField<unknown, unknown> | GraphQLDirective,
@@ -316,9 +314,8 @@ function coerceArgument(
  *
  * If the directive does not exist on the node, returns undefined.
  *
- * Note: The returned value is a plain Object with a prototype, since it is
- * exposed to user code. Care should be taken to not pull values from the
- * Object prototype.
+ * Note: The returned value uses a null prototype to avoid collisions with
+ * JavaScript's own property names.
  */
 export function getDirectiveValues(
   directiveDef: GraphQLDirective,
@@ -326,7 +323,7 @@ export function getDirectiveValues(
   variableValues?: Maybe<VariableValues>,
   fragmentVariableValues?: Maybe<FragmentVariableValues>,
   hideSuggestions?: Maybe<boolean>,
-): undefined | { [argument: string]: unknown } {
+): undefined | ObjMap<unknown> {
   const directiveNode = node.directives?.find(
     (directive) => directive.name.value === directiveDef.name,
   );
--- a/src/execution/__tests__/variables-test.ts
+++ b/src/execution/__tests__/variables-test.ts
@@ -164,6 +164,15 @@ const TestType = new GraphQLObjectType({
       type: TestNestedInputObject,
     }),
     fieldWithJSONScalarInput: fieldWithInputArg({ type: TestJSONScalar }),
+    fieldWithPrototypeNamedArgument: {
+      type: GraphQLString,
+      args: {
+        toString: { type: GraphQLString },
+      },
+      resolve(_, args) {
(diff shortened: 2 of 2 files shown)


abishekgiri:
Status update

- Fixed `execution/values.ts` so `getArgumentValues()` and `getVariableValues()` keep their null-prototype maps instead of reintroducing `Object.prototype` via object spread.
- Added regression coverage for omitted prototype-colliding names like `toString` in both resolver args and coerced variable values.
- Results: targeted `mocha`, `eslint`, and `prettier` checks passed for this change.
- `tsc` in this environment still reports existing external errors from `../node_modules/@types/react-dom`, which are outside this PR.

yaacovCR:
This would be a breaking change, and would have to land in v17, but we are considering doing so in:
 -  #4634  
 
See: https://github.com/graphql/graphql-js/pull/4634#discussion_r2989820453

For background, see:

- https://github.com/graphql/graphql-js/pull/1056
- https://github.com/graphql/express-graphql/issues/177

Note that #1056 does not fully address the problem raised by https://github.com/graphql/express-graphql/issues/177. In particular, while @leebyron gave a prototype to the args map itself, an input object arg still itself has no prototype -- just as ExecutionResult.data and its descendants have no prototype.

yaacovCR:
Actually @abishekgiri -- on my box the new tests pass even without the suggested code changes. Is there an actual bug in this case

(we were considering doing so in the linked PR above just because the re-addition of the prototype didn't seem strictly necessary)

abishekgiri:
@yaacovCR I dug through this end to end and verified that there is an actual bug here on plain \16.x.x`. I checked the new `variables-test.ts` cases against an unmodified `16.x.x` checkout, and two of them fail without the `execution/values.ts` change: omitted resolver arg `toString` resolves to the inherited function instead of behaving as missing, and `getVariableValues(...).coerced.toString` also resolves to the inherited function instead of `undefined`.

So I do think the bug is real. The harder part is branch compatibility: the current fix changes `getArgumentValues()` / `getVariableValues()` from returning plain objects with a prototype to null-prototype maps, and that seems to be the long-standing/documented behavior on `16.x.x`. So my read is: real bug, but the current fix is probably `17.x.x` material rather than a safe `16.x.x` backport.`

yaacovCR:
Ah, my box duplicates it now as well. I wonder if there would be a 16.x.x compatible fix possible, but agree that the direction of this fix would have to wait until 17.x.x.

yaacovCR:
> Ah, my box duplicates it now as well. I wonder if there would be a 16.x.x compatible fix possible, but agree that the direction of this fix would have to wait until 17.x.x.

Hmmm. actually, yet again looked closer, what the tests are testing is just the case @leebyron warns about in the changed comment, i.e. these objects have a prototype, and therefore one must be careful to not draw items from the prototype through an `Object.hasOwn` check or older equivalent. So there can't be a fix per se in 16.x.x -- although we can consider removing the prototype in 17.x.x.

abishekgiri:
@yaacovCR Makes sense to me. Looking closer, this does seem to match the long-standing 16.x.xcontract that these returned objects have a prototype and consumers need an own-property check when reading from them. In that case, I agree there is not really a16.x.xfix to land here without changing behavior. This seems better tracked asv17 work alongside #4634.

src/execution/values.ts
```diff
@@ -138,16 +137,15 @@ function coerceVariableValues(
     );
   }
 
-  return { ...coercedValues };
+  return coercedValues;
```
afurm: The `getArgumentValues` change (removing `{...coercedValues}`) is parallel to `getVariableValues`. Both previously created a plain-object copy via spread, which masked prototype collisions. Now that `coercedValues` is a local variable with no persistent alias, returning it directly is safe. Was the spread in both functions added defensively at the same time, or was there a specific reason it was needed historically that the null-prototype approach now renders unnecessary?

abishekgiri: Tracked this back through history: the spread is there for a specific reason, not just as a defensive copy. Commit 36b3407c (#1056, "Return normal objects for args and vars") explicitly switched these APIs back to plain objects because exposing prototype-less args/vars was burdensome for user code, especially Flow-typed resolvers. The more recent 13f130d2 (#4631) changed the local accumulators to `Object.create(null)` internally, but intentionally kept the final spread to preserve that public `16.x.x` contract. So I agree this null-prototype return change is not a safe `16.x.x` change; it seems like `v17` / #4634 territory rather than something to land here.

abishekgiri:
Following up here as well: after tracing the history in #1056 and #4631, I agree this does not look like a safe `16.x.x` fix. The spread was kept intentionally so `getArgumentValues()` / `getVariableValues()` continue returning plain objects to user code, even after the internal accumulators moved to `Object.create(null)`. That means the current change would alter the long-standing public contract on `16.x.x`, so this seems better handled as `v17` work alongside #4634 rather than something to merge from this PR.

yaacovCR:
@abishekgiri I rebased this on 17.x.x and changed the PR base (after merging #4634). This PR now contains your tests and comment changes (as well as a small type change that could have also been in #4634 along with the similar type changes there).

Let me know if this looks good and will merge to 17.x.x.

abishekgiri:
@yaacovCR This looks good to me on 17.x.x. Rebasing it after #4634 makes the direction much clearer, and the current scope also looks right: the regression tests, the updated comments, and the small type change all seem consistent with the null-prototype behavior there. I don’t see anything else I’d want to change on this PR from my side, so I’d be happy with merging it into 17.x.x.

PR #4666, opened 2026-04-14
Add optional maxDepth and maxAliases execution options

### LINKED ISSUE #4662: [Security] No built-in query depth/complexity limit + alias bombing DoS
## Summary
Two related DoS vectors in the execution engine:

### 1. No query depth or complexity limit
Recursive types + deep queries cause exponential resolver amplification. No built-in depth limit exists at execution time (only MaxIntrospectionDepthRule at validation).

### 2. Alias bombing
Thousands of aliases for the same field in a single flat query bypass all depth-based defenses, multiplying resolver cost linearly.

## Suggested Fix
Add optional `maxDepth` and `maxAliases` parameters to the execution options, similar to `MaxIntrospectionDepthRule`.

eddieran:
## Summary

Fixes #4662 — the execution engine has no built-in depth or complexity limits, allowing deep recursive queries and alias-bombing to cause denial-of-service via resolver amplification.

This PR adds two **opt-in** execution options:

- **`maxDepth`** — limits the field nesting depth during execution. When a field at depth > `maxDepth` is encountered, a `GraphQLError` is raised and the parent field resolves to `null` (standard nullable error handling). List indices do not count toward depth.

- **`maxAliases`** — limits the number of response keys (including aliases) per selection set. This prevents alias-bombing attacks where thousands of aliases for the same field bypass depth-based defenses. When exceeded, a `GraphQLError` is raised before the selection set executes.

Both options are `undefined` by default — no limits are applied and existing behavior is fully preserved. They are passed via `ExecutionArgs.options` alongside the existing `maxCoercionErrors`:

```ts
const result = await execute({
  schema,
  document,
  options: {
    maxDepth: 10,
    maxAliases: 50,
  },
});
```

### Implementation details

- Depth checking happens in `executeField` by walking the `Path` linked list (only counting string keys, not list indices)
- Alias counting happens in `executeOperation` (root fields) and `completeObjectValue` (sub-selections) after field collection
- Errors follow standard GraphQL error propagation — nullable parent fields are nulled, non-null fields propagate upward
- 12 new tests covering: within-limit queries, exceeded limits, no-op when unset, list index handling, nested alias limits, combined usage

## Test plan

- [x] All 12 new tests pass
- [x] Full test suite (1972 tests) passes with no regressions
- [x] TypeScript compilation clean
- [x] ESLint clean

### DIFF
--- a/src/execution/execute.ts
+++ b/src/execution/execute.ts
@@ -115,6 +115,8 @@ export interface ExecutionContext {
   typeResolver: GraphQLTypeResolver<any, any>;
   subscribeFieldResolver: GraphQLFieldResolver<any, any>;
   collectedErrors: CollectedErrors;
+  maxDepth: number | undefined;
+  maxAliases: number | undefined;
 }
 
 /**
@@ -195,6 +197,21 @@ export interface ExecutionArgs {
   options?: {
     /** Set the maximum number of errors allowed for coercing (defaults to 50). */
     maxCoercionErrors?: number;
+    /**
+     * Set the maximum allowed depth for field resolution.
+     * Depth is counted as the number of nested field selections from the root.
+     * When exceeded, a GraphQLError is thrown for the offending field.
+     * No limit is applied when undefined (the default).
+     */
+    maxDepth?: number;
+    /**
+     * Set the maximum number of aliases allowed in any single selection set.
+     * This helps prevent alias-bombing denial-of-service attacks where many
+     * aliases for the same field bypass depth-based protections.
+     * When exceeded, a GraphQLError is thrown before executing the selection set.
+     * No limit is applied when undefined (the default).
+     */
+    maxAliases?: number;
   };
 }
 
@@ -393,6 +410,8 @@ export function buildExecutionContext(
     typeResolver: typeResolver ?? defaultTypeResolver,
     subscribeFieldResolver: subscribeFieldResolver ?? defaultFieldResolver,
     collectedErrors: new CollectedErrors(),
+    maxDepth: options?.maxDepth,
+    maxAliases: options?.maxAliases,
   };
 }
 
@@ -419,6 +438,9 @@ function executeOperation(
     rootType,
     operation.selectionSet,
   );
+
+  checkAliasCount(exeContext, rootFields);
+
   const path = undefined;
 
   switch (operation.operation) {
@@ -531,6 +553,48 @@ function executeFields(
 }
 
 /**
+ * Checks whether the number of response keys (including aliases) in a
+ * selection set exceeds the configured limit. Throws a GraphQLError when
+ * the limit is exceeded.
+ */
+function checkAliasCount(
+  exeContext: ExecutionContext,
+  fields: Map<string, ReadonlyArray<FieldNode>>,
+): void {
+  if (exeContext.maxAliases !== undefined) {
+    const aliasCount = fields.size;
+    if (aliasCount > exeContext.maxAliases) {
+      // Collect nodes for the error location from the first field node of
+      // each response key beyond the limit.
+      const nodes: Array<FieldNode> = [];
+      for (const [, fieldNodes] of fields) {
+        nodes.push(fieldNodes[0]);
+      }
+      throw new GraphQLError(
+        `Aliases limit of ${exeContext.maxAliases} exceeded, found ${aliasCount} aliases.`,
+        { nodes },
+      );
+    }
+  }
+}
+
+/**
+ * Counts the field depth represented by a Path. Only string keys are counted
+ * (numeric keys represent list indices and should not increase the depth).
+ */
+function pathDepth(path: Path | undefined): number {
+  let depth = 0;
+  let current = path;
+  while (current !== undefined) {
+    if (typeof current.key === 'string') {
+      depth++;
+    }
+    current = current.prev;
+  }
+  return depth;
+}
+
+/**
  * Implements the "Executing fields" section of the spec
  * In particular, this function figures out the value that the field returns by
  * calling its resolve function, then calls completeValue to complete promises,
@@ -543,6 +607,17 @@ function executeField(
   fieldNodes: ReadonlyArray<FieldNode>,
   path: Path,
 ): PromiseOrValue<unknown> {
+  // Check depth limit before resolving the field.
+  if (exeContext.maxDepth !== undefined) {
+    const depth = pathDepth(path);
+    if (depth > exeContext.maxDepth) {
+      throw new GraphQLError(
+        `Query depth limit of ${exeContext.maxDepth} exceeded, found depth: ${depth}.`,
+        { nodes: fieldNodes },
+      );
+    }
+  }
+
   const fieldDef = getFieldDef(exeContext.schema, parentType, fieldNodes[0]);
   if (!fieldDef) {
     return;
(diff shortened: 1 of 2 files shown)


yaacovCR (dismissed):
Maybe we could consider adding a `fieldDepth` property to FieldDetails of type `number`? Then we wouldn't have to keep recalculating the path length?

eddieran:
Thanks for the review, @yaacovCR! Since `FieldDetails` doesn't exist in the 16.x.x branch this PR targets (collected fields are plain `Map<string, ReadonlyArray<FieldNode>>`), I took the equivalent approach: thread a `depth: number` parameter through `executeFields`/`executeField`/`completeValue` and its helpers. `completeObjectValue` increments depth when descending into sub-selections; list items and abstract-type resolution keep the same depth as their parent field.

This turns the depth check into O(1) per field resolution instead of O(depth) walk of the Path linked list, while preserving identical semantics (list indices still don't count).

All 12 new tests and the full 1972-test suite still pass. Pushed in f9d67dc. Happy to take further suggestions, e.g. if you'd prefer a different naming or placement.

src/execution/execute.ts
```diff
@@ -542,7 +609,16 @@ function executeField(
   source: unknown,
   fieldNodes: ReadonlyArray<FieldNode>,
   path: Path,
+  depth: number,
 ): PromiseOrValue<unknown> {
+  // Check depth limit before resolving the field.
```
yaacovCR, line 614: If the depth limit check was in `completeObjectValue` (and at the root) rather than at the field, right after field collection, we could return all the collected fields/subfields. We could also with this change allow for setting a limit of 0 which should fail everything, not be so useful, but would protects us in that edge case from an error message in that edge case of uncertain use that would read "Query depth limit of 0 exceeded, found depth: 2." when it might be expected to read: "Query depth limit of 0 exceeded, found depth: 1."

src/execution/execute.ts
```diff
@@ -195,6 +197,21 @@ export interface ExecutionArgs {
   options?: {
     /** Set the maximum number of errors allowed for coercing (defaults to 50). */
     maxCoercionErrors?: number;
+    /**
+     * Set the maximum allowed depth for field resolution.
+     * Depth is counted as the number of nested field selections from the root.
+     * When exceeded, a GraphQLError is thrown for the offending field.
+     * No limit is applied when undefined (the default).
+     */
+    maxDepth?: number;
```
yaacovCR, line 206: We could consider calling this option `maxResultDepth` as we may later wish to introduce a separate maximum inline/named fragment depth with respect to field collection.

src/execution/execute.ts
```diff
@@ -195,6 +197,21 @@ export interface ExecutionArgs {
...
+    /**
+     * Set the maximum number of aliases allowed in any single selection set.
+     * This helps prevent alias-bombing denial-of-service attacks where many
+     * aliases for the same field bypass depth-based protections.
+     * When exceeded, a GraphQLError is thrown before executing the selection set.
+     * No limit is applied when undefined (the default).
+     */
+    maxAliases?: number;
```
yaacovCR, line 214:
```suggestion
    maxResponseNames?: number;
```

The idea is that we want a maximum breadth at an object level, whether or not aliases are used.

I think "response name" would be the official erm => https://spec.graphql.org/draft/#sec-Field-Alias

(This would need to be renamed throughout the code/comments.)

src/execution/execute.ts
```diff
@@ -530,6 +571,32 @@ function executeFields(
...
+ * the limit is exceeded.
+ */
+function checkAliasCount(
+  exeContext: ExecutionContext,
+  fields: Map<string, ReadonlyArray<FieldNode>>,
+): void {
+  if (exeContext.maxAliases !== undefined) {
+    const aliasCount = fields.size;
+    if (aliasCount > exeContext.maxAliases) {
+      // Collect nodes for the error location from the first field node of
```
yaacovCR, line 586: Not sure, but I think we would want to include all nodes, not just those above the maximum, it's not the fault of a later node if a new node is added to a operation prior to it and causes it to fail? And we would not be able to tell which is the real offending node. Alternatively, we could just list the parent with too many sub-response names. Not sure how much value we actually add by including all of the nodes causing the response names limit to be breached, it may be more helpful to highlight the offending parent (considering the list of nodes would be expected to be quite long).

src/execution/execute.ts
```diff
@@ -530,6 +571,32 @@ function executeFields(
...
+      // each response key beyond the limit.
+      const nodes: Array<FieldNode> = [];
+      for (const [, fieldNodes] of fields) {
+        nodes.push(fieldNodes[0]);
```
yaacovCR, line 590: If we include all the offending response names, for consistency, I would assume we would need to include all of the fieldNodes, not just the first. This array is the set of overlapping fieldNodes that have been coalesced into one response name, but if we think we're helping by pointing out all the nodes within the operation that cause this limit to be breached, I think we would just confuse things by not listing all the nodes that are part of the problem. If we include just the first, the removal of that node would not cause the error to be avoided if there were other nodes with that response name.

(As above, not sure if we actually need to include all of the offending nodes rather than just the parent.)

yaacovCR (commented):
This looks great to me, comments/suggestions inline.

Looping in @benjie with regards to naming as he is driving the Golden Path initiative (https://github.com/graphql/graphql-wg/issues/1887 https://github.com/graphql/golden-path-wg).

Gentle nudge also about the CLA!

(CI also notes there is a prettier issue).

From benchmarking on my machine ran from the 17.x.x branch:

`npm run benchmark -- benchmark/introspectionFromSchema-benchmark.js benchmark/list-sync-benchmark.js benchmark/list-async-benchmark.js benchmark/object-async-benchmark.js --revs pr/4666 v16.13.2`

Image

benjie:
Is there a reason for doing this at resolve-time rather than validation-time?

Returning `null` for data past a given depth is unsafe; a client might update their normalized cache thinking `null` is the true data for this. An error must be throw instead.

I'd recommend doing this as a validation rule instead.

I'd strongly recommend that you factor in list depth, it's much more significant a concern than selection set depth - selection sets grow cost linearly, lists grow cost exponentially.

yaacovCR:
> Is there a reason for doing this at resolve-time rather than validation-time?

@eddieran agree with @benjie here, whatever can be done at validation time probably should be. I think I was working under the assumption that it might be "difficult" to know the exact depth at validation time, but I think you could just model the maximum possible depth.

> I'd strongly recommend that you factor in list depth, it's much more significant a concern than selection set depth - selection sets grow cost linearly, lists grow cost exponentially.

I presume that lists might require a separate setting altogether as an operation of selectionSet depth of 1 and listDepth 4 is likely not at all equivalent to an operation of selectionSet depth 4 and listDepth 1 for the reason @benjie mentions....

eddieran:
Both points are well-taken — returning `null` past the limit is genuinely unsafe (the cache-poisoning angle hadn't occurred to me; thanks @benjie). I'll rework this as a validation rule that throws during `validate()`, which also means it composes with the existing `specifiedRules`/custom-rules flow rather than being a separate execution option.

For list depth, agreed it's a different shape — selection-set depth is bounded by schema size, list depth multiplies cost per layer. Modeling them as separate limits (`maxSelectionSetDepth` and `maxListDepth`, defaulting to something conservative) makes sense rather than collapsing them. I'll take a stab at both as validation rules and push.

Re: 16.x specifically — `FieldDetails` lands in 17.x, so the validation-rule approach actually ports more cleanly to 16.x than the resolve-time hack since it operates on the AST. Good outcome either way.

kylin791-sys (approved):

kylin791-sys (approved):

kylin791-sys (approved):

PR #4667, opened 2026-04-15
Add @types/react as a peer dependency

afurm:
Adds @types/react as a peer dependency to match the existing react peer dependency in the package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,9 @@
     "build:npm": "node resources/build-npm.js",
     "build:deno": "node resources/build-deno.js"
   },
+  "peerDependencies": {
+    "@types/react": "*"
+  },
   "devDependencies": {
     "@babel/core": "7.17.9",
     "@babel/plugin-syntax-typescript": "7.16.7",


yaacovCR:
Hi @afurm, thanks for contributing!

Just reading the description, it sounds like you are saying we have an existing peer dependency on React, and so we should add the types, but I don't see the existing peer dependency. Did you mean to make a graphql-js website edit instead? Or, as is often, maybe I am missing something!

Thanks again,
Yaacov

afurm:
Closing this PR — there's no existing `react` peer dependency in graphql-js's `package.json` on either `16.x.x` or `main` branches. Adding `@types/react` as a peer dependency would be unrelated to the library's actual dependencies. If there was a specific use case this was intended to address, please feel free to reopen with more context.

yaacovCR:
But @afurm didn't you open this PR?

afurm:
Yes — I opened it assuming there was a React peer dependency that needed type definitions, but after checking `package.json` I confirmed there is no `react` peer dependency in graphql-js. Closing it since adding `@types/react` without an existing peer dependency would be unrelated to the library. Sorry for the noise!

yaacovCR:
The noise is fine, we are hopeful and grateful for more collaborators! I was just surprised by this line:

> If there was a specific use case this was intended to address, please feel free to reopen with more context.

My interpretation in 2026 based on nothing was that you had directed an AI agent to try to fix bugs of this type, and it got overeager. Or that there possibly is a bug in our website code which lives in this same repo under the `website` folder which I am less familiar with.

Either way, welcome to the repo! We appreciate any help you can give us!

PR #4668, opened 2026-04-16
Move maxTokens check to lexer

benjie:
By checking `maxTokens` inside the `createToken` function we can be sure that it's applied consistently.

I've also moved ParserOptions.lexer to be an internal property (it should only be used by the schema coordinates parser, it was never intended to be a public interface) and implemented an assertion that forbids `lexer` and `maxTokens` together (since if they were both set, `maxTokens` would be ignored). I've also marked the SchemaCoordinateLexer as internal; we don't expose it via `index.ts` anyway, but best to signal the intent anyway.

I'm not super happy about mutating the internal state of the lexer from `createToken` - really I'd rather `createToken` were a method on the lexer class - but I don't want the two Lexers to have to implement it each and we don't like to extend classes so I think we can accept it in this one place. I'm not happy about it though, so if you have ideas for a better approach, please go ahead!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -242,6 +242,7 @@ export {
 
 export type {
   ParseOptions,
+  LexerOptions,
   SourceLocation,
   TokenKindEnum,
   KindEnum,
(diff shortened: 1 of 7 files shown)


andimarek:
You also want to limit the max chars overall + the max white space tokens (if they count differently)

Here are the limits GJ uses by default: https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/parser/ParserOptions.java

They should not be optional or off by default but we should rather rather set them to a high value by default.

PR #4670, opened 2026-04-18
feat(v17): Implement tracing channels

### LINKED ISSUE #4629: Use Tracing Channel for observability
I came across https://github.com/graphql/graphql-js/issues/3133 and since we (Sentry and other APM providers) are driving an ecosystem effort to adopt tracing channels into the server runtimes, I think this is a good time to discuss this again, especially with the concerns noted there no longer being valid.

I'd like to propose adding first-class [`TracingChannel`](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel) support to `graphql-js`, following the pattern established by [`undici`](https://github.com/nodejs/undici) in Node.js core.

## Motivation

graphql-js is deliberately minimal, it provides the execution engine and nothing else. There are no built-in hooks, middleware, plugins, or tracing APIs. This is a strength for a reference implementation, but it means **every** APM tool must resort to monkey-patching to observe what the engine is doing.

Today, `@opentelemetry/instrumentation-graphql` patches `parse`, `validate`, and `execute` using `import-in-the-middle` for esm and `require-in-the-middle` for CJS, then recursively traverses the entire schema at instrumentation time to wrap every field resolver's `resolve` function. Datadog's `dd-trace` does the same. Sentry does the same. Every APM vendor independently patches the same three functions and walks the same schema tree which is fragile, duplicative, and version-coupled.

Not to mention the broader ecosystem concerns:

- **Runtime lock-in:** RITM and IITM rely on Node.js-specific module loader internals (`Module._resolveFilename`, `module.register()`). They don't work on Bun or Deno, which implement the Node.js API surface but not the module loader internals.
- **ESM fragility:** IITM is built on Node.js's module customization hooks, which are still evolving and have been a persistent source of breakage in the OTEL JS ecosystem.
- **Initialization ordering:** Both require instrumentation to be set up before `graphql` is first `require()`'d / `import`'d.
- **Bundling:** Users must ensure instrumented modules are externalized, which is increasingly difficult as frameworks bundle server-side code into single executables or deployment files.
- **Schema walking is expensive.** Every APM tool recursively wraps every resolver on every type in the schema at setup time. This is O(fields) work that runs once per instrumented schema — and if the schema is rebuilt (e.g., in a gateway), it runs again. Native emission eliminates this entirely.

`TracingChannel` solves all of these. It provides structured lifecycle events (`start`, `end`, `asyncStart`, `asyncEnd`, `error`) with built-in async context propagation, zero-cost when no subscribers are attached, and a standardized subscription model that requires no monkey-patching.

Tracing channels already solve many issues with standard event emitters:

- No overhead when no subscribers are listening.
- Tracing channels can be acquired from anywhere, anytime in the code, so timing and load order won't be an issue.
- Compatible with all server-runtimes we have today (more on that below).
- Automatically handles correlation and context propagation, so no need to track executions with requestIds and no need to do the span relationship dance.

## Cross-platform compatibility

A previous discussion ([#3133](https://github.com/graphql/graphql-js/issues/3133)) raised the concern that `diagnostics_channel` is Node.js-specific and graphql-js targets multiple platforms. This concern was valid in 2021 but is no longer accurate:

- **Bun** supports `node:diagnostics_channel` including `TracingChannel` ([Bun docs](https://bun.sh/docs/runtime/nodejs-apis#node-diagnostics-channel))
- **Deno** supports `node:diagnostics_channel` via its Node.js compatibility layer ([Deno docs](https://docs.deno.com/api/node/diagnostics_channel/)), and also supports the `TracingChannel` API.
- **CloudFlare Workers** also have the [same level of compatability](https://developers.cloudflare.com/workers/runtime-apis/nodejs/diagnostics-channel/#tracingchannel).

This means that every server-side JavaScript runtime that runs graphql-js in production now supports `diagnostics_channel`. Browser environments don't need APM tracing at the execution engine level, browsers usually run GraphQL clients, not servers and even then the API can be imported/loaded conditionally if available and only used if so, so you can write isomorphic logic for it.

The standard compatibility pattern used across the ecosystem handles this cleanly:

```js
let dc;
try {
  dc = ('getBuiltinModule' in process)
    ? process.getBuiltinModule('node:diagnostics_channel')
    : require('node:diagnostics_channel');
} catch {
  // No diagnostics_channel available — all tracing is a no-op
}
```

This is zero-cost when `diagnostics_channel` is unavailable, this means no import, no overhead, no behavior change.

## Runtime Overhead

Tracing channels are specifically designed to handle way more events than `EventEmitter`, and it can be optimized to have zero-overhead when there are no listeners. In runtime we have access to `hasSubscribers` which can be checked before constructing any context objects.

## Implementation

We can use OTEL's as the bare minimum we need as it is the most use instrumentation for `graphql` out there by most APM providers.

All channels use the Node.js [`TracingChannel`](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel) API, which provides `start`, `end`, `asyncStart`, `asyncEnd`, and `error` sub-channels automatically.

| TracingChannel | Tracks | Context fields |
|---|---|---|
| `graphql:execute` | `execute()` — full operation execution lifecycle | `operationType`, `operationName`, `document`, `schema`, `variableValues` |
| `graphql:parse` | `parse()` — query string to AST | `source` |
| `graphql:validate` | `validate()` — AST validation against schema | `document`, `schema` |
| `graphql:resolve` | Individual field resolver execution | `fieldName`, `fieldPath`, `fieldType`, `parentType`, `args` |
| `graphql:subscribe` | `subscribe()` — subscription setup | `operationType`, `operationName`, `document`, `schema`, `variableValues` |

We can spec out the types and exact fields in each channel's context in a PR.

### Usage in the Ecosystem

```js
const dc = require('node:diagnostics_channel');

// Subscribe to operation execution — the primary span
dc.tracingChannel('graphql:execute').subscribe({
  start(ctx) {
    ctx.span = tracer.startSpan(`${ctx.operationType} ${ctx.operationName}`, {
      attributes: {
        'graphql.operation.type': ctx.operationType,
        'graphql.operation.name': ctx.operationName,
        'graphql.source': print(ctx.document),
      },
    });
  },
  asyncEnd(ctx) {
    ctx.span?.end();
  },
  error(ctx) {
    ctx.span?.setStatus({ code: SpanStatusCode.ERROR, message: ctx.error?.message });
    ctx.span?.recordException(ctx.error);
  },
});
```

That's it, all interested parties including the user can listen for the other channels like this one and run their own telemetry and logic.

## Prior Art

This approach follows the same pattern already adopted or in progress by other major libraries:

- **`undici`** (Node.js core) — ships `TracingChannel` support since Node 20.12: [`undici:request`](https://nodejs.org/api/diagnostics_channel.html#undici-channels)
- **`fastify`** — ships `TracingChannel` support natively (`tracing:fastify.request.handler`)
- **`node-redis`** — [redis/node-redis#3195](https://github.com/redis/node-redis/pull/3195) (`node-redis:command`, `node-redis:connect`)
- **`ioredis`** — [redis/ioredis#2089](https://github.com/redis/ioredis/pull/2089) (`ioredis:command`, `ioredis:connect`)
- **`pg` / `pg-pool`** — [brianc/node-postgres#3624](https://github.com/brianc/node-postgres/pull/3624) (`pg:query`, `pg:connection`, `pg:pool:connect`)
- **`mysql2`** — [sidorares/node-mysql2#4178](https://github.com/sidorares/node-mysql2/pull/4178) (`mysql2:query`, `mysql2:execute`, `mysql2:connect`, `mysql2:pool:connect`)

_Full disclosure, I'm leading the effort on most of these implementations_

---

I would love to hear from you if there's appetite for this, I'm happy to spec this in a PR and fully own this until it is shipped.

I think this feels like a great milestone for v17, the code changes are extremely minimal and if you check some of those PRs it takes a few minutes, or hours at best to transform a library or a framework to be observable-friendly with tracing channels.

logaretm:
This PR implements tracing channel support for `graphql-js`.

graphql-js resolves `node:diagnostics_channel` at module load through `process.getBuiltinModule`. On runtimes where the built-in is unavailable, such as browsers, channel handles are `undefined` and emission sites short-circuit. APMs subscribe through their own `node:diagnostics_channel` import by channel name; no GraphQL API call or option is required.

The implementation adds the diagnostics channel helper, internal structural typings for the Node tracing-channel surface, subscriber-facing context types exported from the root package, and wrappers around parse, validate, execute, subscribe, root-selection-set execution, variable coercion, and resolver execution. The no-subscriber path is guarded by `shouldTrace`/`hasSubscribers` checks, with a sub-channel fallback for runtimes that do not expose aggregate `TracingChannel.hasSubscribers`. Traced execution uses `runStores` to preserve tracing-channel async context propagation, does not normalize synchronous execution to async, and handles promise-like resolver results without `Promise.resolve` normalization.

Test coverage exercises lifecycle emission, lazy context fields, no-subscriber no-op behavior, missing `node:diagnostics_channel` behavior, exported context types, promise-like resolver results, and runtime compatibility coverage for Bun and Deno.

### API

graphql-js vendors a minimal structural subset of the tracing channel types for internal use so it does not depend on Node typings or on `node:diagnostics_channel` being present at runtime. It also exports subscriber context types from the root package, including `GraphQLChannelContextByName` and the per-channel context interfaces.

Each channel publishes lifecycle events through Node tracing channels. The table lists the base context fields present at `start`; successful sync work adds `result` by `end`, successful async work adds `result` by `asyncEnd`, and thrown or rejected work adds `error` on the `error` lifecycle. Variable coercion is synchronous, so it only emits `start`/`end` and `error` when coercion throws abruptly.

| Channel | Context fields |
|---|---|
| `graphql:parse` | `source` |
| `graphql:validate` | `schema`, `document` |
| `graphql:execute` | `schema`, `document`, `rawVariableValues`, `operationName` *(lazy)*, `operationType` *(lazy)* |
| `graphql:execute:variableCoercion` | `schema`, `document`, `operation`, `rawVariableValues`, `operationName`, `operationType` |
| `graphql:execute:rootSelectionSet` | `schema`, `document`, `operation`, `rawVariableValues`, `operationName`, `operationType` |
| `graphql:subscribe` | `schema`, `document`, `rawVariableValues`, `operationName` *(lazy)*, `operationType` *(lazy)* |
| `graphql:resolve` | `fieldName`, `alias`, `parentType`, `fieldType`, `args`, `isDefaultResolver`, `fieldPath` *(lazy)* |

`graphql:execute:variableCoercion` publishes the caller-provided values as `rawVariableValues`; its successful `result` contains the coerced `{ variableValues }`, or `{ errors }` when coercion returns request errors.

`graphql:execute:rootSelectionSet` fires once for the operation root selection set. For subscriptions, it fires once per emitted subscription event, matching the OTel `graphql.subscription.event` shape; subscribers can distinguish operation kinds through `operationType`.

### Usage

```ts
import dc from 'node:diagnostics_channel';
import type { GraphQLResolveContext } from 'graphql';

dc.tracingChannel('graphql:resolve').subscribe({
  start: (context: GraphQLResolveContext) => {
    // start span
  },
  end: (context: GraphQLResolveContext) => {
    // sync resolver result is available here as context.result
  },
  asyncEnd: (context: GraphQLResolveContext) => {
    // async resolver result is available here as context.result
  },
  error: (context: GraphQLResolveContext) => {
    // report context.error
  },
});
```

Closes #4629


src/diagnostics.ts
```diff
@@ -0,0 +1,243 @@
...
+ * After calling this, graphql-js will publish lifecycle events on the
+ * following tracing channels whenever subscribers are present:
+ *
+ *   - `graphql:parse`
+ *   - `graphql:validate`
+ *   - `graphql:execute`
+ *   - `graphql:subscribe`
+ *   - `graphql:resolve`
+ *
+ * Calling this repeatedly is safe: subsequent calls replace the stored
```
yaacovCR: is a potential failure mode if this is for some reason called by different APMs with different values?

logaretm: As long as it is the actual diagnostic channel module, it will work fine. The failure mode is if one APM provides an implementation and a different one provides another. The diagnostic channel runtime implementations today are idempotent AFAIK so it doesn't matter as long everyone ends up on the same channels.

I think this is a highly specialized case, I initially considered either throwing to force only implementation over the other OR silently NOOP but I that favors whoever comes first which isn't ideal.

WDYT, should this be tweaked? 

looping @qard here as he might have some thoughts here.

yaacovCR: We could support multiple diagnostic channels just to avoid this failure mode. Most of the time there would be just one and we would pay the cost of indexing the one item array on every publish, but it seems safest. Thoughts?

in terms of costs in general the introspection query seems to take a 5% hit.

I wonder if that could be reduced if instead of using wrapping functions, we embedded the published decision within the execution code which already tests for sync/async. I guess we would also need some logic to make sure is Publishing/hasSubscribers is uniform

logaretm: > We could support multiple diagnostic channels just to avoid this failure mode...

Storing two refs means publishing the same event to the same underlying channel twice, which double-fires every subscriber or more depending on how many `enableDiagnosticsChannel` calls with the same dc module were made.

In server environments it is very unlikely they would resort to anything else other than the `diagnostic_channels` module available on the runtime because it is already on all these runtimes, yet it is still possible. I think this failure is not worth the added complexity here.

Dedup isn't easy either, `import/require` would normally resolve the same module, but with mixed `cjs/esm`  cases they would have different references so double fires will happen in that case. A shared global channel is a strength of diag channels, so we should lean on it and trust the users/runtimes do it properly.

We could throw in that case specifically when different dc implementations are detected to make this visible. What do you think?


> I wonder if that could be reduced if instead of using wrapping functions, we embedded the published decision within the execution code...

We can try but the issue is `runStores` will have to wrap everything in the code paths to maintain AsyncLocalStorage store propagation from `start->asyncEnd` which means either we duplicate a lot of code (traced vs. untraced paths) or take the hit 🤔

logaretm: I might've overthinked that last part, pushed 908e1eee to see if it can help. It caches the trace decision on the executor instance and avoids allocating closures if no traced calls are made.

yaacovCR: > Dedup isn't easy either, `import/require` would normally resolve the same module, but with mixed `cjs/esm` cases they would have different references so double fires will happen in that case.

Putting aside that we have `require-esm` in newer versions of node, I ran the following on Node v19:

```ts
// test-diagnostics-channel.cjs
'use strict';

module.exports = require('node:diagnostics_channel');
```
```ts
// test-diagnostics-channel.mjs
import diagnosticsChannel from 'node:diagnostics_channel';

export default diagnosticsChannel;
```
```ts
// test.js
import assert from 'node:assert/strict';

import dcCjs from './test-diagnostics-channel.cjs';
import dcEsm from './test-diagnostics-channel.mjs';

assert.strictEqual(dcEsm, dcCjs);
console.log('CJS and ESM received the same diagnostics_channel object.'); // true
```

For good measure, I tested on deno/bun, and they also pass.

So I think the best option within the current implementation framework given the above is as you suggest to have `enableDiagnosticsChannel(...)` throw if it is called an additional time with a non-identical `dc`.

yaacovCR: But see: https://github.com/graphql/graphql-js/issues/4629#issuecomment-4286776524

yaacovCR:
> The implemented channels are...

Please excuse my limited understanding of the current tracing space, but is there a concept of nesting of channels? In theory, I would expect we would want to:

(1) nest the "execute" channel for subscription events under "subscribe" and
(2) nest "resolve" under "execute"

Naively, I would assume this would be very useful information to understanding the hierarchy.

And I would think this would provide a way to pass the overarching context of the subscribe option to execute (i.e. execute would always have the document, as in your description of the channels/context within the linked issue https://github.com/graphql/graphql-js/issues/4629)?

And similarly, we could also have the overall execute (or subscribe/execute) passed to resolve?

src/diagnostics.ts
logaretm: Thanks for checking, I had AI run similar tests but it seems like it used a different module loader or `createRequire` and they were two different references 🤔 but maybe the test was flawed.

Anyways, I added the throwing logic in b2139c17 and elaborated more in the issue (text wall warning).

yaacovCR: > Thanks for checking, I had AI run similar tests but it seems like it used a different module loader or `createRequire` and they were two different references 🤔 but maybe the test was flawed.

Uh-oh I would be interested in how this could fail? Do you have a way of sharing the failing test code?

logaretm:
@yaacovCR Happy to elaborate more!

> is there a concept of nesting of channels?

From a purely diag/tracing channels API perspective, there's no parent/child relationship in the API. Each channel is independent. `graphql:execute` and `graphql:resolve` don't know about each other at the module level.

That said, the hierarchy you're describing does exist at runtime, it's just reconstructed on the subscriber side rather than on the emitter side. The mechanism is done through `AsyncLocalStorage`:

You can check the image I had in the PR description that represents how it looks like, or check this [trace view](https://teley.dev/shared/60a0a006-ebc) as you can see the spans are nested under one another.

This works because tracing channels propagate the async execution context (unlike plain event emitters).  Subscribers bind an `AsyncLocalStorage` to the channel, and every downstream channel fired under that sees the same store via `async_hooks`. Think of it as an implicit requestId threaded through by the runtime.

It's why I kept the runStores wrapper in the types and impl even after dropping `tracePromise`. `TracingChannel.traceSync/tracePromise` wrap runStores for you, but we roll the lifecycle manually here to handle the mixed sync-or-promise return path cleanly, so we need to call runStores ourselves to preserve ALS propagation.


---

Now, for graphql as the emitter, there are a couple of reasons not to encode the hierarchy ourselves:

- GraphQL wouldn't know if the channels are nested, unless we implement logic that does that or use ALS ourselves but that's strictly a node API that's not polyfillable.
- We'd have to track the hierarchy unconditionally even when nothing is subscribed because we don't know at emit time whether a subscriber cares. That defeats the `hasSubscribers` short-circuit we use on the hot paths today.
- Subscribers can attach to one channel or several or all of them. If graphql baked the hierarchy into the ctx object, we risk breaking it for those partial subs.

The good news is subscribers will have access to `document` and other parent event objects naturally through their ALS if they want to, graphql doesn't have to do it and subscribers get it for free.

Here is a quick example snippet:

```ts
import { AsyncLocalStorage } from 'node:async_hooks';
import dc from 'node:diagnostics_channel';
import { enableDiagnosticsChannel } from 'graphql';

enableDiagnosticsChannel(dc);

const executeCtx = new AsyncLocalStorage();

const executeChannel = dc.tracingChannel('graphql:execute');
const resolveChannel = dc.tracingChannel('graphql:resolve');

// Bind on `start` only, the store activates when execute's runStores frame
// opens and stays active through the entire lifecycle. `end`, `asyncStart`,
// `asyncEnd`, and `error` all fire inside that frame (sync handlers directly,
// async ones via async_hooks propagation), so they inherit the same store
// without any extra wiring.
executeChannel.start.bindStore(executeCtx, ctx => ({
  document: ctx.document,
  operationName: ctx.operationName,
  startedAt: Date.now(),
}));

// Subscribe to the full execute lifecycle. Every handler sees the same store.
executeChannel.subscribe({
  start: () => {
    const { operationName } = executeCtx.getStore();
    console.log(`[execute:start] ${operationName}`);
  },
  end: () => {
    const { operationName, startedAt } = executeCtx.getStore();
    console.log(`[execute:end]   ${operationName} (+${Date.now() - startedAt}ms sync)`);
  },
  asyncStart: () => {
    const { operationName } = executeCtx.getStore();
    console.log(`[execute:asyncStart] ${operationName}`);
  },
  asyncEnd: () => {
    const { operationName, startedAt } = executeCtx.getStore();
    console.log(`[execute:asyncEnd]   ${operationName} (+${Date.now() - startedAt}ms total)`);
  },
  error: ({ error }) => {
    const { operationName } = executeCtx.getStore();
    console.log(`[execute:error] ${operationName} -> ${error.message}`);
  },
});

// Nested channels inherit the parent's store too, because resolve fires
// beneath execute's runStores frame.
resolveChannel.subscribe({
  start: resolveCtx => {
    const parent = executeCtx.getStore();
    console.log(`  [resolve:start] ${resolveCtx.fieldPath} (under ${parent.operationName})`);
  },
  asyncEnd: resolveCtx => {
    const parent = executeCtx.getStore();
    console.log(`  [resolve:asyncEnd] ${resolveCtx.fieldPath} (under ${parent.operationName})`);
  },
  error: ({ error }, name) => {
    const parent = executeCtx.getStore();
    console.log(`  [resolve:error] under ${parent.operationName}: ${error.message}`);
  },
});
```

src/diagnostics.ts
logaretm: Sure I put it up here: 

https://github.com/logaretm/node-dc-identity-repro

yaacovCR: That makes sense, and you also point to how we can deduplicate around the identity of `dc.tracingChannel`, so that seems workable for this hopefully unlikely failure mode.

I would assume `import * as dc from 'node:diagnostics_channel` would be the same as `dc = import('node:diagnostics_channel)`, both would be covered by checking `dc.tracingChannel`

yaacovCR:
> @yaacovCR Happy to elaborate more!

This was hugely helpful. On this background, I am wondering whether instead of:

  | Channel | Context fields |                                                                                                                                                                                                                                                                                             
  |---|---|                                                                                                                                                                                                                                                                                                                
  | `graphql:parse` | `source` |                                                                                                                                                                                                                                                                                           
  | `graphql:validate` | `schema`, `document` |                                                                                                                                                                                                                                                                            
  | `graphql:execute` | `schema`, `document`, `variableValues`, `operationName` *(lazy)*, `operationType` *(lazy)* |                                                                                                                                                                                                       
  | `graphql:subscribe` | `schema`, `document`, `variableValues`, `operationName` *(lazy)*, `operationType` *(lazy)* |                                                                                                                                                                                                     
  | `graphql:resolve` | `fieldName`, `parentType`, `fieldType`, `args`, `isTrivialResolver`, `fieldPath` *(lazy)* |       

Maybe something like the below would be better for our official tracing channels:

  | Channel | Context fields |                                                                                                                                                                                                                                                                                             
  |---|---|                                                                                                                                                                                                                                                                                                                
  | `graphql:parse` | `source` |                                                                                                                                                                                                                                                                                           
  | `graphql:validate` | `schema`, `document` |                                                                                                                                                                                                                                                                            
  | `graphql:execute` | `schema`, `document`, `variableValues`, `operationName` *(lazy)*, `operationType` *(lazy)* |                                                                                                                                                                                                       
  | `graphql:executeQueryMutationOrSubscriptionEvent` | `schema`, `operation`, `variableValues`, `operationName`, `operationType` |                                                                                                                                                                                             
  | `graphql:resolve` | `fieldName`, `parentType`, `fieldType`, `args`, `isTrivialResolver`, `fieldPath` *(lazy)* |       

The idea would be that `subscribe` is not really "a thing" according to the spec, and in theory our separate `execute` and `subscribe` endpoints are just a way to keep track of the relationship between `execute` and `executeSubscriptionEvent`. This would nest `executeQueryMutationOrSubscriptionEvent` under every `execute` call for queries, mutations, and subscriptions, with one `executeQueryMutationOrSubscriptionEvent` for each query/mutation `execute`, and a variable number of `executeQueryMutationOrSubscriptionEvent` for a subscription `execute`.

The downside is you can no longer top-level distinguish between subscriptions and queries/mutations. Do we want to preserve that? Do we want to top-level distinguish between each of our operations? What about between queries and mutations?

Again, I am asking these questions with very little knowledge of how the tracing system is used in practice, hopefully you can guide me here?

Looping in @graphql/tsc if they have any thoughts.

logaretm:
@yaacovCR 

> Maybe something like the below would be better for our official tracing channels:

We could do `graphql:operation` as top level and `graphql:execute` 1x for queries/mutations and N x per sub. Does that track or do we want that lengthy name instead 😅 

> The downside is you can no longer top-level distinguish between subscriptions and queries/mutations...

The distinction can be done with the args already sent to the execute channel via the `operationType` property we have right now. Channel level distinction is unlikely to be worth it given how low the overhead is.

I initially added a `subscribe` channel because it felt like its own thing, but it is better to name channels from GQL's POV, like "I'm doing this thing right here....".

What do u think?

yaacovCR:
> What do u think?

Naming = so hard. Definitely like shorter. But I do prefer consistency with the other verbs we have for our channels.

We could consider:

- `execute` vs. `executeQueryMutationOrSubscriptionEvent` (not great if another operation type lands)
- `operation` vs. `execute` (not great as operation = noun)
- `execute` vs. `operation` (ha! both are not great as it's not clear really which should be the wrapper, although this is incorrect as the nested `executeSubscriptionEvent` is not an operation, it's more like a sub-operation which is not a thing.)
- `executeRequest` vs. `executeOperation` (not correct, a subscription operation does not contain multiple underlying operations, as above)
- `executeRequest` vs. `executeRootSelectionSet` (correct and technically disambiguated, as when we execute a single root field to create the source event stream for a subscription, we do not [currently] allow for merging multiple fields by utilizing multiple root fields, so on the subscription side it is [presently] `executeSingleAllowedRootField` and if we ever change that we could just call that part `createSourceEventStream`)

I am leaning toward the last option, again looping in better namers @graphql/tsc @graphql/tsc-emeriti

yaacovCR:
@logaretm 

From my machine on `17.x.x` I ran:

`npm run benchmark -- benchmark/introspectionFromSchema-benchmark.js benchmark/list-sync-benchmark.js benchmark/list-async-benchmark.js benchmark/list-asyncIterable-benchmark.js benchmark/object-async-benchmark.js --revs pr/4670 17.x.x`

And got:

image

1. On our real-world large sync query we got a 7% slowdown. This is the most concerning result.
2. On a synthetic sync query with a lot of list items we got a 23% slowdown, also concerning, although the query is artificial.
3. On synthetic async queries, we were at parity or faster, somehow thanks to this PR and presumably v8 magic!

Not sure if you have any insight. I would assume most of this is secondary to graphql:resolve. Ideally we would of course have no slowdowns at all, at least on these benchmarks when tracing is not enabled.

src/diagnostics.ts
yaacovCR: Actually, @logaretm on further reflection, because APMs would own the process and instructions for browser bundlers, I think we can use the original mechanism you suggested  which implicitly enables directly from node. Apologies for the back and forth.

logaretm: No worries, happy to re-work the PR and thanks for discussing it!

logaretm:
@yaacovCR I will re-work the PR to use the autoloading snippet and I will dig into the benchmarks and see what I can do, ideally it should offer close to zero slowdown so I must be doing something wrong.

yaacovCR:
> * `executeRequest` vs. `executeRootSelectionSet` (correct and technically disambiguated, as when we execute a single root field to create the source event stream for a subscription, we do not [currently] allow for merging multiple fields by utilizing multiple root fields, so on the subscription side it is [presently] `executeSingleAllowedRootField` and if we ever change that we could just call that part `createSourceEventStream`)
> 
> I am leaning toward the last option, again looping in better namers @graphql/tsc @graphql/tsc-emeriti

Although maybe we should do instead:
`graphql:execute` and `graphql:execute:rootSelectionSet`
or even
`graphql:execute` and `graphql:execute:root-selection-set`

is there a convention in the diagnostic channels world?

logaretm:
@yaacovCR Not a strict convention, but this is [what is currently recommended](https://arc.net/l/quote/jrfctjds).

For the names, I think we can keep the inner one called `graphql:execute` for the outer one, how about:

- `graphql:request`
- `graphql:run`
- `graphql:process`
- `graphql:handle`

Naming is hard 😅

yaacovCR:
Not necessarily related, but possibly, what would a trace look like for incremental delivery?

Options abound because of the disconnect between execution groups and delivery groups.

logaretm:
@yaacovCR So I pushed a couple of optimizations, I think it is down significantly after running the benchmark u gave me. It was at 6% overhead at the time of your posting on my machine and now its down to 2.3%-3%.

Can u have a go and see if that's acceptable? I have more ideas but will require juggling alternate Executor class implementations.

----

> Not necessarily related, but possibly, what would a trace look like for incremental delivery?

Right now they aren't handled and any deferred executions will be orphaned (up to the APM how to deal with that) since they will still fire the `resolve` channels.

So it would look like this:

```
  time ──────────────────────────────────────────────────────►

  execute   ◄────── runStores frame #1 ──────►
  ├─ resolve fieldA        (#2, parented to #1)  ✓
  ├─ resolve fieldB        (#3, parented to #1)  ✓
  └─ returns {initialResult, subsequentResults}
                      ▲ frame #1 ends here ─ generator has not run yet

          ─── gap: framework owns the generator ───

           await subsequentResults.next()   (fresh async context, no ALS)
                                                                       
  resolve fieldC (@defer)  (#4, ⚠ ORPHANED — no parent span)
  resolve fieldD (@stream) (#5, ⚠ ORPHANED — no parent span)

```

I think we can keep the current PR scoped to this at the moment, and I can try a follow up if this one goes through. I'm thinking we can add another `:group` channel, so it looks like:

```
 time ──────────────────────────────────────────────────────►

  execute   ◄────── runStores frame #1 ──────►
  ├─ resolve fieldA        (#2, parented to #1)  ✓
  ├─ resolve fieldB        (#3, parented to #1)  ✓
  └─ returns {initialResult, subsequentResults}
                      ▲ frame #1 ends

          ─── gap: framework drives the generator ───

           await subsequentResults.next()

  execute:group 'slowDefer'  ◄── runStores frame #4 ──►
  └─ resolve fieldC (@defer) (#5, parented to #4)  ✓

  execute:group 'feed@stream' ◄── runStores frame #6 ──►
  └─ resolve fieldD (@stream) (#7, parented to #6)  ✓
  ```
  
The group events look like extra spans, but their real job is to keep an ALS frame alive while the deferred resolvers run, so anything the APM bound stays readable. I'll verify end-to-end in the follow-up

**Comment by logaretm** [comment]
Squeezed a bit of performance with the last couple of commits 🙌   for the channel names happy to pick whatever works for us here to make this ready to go.

**Review by Qard — APPROVED** [review]
I'm the creator of diagnostics_channel, for those that don't know me. 👋🏻 

Generally, LGTM. I'd like to know more about the exact timing on resolve tracing. Can those be effectively linked together via `channel.bindStore(...)` triggering nested resolvers within the scope of their parent?

**Comment by logaretm** [comment]
@Qard ATM nested resolvers are not linked together as children, they instead execute and appear as siblings. However they all appear as a child of the `execute` channel invocation.

CleanShot 2026-04-26 at 20 05 59@2x

I realize this isn't the ideal outcome, I will experiment with expanding the runStores callback to include resolve child recursion and see how it fares in benchmarks.

**Comment by Qard** [comment]
Ideally we'd want that causality. Might be sufficient to pass the parent resolver info into the child events as that is past into the resolvers themselves. 🤔

**Comment by logaretm** [comment]
@Qard I widened the runStores frame to include the nested resolvers, and got something looking like this

CleanShot 2026-04-26 at 21 10 16@2x

Perf wise, it has the same perf with no subs as the current state, all near 0% diff so it didn't introduce cold paths. Now comparing it against the current impl (no nesting) with subs attached, the numbers are very similar.

  | Benchmark | Δ |                                                               
  |---|---|
  | Introspection | −0.2% |                                                       
  | Sync list | −1.0% |                                                
  | Async list | −1.8% |
  | Async iterable | −2.8% |
  | Object-async | **−6.8%** |
  
  So only one case has seen a slow down which I think would come to a wash once an APM does somethings anyways. So we basically get it for free here, WDYT? worth including or maybe let the subscriber stitch it together? (they can with the path information in each call).

**Comment by Qard** [comment]
I would generally go for letting the consumer stitch it together themselves, just ensuring the run store boundaries are sufficient that they _can_ do that when desired. Not all vendors will want to use the same structure.

**Comment by logaretm** [comment]
Cool, I have verified it and yes consumers can stitch it in the exact same way but way cheaper on the receiving end. Same trace is produced.

CleanShot 2026-04-26 at 21 35 30@2x

**Comment by logaretm** [comment]
@yaacovCR bun fails because `hasSubscribers` on Bun isn't supported yet, which is why we had `shouldTrace` that checked `hasSubscribers !== false` specifically, it acted as a compatibility and gate checks at the same time.

It means in older or less specced environments we publish since we cannot tell if there is a subscriber or not, which is still low overhead. For Node > 18 it is supported and should be patched up by bun as well soon.

Should I put it back? I noticed it got changed in 626813d8 and then later removed.

**Comment by yaacovCR** [comment]
> Should I put it back? I noticed it got changed in [626813d](https://github.com/graphql/graphql-js/commit/626813d8d2df048d89c4ed7de066cc1175ec3298) and then later removed.

Aha.

Supposedly bun supports `channel.start/stop/etc.hasSubscribers`, so we could temporarily make `shouldTrace` equivalent to something like:

`channel?.hasSubscribers || ['start', 'end', 'asyncStart', 'asyncEnd', 'error'].some(tc => channel?.tc.hasSubscribers`

Also, apologies, I think I sloppily may have clobbered over some of your other changes re: discussion above with nesting and additional context if those were additional commits? or did you land on not making an actual change, but just having subscriber stitch the resolver waterfall via paths?

I would just add that there actually is no actual nesting of `resolve` calls whatsoever. `executeField` calls are nested, `resolve` calls appear in a waterfall. But this is probably again exposing my tracing ignorance!

**Comment by yaacovCR** [comment]
Otherwise, I am basically done with my additions/changes for the moment. In terms of the names, I think considering I can't think of a great alternative schema, I am temped to stick with the original names.

**Comment by logaretm** [comment]
@yaacovCR 

> Supposedly bun supports channel.start/stop/etc.hasSubscribers

We could apply the `some` branch if `hasSubscribers` is undefined, so yep. I will add that back in and run the benches to see if branching is safe there.

> Also, apologies, I think I sloppily may have clobbered over some of your other changes re: discussion above with nesting and additional context if those were additional commits? 

Not doing it at the moment, right now subscribers can use the path to stitch the waterfall properly which is easy enough. We can later change that since it doesn't change how each event fires.

> I would just add that there actually is no actual nesting of `resolve` ...

Oh yea, we were just discussing it from the subscriber POV, not the impl. Sorry for the confusion.

> I am temped to stick with the original names.

Haha, okay then. I will just push the `shouldTrace` and this would be ready to go.

**Comment by logaretm** [comment]
@yaacovCR Thanks for all those polishes ❤️  Do we think we can merge this soon?

**Comment by yaacovCR** [comment]
@logaretm 

This PR is a great effort in itself -- and also helped clarify some internals. Thank you!

In brief (and after the refinements this PR prompted!), for v17, we are encouraging interested frameworks to call some of the internal helpers:

1. `validateExecutionArguments` and `executeRootSelectionSet` instead of `execute`.
2. `validateSubscriptionArguments`, `createSourceEventStream`, and `ValidatedSubscriptionArgs.perEventExecutor`.

This would be a problem if we only trace `execute` or `subscribe`

My solution is to introduce `execute:rootSelectionSet` which is called by anything that hits our internal `Executor.executeRootSelectionSet` method.

Then we can drop how `graphql:subscribe` wraps a `graphql:execute` trace for each event, because there will already be an internal `graphql:execute:rootSelectionSet` event, and if someone is using a custom `perEventExecutor` instead of wrapping ours, they can presumably add their own custom tracing. (We could also think about also introducing a `subscribe:perEventExecutor` within `mapFn`, but I think we can add that later on request.) 

What do you think?

**Comment by logaretm** [comment]
@yaacovCR I think that makes sense! I see you already implemented them. As long as we can use the channels to trace, the naming and structure doesn't matter much to APMs as they map them to spans/metrics/whatever.

So naming and "shape" should be owned by GraphQL,  I think it looks good as it stands right now. We can use this immediately once it gets merged in our SDKs.

I tested the lates changes in a real app and it worked well, producing the expected traces. We can keep iterating on this after an initial release as graphql grows more and more into owning its own observability.

**Comment by logaretm** [comment]
@yaacovCR noticed a v17 release going out with some changes. I rebased and aligned this PR with v17 branch.

I also noticed we had double instrumentations firing for the rootSelectionSet, I fixed it and added a test for both use-cases you mentioned.

**Comment by yaacovCR** [comment]
Ty ty ty. Last thing I want to change is that I am convinced in the end that ValidatedExecutionArgs should save a reference to the original Execution Args and instead of reconstructing the original variable values, then we will have access to them. Then this seems good to go from my perspective.

another question I have — I remembered only recently that we have a graphql otel wg. This PR would seem to my instrumentation unaware self somewhat related to their work… are you at all involved in that work? Are you score to solicit any feedback from that group?

That doesn’t necessarily have to block merging, but I imagine we would want feedback before v17 is released

**Comment by logaretm** [comment]
@yaacovCR I pushed 1fcccd82436e3a87b768e4cb197d401a95db4477, is that what you had in mind?

> I remembered only recently that we have a graphql otel wg [...]

It could be foundational to theirs, they can use the same channels here to power the OTEL instrumentation as many others in the ecosystem are doing. Happy to talk to them or help them out.

**Comment by logaretm** [comment]
@yaacovCR Friendly ping, anything missing to push this further?

**Comment by yaacovCR** [comment]
Just exploring some additional work I’m doing in other areas and hoping it doesn’t require changing the shape of anything here, so far all good as far as I can tell.

If you are able to make next otel wg meeting maybe they might have feedback

https://github.com/graphql/otel-wg/blob/main/agendas/2026/05-May/28-otel-wg-may-2026.md

**Comment by logaretm** [comment]
Glad to hear it, will attend the next meeting.

File: src/diagnostics.ts [hunk 2]
```diff
@@ -0,0 +1,328 @@
...
+ * already verified that a subscriber is attached.
+ *
+ * @internal
+ */
+export function traceMixed<T, TContext = unknown>(
+  channel: MinimalTracingChannel<TContext>,
+  contextInput: TContext extends object ? TContext : object,
+  fn: () => T | Promise<T>,
+): T | Promise<T> {
+  const context = contextInput as TContext & {
```
PascalSenn:
```suggestion
export function traceMixed<T, TContext = unknown, TResult =  T | Promise<T>>(
  channel: MinimalTracingChannel<TContext>,
  contextInput: TContext extends object ? TContext : object,
  fn: () => TResult,
): TResult {
  const context = contextInput as TContext & {
```

logaretm: Thanks for the suggestion, this version would be more accurate and preserve the caller's result type in call sites. I Applied it.

logaretm:
@PascalSenn I have reviewed the [OTel WG semantic conventions](https://github.com/open-telemetry/semantic-conventions/pull/3515/changes) and addressed a couple of gaps that were left out.

### OTel Alignment

I believe now it is fully aligned, with minor exceptions that are outside of `graphql-js` responsibilities.

  | OTel span (`graphql.processing.type`) | graphql-js channel | Status |
  |---|---|---|
  | `graphql.server` (`request`) | N/A |  N/A (Usually done by the HTTP instrumentation) |
  | `graphql.document.parsing` (`parse`) | `graphql:parse` | ✅ |
  | `graphql.document.validation` (`validate`) | `graphql:validate` | ✅ |
  | `graphql.document.variable_coercion` (`variable_coercion`) | `graphql:execute:variableCoercion` | ✅ |
  | `graphql.operation.planning` (`plan`) | N/A | N/A (plan-based engines) |
  | `graphql.operation.execution` (`execute`) | `graphql:execute` / `graphql:execute:rootSelectionSet` | ✅ |
  | `graphql.step.execution` (`step_execute`) | N/A | N/A (plan-based engines) |
  | `graphql.field.execution` (`resolve`) | `graphql:resolve` | ✅ |
  | `graphql.subscription.event` (`subscription_event`) | `graphql:execute:rootSelectionSet` (fires once per emitted event) | ✅ |
  | `graphql.dataloader.*` | N/A | N/A (No a `graphql-js` concern) |
  | `graphql.client` (CLIENT) | N/A | N/A (Client-side concern, unrelated to this work) |

For subscriptions, `graphql:execute:rootSelectionSet` maps cleanly onto `subscription.event` (one per event); for queries/mutations it nests under `graphql:execute` as the execution phase. A subscriber distinguishes the two via `operationType`.

All recommended attributes are published directly or trivially derivable:

| OTel attribute | Source |
  |---|---|
  | `graphql.operation.name` / `graphql.operation.type` | published on execute / rootSelectionSet / variableCoercion / subscribe |
  | `graphql.document.hash` / `graphql.document.id` | derive from published `source` / `document` |
  | `graphql.field.name`, `graphql.field.parent_type` | `fieldName`, `parentType` |
  | `graphql.field.schema_coordinate` (`{Parent}.{field}`) | derive from the two above |
  | `graphql.field.alias` | `alias` **(added)** |
  | `graphql.field.path` | `fieldPath` (already uses response keys/aliases, per spec) |
  | `error.type` / `graphql.error.*` (message, locations, path, code) | derive from published `result.errors` / `error` |

### Benchmarks

I Re-ran the benchmark suite comparing local (with change) vs HEAD, I'm benchmarking against the base when there are _no subscribers_.

  | Benchmark | Δ ops/sec | 95% CI |
  |---|---|---|
  | Introspection | -0.55% | -1.62% to +0.54% |
  | Sync list | +0.27% | -0.78% to +1.33% |
  | Async list | -2.20% | -4.15% to -0.21% |
  | Async iterable | +0.08% | -1.79% to +1.99% |
  | Object async | +0.48% | -2.00% to +3.03% |

yaacovCR:
@logaretm

1. rebased on 17.x.x
2. stored `document` along with `rawVariableValues` on `ValidatedExecutionArgs` so that `execute:rootSelectionSet` has access to it.
3. dropped node 20.x support
4. added some coverage
5. pushes a cosmetic TS change

What do you think, anything else you want to round out?

logaretm:
@yaacovCR Nope, pretty happy with this!

yaacovCR:
@logaretm one last tweak above in [015b959](https://github.com/graphql/graphql-js/pull/4670/commits/015b95992dc6d3a9649ef421fd54b35a53edd06c) in direction of your earlier hoisting of `shouldTrace` for `graphql:resolve` to `executeFields`.

I didn't realize when first reviewing this that hoisting in this way should not affect correctness at all, as the value of `shouldTrace` should not change during this call, so I went back to it. For `executeFieldsSerially`, the value could change, because we are doing potentially async serial execution, so I call `shouldTrace` again along that async boundary.

And added a test.

Thoughts?

With this, I'm good to go, from my end.

Next thing would be to add an article to the website discussing it. One central piece for the article would be an updated version of the table above with the channels and context?

The website is published (still) from the 16.x.x branch, but covers both 16.x.x and 17.x.x and we would need to update the What's new in v17 guide perhaps as well. This would also belong in our growing list of "graphql-js runtime features."

Would you like to take a stab at it?

logaretm:
@yaacovCR 

> I didn't realize when first reviewing this that hoisting in this way [...]

Yep that should be safe, no APM/user should expect to subscribe mid-call or during async continuation ticks as they will get busted async contexts anyways

> Next thing would be to add an article to the website discussing it [...]

Yea, I'm happy to. I will create a new PR against `16.x.x` with the website/docs updates. Sounds good?

yaacovCR:
@logaretm 

sounds great!

some more changes for your review

1. preserve ancient option of resolvers to return Promise/A+ thenables without `finally`
2. rename variableValues => rawVariableValues throughout
3. preserve type of rawVariableValues as `{ [value]: unknown} | null | undefined` rather than just `{ [value]: unknown} | undefined`

yaacovCR:
And here is my update of the overall PR description for use in the merge commit:

This PR implements tracing channel support for `graphql-js`.

graphql-js resolves `node:diagnostics_channel` at module load through `process.getBuiltinModule`. On runtimes where the built-in is unavailable, such as browsers, channel handles are `undefined` and emission sites short-circuit. APMs subscribe through their own `node:diagnostics_channel` import by channel name; no GraphQL API call or option is required.

The implementation adds the diagnostics channel helper, internal structural typings for the Node tracing-channel surface, subscriber-facing context types exported from the root package, and wrappers around parse, validate, execute, subscribe, root-selection-set execution, variable coercion, and resolver execution. The no-subscriber path is guarded by `shouldTrace`/`hasSubscribers` checks, with a sub-channel fallback for runtimes that do not expose aggregate `TracingChannel.hasSubscribers`. Traced execution uses `runStores` to preserve tracing-channel async context propagation, does not normalize synchronous execution to async, and handles promise-like resolver results without `Promise.resolve` normalization.

Test coverage exercises lifecycle emission, lazy context fields, no-subscriber no-op behavior, missing `node:diagnostics_channel` behavior, exported context types, promise-like resolver results, and runtime compatibility coverage for Bun and Deno.

### API

graphql-js vendors a minimal structural subset of the tracing channel types for internal use so it does not depend on Node typings or on `node:diagnostics_channel` being present at runtime. It also exports subscriber context types from the root package, including `GraphQLChannelContextByName` and the per-channel context interfaces.

Each channel publishes lifecycle events through Node tracing channels. The table lists the base context fields present at `start`; successful sync work adds `result` by `end`, successful async work adds `result` by `asyncEnd`, and thrown or rejected work adds `error` on the `error` lifecycle. Variable coercion is synchronous, so it only emits `start`/`end` and `error` when coercion throws abruptly.

| Channel | Context fields |
|---|---|
| `graphql:parse` | `source` |
| `graphql:validate` | `schema`, `document` |
| `graphql:execute` | `schema`, `document`, `rawVariableValues`, `operationName` *(lazy)*, `operationType` *(lazy)* |
| `graphql:execute:variableCoercion` | `schema`, `document`, `operation`, `rawVariableValues`, `operationName`, `operationType` |
| `graphql:execute:rootSelectionSet` | `schema`, `document`, `operation`, `rawVariableValues`, `operationName`, `operationType` |
| `graphql:subscribe` | `schema`, `document`, `rawVariableValues`, `operationName` *(lazy)*, `operationType` *(lazy)* |
| `graphql:resolve` | `fieldName`, `alias`, `parentType`, `fieldType`, `args`, `isDefaultResolver`, `fieldPath` *(lazy)* |

`graphql:execute:variableCoercion` publishes the caller-provided values as `rawVariableValues`; its successful `result` contains the coerced `{ variableValues }`, or `{ errors }` when coercion returns request errors.

`graphql:execute:rootSelectionSet` fires once for the operation root selection set. For subscriptions, it fires once per emitted subscription event, matching the OTel `graphql.subscription.event` shape; subscribers can distinguish operation kinds through `operationType`.

### Usage

```ts
import dc from 'node:diagnostics_channel';
import type { GraphQLResolveContext } from 'graphql';

dc.tracingChannel('graphql:resolve').subscribe({
  start: (context: GraphQLResolveContext) => {
    // start span
  },
  end: (context: GraphQLResolveContext) => {
    // sync resolver result is available here as context.result
  },
  asyncEnd: (context: GraphQLResolveContext) => {
    // async resolver result is available here as context.result
  },
  error: (context: GraphQLResolveContext) => {
    // report context.error
  },
});
```

Closes #4629

logaretm:
Changes look good, especially the promise-like preservation. I updated the PR description and will work on the docs/website PR and submit it tomorrow morning.

PR #4692, opened 2026-04-26
fix incremental label null validation

jbellenger:
DeferStreamDirectiveLabelRule rejects directives whose label argument is a `null` literal, resulting in unexpected validation errors.

Example:
```graphql
# query
{ ... @defer(label:null) { __typename } }

# response
{
  "errors": [
    {
      "message": "Argument \"@defer(label:)\" must be a static string.",
      "locations": [
        {
          "line": 1,
          "column": 7
        }
      ]
    }
  ]
}
```
More failures can be found at [graphql-conformance](https://jbellenger.github.io/graphql-conformance/#/runs/79d8d08e-da02-4013-9dc2-8be1e9804b5f/impl/graphql-js-17)

These values are allowed by the spec, which defines these arguments as a nullable String. 

This PR updates the validator to exit early when it encounters a null literal.

### DIFF
--- a/src/execution/incremental/__tests__/defer-test.ts
+++ b/src/execution/incremental/__tests__/defer-test.ts
@@ -260,6 +260,76 @@ describe('Execute: defer directive', () => {
       },
     ]);
   });
+  it('Returns label from defer directive', async () => {
+    const document = parse(`
+      query HeroNameQuery {
+        hero {
+          id
+          ...NameFragment @defer(label: "defer-label")
+        }
+      }
+      fragment NameFragment on Hero {
+        name
+      }
+    `);
+    const result = await complete(document);
+
+    expectJSON(result).toDeepEqual([
+      {
+        data: {
+          hero: {
+            id: '1',
+          },
+        },
+        pending: [{ id: '0', path: ['hero'], label: 'defer-label' }],
+        hasNext: true,
+      },
+      {
+        incremental: [
+          {
+            data: {
+              name: 'Luke',
+            },
+            id: '0',
+          },
+        ],
+        completed: [{ id: '0' }],
+        hasNext: false,
+      },
+    ]);
+  });
+  it('Treats null defer label the same as no label', async () => {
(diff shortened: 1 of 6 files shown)


yaacovCR:
@jbellenger 

Nice catch! I think there is an inconsistency between the spec and implementation, but I would just tweak your wording a bit. 

Your wording suggests that we MUST not forbid a label from having the `null` value because it is nullable by type. But in the validation phase, we can enforce additional rules, and it may indeed make sense to prohibit this value, as the purpose of the label is to uniquely label a defer/stream within the response, and claiming the `null` label is  a bit rude/uninformative, we are hoping to encourage labels to be not just unique, but unique by design, descriptively unique.

So we could prohibit the explicit `null` and maybe we should, but still have the type be nullable, which is useful because in GraphQL (unfortunately?), there is no distinction between optional/nullable, so in order to not require this argument, by the type system alone we have to allow `null`, although we may in a validation rule prohibit it. At least that's my take!

But you are correct anyways on two other important levels:
1. The code as currently written assumes that if the value is present, it is either of `Kind.VARIABLE` or `Kind.STRING` and so only checks for the latter, so is confusing `Kind.NULL` for `Kind.VARIABLE` and giving the wrong error message about static/dynamic.
2. [The spec changes](https://github.com/graphql/graphql-spec/pull/1223/changes#diff-607ee7e6b71821eecadde7d92451b978e8a75e23d596150950799dc5f8afa43eR1786) do not prohibit an explicit `null`.

So, this has to be fixed both in the implementation and I think should also be clarified in the spec!

Tagging @robrichard

robrichard (approved):
I think this change make sense. Since there is no distinction between optional/nullable I think we should treat a null literal the same as if the label is not present. ~And this aligns with the current spec proposal so no change is necessary there.~ I will also update the spec to ensure null is treated the same as the label argument not being present.

yaacovCR:
Thanks @jbellenger !!!!

PR #4699, opened 2026-04-29
Validate SDL directive argument values

### LINKED ISSUE #3912: Missing validation of schema directive argument types
It seems there is no validation for schema directive argument types.

This makes the `kind` of the directive end up with the wrong type. Notably, if the type doesn't exist, it ends up as `EnumValue` as this seems to be the default type.

This is different from field argument types which correctly throw an error message if you pass the wrong type.

## Steps to reproduce

1. Run with deno, e.g. open `deno repl` and paste in, or run from file with `deno run file.ts`

```
import { buildSchema } from "npm:graphql@16.6.0";

const schemaSource = `
  type Query {
    # notice the missing quotes around the string!
    age: Int @foo(bar: FOOBAR)
  }

  directive @foo(
    bar: String!
  ) on FIELD_DEFINITION
`;

const schema = buildSchema(schemaSource);
// expected throw

const kind = schema.getQueryType()!.getFields()["age"].astNode!.directives![0].arguments![0].value.kind;

console.log(kind);
// EnumValue
```

## Expected result

The `buildSchema` function throws.

## Actual result

The `buildSchema` doesn't throw.

afurm:
## Summary
- Add SDL validation for directive argument value types, including custom directive usages.
- Reuse `ValuesOfCorrectTypeRule` for SDL directive nodes through a directive-aware validation context.
- Add coverage for `validateSDL` and `buildSchema` rejecting invalid directive argument values.

## Why
Fixes #3912.

SDL already rejects invalid values for built-in directives, but custom directive usages could accept invalid argument values such as `@foo(bar: FOOBAR)` when `bar` is declared as `String!`.

## Validation
- `npm run lint`
- `npm run check`
- `npm run testonly`
- `npm run prettier:check`
- `git diff --check`

Full `npm test` was attempted, but the Docker-backed supported-node integration could not run locally because the Docker daemon was unavailable.

### DIFF
--- a/src/utilities/__tests__/buildASTSchema-test.ts
+++ b/src/utilities/__tests__/buildASTSchema-test.ts
@@ -1078,6 +1078,19 @@ describe('Schema Builder', () => {
     expect(() => buildSchema(sdl)).to.throw('Unknown directive "@unknown".');
   });
 
+  it('Rejects SDL with invalid directive argument values', () => {
+    const sdl = `
+      type Query {
+        foo: String @test(arg: UNQUOTED_STRING)
+      }
+
+      directive @test(arg: String!) on FIELD_DEFINITION
+    `;
+    expect(() => buildSchema(sdl)).to.throw(
+      'String cannot represent a non string value: UNQUOTED_STRING',
+    );
+  });
+
   it('Allows to disable SDL validation', () => {
     const sdl = `
       type Query {
--- a/src/validation/__tests__/ValuesOfCorrectTypeRule-test.ts
+++ b/src/validation/__tests__/ValuesOfCorrectTypeRule-test.ts
@@ -11,10 +11,16 @@ import { GraphQLObjectType, GraphQLScalarType } from '../../type/definition';
 import { GraphQLString } from '../../type/scalars';
 import { GraphQLSchema } from '../../type/schema';
 
-import { ValuesOfCorrectTypeRule } from '../rules/ValuesOfCorrectTypeRule';
+import { buildSchema } from '../../utilities/buildASTSchema';
+
+import {
+  ValuesOfCorrectTypeOnDirectivesRule,
+  ValuesOfCorrectTypeRule,
+} from '../rules/ValuesOfCorrectTypeRule';
 import { validate } from '../validate';
 
 import {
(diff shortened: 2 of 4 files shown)


yaacovCR:
Hi @afurm !

Thanks for submitting this PR.

You wrote:

> SDL already rejects invalid values for built-in directives, but custom directive usages could accept invalid argument values such as `@foo(bar: FOOBAR)` when `bar` is declared as `String!`.

I would clarify that prior to this PR:
1. During validation of operations, directives included within those operations have their arguments via `ValuesOfCorrectTypesRule`.
2. Directive definitions within schemas, including custom schemas, have their ~~arguments~~ default values for arguments checked post-schema construction (when type information from the schema is available) by`validateSchema()`.
3. The built-in schema directives `@deprecated` and `@specifiedBy` coerce their arguments during schema construction, and should give well-formatted errors if this coercion fails. Even if this could be ~~extended to for custom directives~~ done during or post-construction for custom directives, it's not exactly the gap you're filling here as...

...your PR fills an existing gap, for both built in and custom directives meant for use on schema SDL, there is no SDL Validation Rule that forbids invalid arguments, because the typed schema is not yet available. It seems like you cleverly built a schema subset to get around this? Can you describe some more the method you use, and whether there are any dangerous of a potentially invalid schema subset being used by this rule, etc.

Thanks!

yaacovCR:
Correction: directive default values are audited by `validateSchema` but there is still no `ValuesOfCorrectTypeRule` yet for directives used within schema SDL, except for the built-in-directives that will throw during schema construction. Tried to correct above as well. Either way, this PR goes beyond just "any validation" moving the stage to the earliest stage possible, the SDL.

yaacovCR:
Compare to #4332 which just assumes we have a valid schema, and uses that with `validate` (normally used for operations) and the `ValuesOfCorrectType` rule to validate the directive arguments. The slowest alternative to this PR #4699 would do something similar, call `buildSchema` on the SDL, call `validateSchema` to make sure the schema is valid, and then call `validate` on the SDL with the schema. Presumably this PR is about paring that down to the minimum necessary, but we some discussion on how that can be/is done safely would be helpful.

PR #4705, opened 2026-05-01
Fix issue with recursive input object definitions in SDL

### LINKED ISSUE #3871: `RangeError` for combination of recursive input and default value
Hi, I was working with input fields' default values in [graphql-core](https://github.com/graphql-python/graphql-core/issues/196) and I encountered a problem with a combination of recursive inputs and default values which also occurs in `graphql-js`.
Let's take this SDL as an example:
```gql
"""Example 1"""
type Query { field(recursiveInput: RecursiveInput): String }

input RecursiveInput {
  self: RecursiveInput = {value: "foo"}
  value: String 
}
```
If I understand correctly that example is invalid, because `{value: "foo"}` is lacking the `self` field, then the default value for self is looked up, which leads to a recursion error - `RangeError: Maximum call stack size exceeded`.  

When the default value is removed, everything works fine:
```gql
"""Example 2"""
type Query { field(recursiveInput: RecursiveInput): String }

input RecursiveInput {
  self: RecursiveInput
  value: String 
}
```

However, the following also looks fine, because `self` doesn't need to be looked up, but it still causes a recursion error:
```gql
"""Example 3"""
type Query { field(recursiveInput: RecursiveInput): String }

input RecursiveInput {
  self: RecursiveInput = {value: "foo", self: null}
  value: String 
}
```

On the other hand, the following actually invalid example does not give an error:
```gql
"""Example 4"""
type Query { field(recursiveInput: RecursiveInput): String }

input RecursiveInput {
  self: RecursiveInput!
  value: String 
}
```

If I'm not mistaken:
- Example 1 should produce a proper error instead of creating a stack overflow
- Example 3 should not produce an error
- Example 4 should produce an error showing that this is an invalid definition

To test all examples I used `graphql@16.6.0` and following code:
```js
const { buildSchema } = require('graphql');
const schemaString = `...`;
const schema = buildSchema(schemaString);
```

benjie:
Prior to this, the following would throw "Maximum call stack size exceeded":

```ts
import { buildSchema } from './npmDist/index.mjs';
const sdl = `
type Query {
  depth(person: PersonInput): Int
}

input PersonInput {
  parent: PersonInput = {parent: null}
}
`;

buildSchema(sdl);
```

The equivalent in native code would not:

```ts
import {
  GraphQLSchema,
  GraphQLInt,
  GraphQLObjectType,
  GraphQLInputObjectType,
  printSchema,
} from './npmDist/index.mjs';

const PersonInput = new GraphQLInputObjectType({
  name: 'PersonInput',
  fields: () => ({
    parent: {
      type: PersonInput,
      defaultValue: { parent: null },
    },
  }),
});

    name: 'Query',
    fields: {
      depth: {
        args: {
          person: {
            type: PersonInput,
          },
        },
        type: GraphQLInt,
      },
    },
  }),
});

console.log(printSchema(schema));
```

This means if you `printSchema` on a valid schema and try to rebuild it from SDL, the rebuild _might_ fail (it will for the above schema). The reason being somewhat related to https://rfcs.graphql.org/rfcs/793/ - in code we treat the `defaultValue` as coerced, but in SDL we do the coercion of the default value ourselves, which requires us to know the definition of the types, which requires us to know the default value, which requires us to know the types, which [...infinite recursion...]

v17 solves this properly but required a breaking change. To address the issue in v16 I've made it so that the `fields()` thunk will not allow recursion (it'll return the currently-being-resolved value instead), and I've written everything except the defaultValue to the being-resolved object first, and finally I follow up with a "apply default value" pass which then adds the default values in.

I could not come up with SDL that constructs in the current v16 (i.e. doesn't throw the max depth error) that this would cause a divergence in behavior for, so I _think_ it's safe, but I could do with another pair of eyes.

- Fixes #3871

### DIFF
--- a/src/utilities/__tests__/buildASTSchema-test.ts
+++ b/src/utilities/__tests__/buildASTSchema-test.ts
@@ -509,6 +509,26 @@ describe('Schema Builder', () => {
     expect(cycleSDL(sdl)).to.equal(sdl);
   });
 
+  it('builds recursive input object field defaults', () => {
+    const sdl = dedent`
+      type Query {
+        depth(person: PersonInput): Int
+      }
+
+      input PersonInput {
+        parent: PersonInput = {parent: {parent: {parent: null}}}
+      }
+    `;
+
+    const schema = buildSchema(sdl);
+    const personInput = assertInputObjectType(schema.getType('PersonInput'));
+
+    expect(personInput.getFields().parent.defaultValue).to.deep.equal({
+      parent: { parent: { parent: null } },
+    });
+    expect(printSchema(schema)).to.equal(sdl);
+  });
+
   it('Simple argument field with default', () => {
     const sdl = dedent`
       type Query {
--- a/src/utilities/__tests__/extendSchema-test.ts
+++ b/src/utilities/__tests__/extendSchema-test.ts
@@ -271,6 +271,38 @@ describe('extendSchema', () => {
     `);
   });
 
+  it('extends inputs with recursive field defaults', () => {
+    const schema = buildSchema(`
+      type Query {
+        someInput(arg: PersonInput): String
+      }
+
+      input PersonInput {
+        parent: PersonInput
+      }
+    `);
+    const extensionSDL = dedent`
+      extend input PersonInput {
+        ancestor: PersonInput = {parent: { parent: { parent: null } } }
+      }
+    `;
+    const extendedSchema = extendSchema(schema, parse(extensionSDL));
+    const personInput = assertInputObjectType(
+      extendedSchema.getType('PersonInput'),
+    );
+
+    expect(personInput.getFields().ancestor.defaultValue).to.deep.equal({
+      parent: { parent: { parent: null } },
+    });
+    expect(validateSchema(extendedSchema)).to.deep.equal([]);
+    expectSchemaChanges(schema, extendedSchema).to.equal(dedent`
+      input PersonInput {
+        parent: PersonInput
+        ancestor: PersonInput = {parent: {parent: {parent: null}}}
+      }
+    `);
+  });
+
   it('extends scalars by adding new directives', () => {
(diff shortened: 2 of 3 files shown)


benjie:
Hmmm... no it's not safe because it can lead to unexpected outputs:

```ts
import { buildSchema, printSchema } from './npmDist/index.mjs';

const sdl = `
type Query {
  field(in: A): Int
}

input A {
  b: B = {}
  label: String = "A"
}

input B {
  a: A = {b: null}
  tail: String = "B"
}
`;

const schema = buildSchema(sdl);
console.log(printSchema(schema));
```

Produces:

```graphql
type Query {
  field(in: A): Int
}

input A {
  b: B = {a: {b: null}, tail: "B"}
  label: String = "A"
}

input B {
  a: A = {b: null}
  tail: String = "B"
}
```

I.e. the default value for `B.a` does not include the expected `label: "A"`

Worse when you start adding non-nulls things get even weirder:

```ts
import { buildSchema, printSchema } from './npmDist/index.mjs';
const sdl = `
type Query {
  field(in: A): Int
}

input A {
  b: B = {}
  label: String! = "A"
}

input B {
  a: A! = {b: null}
  tail: String = "B"
}
`;

const schema = buildSchema(sdl);
console.log(printSchema(schema));
```

produces:

```graphql
type Query {
  field(in: A): Int
}

input A {
  b: B
  label: String! = "A"
}

input B {
  a: A!
  tail: String = "B"
}
```

The `defaultValue`s have gone missing! This is due to the `field.defaultValue !== undefined` check in `valueFromAST` - the default value for this hasn't been resolved yet.

benjie:
I think we're better off leaving the behavior as-is and waiting for v17 to fix it - we'll leave the current schema definitions as invalid.

(Might be worth turning the `Maximum call stack size exceeded` error into a pro-active error, but that's a separate concern.)

PR #4721, opened 2026-05-06
move to beta!

yaacovCR:
I think we still have to do this separately from the release commit.


yaacovCR:
integration tests fail because it looks like current tests are in a catch-22 here.

PR #4737, opened 2026-05-11
docs: refresh upgrade guide and v17 topics

yaacovCR:


website/pages/upgrade-guides/v16-v17.mdx
```diff
@@ -1,221 +1,552 @@
...
+to select the right build. Use public entry points such as `graphql`,
+`graphql/execution`, `graphql/language`, `graphql/type`, `graphql/utilities`,
+and `graphql/validation`.
 
-GraphQL.js development mode no longer depends on the `NODE_ENV` environment variable; build tools other than Node.js no longer need
-to replace this Node.js specific code. See [Development Mode](./development-mode) for further details regarding how to enable these
-checks in v17.
+Deep imports into GraphQL.js internals may still work in some environments, but
+they are not officially supported. Prefer the public entry points above for
+application and library code.
```
jaydenseric, line 65: What exactly are "GraphQL.js internals" and what is a "public entry point"? Technically, according to npm convention anything importable via the package `exports` field is part of the declared public API and subject to SemVer.

The wording "such as" implies the list mentioned is incomplete. It’s not specifically said if there is a concrete rule that all current first level deep imports without file extensions are the SemVer stable API?

Specifically, I need to know for work in `graphql-upload` if `import { graphql } from "graphql/graphql"` is SemVer stable, even tho `graphql/graphql` isn't explicitly mentioned here. It would be strange if most of the GraphQL.js API has deep imports, but `graphql` is only available from the package main index (i.e. `import { graphql } from "graphql"`).

yaacovCR: We consider the public api to be anything exported from a barrel file either from the main graphql import or the individual submodules. The entire package is available via deep imports, however, and we do not plan on changing the availability of any of these deeper exports, and certainly not the internals that are part of the public API.

jaydenseric: Thanks for the quick reply! So you're saying, unless the package export maps to a file with `index` in the name, it's not part of the GraphQL.js SemVer stable API? In that case, `graphql/graphql` should not be used because it maps to [graphql/graphql.js](https://unpkg.com/graphql@17.0.2/graphql.js).

jaydenseric: Because in IDE's you can import from a lot of `graphql/*` modules without a file extension, it's pretty hard for users to tell which under the hood map to an index module and which map to modules named something else; they have to literally reverse engineer the package source code to see what is the public SemVer stable API. It would be good to update the wording to have a concrete bullet list of what import paths are stable, and also have the same documentation about the public API in the `graphql` npm package readme.

PR #4740, opened 2026-05-14
refactor: export enums as module namespaces

yaacovCR:
TypeKind and DirectiveLocation now use the same representation as Kind: each enum member is a standalone ESM export with a matching type export, re-exported through a module namespace.

This preserves the public TypeKind.SCALAR and DirectiveLocation.FIELD access pattern while avoiding a single object literal that makes all members part of the same runtime value. For example, TypeKind.SCALAR now resolves through the SCALAR named export from typeKind_.ts, which gives bundlers a smaller unit to analyze and tree-shake.

BREAKING CHANGE: TypeKind and DirectiveLocation are now module namespace objects rather than plain object literals.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -86,8 +86,6 @@ export {
   GraphQLDeprecatedDirective,
   GraphQLSpecifiedByDirective,
   GraphQLOneOfDirective,
-  // "Enum" of Type Kinds
-  TypeKind,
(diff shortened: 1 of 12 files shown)


yaacovCR:
The optimization was originally specifically requested just for `Kind` as this was what was mostly used on the client. We'll keep it that way for now!

PR #4767, opened 2026-05-23
docs: fix Post_body example variable

### LINKED ISSUE #4765: docs: `Post_body` example in production guide uses undefined `obj` variable
## Description

The example in `website/pages/docs/going-to-production.mdx` under the authentication/authorization section defines the resolver as `function Post_body(source, args, context, info)`, but the body passes `post: obj` into `postRepository.getBody(...)`.

Because `obj` is not defined in the snippet, copying the example throws immediately. It looks like the intended variable is `source`.

## Current snippet

```js
function Post_body(source, args, context, info) {
  // return the post body only if the user is the post's author
  return postRepository.getBody({ user: context.user, post: obj })
}
```

## Expected snippet

```js
function Post_body(source, args, context, info) {
  // return the post body only if the user is the post's author
  return postRepository.getBody({ user: context.user, post: source })
}
```

## History

This appears to have been introduced in PR #4398 (`docs: update "going to production" guide`).

fallintoplace:
## Summary

- replace the undefined `obj` variable with `source` in the `Post_body` resolver example
- keep the change scoped to the production guide docs example only

## Why

The current snippet defines `function Post_body(source, args, context, info)` but passes `post: obj`, which is undefined and causes the copied example to fail immediately.

Closes #4765.

## Validation

- `npx --yes prettier@2.6.2 --check website/pages/docs/going-to-production.mdx`
- `git diff --check`

### DIFF
--- a/website/pages/docs/going-to-production.mdx
+++ b/website/pages/docs/going-to-production.mdx
@@ -149,7 +149,7 @@ const postRepository = {
 // Resolver for the `Post.body` field:
 function Post_body(source, args, context, info) {
   // return the post body only if the user is the post's author
-  return postRepository.getBody({ user: context.user, post: obj })
+  return postRepository.getBody({ user: context.user, post: source })
 }
 ```
 


yaacovCR:
Thanks for this fix!

PR #4784, opened 2026-06-02
fix: do not copy over stack from error cause

Malien:
I've authored #4485 long time ago. Since that PR sat there for a while, I didn't get a chance to amend the design to not copy over the stack from the `cause`. 

Copying over the stack will just double up the same stack trace twice in the reporting tooling / console logs.


yaacovCR:
@Malien fantastic!

Thanks for the original PR, and thanks for these corrections. I pushed a few cosmetic test changes, looks ok to you?

Malien:
@yaacovCR Yeah. Looks great.

PR #4798, opened 2026-06-04
feat: add validateFullSchema and assertValidFullSchema

### LINKED ISSUE #4415: Allow to validate a full schema
GraphQL JS has a function to validate a user schema, [`validateSchema()`](https://github.com/graphql/graphql-js/blob/b0f467d6d00da71d182e50280ac180b694f1cd63/src/type/validate.ts#L75).

This works well if the schema is a user schema not containing any of the builtin types.

In some cases though, we might want to validate a schema coming from introspection. This is typically the case in GraphiQL but other clients might also want to do this as a sanity check before running code generation or anything else.

In the case of a remote service, there is a high chance that the results of introspection do not match the version of GraphQL JS being used (because the tools haven't been updated).

In those cases, validation may fail for otherwise valid schemas.

Reproducer:

```ts
// This is valid SDL returned from an hypothetical newer version of GraphQL
const typeDefs = `
    type Query {
      hello: String
    }
    
    enum __ErrorBehaviour {
      NULL
      PROPAGATE
      HALT
    }

    directive @behavior(onError: __ErrorBehaviour) on SCHEMA
`
let schema = buildSchema(typeDefs)
let errors = validateSchema(schema)
```

Output:

```
  GraphQLError: Name "__ErrorBehaviour" must not begin with "__", which is reserved by GraphQL introspection.
      at SchemaValidationContext.reportError (/Users/martinbonnin/git/graphql-js-playground/node_modules/graphql/type/validate.js:73:7)
      at validateName (/Users/martinbonnin/git/graphql-js-playground/node_modules/graphql/type/validate.js:216:13)
      at validateTypes (/Users/martinbonnin/git/graphql-js-playground/node_modules/graphql/type/validate.js:239:7)
      at validateSchema (/Users/martinbonnin/git/graphql-js-playground/node_modules/graphql/type/validate.js:43:3)
      at Object.<anonymous> (/Users/martinbonnin/git/graphql-js-playground/dist/index.js:18:43)
      at Module._compile (node:internal/modules/cjs/loader:1723:14)
      at Object..js (node:internal/modules/cjs/loader:1888:10)
      at Module.load (node:internal/modules/cjs/loader:1458:32)
      at Function._load (node:internal/modules/cjs/loader:1275:12)
      at TracingChannel.traceSync (node:diagnostics_channel:322:14) {
    path: undefined,
    locations: [ [Object] ],
    extensions: [Object: null prototype] {}
  }
```

It would be nice to have a `validateFullSchema()` that is able to deal with [Full Schemas](https://github.com/graphql/graphql-wg/blob/main/rfcs/FullSchemas.md), including built in types.

Context: https://github.com/graphql/graphiql/issues/3968

vishwakt:
## Motivation

Closes #4415.

`validateSchema()` validates a *source* schema and rejects any name beginning with `__`, since those are reserved by GraphQL introspection. However, tools that work with a *full* schema (for example one reconstructed from an introspection result, as in GraphiQL or codegen) may legitimately encounter reserved names, including ones added by a newer version of GraphQL than the running version of graphql-js implements.

The reproducer from the issue fails today purely on the reserved-name rule, even though the SDL is otherwise valid:

```ts
const schema = buildSchema(`
  type Query { hello: String }

  enum __ErrorBehavior { NULL PROPAGATE HALT }

  directive @behavior(onError: __ErrorBehavior) on SCHEMA
`);

validateSchema(schema);
// GraphQLError: Name "__ErrorBehavior" must not begin with "__", which is reserved by GraphQL introspection.
```

See the [Full Schemas RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/FullSchemas.md) for the broader context on source vs. full schemas.

## Changes

- Add `experimentalValidateFullSchema(schema)` and `experimentalAssertValidFullSchema(schema)`, mirroring `validateSchema` / `assertValidSchema`.
- They run every existing type-system check **except** the reserved-name (`__`) restriction, so a full schema validates without spurious errors.
- Internally, both modes share the existing validation pipeline via a single `allowReservedNames` flag threaded through `SchemaValidationContext`.
- The two modes use independent validation caches (`__validationErrors` vs. `__fullSchemaValidationErrors`) so their results never collide regardless of call order.
- The functions use the `experimental` name prefix (matching `experimentalExecuteIncrementally`) to signal the API may change while behavior and test coverage for full schemas converge.

## Tests

Added coverage for: the issue's reproducer, reserved names on types/fields/args/enum values/input fields, that non-name errors are still reported when validating a full schema, independent caching, and the assert variant.

`npm run check`, `npm run lint`, and the full test suite pass locally.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -172,8 +172,8 @@ export {
   // Validate GraphQL schema.
   validateSchema,
   assertValidSchema,
-  validateFullSchema,
-  assertValidFullSchema,
+  experimentalValidateFullSchema,
+  experimentalAssertValidFullSchema,
   // Upholds the spec rules about naming.
   assertName,
   assertEnumValueName,
--- a/src/index.ts
+++ b/src/index.ts
@@ -172,6 +172,8 @@ export {
   // Validate GraphQL schema.
   validateSchema,
   assertValidSchema,
+  validateFullSchema,
+  assertValidFullSchema,
   // Upholds the spec rules about naming.
   assertName,
   assertEnumValueName,
--- a/src/type/__tests__/validation-test.ts
+++ b/src/type/__tests__/validation-test.ts
@@ -41,9 +41,9 @@ import { assertDirective, GraphQLDirective } from '../directives.ts';
 import { GraphQLInt, GraphQLString } from '../scalars.ts';
 import { GraphQLSchema } from '../schema.ts';
 import {
-  assertValidFullSchema,
   assertValidSchema,
-  validateFullSchema,
+  experimentalAssertValidFullSchema,
+  experimentalValidateFullSchema,
   validateSchema,
 } from '../validate.ts';
 
@@ -3433,8 +3433,8 @@ describe('Type System: A full schema may contain reserved names', () => {
       },
     ]);
 
-    // validateFullSchema permits it.
-    expectJSON(validateFullSchema(schema)).toDeepEqual([]);
+    // experimentalValidateFullSchema permits it.
+    expectJSON(experimentalValidateFullSchema(schema)).toDeepEqual([]);
   });
 
   it('accepts reserved names on types, fields, args, enum values and input fields', () => {
@@ -3453,7 +3453,7 @@ describe('Type System: A full schema may contain reserved names', () => {
     );
 
     expect(validateSchema(schema)).to.not.deep.equal([]);
-    expectJSON(validateFullSchema(schema)).toDeepEqual([]);
+    expectJSON(experimentalValidateFullSchema(schema)).toDeepEqual([]);
   });
 
   it('still reports non-name errors when validating a full schema', () => {
@@ -3467,7 +3467,7 @@ describe('Type System: A full schema may contain reserved names', () => {
       }
     `);
 
-    expectJSON(validateFullSchema(schema)).toDeepEqual([
+    expectJSON(experimentalValidateFullSchema(schema)).toDeepEqual([
       {
         message: 'Query root type must be provided.',
       },
@@ -3491,18 +3491,18 @@ describe('Type System: A full schema may contain reserved names', () => {
 
     // Populate the regular cache first, then the full schema cache.
     const regularErrors = validateSchema(schema);
-    const fullErrors = validateFullSchema(schema);
+    const fullErrors = experimentalValidateFullSchema(schema);
 
     expect(regularErrors).to.have.lengthOf(1);
     expect(fullErrors).to.deep.equal([]);
 
     // Repeated calls return the cached results for each mode.
     expect(validateSchema(schema)).to.equal(regularErrors);
-    expect(validateFullSchema(schema)).to.equal(fullErrors);
+    expect(experimentalValidateFullSchema(schema)).to.equal(fullErrors);
   });
 });
 
-describe('assertValidFullSchema', () => {
+describe('experimentalAssertValidFullSchema', () => {
   it('does not throw on a full schema containing reserved names', () => {
     const schema = buildSchema(`
       type Query {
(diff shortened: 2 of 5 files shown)


martinbonnin (commented):
LGTM although not a graphql-js expert. 

Might be worth marking this function as experimental until we collectivelly converge on a bigger test suite?

vishwakt:
Good idea, thanks. I marked both functions as experimental by adopting the existing `experimental` name prefix convention (as used by `experimentalExecuteIncrementally` / `experimentalExecuteRootSelectionSet`), since that seems to be the established pattern in the repo for signalling an unstable API rather than a JSDoc tag. They're now `experimentalValidateFullSchema` and `experimentalAssertValidFullSchema`, and I added a note in the JSDoc too. Happy to revisit the names once the behavior and test coverage for full schemas converge.

martinbonnin (approved):
LGTM. Thanks for addressing the feedback. 

Pinging @yaacovCR @benjie @jerelmiller for more expert eyes.

benjie (changes requested):
I'm not sure this is the right solution. For example, the schema:

```graphql
scalar __S
type __Q {
  __i: __S
}
schema {
  query: __Q
}
```

should be invalid.

Similarly, the less ridiculous schema:

```graphql
type Query {
  int: __Type
}
```

should also be invalid.

I understand what we're trying to achieve here, but I don't think simply turning off the "no double underscore" validation rule is sufficient, there's a lot of assumptions baked into that rule that if we turn them off we'll need to re-apply other ways (e.g. "starting at the root operation types and without traversing or including metafields, no `__`-prefixed field, argument or type should be reachable").

I think we need to invest a lot of thought into what a "full schema" means for GraphQL.js

A simpler solution might be for GraphiQL to accept an option that tells it to assume the schema is valid.

vishwakt:
Thanks @benjie, that's a good catch and I agree the blanket approach was wrong. I've reworked it to implement the rule you described.

Instead of disabling the reserved-name check, a full schema now reports a reserved (`__`-prefixed) name only when it is **reachable from the root operation types through ordinary fields, arguments, and their types**. Because the introspection types (and other implementation-provided definitions) are only reachable via the `__schema` / `__type` / `__typename` meta-fields, which are never part of a type's `getFields()`, they are never traversed and remain valid.

Both of your examples are now rejected:

```graphql
scalar __S
type __Q { __i: __S }
schema { query: __Q }
```
```graphql
type Query { int: __Type }
```

while a reserved definition that is not part of the queryable surface (e.g. the `__ErrorBehavior` enum used only by a directive argument in the original issue, or an unreachable reserved type) is still accepted.

Added tests for both of your examples plus reachable reserved field/argument names, unreachable reserved definitions, and independent caching. This still doesn't settle the larger "what is a full schema" design question, but it should at least make the validation sound rather than simply permissive. Happy to keep iterating.

benjie:
In case you didn't already, you should also target all non-type-system directives (e.g. `  directive @behavior(onError: __ErrorBehavior) on SCHEMA` should be ignored, because SCHEMA is type system, but `directive @flibble(__foo: __Bar) on FIELD` should be rejected).

martinbonnin:
```graphql
type Query {
  int: __Type
}
```

I personally don't mind this being a valid schema. 
Sure it feels weird but if people want to reuse the introspection types, I don't see the harm in doing so. And if it can make the validation code samaller, it's even better.

vishwakt:
@benjie Done. The reachability pass now also covers directives usable in executable documents: if a directive's locations include an executable location (FIELD, QUERY, etc.), its name, argument names, and argument types are checked for reserved names. A directive limited to type-system locations may still reference reserved definitions, so `directive @behavior(onError: __ErrorBehavior) on SCHEMA` remains valid while `directive @flibble(__foo: __Bar) on FIELD` is rejected (both the `__foo` argument and the `__Bar` type). A mixed case like `on SCHEMA | FIELD` is treated as executable and rejected. Tests added for all three.

@martinbonnin On allowing `type Query { int: __Type }`: I kept the stricter behavior for now since it follows the rule Benjie outlined and his review is the blocking one, but relaxing it later would be a small change (the traversal would simply skip reporting known introspection types). If you two converge on allowing reuse of the built-in introspection types from the user surface, I'm happy to adjust.

PR #4805, opened 2026-06-09
docs: publish fixed extensions comments

yaacovCR:
:( this could have been rolled into #4802 as the website is now published from 17.x.x


yaacovCR:
> :( this could have been rolled into #4802 as the website is now published from 17.x.x

actually maybe it couldn't because the build script as of now reads from merged 17.x.x tip and not the local checkout.

...

PR #4824, opened 2026-06-16
fix(diagnostics): emit asyncStart after promise settlement to match native `tracePromise`

logaretm:
@yaacovCR This is a minor correction for `traceMixed` that was introduced in #4670 

Once v17 RC was released, I made a round of tests against the observability ecosystem. One thing that I noticed was that OTEL's proposed API for tracing channels in https://github.com/open-telemetry/opentelemetry-js/pull/6387 was mis-parenting async resolvers sometimes.

The reason for this was `asyncStart` was emitted when the promise was first returned, rather than when the promise settled. Emitting it early meant subscribers that temporarily activate context between asyncStart and asyncEnd could leak that context into overlapping sibling operations.

This did not show up in earlier tests because most subscribers only used `start` for context creation and `asyncEnd` for completion. This should not be a breaking change because the terminal `end/asyncEnd` semantics and result/error payloads remain the same. This is an invisible change for most users, and is only correcting a slight mismatch.

We had to manually implement the async lifecycle in `traceMixed` here because we needed to preserve the `sync/async` return paths in resolvers.

I discussed this with @qard and this should be a more faithful implementation to how `tracePromise` works under the hood timing-wise.

### DIFF
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -427,17 +427,18 @@ export function traceMixed<TResult, TContext extends TraceLifecycleContext>(
     }
 
     channel.end.publish(context);
-    channel.asyncStart.publish(context);
 
     return result.then(
       (value) => {
         context.result = value;
+        channel.asyncStart.publish(context);
         channel.asyncEnd.publish(context);
         return value;
       },
       (err: unknown) => {
         context.error = err;
         channel.error.publish(context);
+        channel.asyncStart.publish(context);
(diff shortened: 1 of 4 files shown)


yaacovCR:
@logaretm => agreed, this fix [matches the docs](https://nodejs.org/api/diagnostics_channel.html#asyncstartevent).

yaacovCR:
@logaretm released in 17.0.1 => thanks for catching this so quickly!

Issue #4830, opened 2026-06-28
apollo server cache control error with multiple versions of graphql appears to surface only in v17


DenyVeyten:
E.g. `.filter(isCompositeType).length` below always returns 0 because `Symbol('Object');` may be created twice at some point:
https://github.com/apollographql/apollo-server/blob/cfbfe71a936a749b68b802e6eb8e8477fe19d144/packages/server/src/plugin/cacheControl/index.ts#L80

Commit message here probably talks about the root cause https://github.com/graphql/graphql-js/commit/ffcfa3756b42cf892f9962480aaba0bef3004aa6

v16 does not have this issue

yaacovCR:
Hi, thanks for trying out v17. graphql-js in v17 as before does not support multiple versions of the package within a repository; in v17 we use symbol checks to manage this whereas previously we used `instanceof` but the hard requirement to use only a single version of the `graphql` package has not been lifted.

This is supposed to -- in development mode -- lead to a helpful error, and in production, to fail silently. Previously, development mode was managed by the `NODE_ENV` environment variable, and enabled by default unless `NODE_ENV` was set to `production`. In v17, development mode is disabled by default, but can be enabled by using the development entry points managed by the `development` condition for runtimes that support it, or enabled explicitly via the `enableDevMode` export.

You seem to be reporting a problem in production mode, with no helpful error message. Surprisingly, you write that this did not occur in v16, which is not in line of my understanding how things did/should work, is intriguing, and deserves investigation!

Are you able to include a minimal reproduction of your setup in a small github repository or other format?

Thanks again for raising this issue!

yaacovCR:
I tried to reproduce a successful case directly with v16, and was unable. My suspicion is that when using v16, you are using a single version of your setup, but when you switch to v17, you do so only partially, i.e. the "partial" switch to v17 is what is causing multiple conflicting versions of graphql.

I am going to close this issue for now, but will reopen immediately if you can provide an actionable minimal reproduction.

PR #4831, opened 2026-06-29
test: cover isCompositeType identity across package entrypoints

dosacha:
## Summary

Adds integration coverage for `isCompositeType` identity across GraphQL.js package entrypoints and module conditions.

This is related to #4830, where upstream code using:

```ts
Object.values(schema.getTypeMap()).filter(isCompositeType)
```

can observe no composite types if GraphQL type instances and predicates come from different evaluated `graphql` module instances.

## What changed

- Extended the conditional exports integration check to verify that:
  - `graphql` and `graphql/type` share the same `GraphQLObjectType`
  - `graphql` and `graphql/type` share the same `GraphQLString`
  - `graphql` and `graphql/type` share the same `isCompositeType`
  - ESM and CJS imports share the same `isCompositeType`
  - object types constructed through root, subpath, ESM, and CJS entrypoints are accepted by all `isCompositeType` predicates
- Added `require('graphql/type')` path checks alongside the existing `require('graphql')` path check.

## Why

GraphQL.js v17 type predicates rely on internal `__kind` brand symbols. If supported package entrypoints or module conditions accidentally evaluate separate GraphQL.js module instances, a type created by one entrypoint can fail a predicate imported from another entrypoint.

This test does not change duplicate-module semantics. It only ensures that the supported package entrypoints and CJS/ESM conditions preserve predicate identity for the same built package.

## Tests

Ran:

```sh
npm run node:test -- "src/type/__tests__/predicate-test.ts"
npm run node:test -- "src/jsutils/__tests__/instanceOf-test.ts"
npm run build:npm
npx prettier --check integrationTests/conditions/check.mjs integrationTests/conditions/cjs-importer.cjs
npx eslint --max-warnings 0 integrationTests/conditions/check.mjs integrationTests/conditions/cjs-importer.cjs
node --check integrationTests/conditions/check.mjs
node --check integrationTests/conditions/cjs-importer.cjs
```

Also ran a smoke check against the built npm package:

```text
Module identity, subpath identity, and path checks passed.
```

Not run:

```sh
npm run check:integrations
```

because it requires the full Docker-based integration test environment.

Refs #4830

### DIFF
--- a/integrationTests/conditions/check.mjs
+++ b/integrationTests/conditions/check.mjs
@@ -1,15 +1,40 @@
 import assert from 'node:assert';
 
-import { GraphQLObjectType as ESMGraphQLObjectType } from 'graphql';
+import {
+  GraphQLObjectType as ESMGraphQLObjectType,
+  GraphQLString as ESMGraphQLString,
+  isCompositeType as ESMIsCompositeType,
+} from 'graphql';
+import {
+  GraphQLObjectType as ESMTypeGraphQLObjectType,
+  GraphQLString as ESMTypeGraphQLString,
+  isCompositeType as ESMTypeIsCompositeType,
+} from 'graphql/type';
 
-import { CJSGraphQLObjectType, cjsPath } from './cjs-importer.cjs';
+import {
+  CJSGraphQLObjectType,
+  CJSGraphQLString,
+  CJSIsCompositeType,
+  cjsPath,
+  CJSTypeGraphQLObjectType,
+  CJSTypeGraphQLString,
+  CJSTypeIsCompositeType,
+  cjsTypePath,
+} from './cjs-importer.cjs';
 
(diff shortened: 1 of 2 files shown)


yaacovCR:
Hi, thanks for raising this PR and engaging with the repository. We appreciate it. In terms of this PR, I'm confused by #4830 => in the sense that although the internal mechanism of GraphQL schema element identify are now managed by symbols rather than `instanceof`, the underlying requirement to have a single `graphql` package in a given repo have not changed, see my comments there.

In terms of this PR, I do not think we need to add these tests. We need only a single test to prove we have imported the same package, which means our internal symbols are the same. Adding extra tests for just a single function does not do too much. Another way of thinking about this is that for completeness, we would then need to add every predicate.

Do you have any insight as to why #4830 apparently did not hit v16? That's where I'm at with this one, I have asked for a reproduction there.

Thanks again for contributing!

PR #4832, opened 2026-07-02
fix: detect default-value changes on input object fields

spokodev:
`findSchemaChanges`, `findDangerousChanges` and `findBreakingChanges` detect a default-value change on a field argument and on a directive argument, but not on an input object field. Changing the default value of an input-object field is a dangerous change: a client that omits that field gets different behavior after the change. Tooling and CI gates that rely on these functions (schema registries, the graphql-inspector action, custom "is this change safe" checks) ship such a change with no warning.

## Cause

`src/utilities/findSchemaChanges.ts`, `findInputObjectTypeChanges` compares the field type and description but never compares the default value, unlike the sibling `findArgChanges`. The gap is long-standing (v16 has the same omission).

## Fix

Compare the input-field default value the same way `findArgChanges` does (via `getDefaultValue`, which runs `sortValueNode` so object-field reordering is ignored), and report:

- `INPUT_FIELD_DEFAULT_VALUE_CHANGE` (dangerous) when a default is changed or removed,
- `INPUT_FIELD_DEFAULT_VALUE_ADDED` (safe) when a default is added.

These are new members alongside the existing `REQUIRED_INPUT_FIELD_ADDED` / `OPTIONAL_INPUT_FIELD_ADDED`, so an input-field change is reported under an input-field type rather than reusing the argument type.

Added a test in `findBreakingChanges-test.ts` covering a removed, changed, and list default. It fails before this change (the input-field changes are not reported); the `findBreakingChanges` and `findSchemaChanges` suites (55) stay green, and tsc, eslint and prettier pass.

### DIFF
--- a/src/utilities/__tests__/findBreakingChanges-test.ts
+++ b/src/utilities/__tests__/findBreakingChanges-test.ts
@@ -979,6 +979,55 @@ describe('findDangerousChanges', () => {
     ]);
   });
 
+  it('should detect if a defaultValue changed on an input field', () => {
+    const oldSDL = `
+      input InputType1 {
+        withDefaultValue: String = "TO BE DELETED"
+        stringField: String = "test"
+        emptyArray: [Int!] = []
+        withoutDefaultValue: String
+      }
+
+      type Query {
+        field1(arg: InputType1): String
+      }
+    `;
+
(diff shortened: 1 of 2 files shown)


yaacovCR:
Thanks so much for contributing. Will merge this as a bug-fix.

PR #4835, opened 2026-07-07
feat: introduce validateWithRules for complete document (sdl and/or executable) and/or existing schema validation

yaacovCR:
Previously, we could use `validateSDL` to validate type system AST against a limited set of SDL rules, `validate` when given a document against a schema to validate the document against the schema, or `validateSchema` to validate a built `GraphQLSchema` in a non rules-based manner.

This PR introduces `validateWithRules` that performs integrated rules-based validation on a document containing anything (executable definitions and/or SDL) and/or an existing built `GraphQLSchema`. The validation of a built schema is also divided up into rules.

By default, when `validateWithRules` is supplied both a document and a `GraphQLSchema`, existing schema-only errors are not reported; validation focuses on the document and any type-system definitions it contains. Set `includeExistingSchemaErrors: true` to also report errors from the existing schema, evaluated together with the document’s type-system definitions. When only a schema is passed, schema errors are always reported, and `includeExistingSchemaErrors: false` throws.

WIP.


yaacovCR:
image

HEAD~1 uses fallbacks to old workflows to "kind of" compare the new workflow to the old for rough ballpark estimates of feasability.

PR #4846, opened 2026-08-13
`onError` implementation (GraphQL v17)

benjie:
codex-performed port of #4364 to the GraphQL v17 branch

```
git checkout on-error
git diff 16.x.x > on-error.diff
```

Then codex with prompt:

> on-error.diff contains a diff from 16.x.x to `on-error`, implementing the new onError request argument. Implement these same changes against the current branch, adjusting for the new features in 17.x.x as necessary. At the end `npm run test` should pass, as should `npm run lint`.

Then telling codex off a bit for cheating... ;)


benjie:
Took a while to go through and fix all the shortcuts codex took, but I think we're there.

PR #4852, opened 2026-08-31
Allow empty selection sets - flag version

martinbonnin:
See https://github.com/graphql/graphql-spec/pull/1227

There is a new (and already deprecated!) parser/validation option `allowEmptySelectionSets`. It is true by default (new behaviour). Set it to `false` to switch back to the old behaviour.


magicmark (approved):
🥳

src/validation/validate.ts
```diff
@@ -154,6 +167,7 @@ function validateImpl(
       errors.push(error);
     },
     hideSuggestions,
+    experimentalEmptySelectionSets,
```
benjie: In V18 these should maybe become an options object. In fact, the entire `ValidationContext` class should probably take just one argument: an options object.

src/validation/ValidationContext.ts
```diff
@@ -225,19 +227,23 @@ export class ValidationContext extends ASTValidationContext {
...
    */
+  // eslint-disable-next-line max-params
   constructor(
     schema: GraphQLSchema,
     ast: DocumentNode,
     typeInfo: TypeInfo,
     onError: (error: GraphQLError) => void,
     hideSuggestions?: Maybe<boolean>,
+    experimentalEmptySelectionSets?: Maybe<boolean>,
   ) {
```
benjie, line 263: Maybe we should make this an options object instead:

```suggestion
  constructor(
    schema: GraphQLSchema,
    ast: DocumentNode,
    typeInfo: TypeInfo,
    onError: (error: GraphQLError) => void,
    options?: ValidationContextOptions
  );
  /** @deprecated Please pass options object; this signature will be removed in V18 */
  constructor(
    schema: GraphQLSchema,
    ast: DocumentNode,
    typeInfo: TypeInfo,
    onError: (error: GraphQLError) => void,
    hideSuggestions?: Maybe<boolean>,
  );
  constructor(
    schema: GraphQLSchema,
    ast: DocumentNode,
    typeInfo: TypeInfo,
    onError: (error: GraphQLError) => void,
    hideSuggestionsOrOptions?: Maybe<boolean> | ValidationContextOptions
  ) {
    const options = typeof hideSuggestionsOrOptions === 'boolean' ? { hideSuggestions: hideSuggestionsOrOptions } : hideSuggestionsOrOptions ?? {};
    const { hideSuggestions = false, experimentalEmptySelectionSets = false } = options;
```

benjie (approved):
Solid changes 👍

src/validation/ValidationContext.ts
martinbonnin: @benjie I had to massage that change a little bit to make tests pass etc. GIven I'm not super familiar with the intrications of API compatibility in JS, I'd love a second review pass when you get a chance.

benjie (approved):
LGTM; the options object does add some garbage collection overhead so @yaacovCR might want to trigger the benchmarks, but I imagine it's an incredibly lightweight cost so unlikely to be a big deal.

glen-84:
With the parser now accepting `{ }`, `subscription { }` parses and validates cleanly, but I think it should still fail validation.

The spec's [Single Root Field](https://spec.graphql.org/draft/#sec-Single-Root-Field) rule says the collected fields map "must have exactly one entry", and this PR doesn't change that rule. `SingleFieldSubscriptionsRule` only checks the upper bound though:

```ts
if (groupedFieldSet.size > 1) {
```

That was enough while the grammar guaranteed at least one selection. Now it lets an empty subscription through, and the same goes for `subscription { ...Empty }` with an empty fragment, since that also collects to zero fields.

Suggestion: report when `groupedFieldSet.size === 0` as well, e.g. `Subscription "X" must select a top level field.` (or fold both into `size !== 1`), plus a case in `SingleFieldSubscriptionsRule-test.ts` for the empty operation and the empty-fragment variant.

For reference, Hot Chocolate's implementation of this change ([ChilliCream/graphql-platform#10337](https://github.com/ChilliCream/graphql-platform/pull/10337)) allows empty selection sets on query and mutation roots and on composite fields, but keeps rejecting an empty subscription root.

_Drafted by Claude (Anthropic AI assistant)._

src/validation/rules/SingleFieldSubscriptionsRule.ts
```diff
@@ -104,7 +104,14 @@ export function SingleFieldSubscriptionsRule(
...
           }
-          if (groupedFieldSet.size > 1) {
+          if (node.selectionSet.selections.length === 0) {
+            context.reportError(
+              new GraphQLError(
+                'Subscription must select exactly one top level field.',
+                { nodes: node.selectionSet },
+              ),
+            );
+          } else if (groupedFieldSet.size > 1) {
```
glen-84, line 116: I would just do `if (groupedFieldSet.size !== 1) {`, and update the error message to "must select exactly one".

martinbonnin: Agreed, I like the "exactly one" too. I would also always put the selection set in the `GraphQLErrorOptions` 

```
export interface GraphQLErrorOptions {
  /** AST node or nodes associated with this error. */
  nodes?: ReadonlyArray<ASTNode> | ASTNode | null | undefined;
```

(compared to now the list of extra fields).

Would that be an OK thing to change @yaacovCR @benjie ? Or do we prefer to keep things as is?

benjie: The spec says "exactly one entry, which must not be an introspection field", so I'd just echo that, personally. :+1: 

https://spec.graphql.org/draft/#sec-Single-Root-Field

benjie: Ah... the difference is the error reporting. If >1 then the additional fields are the error, if <1 then the selection set itself is the error.

Scratch what I said before; I'd keep it as-is.

martinbonnin:
@glen-84 sorry I forgot to reply here but the case where a subscription has no root field is now handled in this PR. Thank you and Claude for catching it.

martinbonnin:
@yaacovCR the spec is merged, can we merge this one?

PR #4854, opened 2026-09-04
Allow empty selection sets - no flag version

martinbonnin:
Same as https://github.com/graphql/graphql-js/pull/4852/ but without the experimental flag.


martinbonnin:
Closing in favor of #4852 as discussed in https://github.com/graphql/graphql-spec/pull/1227#issuecomment-5542588216. We want to have a flag to opt out.

